PR #3, opened 2018-10-30
HoTT-UF core library with cubical machinery hidden

martinescardo:
An interface file for users who wish to compute with funext, propositional truncation and univalence in cubical Agda with the cubical machinery hidden.

### DIFF
--- a/Cubical/HoTT-UF.agda
+++ b/Cubical/HoTT-UF.agda
@@ -1,0 +1,101 @@
+{-
+
+This is a HoTT-UF core library based on cubical type theory, where the
+cubical machinery is hidden, using the HoTT Book terminology and
+notation.
+
+The point is that function extensionality, propositional truncation
+and univalence compute (an example is given below).
+
+For the moment, this requires the development version of Agda.
+
+-}
+
+{-# OPTIONS --cubical --exact-split --safe #-}
+
+module Cubical.HoTT-UF where
+
+open import Cubical.Id public
+     using ( _≡_            -- The identity type.
+           ; refl           -- Unfortunately, pattern matching on refl is not available.
+           ; J              -- Until it is, you have to use the induction principle J.
+
+           ; transport      -- As in the HoTT Book.
+           ; ap
+           ; _∙_
+           ; _⁻¹
+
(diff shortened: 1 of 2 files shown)


Cubical/HoTT-UF.agda
```diff
@@ -0,0 +1,101 @@
...
+notation.
+
+The point is that function extensionality, propositional truncation
+and univalence compute (an example is given below).
+
+For the moment, this requires the development version of Agda.
+
+-}
+
+{-# OPTIONS --cubical --exact-split --safe #-}
```
mortberg: What is `--exact-split` and `--safe`? Should these be added to the other files?

Cubical/HoTT-UF.agda
```diff
@@ -0,0 +1,101 @@
...
+  c : pr₁ b ≡ five
+  c = refl
+
+{-
+
+If we had funExt as a postulate, then the definition of c would not
+type check. Moreover, the term pr₁ b would not evaluate to five, as it
+does with the cubical type theory implementation of funext.
+
+TODO. A similar computational example with univalence.
```
mortberg: Do you have any idea for this?

mortberg:
I'm merging this now. I guess we can still discuss my two questions afterwards. Thanks a lot for contributing!

Cubical/HoTT-UF.agda
martinescardo: * [safe](https://agda.readthedocs.io/en/v2.5.4.2/language/safe-agda.html): "By using the command-line option --safe, a user can specify that Agda should ensure that features leading to possible inconsistencies should be disabled."
* [exact split](https://agda.readthedocs.io/en/v2.5.4/tools/command-line-options.html): "Require all clauses in a definition to hold as definitional equalities ". See [case trees](https://agda.readthedocs.io/en/v2.5.4/language/function-definitions.html#case-trees). In practice you get this if you construct your terms using case split in the emacs mode.

mortberg: Thanks! I'll add `--safe` to all of the files.

PR #6, opened 2018-11-02
pathToEquiv and finish proof that Omega(S1) = Int

mortberg:


Cubical/HITs/Circle.agda
```diff
@@ -39,11 +37,59 @@ intLoop (pos (suc n))    = compPath (intLoop (pos n)) loop
 intLoop (negsuc zero)    = sym loop
 intLoop (negsuc (suc n)) = compPath (intLoop (negsuc n)) (sym loop)
 
+-- Why is this not proved with refl?
```
mortberg: @Saizan : For some reason is hcomp with empty systems in Int not the judgmentally the identity function. This makes the proof of windingIntLoop really bad, see below. For normal datatypes like Int we can just remove all hcomps with empty systems. 

I think that we want this more generally, so we should probably implement your `ghcomp` trick.

mortberg:
I'm merging this now, but the thing with hcomp with empty systems for Int should really be fixed.

PR #8, opened 2018-11-16
Another HIT encoding of integers

loic-p:


mortberg:
Great! Thanks for contributing with this example!

PR #10, opened 2018-11-27
S3 is equivalent to the join of two copies of S1

loic-p:
After https://github.com/RedPRL/redtt/blob/master/library/cool/s3-to-join.red

### DIFF
--- a/Cubical/HITs/Join.agda
+++ b/Cubical/HITs/Join.agda
@@ -5,6 +5,7 @@ open import Cubical.Core.Primitives
 open import Cubical.Core.Prelude
 open import Cubical.Core.Glue
 
+open import Cubical.Basics.Equiv
 open import Cubical.HITs.S1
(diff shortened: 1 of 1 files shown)


Cubical/HITs/Join.agda
```diff
@@ -5,34 +5,77 @@ open import Cubical.Core.Primitives
 open import Cubical.Core.Prelude
 open import Cubical.Core.Glue
 
+open import Cubical.Basics.Equiv
 open import Cubical.HITs.S1
 open import Cubical.HITs.S3
 
+-- RedPRL version : https://github.com/RedPRL/redtt/blob/master/library/cool/s3-to-join.red
```
mortberg: This should say "redtt version" (redtt is different from RedPRL...)

mortberg:
Thanks a lot, this is really nice!

PR #11, opened 2018-11-30
More direct Hopf definition

ecavallo:
Not sure if it's faster here, but it's much faster in redtt (https://github.com/RedPRL/redtt/pull/473).


mortberg:
Very nice!

PR #14, opened 2019-01-07
Show that Pi preserves all h-levels

dolio:
- Includes a used lemma that funExt is an equivalence/path

### DIFF
--- a/Cubical/Basics/Function.agda
+++ b/Cubical/Basics/Function.agda
@@ -1,0 +1,29 @@
+{-# OPTIONS --cubical #-}
+
+module Cubical.Basics.Function where
+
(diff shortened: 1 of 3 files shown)


mortberg:
Thank you for this nice contribution! I would be happy to merge if you just move the changes from Core/Glue to somewhere in Basics (maybe a new file or Basics/Equiv).

My plan is to keep the Core library as minimal as possible and keep all the basic theory in Basics.

dolio:
Okay. Will do.

dolio:
I moved it to Cubical.Basics.Function. Equiv would have created a cycle with NTypes.

mortberg:
Great, thanks!

PR #15, opened 2019-01-16
Add suspension of S1 and S2

loic-p:
…and show their equivalence to S2 and S3, respectively.


mortberg:
Thanks, very nice! 

I wish we could give a direct definition of S^n and prove the equivalence in general...

Issue #22, opened 2019-01-17
Make use of the generalization feature?


ice1000:
Now, with `variable` keyword, codes like

```agda
f0 : forall {i} {A : Set i} -> blabla
f1 : forall {i} {A : Set i} -> blabla
```

can be replaced with

```agda
variable
  i : Level
  A : Set i
f0 : blabla
f1 : blabla
```

which makes the code more readable.

PR #24, opened 2019-01-18
Make use of `variable`

ice1000:


ice1000:
#22

Cubical/Core/Glue.agda
```diff
@@ -45,44 +45,50 @@ open import Agda.Builtin.Cubical.Glue public
...
 open isEquiv public
 
-fiber : ∀ {ℓ ℓ'} {A : Set ℓ} {B : Set ℓ'} (f : A → B) (y : B) → Set (ℓ-max ℓ ℓ')
+private
+  variable
+    ℓ ℓ' : Level
+    A : Set ℓ
+    B : Set ℓ'
+
+fiber : ∀ {A : Set ℓ} {B : Set ℓ'} (f : A → B) (y : B) → Set (ℓ-max ℓ ℓ')
```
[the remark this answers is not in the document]
mortberg: I guess it might be because you need to quantify over arbitrary A and B here and not use the A and B from the top variables? If that's the case then I think it would be bit nicer to only use `variable` for the Levels and not A and B.

Cubical/Core/Id.agda
```diff
@@ -64,25 +64,30 @@ open import Cubical.Core.PropositionalTruncation public
 
 {- BUILTIN ID Id -}
 
-_≡_ : ∀ {ℓ} {A : Set ℓ} → A → A → Set ℓ
+private
+  variable
+    ℓ ℓ' : Level
+    A : Set ℓ
+    x y z : A
```
mortberg: I would not declare x y z up here as y is used for other things later.

Cubical/Core/Prelude.agda
```diff
@@ -28,102 +28,104 @@ open import Cubical.Core.Primitives public
...
+_≡⟨⟩_ : (x : A) → x ≡ y → x ≡ y
+_ ≡⟨⟩ x≡y = x≡y
 
-  _≡-qed _∎ : (x : A) → x ≡ x
-  _ ≡-qed  = refl
-  _∎ = _≡-qed
+_≡⟨_⟩_ : (x : A) → x ≡ y → y ≡ z → x ≡ z
+_ ≡⟨ x≡y ⟩ y≡z = compPath x≡y y≡z
 
+_≡-qed _∎ : (x : A) → x ≡ x
```
mortberg: Why is x being parametrized over explicitly here? Same for some of the above definitions.

Cubical/Core/Primitives.agda
```diff
@@ -140,25 +140,31 @@ ouc = primSubOut
 -- When calling "hcomp A φ u a" Agda makes sure that "a" agrees with "u i0" on "φ".
 -- hcomp : ∀ {ℓ} {A : Set ℓ} {φ : I} (u : I → Partial A φ) (a : A) → A
 
+private
+  variable
+    ℓ  : Level
+    ℓ′ : I → Level
+    A′ : Set ℓ
+    φ : I
```
mortberg: I would put A' and phi back below so that the rules read exactly like the typing rule from the papers. Also, why is it A' and not A?

Cubical/Core/PropositionalTruncation.agda
```diff
@@ -19,15 +19,20 @@ data ∥_∥ {ℓ} (A : Set ℓ) : Set ℓ where
   ∣_∣ : A → ∥ A ∥
   squash : ∀ (x y : ∥ A ∥) → x ≡ y
 
-recPropTrunc : ∀ {ℓ} {A : Set ℓ} {P : Set ℓ} → isProp P → (A → P) → ∥ A ∥ → P
+private
+  variable
+    ℓ : Level
+    A : Set ℓ
+
+recPropTrunc : ∀ {A : Set ℓ} {P : Set ℓ} → isProp P → (A → P) → ∥ A ∥ → P
```
mortberg: Why is A explicitly quantified here?

Cubical/Core/Glue.agda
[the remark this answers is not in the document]
ice1000: Because if `A` is generalized, the level of A and `l` will become separated variables, causing some problems.

Cubical/Core/Prelude.agda
ice1000: I think I didn't change the semantics, this question should be asked to the original author...

ice1000:
@mortberg all generalized variables are implicit.

Cubical/Core/PropositionalTruncation.agda
ice1000: Oh! I forgot about that. Will update soon

Cubical/Core/Primitives.agda
ice1000: Because `A` is already used a lot in the rest of the file and they're not `Set l`.

Cubical/Core/Id.agda
ice1000: Considering that unused/shadowed variables are not gonna be generalized, do you still think so?

I'm fine with writing x y z explicitly.

Cubical/Core/PropositionalTruncation.agda
ice1000: No. For the same reason as mentioned above, we need to ensure that the level of `A` and `P` are the same variable.

ice1000: Tried to make them different and it doesn't cause any problem.
Will remove the explicit `A` here.

Cubical/Core/Glue.agda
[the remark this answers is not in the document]
mortberg: I see. Let's remove A and B from the variables and have them explicitly then. It's a bit confusing if we generalize A and B at the top and then still have to abstract over A's and B's in some of the definitions.

Cubical/Core/Id.agda
mortberg: Yes, I'm not a big fan of shadowing the global variables. When reading the file top-down I would get confused if I first think that "y : A" but then y suddenly has another type in some definitions.

Cubical/Core/Prelude.agda
mortberg: I just copy-pasted this code from the standard library :) 

I don't know if this is used anywhere actually... Maybe it should just be removed.

Cubical/Core/Primitives.agda
mortberg: Ok, let's put the types back so that the definitions later look like the typing rules in our papers. Maybe one should format the types of the Kan operations like in https://www.cl.cam.ac.uk/~rio22/agda/cubical-topos/fibrations.html#3142

Cubical/Core/Glue.agda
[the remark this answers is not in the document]
ice1000: Ok.

Cubical/Core/Id.agda
ice1000: Works for me.

(BTW if we have the HTML website, people can click on the generalized variable and jump to their `variable` definitions)

Cubical/Core/Prelude.agda
ice1000: Oh. That's because the `EqReasoning` style reasoning explicitly specifies the last step of the equation transformation, so `x : A` here.

Cubical/Core/Glue.agda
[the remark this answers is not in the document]
ice1000: Done.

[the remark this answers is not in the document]
mortberg: Is it? I might be confused, but A and B are still in the variable list?

Cubical/Core/Id.agda
ice1000: Done.

ice1000:
I think I've got everything done, this PR is now ready-to-merge.

mortberg:
Thanks, merging.

Issue #18, opened 2019-01-17
Make example definitions private?


ice1000:
I see there're definitions of 1024 in different representations of int/nats, it's good to have them. But they're exposed as a public function, it's not what we expect for a library.

Is it better to make them private?

mortberg:
Yes, I think the various experiments and examples should be made private.

PR #25, opened 2019-01-19
Use `private`

ice1000:
#18


Cubical/Basics/Int.agda
```diff
@@ -55,8 +55,7 @@ coherence (negsuc (suc zero)) = refl
 coherence (negsuc (suc (suc n))) = refl
 
 -- Some tests
-module _ where
```
mortberg: Aha, this is how one does this!

mortberg:
Looks good to me. Thanks

Cubical/Basics/Int.agda
ice1000: Yes :)

Issue #20, opened 2019-01-17
Generate the "Everything" modules directly?


ice1000:
I don't think they should be hand written. We can have a script for that purpose.

mortberg:
This would be good to have for the `Basics` and `HITs` folders. In the `Core` folder I prefer to manage it myself because of name clashes and comments. 

I guess the best is if this script would be called from the Makefile before compiling the library? We should then also not track the Everything files in git.

ice1000:
Yes. The Everything modules in the stdlib is generated, I think we can copy them directly and do some slight modification.

mortberg:
Sounds good! Can you do it?

mortberg:
An attempt to solve this was done in: https://github.com/agda/cubical/pull/26

See the discussion for why it wasn't merged. Contributions fixing it are welcome

mortberg:
With https://github.com/agda/cubical/pull/243/files many of the Everything files won't re-export their content. These could be automatically generated.

mortberg:
Solved via https://github.com/agda/cubical/pull/268. Thanks a lot @m-yac !

PR #26, opened 2019-01-20
Add GenerateEverything.hs, remove hand-written everything modules

ice1000:
#20


ice1000:
I added `AllNonAsciiChars` by the way, because it looks good.

GenerateEverything.hs
```diff
@@ -0,0 +1,132 @@
...
+import System.IO
+import System.Exit
+import System.FilePath
+import System.FilePath.Find
+
+headerFile     = "Header"
+allOutputFile  = "Everything"
+srcDirs        = ["Cubical/HITs", "Cubical/Basics"]
+
+excludedModules :: [FilePath]
```
mortberg: I guess the modules listed here won't be listed in the Everything file?

Note that HInt has moved to Experiments now, so I don't think we have to list it. I would also not list HitInt and SmashProduct.

GenerateEverything.hs
```diff
@@ -0,0 +1,132 @@
...
+  , "headers from library modules. The output is written to " ++ allOutputFile
+  , "with the file " ++ headerFile ++ " inserted verbatim at the beginning."
+  ]
+
+-- | Returns 'True' for all Agda files except for core modules.
+
+isLibraryModule :: FilePath -> Bool
+isLibraryModule f =
+  takeExtension f `elem` [".agda", ".lagda"]
+  -- && dropExtension (takeFileName f) /= "Core"
```
mortberg: Why is this commented here? The comment above seems to indicate that it shouldn't be...

GenerateEverything.hs
ice1000: > Why SmashProduct

I see now. Updated.

GenerateEverything.hs
mortberg: So the "Returns ..." comment above is not correct anymore?

mortberg:
I tried to build this locally, but it didn't work because I missed the `FileManip` package. The problem was then that I couldn't get the `FileManip` package to `cabal install` for me... I would prefer if the `agda/cubical` library did not rely on any external Haskell packages. It is already hard enough for people to get the development version of Agda up and running, so requiring them to install even more Haskell packages just to compile the library seems like a bad idea. 

Can you rewrite `GenerateEverything.hs` to be completely self-contained and not rely on any additional Haskell packages?  If not I prefer to have hand-written `Everything.hs` modules and just a simple bash script that checks that everything is included in the `Basics` and `HITs` directories.

GenerateEverything.hs
ice1000: Yes, I'll remove that.

gallais:
Note that in the stdlib we have a [.cabal file](https://github.com/agda/agda-stdlib/blob/master/lib.cabal) which makes installing the required dependencies
to generate the `Everything` module fairly easy.

ice1000:
@mortberg What if I add the .cabal file to this PR?

mortberg:
> @mortberg What if I add the .cabal file to this PR? Is that gonna make this PR acceptable?

Please do it and I'll try it.  My fear is that it will still be too hard to compile things... I've seen how hard it is to just compile the development version of Agda on a variety of platforms and I don't want this library to be yet another obstacle for newcomers.

I think the ideal solution would be to have a file that checks if all the files has been added to the Everything.agda files. This can then be made part of the CI so that the developers can deal with potential issues without adding any extra burden for new users.

gallais:
Potential solution: commit the `Everything.agda` generated by the haskell program?
Then only devs need to bother installing it.

mortberg:
> Potential solution: commit the `Everything.agda` generated by the haskell program?
> Then only devs need to bother installing it.

How does the generated `Everything.agda` files deal with name clashes between modules?

mortberg:
I closed this now as there has been no progress for many months. Feel free to open again if someone fixes it

PR #28, opened 2019-01-21
Implement a cylinder type as a HIT.

dolio:


ice1000:
You may want to use the keyword `variable`.
Referring to the discussion in #24

dolio:
I believe it is the cylinder in this sense: https://ncatlab.org/nlab/show/cylinder+object

They are a dual of the path space, which is sometimes called the cocylinder. It seems like there should be an adjunction between them, too, but maybe that's just intuition about cubical tt as a special case. (I could add some of this to the file, although some of the stuff I tried didn't work. Like showing how `Cylinder A` is like `A × I` didn't work, because you can't eliminate into `I`.)

I did try variables, but it didn't work with the data type (the level is quantified the wrong way). I could use them for the elimination, but it doesn't seem like it'd make a huge difference. I'll change it if desired, though.

dolio:
@ice1000 So, I tried moving `elimCyl` to variables, and it didn't work there, either. It complained about `f`. Is my Agda too old?

mortberg:
Some comments would be great @dolio! (There are also cylinder functors https://ncatlab.org/nlab/show/cylinder+functor ...) Maybe the file should be called CylinderObject?

You could prove that it is equivalent to the pushout of id and id. This should be really easy, but good to have. (You could even take this as the definition, but maybe it is easier to work with a concrete definition)

Also, could you prove that `Cylinder A` is like `A × Interval`? There should be no problem eliminating to and from the HIT interval (https://github.com/agda/cubical/blob/master/Cubical/HITs/Interval.agda)

Cubical/HITs/Cylinder.agda
```diff
@@ -0,0 +1,71 @@
...
+  int-cyl zero = refl
+  int-cyl one = refl
+  int-cyl (seg i) = refl
+
+  cyl-int : ∀ c → cyl (int c) ≡ c
+  cyl-int (inl _) = refl
+  cyl-int (inr _) = refl
+  cyl-int (cross _ i) = refl
+
+  cylIntEquiv : Cylinder Unit ≃ Interval
```
mortberg: I like using ≃ in the name of an equivalence, so this would be: CylinderUnit≃Interval. That way it's very easy to guess the name (even though they get a bit long).

mortberg:
Right, this should be a cylinder functor internally. It might not be too hard to prove that it is left adjoint to exponentiating by the interval (you might have to use `Interval` and not `I`) ?

ice1000:
> @ice1000 So, I tried moving `elimCyl` to variables, and it didn't work there, either. It complained about `f`. Is my Agda too old?

`variable`s are implicit arguments. If you need common explicit arguments you'll need anonymous modules, of course.

So, `ℓ`, `ℓ'` and `B` can be made `variable`s. However, I don't see any points of making `A` an explicit argument (so you may want to make it a `variable` too).

dolio:
Okay, I see. I'll try incorporating more variables.

You're right that `A` shouldn't be an explicit argument to `elimCyl`. If I make it a variable, will it be explicit in `Cylinder A` or do I need to put `elimCyl` in a separate module?

I think I still can't make `ℓ` a variable before `Cylinder` because Agda turns it into an index of the data type for some reason, and it doesn't like that.

dolio:
Okay, I added a bunch of stuff and shuffled other stuff around.

In the end, I decided not to use variables except in one case, because they ended up being more verbose when the file was rearranged.

Let me know if there are any further tweaks you think would be good.

ice1000:
You can still make `ℓ` a `variable` and keep the one after `data` to avoid the issue you've mentioned.

Cubical/HITs/Cylinder.agda
```diff
@@ -7,65 +7,263 @@ open import Cubical.Core.Everything
...
+-- e₀ = inl
+-- e₁ = inr
+-- σ = out
+module Functorial where
+  private
+    variable
+      ℓa ℓb ℓc : Level
+      A : Set ℓa
+      B : Set ℓb
+      C : Set ℓc
```
ice1000: This can be made
```agda
    variable
      ℓ : Level
      A B C : Set ℓ
```

`A B C` will have different levels.

Cubical/HITs/Cylinder.agda
```diff
@@ -7,65 +7,263 @@ open import Cubical.Core.Everything
...
+    isoToEquiv
+      CylinderUnit→Interval
+      Interval→CylinderUnit
+      Interval→CylinderUnit→Interval
+      CylinderUnit→Interval→CylinderUnit
+
+
+  -- More generally, there is an equivalence between the cylinder
+  -- over any type A and the product of A and the interval.
+  module _ {ℓ} {A : Set ℓ} where
```
ice1000: This `ℓ` and `A` can be a `variable` too, I believe.

Cubical/HITs/Cylinder.agda
```diff
@@ -7,65 +7,263 @@ open import Cubical.Core.Everything
...
+    CylinderA≃A×Interval : Cylinder A ≃ Cyl
+    CylinderA≃A×Interval =
+      isoToEquiv
+        CylinderA→A×Interval
+        A×Interval→CylinderA
+        A×Interval→CylinderA→A×Interval
+        CylinderA→A×Interval→CylinderA
+
+-- The cylinder is also the pushout of the identity on A with itself.
+module Push {ℓ} {A : Set ℓ} where
```
ice1000: Ditto.

Cubical/HITs/Cylinder.agda
dolio: What is the point of this, though? It's just adding 4 extra lines for something that already fits easily on an existing line.

dolio: Oh, also, it actually can't be made a variable. I forgot, but I tried it, and Agda complained about it being used in e.g. `Cyl` below.

Saizan: @ice1000 I don't think we should push `variable` as much as you are suggesting. Modules are great at abstracting some parameters out of a bunch of definitions. 

In general I would use `variable` to establish specific naming conventions, not for local arguments.

>  I just want to avoid indenting definitions

Do you have a specific reason for that? Other than saving horizontal space?

ice1000: > @ice1000 I don't think we should push variable as much as you are suggesting.

Yes, I agree. My final suggestion is to push the levels to `variable` but Agda does not support using `variable`s in GADT definitions, which has probably confused OP.

> Do you have a specific reason for that?

Yes,

> Other than saving horizontal space?

Then no.

ice1000:
You may consider adding your module to the list in `Cubical.HITs.Everything`.

Cubical/HITs/Cylinder.agda
dolio: By the way, is there documentation on the `variable` feature somewhere? I would not have guessed that it works this way.

mortberg: Same here, this behavior is very surprising to me

Saizan: Currently it's only documented in the CHANGELOG it seems:

https://github.com/agda/agda/blob/master/CHANGELOG.md

Issue #12, opened 2018-12-13
Include clearer installation instructions


Alizter:
Currently the readme talks about making an agda sandbox, but doesn't specify what to do afterwards. For those who aren't so familiar with agda, this isn't completely obvious. Could you add some clearer installation instructions?

Alizter:
Its not entirely clear how to compile and install the cubical library. The makefile talks about "agda" but it hasn't been fully installed in the sandbox anyway. I don't know how to do this in the agda sandbox, so I have run `make install` after making the sandbox, making the development version my main version of agda.

Saizan:
For those not familiar with agda we could certainly have more detailed instructions.

If you don't mind, could you tell us if the following helps?

If you have agda master installed by following these instructions, given in the readme:
```
git clone https://github.com/agda/agda
cd agda
cabal sandbox init
cabal install
```

Then you want to "install" the emacs mode by running

```
agda-mode setup
```

Then you can just open the files from this repo in emacs and work with them. The emacs mode is briefly described here
https://agda.readthedocs.io/en/latest/getting-started/quick-guide.html#quick-guide-introduction


If you want to use cubical as a library in your own development you have to "register" its .agda-lib file as described here:
https://agda.readthedocs.io/en/latest/tools/package-system.html?highlight=library

Alizter:
After running `cabal install` I cannot run `agda-mode setup`. There is no command, nor has agda been installed.

Saizan:
`cabal install` by default installs things in a local directory, on linux it's `~/.cabal/bin`, maybe you do not have that in your `$PATH`.

Alizter:
I've checked inside `~/.cabal/bin` there is no agda or agda-mode.

Saizan:
I see, I don't use `cabal sandbox` myself, it seems you indeed have to copy the files from the sandbox to somewhere in your `$PATH`

https://stackoverflow.com/questions/27274475/cabal-install-an-executable-from-hackage-via-a-sandbox

@mortberg what do you do?

mortberg:
I just update my PATH in `~/.bashrc`:
```
PATH=/home/mortberg/dev/agda/.cabal-sandbox/bin:$PATH
```

Writing some detailed install instructions in an INSTALL file sounds like a good idea. Maybe @Alizter can write a first draft once you get it to work? 

We should not forget to include some instructions for how to set it up as a library.

Alizter:
[...]

PR #30, opened 2019-01-23
Add INSTALL instructions

mortberg:
This resolves issue https://github.com/agda/cubical/issues/12

@Alizter and @Saizan: Can you please take a look and let me know if they are correct and sufficiently detailed.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -80,9 +80,11 @@ works in emacs run
 and then type C-c C-l. This should now load the file and you can start
 developing your own cubical files.
 
-You can also register cubical as a library to depend on it in your own Agda developments:
+You can also register cubical as a library to depend on it in your own
+Agda developments:
 
 https://agda.readthedocs.io/en/latest/tools/package-system.html
+
 
 stack install instructions
 ==========================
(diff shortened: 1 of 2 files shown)


Alizter:
LGTM

INSTALL
```diff
@@ -0,0 +1,137 @@
...
+> make
+
+This should compile all of the agda/cubical files. To test that it
+works in emacs run
+
+> emacs Cubical/Core/Primitives.agda
+
+and then type C-c C-l. This should now load the file and you can start
+developing your own cubical files.
+
```
Saizan:
```suggestion
You can also register cubical as a library to depend on it in your own Agda developments:

https://agda.readthedocs.io/en/latest/tools/package-system.html
```

Saizan:
I made a suggestion about referencing the library system, but I guess it should be copied to the stack instructions too.

Otherwise it looks good.

mortberg:
Thanks Andrea, that's something I had to help someone with as well!

ice1000:
Why it's not a `.md`?

PR #32, opened 2019-01-23
Add some operations and properties for Stream

ice1000:


ice1000:
I've updated the Makefile file by the way so it cleans `*.Rewrite` modules.

Alizter:
LGTM

mortberg:
LGTM, merging

PR #33, opened 2019-01-23
Trying to cache Agda's dist folder

ice1000:
I noticed that even if we cache all the dependencies, the master Agda still takes 10+ minutes to build.
This is too slow. Let's try cache Agda's dist folder as well.


ice1000:
This does not seem to work. Closing.

PR #34, opened 2019-01-25
Derive some of the cartesian coe operations using transp

mortberg:
I mainly did this for fun to clarify the connection between transp and the generalized coe operations of Cartesian Cubical Type Theory. It might be easier for some people to use the cartesian style operations instead of the transp operation as they are maybe a bit more intuitive.


mortberg:
I'm merging this now. 

@ecavallo Can you please make a new PR with your additional results and fixes?

PR #37, opened 2019-01-29
Update the binary nat  examples and prove more induction principles for equivalences and isomorphisms

mortberg:
The `propDoubleℕ` lemma does not typecheck as fast as I'm expecting and I'm hoping that some Agda expert can help me

### DIFF
--- a/Cubical/Basics/BinNat.agda
+++ b/Cubical/Basics/BinNat.agda
@@ -80,7 +80,7 @@ oddℕ : ℕ → Bool
 oddℕ = transp (λ i → ℕ≡binnat (~ i) → Bool) i0 oddbinnat
 
 module _ where
-  -- Define what it means to be an interface for naturals
+  -- Define what it means to be an interface for natural numbers
   record impl (A : Set) : Set where
     field
       z : A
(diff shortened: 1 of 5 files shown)


Cubical/Basics/BinNat.agda
```diff
@@ -174,165 +44,330 @@ posInd {P} h1 hs ps = f ps
...
+  -- propDoubleℕ : propDouble Doubleℕ
+  -- propDoubleℕ = refl
 
   -- With binary numbers it is instant
-  propBin : propDouble DoubleBinN
-  propBin = refl
+  propDoubleBinℕ : propDouble DoubleBinℕ
+  propDoubleBinℕ = refl
+
+  -- Why is this not typechecking fast?
```
mortberg: Can some Agda expert help me understand why Agda so eagerly unfolds `propDouble Doubleℕ` in the below piece of code? Is there some way to locally block reduction of some definition without making it globally blocked? 

@Saizan @andreasabel

Saizan: @mortberg Agda does not take advantage of the fact that `propDouble` does not need to be reduced to typecheck this term.

Two terms are either syntactically equal or they get reduced to whnf and compared.
This is because in principle `propDouble foo` and `propDouble bar` could be equal even if `foo` and `bar` are not.

I have tried to set things up so that we could rely on the syntactic equality, but I end up with terms that internally have different representation of universe levels, unfortunately. Maybe syntactic equality should be a little more lenient here.

A simple way around reducing `propDouble` is to turn it into a record type with a single field though.

mortberg: Thanks! Turning it into a record worked! Can you explain why?

Saizan: Because record (or data) types do not reduce further, so they get compared by name and equality of arguments.

e.g. if `R` is a record type with one argument, then `R A` and `R B` are already in WHNF, and their equality is determined by comparing `A` and `B`.

mortberg: Excellent, thanks! So whenever I have computationally expensive term that shouldn't be reduced I'll wrap it in a record (in Coq we do something similar using Modules).

ice1000: FYI you can also put functions under an `abstract` block to prevent them from being reduced.

mortberg: Yes, but that makes them globally blocked without the possibility of being un-blocked again right?

mortberg:
I fixed the thing that didn't work and polished the file. If @Saizan has no more comments I suggest we merge?

Saizan:
@mortberg Looks ok to me, maybe squash the commits when merging

Cubical/Basics/BinNat.agda
ice1000: IIUC yes.

PR #38, opened 2019-01-30
Correct a typo

3abc:


Saizan:
Thanks for the pull request, but it is actually not a typo, `IsOne` does not support `transp` so it's relegated into `Setω` (at least until we have a better universe for it).

https://agda.readthedocs.io/en/latest/language/universe-levels.html#expressions-of-kind-set

3abc:
@Saizan Did I misunderstand? The current code says in comment `IsOne : I -> Set` and I think what is correct is `IsOne : I -> Setω`. That's why I want to correct the typo. (I want to ω in instead of removing it.)

Saizan:
@3abc oh you're right, i read the diff wrong!

Issue #39, opened 2019-01-30
Library structure - follow standard library?


MatthewDaggitt:
This library seems to be getting a lot of development effort at the moment, and it looks like it's going to develop into a useful base library for those interested in using cubical Agda. I'm not sure how much thought has been put into its layout? I was wondering if you'd considered using the structure of the standard library (where applicable)?

The advantages would be:
1. It would make it far easier for many users to make the switch from non-cubical to cubical Agda (and visa-versa).
2. In the standard library we've already put quite a lot of effort into designing a structure that is reasonably extensible. Adopting it now would hopefully save quite a lot of non-backwards compatible content shuffling later on in this library's life.

Obviously I realise it wouldn't be possible to mimic everything!

Saizan:
I think the main requirement is to keep `Cubical.Core` quite minimal. The rest is quite flexible I think.

Can you elaborate on how it would look if we followed the structure of the standard library and how that would help with extensibility?

MatthewDaggitt:
> I think the main requirement is to keep Cubical.Core quite minimal. The rest is quite flexible I think.

Yup no need to enlarge that.

> Can you elaborate on how it would look if we followed the structure of the standard library

So for example much of the `Basics` folder would be rearranged as follows:

- Data
  - Bool
    - Properties
  - BinNat
    - Properties
  - Empty
  - Int
    - Properties
  - Nat
    - Properties
  - Unit
  - Sum (currently in `Cubical.Basics.NTypes`)
  - Product (currently in `Cubical.Core.Prelude`)
- Codata
  - Stream
- Function

Not sure where `Univalence` and `Equiv` fit. Some other folder?

The major point is that in the standard library we've found that mixing definitions and proofs is a common cause of dependency cycles. The inclusion of the `Properties` files above tries to address this. 

At the moment the definition, various operations and the properties of `Nat` are all mixed together. This will almost certainly cause problems when you want to use machinery that depends on the definition of `Nat` but not the properties. I hypothesis that you're **already** running into this problem and that's why the definition of `Nat` is not currently actually exported by `Cubical.Basics.Nat` but is exported in `Cubical.Core.Prelude`...

> How that would help with extensibility?

[...]

PR #49, opened 2019-02-01
Reorganize datatypes

mortberg:
This PR partially resolves https://github.com/agda/cubical/issues/39 by making `Cubical.Data` and `Cubical.Codata` folders. These folders then contain all of the datatypes in separate subfolders. All of the files about datatypes are split into `Base` and `Properties` where `Base` contains the definition of the type and all operations on it and `Properties` contain all the properties of the types and operations. 

This made it possible to move some stuff from the `Core`, in particular the import of Nat which was problematic as both the file with theory about Nat and NTypes needed it, but these couldn't import one another which was why I put the import of Nat in the Core. 

The patch also renames the Basics library to Foundations.

The only file that has not been split into Base/Properties is the BinNat file. The reason is that the file is written in a tutorial style fashion. I might make a blog post based on it some day and then I'd be happy to split it and reorganize, but for now I want to keep it as it is.

I will do more reorganization in a subsequent patch, in particular I am planning to make a `Relation` folder and reorganize the `HITs` folder as the `Data` folder.


ice1000:
Please be careful: the merge of this PR will cause issue #39 to be closed

Cubical/Data/Empty.agda
```diff
@@ -0,0 +1,6 @@
+{-# OPTIONS --cubical --safe #-}
+module Cubical.Data.Empty where
+
+open import Cubical.Data.Empty.Base public
+
+open import Cubical.Data.Empty.Properties public
```
ice1000: Maybe you've forgotten to track this file with git.

PR #50, opened 2019-02-01
Reorganize HITs folder

mortberg:
I did a similar reorganization to the HITs folder as in the Data folder. However the distinction between definitions and properties is a lot more blurry when working with HITs, so at the moment none of the HIT files have been splits into Base/Properties. For example, the fact that the map from "Susp S1" to "S2" is an equivalence is a property, but we need that proof in order to construct the equality between the types using univalence (and because of this the equality itself will be computational). Similar things happen when doing more HoTT inspired proofs about basic datatypes.

Instead of having Base/Properties it might be more reasonable to split the files into Def which just contains the datatype declaration and then a Properties file that contain everything else (or more fine-grained files with specific results about the types). Any opinions?


mortberg:
I'll merge this now, but let's still have the discussion about Base/Properties vs Def/Properties.

Saizan:
I suppose for extensibility what we want is being able to refer to the type without creating an import loop? Though there might be other things that are part of the core interface of a type and will be needed wherever its imported.

@MatthewDaggitt what do you think should be in .Base vs. .Properties for exensibility's sake?

mortberg:
Ok, I think I might implement my suggestion. I do think that it is often good to distinguish programs and their properties in vanilla type theory, but in HoTT the distinction is very blurry because everything is so proof relevant. In particular non-trivial equalities can be constructed using univalence and transporting along these are then programs that will do things in cubical type theory.

PR #53, opened 2019-02-02
Define a cool version of addition for integers using univalence

mortberg:
This a nice and simple example of programming with univalence. The addition function also seems to compute reasonably fast! However proving properties about it was a pain because of hcomp's with empty systems, so I decided not to do it now. I might try to finally implement ghcomp first which should make it a lot less painful (and hopefully even more efficient) to work with definitions like this.

### DIFF
--- a/Cubical/Data/Everything.agda
+++ b/Cubical/Data/Everything.agda
@@ -4,8 +4,8 @@ module Cubical.Data.Everything where
 open import Cubical.Data.BinNat public
 open import Cubical.Data.Bool public
 open import Cubical.Data.Empty public
-open import Cubical.Data.Int public
 open import Cubical.Data.Nat public
+open import Cubical.Data.Int public renaming (_+_ to _+Int_)
 open import Cubical.Data.Sum public
 open import Cubical.Data.Prod public
 open import Cubical.Data.Unit public
--- a/Cubical/Data/EverythingSafe.agda
+++ b/Cubical/Data/EverythingSafe.agda
@@ -4,7 +4,7 @@ module Cubical.Data.EverythingSafe where
 open import Cubical.Data.BinNat public
 open import Cubical.Data.Bool public
 open import Cubical.Data.Empty public
-open import Cubical.Data.Int public
+open import Cubical.Data.Int renaming (_+_ to _+Int_) public
 open import Cubical.Data.Nat public
 open import Cubical.Data.Sum public
 open import Cubical.Data.Prod public
--- a/Cubical/Data/Int/Properties.agda
+++ b/Cubical/Data/Int/Properties.agda
@@ -7,7 +7,7 @@ open import Cubical.Foundations.Equiv
 open import Cubical.Foundations.HLevels
 
 open import Cubical.Data.Empty
-open import Cubical.Data.Nat
+open import Cubical.Data.Nat hiding (_+_ ; +-assoc)
 open import Cubical.Data.Sum
 open import Cubical.Data.Int.Base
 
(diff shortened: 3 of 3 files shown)


3abc:
Probably

```
addEq' : ℕ → Int ≡ Int
addEq' zero = refl
addEq'(suc n) = compPath (addEq n) sucPathInt

subEq' : ℕ → Int ≡ Int
subEq' n i = addEq n (~ i)

_+'_ : Int → Int → Int
m +' pos n    = transport (addEq' n) m
m +' negsuc n = transport (subEq' (suc n)) m
```

is a better definition. (The positions of `(addEq n)` and `sucPathInt` are swapped). 

It's tricky because `transport (compPath p q) a` becomes `transport q (transport p a)`, where `p` and `q` are swapped.

mortberg:
Did you notice any difference in performance when adding bigger numbers between the two definitions? Last summer I tried similar things in cubicaltt with very different performance depending on how I nested the hcomp's. 

Feel free to redefine the operation however you want. I bet many of the properties follow by `J` and then some fiddling with `transportRefl`

3abc:
I didn't notice anything particular. (Is there a way I can time how long does Agda take to normalize an expression in command line?) Performance seems similar (both takes some time to compute `pos 3000 + negsuc 2000`) but it seems the `+'` version is slightly better.

To prove associativity I expected to use `uabeta` since `_+_` is defined by `ua`, however I found it now doable without any explicit use of `uabeta`.

I did found many things easier but mostly are just `refl`, and the proof is still not very trivial.  Doesn't seems to need `transportRefl` at all.

mortberg:
That sounds very good! Can you make a PR to this PR?

mortberg:
PS: I have no clue how to time things with Agda, but I would very much like to have it

Saizan:
> PS: I have no clue how to time things with Agda, but I would very much like to have it

There's the flag `-vprofile.interactive:10` that will give give you a time for commands like `C-c C-n`

3abc:
> flag `-vprofile.interactive:10` that will give give you a time for commands like `C-c C-n`

I just tried this. Turns out the original `_+_` was a lot faster.

`pos 1000 + negsuc 999` ------> `Time: 3,084ms`
`pos 1000 +' negsuc 999` ------> `Time: 77,788ms`

3abc:
Of course both are a lot slower than

```
+pos : ℕ → Int → Int
+pos 0 z = z
+pos (suc n) z = sucInt (+pos n z)

+negsuc : ℕ → (Int → Int)
+negsuc 0 = predInt
+negsuc (suc n) z = predInt (+negsuc n z)

_+''_ : Int → Int → Int
m +'' pos n = +pos n m
m +'' negsuc n = +negsuc n m
```

mortberg:
I see, can you prove that the various addition functions are equal in your other PR?

3abc:
> I see, can you prove that the various addition functions are equal in your other PR?

Sure. How about calling the one defined via `+pos` and `+negsuc` by `_+_` and calling the one defined via `addEq n` by `_+'_`? So that when people use `_+_` they will get the fast version. And `_+'_` is used to prove addition by a fixed number is an equivalence.

mortberg:
> > I see, can you prove that the various addition functions are equal in your other PR?
> 
> Sure. How about calling the one defined via `+pos` and `+negsuc` by `_+_` and calling the one defined via `addEq n` by `_+'_`? So that when people use `_+_` they will get the fast version. And `_+'_` is used to prove addition by a fixed number is an equivalence.

Sure, that sounds good to me! I guess you noticed that I merged the other PR, so you make a new one once you have everything implemented.

3abc:
> There's the flag `-vprofile.interactive:10` that will give give you a time for commands like `C-c C-n`
> 
> In emacs, the Agda2 customize-group has a `Agda2 Program Args` variable for flags if you want it always on.

Is there a flag that allows me to profile how long it takes to type check a file (or a particular term)?

PR #57, opened 2019-02-04
Add Conat

ice1000:
Add conat, prove that mugen plus one equals mugen :)

### DIFF
--- a/Cubical/Codata/Conat/Base.agda
+++ b/Cubical/Codata/Conat/Base.agda
@@ -1,0 +1,13 @@
+{-# OPTIONS --cubical --safe --guardedness #-}
+module Cubical.Codata.Conat.Base where
+
+open import Cubical.Core.Everything
+
+data Pred (A : Set) : Set where
+  pzero : Pred A
+  psuc  : A → Pred A
+
+record Conat : Set where
+  coinductive
+  constructor conat
+  field pred : Pred Conat
--- a/Cubical/Codata/Conat/Base.agda
+++ b/Cubical/Codata/Conat/Base.agda
@@ -10,4 +10,4 @@ data Pred (A : Set) : Set where
 record Conat : Set where
   coinductive
   constructor conat
-  field pred : Pred Conat
+  field prev : Pred Conat
--- a/Cubical/Codata/Conat/Base.agda
+++ b/Cubical/Codata/Conat/Base.agda
@@ -1,3 +1,19 @@
+{- Conatural numbers (Tesla Ice Zhang, Feb. 2019)
+
+This file defines a coinductive natural number representation which is dual to
+the inductive version (zero | suc Nat → Nat) of natural numbers and some
+trivial operations (succ, pred) and the pattern synonyms on it.
+
+While this definition can be seen as a coinductive wrapper of an inductive
+family, another way of definition is to define an inductive family that wraps
+a coinductive thunk of Nat. The standard library uses the second approach:
+
+https://github.com/agda/agda-stdlib/blob/master/src/Codata/Conat.agda
+
+The first approach is chosen to exploit guarded recursion and to avoid the use
+of Sized Types.
+-}
+
 {-# OPTIONS --cubical --safe --guardedness #-}
 module Cubical.Codata.Conat.Base where
 
--- a/Cubical/Codata/Conat/Base.agda
+++ b/Cubical/Codata/Conat/Base.agda
@@ -18,4 +18,11 @@ pattern zero  = inl tt
 pattern suc n = inr n
 
 succ : Conat → Conat
-force (succ a) = inr a
+force (succ a) = suc a
+
+succ′ : Conat′ → Conat′
+succ′ n = suc (conat n)
+
+pred′ : Conat′ → Conat′
+pred′  zero   = zero
+pred′ (suc x) = force x
--- a/Cubical/Codata/Conat/Base.agda
+++ b/Cubical/Codata/Conat/Base.agda
@@ -1,8 +1,11 @@
 {- Conatural numbers (Tesla Ice Zhang, Feb. 2019)
 
-This file defines a coinductive natural number representation which is dual to
-the inductive version (zero | suc Nat → Nat) of natural numbers and some
-trivial operations (succ, pred) and the pattern synonyms on it.
+This file defines:
+
+- A coinductive natural number representation which is dual to
+  the inductive version (zero | suc Nat → Nat) of natural numbers.
+
+- Trivial operations (succ, pred) and the pattern synonyms on conaturals.
 
 While this definition can be seen as a coinductive wrapper of an inductive
 family, another way of definition is to define an inductive family that wraps
--- a/Cubical/Codata/Conat/Base.agda
+++ b/Cubical/Codata/Conat/Base.agda
@@ -21,8 +21,12 @@ succ : Conat → Conat
 force (succ a) = suc a
 
 succ′ : Conat′ → Conat′
-succ′ n = suc (conat n)
+succ′ n = suc λ where .force → n
 
 pred′ : Conat′ → Conat′
 pred′  zero   = zero
 pred′ (suc x) = force x
+
+pred′′ : Conat′ → Conat
+force (pred′′ zero) = zero
+pred′′ (suc x) = x
--- a/Cubical/Codata/Conat/Base.agda
+++ b/Cubical/Codata/Conat/Base.agda
@@ -7,16 +7,15 @@ open import Cubical.Data.Sum
 open import Cubical.Core.Everything
 
 record Conat : Set
-Prev = Unit ⊎ Conat
+Conat′ = Unit ⊎ Conat
 record Conat where
   coinductive
   constructor conat
-  field prev : Prev
+  field force : Conat′
 open Conat public
 
 pattern zero  = inl tt
 pattern suc n = inr n
 
 succ : Conat → Conat
-prev (succ a) = inr a
-
+force (succ a) = inr a
--- a/Cubical/Codata/Conat/Base.agda
+++ b/Cubical/Codata/Conat/Base.agda
@@ -8,8 +8,9 @@ This file defines:
 - Trivial operations (succ, pred) and the pattern synonyms on conaturals.
 
 While this definition can be seen as a coinductive wrapper of an inductive
-family, another way of definition is to define an inductive family that wraps
-a coinductive thunk of Nat. The standard library uses the second approach:
+datatype, another way of definition is to define an inductive datatype that
+wraps a coinductive thunk of Nat.
+The standard library uses the second approach:
 
 https://github.com/agda/agda-stdlib/blob/master/src/Codata/Conat.agda
 
--- a/Cubical/Codata/Conat/Base.agda
+++ b/Cubical/Codata/Conat/Base.agda
@@ -10,12 +10,15 @@ record Conat : Set
 Conat′ = Unit ⊎ Conat
 record Conat where
   coinductive
-  constructor conat
+  constructor conat′
   field force : Conat′
 open Conat public
 
 pattern zero  = inl tt
 pattern suc n = inr n
+
+conat : Conat′ → Conat
+force (conat a) = a
 
 succ : Conat → Conat
 force (succ a) = suc a
--- a/Cubical/Codata/Conat/Base.agda
+++ b/Cubical/Codata/Conat/Base.agda
@@ -1,13 +1,22 @@
 {-# OPTIONS --cubical --safe --guardedness #-}
 module Cubical.Codata.Conat.Base where
 
+open import Cubical.Data.Unit
+open import Cubical.Data.Sum
+
 open import Cubical.Core.Everything
 
-data Pred (A : Set) : Set where
-  pzero : Pred A
-  psuc  : A → Pred A
-
-record Conat : Set where
+record Conat : Set
+Prev = Unit ⊎ Conat
+record Conat where
   coinductive
   constructor conat
-  field prev : Pred Conat
+  field prev : Prev
+open Conat public
+
+pattern zero  = inl tt
+pattern suc n = inr n
+
+succ : Conat → Conat
+prev (succ a) = inr a
+
--- a/Cubical/Codata/Conat/Properties.agda
+++ b/Cubical/Codata/Conat/Properties.agda
@@ -1,0 +1,44 @@
+{-# OPTIONS --cubical --safe --guardedness #-}
+module Cubical.Codata.Conat.Properties where
+
+open import Agda.Builtin.Unit
+
+open import Cubical.Core.Everything
+
+open import Cubical.Codata.Conat.Base
+
+open Conat
+
+succ : Conat → Conat
+succ a = conat (psuc a)
+
+Unwrap-pred : Pred Conat -> Set
+Unwrap-pred  pzero   = ⊤
+Unwrap-pred (psuc _) = Conat
+
+unwrap-pred : (n : Pred Conat) -> Unwrap-pred n
+unwrap-pred  pzero   = _
+unwrap-pred (psuc x) = x
+
+private -- tests
+  zero = conat pzero
+  one  = succ zero
+  two  = succ one
+
+  succOne≡two : succ one ≡ two
+  succOne≡two i = two
+
+  predTwo≡one : unwrap-pred (pred two) ≡ one
+  predTwo≡one i = one
+
+∞ : Conat
+pred ∞ = psuc ∞
+
+∞+1≡∞ : succ ∞ ≡ ∞
+pred (∞+1≡∞ _) = psuc ∞
+
+∞+2≡∞ : succ (succ ∞) ≡ ∞
+∞+2≡∞ = compPath (cong succ ∞+1≡∞) ∞+1≡∞
+
+-- TODO: plus for conat, ∞ + ∞ ≡ ∞
+
--- a/Cubical/Codata/Conat/Properties.agda
+++ b/Cubical/Codata/Conat/Properties.agda
@@ -42,7 +42,34 @@ force (∞+1≡∞ _) = suc ∞
 
 -- TODO: plus for conat, ∞ + ∞ ≡ ∞
 
+conat-absurd : ∀ {y : Conat} {ℓ} {Whatever : Set ℓ} → zero ≡ suc y → Whatever
+conat-absurd eq = ⊥-elim (transport (cong diag eq) tt)
+  where
+  diag : Conat′ → Set
+  diag zero = Unit
+  diag (suc _) = ⊥
+
+module EqStable where
+  ≡-stable  : {x y : Conat} → Stable (x ≡ y)
+  ≡′-stable : {x y : Conat′} → Stable (x ≡ y)
+
+  force (≡-stable ¬¬p i) = ≡′-stable (λ ¬p → ¬¬p (λ p → ¬p (cong force p))) i
+  ≡′-stable {zero}  {zero}  ¬¬p′ = refl
+  ≡′-stable {suc x} {suc y} ¬¬p′ =
+     cong′ suc (≡-stable λ ¬p → ¬¬p′ λ p → ¬p (cong pred′′ p))
+  ≡′-stable {zero}  {suc y} ¬¬p′ = ⊥-elim (¬¬p′ conat-absurd)
+  ≡′-stable {suc x} {zero}  ¬¬p′ = ⊥-elim (¬¬p′ λ p → conat-absurd (sym p))
+
+  ≡-unique : {m n : Conat} (p q : m ≡ n) → p ≡ q
+  ≡-unique = Stable≡→isSet (λ _ _ → ≡-stable) _ _
+
+  ≡′-unique : {m n : Conat′} (p q : m ≡ n) → p ≡ q
+  ≡′-unique {m′} {n′} p′ q′ = cong (cong force) (≡-unique {conat m′} {conat n′} p q)
+    where p = λ where i .force → p′ i
+          q = λ where i .force → q′ i
+
 module Bisimulation where
+  open EqStable using (≡-unique)
 
   record _≈_ (x y : Conat) : Set
   data _≈′_ (x y : Conat′) : Set
@@ -61,13 +88,6 @@ module Bisimulation where
 
   open _≈_ public
 
-  zsuc-inv : ∀ {y : Conat} {ℓ} {Whatever : Set ℓ} → zero ≡ suc y → Whatever
-  zsuc-inv eq = ⊥-elim (transport (cong diag eq) tt)
-    where
-    diag : Conat′ → Set
-    diag zero = Unit
-    diag (suc _) = ⊥
-
   bisim : ∀ {x y} → x ≈ y → x ≡ y
   bisim′ : ∀ {x y} → x ≈′ y → x ≡ y
 
@@ -81,8 +101,8 @@ module Bisimulation where
   misib′ : ∀ {x y} → x ≡ y → x ≈′ y
 
   misib′ {zero} {zero} _ = con tt
-  misib′ {zero} {suc x} = zsuc-inv
-  misib′ {suc x} {zero} p = zsuc-inv (sym p)
+  misib′ {zero} {suc x} = conat-absurd
+  misib′ {suc x} {zero} p = conat-absurd (sym p)
   -- misib′ {suc x} {suc y} p = con λ where .prove → misib′ (cong pred′ p)
   misib′ {suc x} {suc y} p = con (misib (cong pred′′ p))
   prove (misib x≡y) = misib′ (cong force x≡y)
@@ -95,24 +115,6 @@ module Bisimulation where
   iso′ {suc x} {zero} (con ())
   iso′ {suc x} {suc y} (con p) = cong con (iso p)
   prove (iso p i) = iso′ (prove p) i
-
-  ≡-stable  : {x y : Conat} → Stable (x ≡ y)
-  ≡′-stable : {x y : Conat′} → Stable (x ≡ y)
-
-  force (≡-stable ¬¬p i) = ≡′-stable (λ ¬p → ¬¬p (λ p → ¬p (cong force p))) i
-  ≡′-stable {zero}  {zero}  ¬¬p′ = refl
-  ≡′-stable {suc x} {suc y} ¬¬p′ =
-     cong′ suc (≡-stable λ ¬p → ¬¬p′ λ p → ¬p (cong pred′′ p))
-  ≡′-stable {zero}  {suc y} ¬¬p′ = ⊥-elim (¬¬p′ zsuc-inv)
-  ≡′-stable {suc x} {zero}  ¬¬p′ = ⊥-elim (¬¬p′ λ p → zsuc-inv (sym p))
-
-  ≡-unique : {m n : Conat} (p q : m ≡ n) → p ≡ q
-  ≡-unique = Stable≡→isSet (λ _ _ → ≡-stable) _ _
-
-  ≡′-unique : {m n : Conat′} (p q : m ≡ n) → p ≡ q
-  ≡′-unique {m′} {n′} p′ q′ = cong (cong force) (≡-unique {conat m′} {conat n′} p q)
-    where p = λ where i .force → p′ i
-          q = λ where i .force → q′ i
 
   osi : ∀ {x y} → (p : x ≡ y) → bisim (misib p) ≡ p
   osi p = ≡-unique _ p
--- a/Cubical/Codata/Conat/Properties.agda
+++ b/Cubical/Codata/Conat/Properties.agda
@@ -40,15 +40,19 @@ force (∞+1≡∞ _) = suc ∞
 -- TODO: plus for conat, ∞ + ∞ ≡ ∞
 
 record _~_ (x y : Conat) : Set
-_~′_ : Conat′ → Conat′ → Set
-zero  ~′ zero  = Unit
-suc x ~′ suc y = x ~ y
+data _~′_ (x y : Conat′) : Set
+_~′′_ : Conat′ → Conat′ → Set
+zero  ~′′ zero  = Unit
+suc x ~′′ suc y = x ~ y
 -- So impossible proofs are preserved
-x ~′ y = x ≡ y
+x ~′′ y = x ≡ y
 
 record _~_ x y where
   coinductive
   field prove : force x ~′ force y
+
+data _~′_  x y where
+  con : x ~′′ y → x ~′ y
 
 open _~_ public
 
@@ -56,16 +60,16 @@ bisim : ∀ {x y} → x ~ y → x ≡ y
 force (bisim eq i) = bisim′ (prove eq) i
   where
   bisim′ : ∀ {x y} → x ~′ y → x ≡ y
-  bisim′ {zero} {zero} tt = refl
-  bisim′ {zero} {suc x} p = p
-  bisim′ {suc x} {zero} p = p
-  bisim′ {suc x} {suc y} eq i = suc (bisim eq i)
+  bisim′ {zero} {zero}  (con tt) = refl
+  bisim′ {zero} {suc x} (con p) = p
+  bisim′ {suc x} {zero} (con p) = p
+  bisim′ {suc x} {suc y} (con eq) i = suc (bisim eq i)
 
 misib : ∀ {x y} → x ≡ y → x ~ y
 prove (misib x≡y) = misib′ (cong force x≡y)
   where
   misib′ : ∀ {x y} → x ≡ y → x ~′ y
-  misib′ {zero} {zero} p = _
-  misib′ {zero} {suc x} p = p
-  misib′ {suc x} {zero} p = p
-  misib′ {suc x} {suc y} p = λ where .prove → misib′ (cong pred′ p)
+  misib′ {zero} {zero} p = con tt
+  misib′ {zero} {suc x} p = con p
+  misib′ {suc x} {zero} p = con p
+  misib′ {suc x} {suc y} p = con λ where .prove → misib′ (cong pred′ p)
--- a/Cubical/Codata/Conat/Properties.agda
+++ b/Cubical/Codata/Conat/Properties.agda
@@ -13,6 +13,8 @@ This file defines operations and properties on conatural numbers:
 - Proof that bisimulation is equivalent to equivalence (Coinductive Proof
   Principle).
 
+- Proof that this bisimulation is an hProp
+
 The standard library also defines bisimulation on conaturals:
 
 https://github.com/agda/agda-stdlib/blob/master/src/Codata/Conat/Bisimilarity.agda
@@ -145,3 +147,11 @@ module Bisimulation where
   path≡bisim : ∀ {x y} → (x ≡ y) ≡ (x ≈ y)
   path≡bisim = ua path≃bisim
 
+  isProp≈ : ∀ {x y} → isProp (x ≈ y)
+  prove (isProp≈ p q i) = isProp≈′ (prove p) (prove q) i
+    where
+    isProp≈′ : ∀ {x y} → isProp (x ≈′ y)
+    isProp≈′ {zero} {zero} (con tt) (con tt) = refl
+    isProp≈′ {zero} {suc x} (con ()) (con ())
+    isProp≈′ {suc x} {zero} (con ()) (con ())
+    isProp≈′ {suc x} {suc y} (con p) (con q) = cong con (isProp≈ p q)
--- a/Cubical/Codata/Conat/Properties.agda
+++ b/Cubical/Codata/Conat/Properties.agda
@@ -8,6 +8,8 @@ open import Cubical.Data.Empty
 open import Cubical.Core.Everything
 open import Cubical.Foundations.Equiv
 
+open import Cubical.Relation.Nullary
+open import Cubical.Relation.Nullary.DecidableEq
 open import Cubical.Codata.Conat.Base
 
 Unwrap-prev : Conat′ -> Set
@@ -46,10 +48,10 @@ module Bisimulation where
   record _≈_ (x y : Conat) : Set
   data _≈′_ (x y : Conat′) : Set
   _≈′′_ : Conat′ → Conat′ → Set
-  -- zero  ≈′′ zero  = Unit
+  zero  ≈′′ zero  = Unit
   suc x ≈′′ suc y = x ≈ y
   -- So impossible proofs are preserved
-  x ≈′′ y = x ≡ y
+  x ≈′′ y = ⊥
 
   record _≈_ x y where
     coinductive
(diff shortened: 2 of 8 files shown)


Cubical/Codata/Stream/Base.agda
```diff
@@ -3,8 +3,6 @@ module Cubical.Codata.Stream.Base where
 
 open import Cubical.Core.Everything
 
-open import Cubical.Data.Nat
```
ice1000: I've found that this is unused so I deleted it

Cubical/Codata/Conat/Properties.agda
```diff
@@ -0,0 +1,44 @@
+{-# OPTIONS --cubical --safe --guardedness #-}
+module Cubical.Codata.Conat.Properties where
+
+open import Agda.Builtin.Unit
```
mortberg: Why not import https://github.com/agda/cubical/blob/master/Cubical/Data/Unit.agda ?

mortberg:
The code looks fine to me, but is there anything univalent you can do with the co-natural numbers?

ice1000:
I can't think of any. Maybe proving bisimulation on conat is equivalent to path?
I'll have a try.

Happy Chinese new year!

Cubical/Codata/Conat/Properties.agda
ice1000: I didn't know that. Will fix soon

mortberg:
That sounds like a good idea! Maybe you can construct some non-trivial equivalence on conat as well? Or to some other type. You can then transport some things back and forth just to check that transp works for coinductives

mortberg:
Xīnnián kuàilè! (I hope Google translate got that right)

ice1000:
Yes the pronunciation is correct!

I'm truggling with the definition of plus. TBH, I don't like Sized Types. I'm trying to get rid of it (while the Conat in the stdlib is completely based on Sized Types)

Cubical/Codata/Conat/Base.agda
```diff
@@ -0,0 +1,13 @@
+{-# OPTIONS --cubical --safe --guardedness #-}
+module Cubical.Codata.Conat.Base where
+
+open import Cubical.Core.Everything
+
+data Pred (A : Set) : Set where
```
Blaisorblade: Is `pred` a good name here? It's short for `predecessor`, but you write `zero = conat pzero`, and the predecessor of 0 is not 0... the predecessor function here is `\ x -> unwrap-pred (pred x)`, as shown by  `unwrap-pred (pred two) ≡ one`...

martinescardo:
A univalent thing you can do is to show that conat is equivalent and hence equal to the type of decreasing binary sequences, which is how I define conat here: 

http://www.cs.bham.ac.uk/~mhe/agda-new/CoNaturals.html
http://www.cs.bham.ac.uk/~mhe/agda-new/CoNaturalsExercise.html

Cubical/Codata/Conat/Base.agda
ice1000: Do you have any recommendations? I agree with you but I failed to find a better name.

ice1000:
I've just found that bisimulation on a sum type is quite hard to define since some definitional equalities are lost due to guarded recursion. I'll take a look at @martinescardo@'s code.

martinescardo:
When I did this in 2010-2011 (then extended in 2012 with coinduction), I was not satisfied with Agda's native treatment of coinduction/corecursion, and this is why I implemented conat without Agda's coinduction features.  Also, the way I am using to encode conat comes from traditional constructive mathematics. However, it would be nice to do everything I did with native support for coinduction/corecursion. Perhaps you can suggest an idiom to expresses what you need as a feature request. But I believe it should be possible, without additional support, to show that my conat is equivalent (and hence equal) to your conat, no?

martinescardo:
Can you be more precise in what you would like to be able to write, with a concrete example involving conat?

ice1000:
You can see my unfinished proof in https://github.com/Agda-zh/cubical/tree/conat-bisi

martinescardo:
Thanks. But your link is broken.

ice1000:
### Complete explanation

A little prelude to this comment, just plain coproduct and product:

```agda
open import Agda.Builtin.Sigma
_×_ : ∀ {ℓ ℓ'} (A : Set ℓ) (B : Set ℓ') → Set (ℓ-max ℓ ℓ')
A × B = Σ A (λ _ → B)

data _⊎_ {ℓ ℓ'} (P : Set ℓ) (Q : Set ℓ') : Set (ℓ-max ℓ ℓ') where
  inl : P → P ⊎ Q
  inr : Q → P ⊎ Q
```

I have a conat which is a coinductive wrapper on a `A → Unit ⊎ A` instead of an inductive wrapper on a Thunk (to avoid using Sized Types):

```agda
record Conat : Set
Prev = Unit ⊎ Conat
record Conat where
  coinductive
  constructor conat
  field prev : Prev

pattern zero  = inl tt
pattern suc n = inr n

succ : Conat → Conat
prev (succ a) = inr a
```

I'm trying to define a bisimulation relation on it, so I came up with this definition:

```agda
module Bisimulation where
 data _≠0 : (a : Prev) → Set where
    no0 : ∀ {a} → suc a ≠0

  pred : Prev → Conat
  prev (pred  zero  ) = zero
  prev (pred (suc x)) = prev x

  -- Maybe this definition is not good enough?
  record _~_ (a b : Conat) : Set where
    coinductive
    constructor intro
    field proof : (a ≡ b) ⊎ ((pred (prev a) ~ pred (prev b)) × (prev a ≠0 × prev b ≠0))
  open _~_
```

With `with` abstraction and copatterns, I can easily define an identity function:

```agda
  identity : ∀ n → n ~ n
  proof (identity n) with prev n
  ... | zero  = inl refl
  ... | suc x = inr (identity (pred (suc x)) , no0 , no0)
```

And I failed to define a `bisim` with a signature like this:

```agda
bisim : ∀ {a b} → a ~ b → a ≡ b
```

Because with abstraction cannot be used with path patterns, I can't use a seem-to-better definition like the one in `identity`:

```agda
  -- Fake code, rejected because copattern is not supported in path patterns
  proof (bisim {a} {b} p i) with proof p
```

I tried to use with abstraction without path types (and of course, without copatterns), and I come up with a code like this (have to pattern match on `prev a` and `prev b` to indicate that they are not `zero`):

```agda
 bisim {a} {b} p with prev a | prev b | proof p
  ... | _     | _     | inl x = x
  ... | suc x | suc y | suc (bi , no0 , no0) = cong {!succ !} (bisim bi)
```

Agda rejects the hole because `succ (pred (suc x)) != a`.

Since `suc x` is equal to `prev a` definitionally, we can use Agda's `inspect` idiom, which becomes the code you see in the unfinished branch. However, Agda no longer knows that `proof p`'s second member is `no0 , no0` because all `suc x`s becomes `prev a`s so the information that "a is never 0" is lost (which causes incomplete pattern error).

I tried a little to use the eq-reasoning style proving, as in the unfinished branch, and it causes termination error.

I have no idea about how to prove this. I hope I'm making stupid mistake here.

ice1000:
I have no problem viewing that. Is it private? I don't think it's private:
image

martinescardo:
Ah, it is here: https://github.com/Agda-zh/cubical/blob/conat-bisi/Cubical/Codata/Conat/Properties.agda

Cubical/Codata/Conat/Base.agda
ice1000: Renamed.

Cubical/Codata/Conat/Base.agda
```diff
@@ -0,0 +1,22 @@
+{-# OPTIONS --cubical --safe --guardedness #-}
+module Cubical.Codata.Conat.Base where
+
+open import Cubical.Data.Unit
+open import Cubical.Data.Sum
+
+open import Cubical.Core.Everything
+
+record Conat : Set
+Prev = Unit ⊎ Conat
```
Saizan: This type is isomorphic to `Conat`, so `Prev` seems like a bad name.

Cubical/Codata/Conat/Base.agda
```diff
@@ -0,0 +1,22 @@
...
+record Conat where
+  coinductive
+  constructor conat
+  field prev : Prev
```
Saizan: Here I would stick to `force` for the field, because indeed it's not the previous natural number that you are projecting out.

Saizan:
@ice1000 Here's a way to do bisimulation for `Conat` 31f7de105f76a9e7de5efecd44943dcd8cd78219

You can pull that commit into your branch if you like it, so we can stay on this PR. (or is there a better way?)

I defined `_~′_` by recursion to avoid inductive families since they are not supported by cubical yet, otherwise an inductive definition would be better style.

Also, IMO `Conat′` should be its own datatype with `zero` and `suc` constructors, as that works better for case splitting and overloading.

Cubical/Codata/Conat/Base.agda
ice1000: Maybe there should not be an alias for this simple type?

Saizan: That might be the best option, but see my comment in the conversation about this.

ice1000: Well having a name for it will be helpful for refactoring though. I'll keep it, thanks!

Cubical/Codata/Conat/Base.agda
ice1000: Looks good!

ice1000:
I feel like what you're doing is manually lifting the `with` auxiliary function and it sounds like a good idea!

I like the new style mutual recursion so I'll refactor your code a little bit and do the rest of the work.

ice1000:
> Also, IMO `Conat′` should be its own datatype with `zero` and `suc` constructors, as that works better for case splitting and overloading.

This is very opinion-based and actually I was using a standalone datatype and then switched to a sigma type. I'm gonna keep this anyway.

Cubical/Codata/Conat/Properties.agda
```diff
@@ -0,0 +1,71 @@
...
+  bisim′ {suc x} {suc y} eq i = suc (bisim eq i)
+
+misib : ∀ {x y} → x ≡ y → x ~ y
+prove (misib x≡y) = misib′ (cong force x≡y)
+  where
+  misib′ : ∀ {x y} → x ≡ y → x ~′ y
+  misib′ {zero} {zero} p = _
+  misib′ {zero} {suc x} p = p
+  misib′ {suc x} {zero} p = p
+  misib′ {suc x} {suc y} p = λ where .prove → misib′ (cong pred′ p)
```
ice1000: I'm having a termination error here and I don't expect one (this should work! It's guarded recursion!).

ice1000: image
Still.

Saizan: Seems like Agda is not quite seeing `.prove` as guarding enough.
It works if you add a datatype in the mix, like in this commit  dba72a633971ca593017864a3d1641a2daa9d9e0

potato4444:
I was also working on Conats. I haven't pushed them to github yet. I did have `bisim` and `misib` defined until I was reminded by Saizan's comment that inductive families aren't supported, so I changed am in the process of changing my implementation.

Cubical/Codata/Conat/Properties.agda
potato4444: I tried defining `misib` via a `refl~` but that also fails the termination checker. It seems a extra layer of data is the way to go.

ice1000: I would say there should be some improvements on Agda's termination checking on guarded recursion...

Cubical/Codata/Conat/Properties.agda
```diff
@@ -39,37 +41,64 @@ force (∞+1≡∞ _) = suc ∞
...
+  iso1′ {suc x} {suc y} p i = p
+  prove (iso1 p i) = iso1′ (prove p) i
+
+  iso2 : ∀ {x y} → (p : x ≡ y) → bisim (misib p) ≡ p
+  iso2′ : ∀ {x y} → (p : x ≡ y) → bisim′ (misib′ p) ≡ p
+
+  iso2′ {zero} {zero} p _ = p
+  iso2′ {zero} {suc x} p _ = p
+  iso2′ {suc x} {zero} p _ = p
+  iso2′ {suc x} {suc y} p i j = suc (iso2 (cong pred′′ p) i j)
```
ice1000: Now I'm having a unification error here (CI should show that) and I'm kind of stucking on it.

The target Path should be a Path between `p` and `suc (bisim (misib (cong pred‘’ p)) i)`, if I understand it correctly.

gallais:
This is what I would do:

```agda
  record _≈_ (x y : Conat) : Set
  data _≈′_ : (x y : Conat′) → Set

  record _≈_ x y where
    coinductive
    field force : force x ≈′ force y

  data _≈′_ where
    zero' : zero ≈′ zero
    suc'  : ∀ {x y} → x ≈ y → suc x ≈′ suc y

  open _≈_

  cong′ : ∀ {a b} {A : Set a} {B : Set b} (f : A → B) {x y : A} (p : x ≡ y)
       → PathP (λ i → B) (f x) (f y)
  cong′ f p = λ i → f (p i)

  zero-suc-inv : ∀ {y : Conat} → zero ≡ suc y → ⊥
  zero-suc-inv eq = transport (cong′ diag eq) tt where

    diag : Conat′ → Set
    diag zero    = Unit
    diag (suc _) = ⊥

  bisim : ∀ {x y} → x ≈ y → x ≡ y
  bisim′ : ∀ {x y} → x ≈′ y → x ≡ y

  force (bisim p i) = bisim′ (p .force) i
  bisim′ zero'    = refl
  bisim′ (suc' p) = cong′ suc (bisim p)

  misib : ∀ {x y} → x ≡ y → x ≈ y
  misib′ : ∀ {x y} → x ≡ y → x ≈′ y

  force (misib p) = misib′ (cong′ force p)
  misib′ {zero}  {zero}  p = zero'
  misib′ {suc x} {suc y} p = suc' (misib (cong′ pred′′ p))
  misib′ {zero}  {suc y} p = ⊥-elim (zero-suc-inv p)
  misib′ {suc x} {zero}  p = ⊥-elim (zero-suc-inv (sym p))

  iso  : ∀ {x y} → (p : x ≈ y) → misib (bisim p) ≡ p
  iso′ : ∀ {x y} → (p : x ≈′ y) → misib′ (bisim′ p) ≡ p

  force (iso p i) = iso′ (p .force) i
  iso′ zero'    = refl
  iso′ (suc' p) = cong′ suc' (iso p)
```

potato4444:
@gallais I had essentially this definition until I saw @Saizan message about inductive families. I don't know what problems they cause if any. I also tried emulating inductive families by storing equalities, but that just seemed to anger the termination checker in the definition of `bisim`

Cubical/Codata/Conat/Properties.agda
ice1000: It's expanded by the compiler to `p != suc (iso2 (λ i → pred′′ (p i)) j)`

ice1000:
@gallais Yeah that looks quite good. I'll go through the definition to study it

Saizan:
@gallais @potato4444 For now `transp` and `hcomp` do not compute for Inductive families, and definitions by pattern matching would get stuck on those.

ice1000:
Yes, I have found some code accepted by goal-filling but does not typecheck afterwards:

```agda
 iso2  : ∀ {x y} → (p : x ≡ y) → bisim (misib p) ≡ p
  iso2′ : ∀ {x y} → (p : x ≡ y) → bisim′ (misib′ p) ≡ p

  force (iso2 p i j) = iso2′ (cong force p) i j
  iso2′ {zero} {zero} p = refl
  iso2′ {suc x} {suc y} p = cong (cong suc) (iso2 (cong pred′′ p))
  iso2′ {zero} {suc x} p = ⊥-elim (zero-suc-inv p)
  iso2′ {suc x} {zero} p = ⊥-elim (zero-suc-inv (sym p))
```

gallais:
I've tried to have a look at the other side of the iso and I'm stuck. :/

ice1000:
My development branch: https://github.com/Agda-zh/cubical/blob/inductive-conat/Cubical/Codata/Conat/Properties.agda

gallais:
I am down to a termination error in `osi` which I don't understand because e.g. `stable`
seems to use the same induction strategy?

```agda
{-# OPTIONS --cubical --safe --guardedness #-}
module Cubical.Codata.Conat.Properties where

open import Cubical.Data.Unit
open import Cubical.Data.Sum
open import Cubical.Data.Empty

open import Cubical.Foundations.Equiv

open import Cubical.Relation.Nullary
open import Cubical.Relation.Nullary.DecidableEq
open import Cubical.Codata.Conat.Base

Unwrap-prev : Conat′ -> Set
Unwrap-prev  zero   = Unit
Unwrap-prev (suc _) = Conat

unwrap-prev : (n : Conat′) -> Unwrap-prev n
unwrap-prev  zero   = _
unwrap-prev (suc x) = x

private -- tests
  𝟘 : Conat
  force 𝟘 = zero
  one  = succ 𝟘
  two  = succ one

  succOne≡two : succ one ≡ two
  succOne≡two i = two

  predTwo≡one : unwrap-prev (force two) ≡ one
  predTwo≡one i = one

∞ : Conat
force ∞ = suc ∞

∞+1≡∞ : succ ∞ ≡ ∞
force (∞+1≡∞ _) = suc ∞

∞+2≡∞ : succ (succ ∞) ≡ ∞
∞+2≡∞ = compPath (cong succ ∞+1≡∞) ∞+1≡∞

-- TODO: plus for conat, ∞ + ∞ ≡ ∞

cong′ : ∀ {a b} {A : Set a} {B : Set b} (f : A → B) {x y : A} (p : x ≡ y)
      → PathP (λ i → B) (f x) (f y)
cong′ f p = λ i → f (p i)

zero-suc-inv : ∀ {y : Conat} → zero ≡ suc y → ⊥
zero-suc-inv eq = transport (cong′ diag eq) tt where

  diag : Conat′ → Set
  diag zero    = Unit
  diag (suc _) = ⊥

module Bisimulation where

  _≈′?_ : (x y : Conat′) → Set
  data _≈′_ (x y : Conat′) : Set
  record _≈_ (x y : Conat) : Set

  zero  ≈′? zero  = Unit
  suc x ≈′? suc y = x ≈ y
  x     ≈′? y     = ⊥

  data _≈′_ x y where
    [_] : x ≈′? y → x ≈′ y

  record _≈_ x y where
    coinductive
    field force : force x ≈′ force y
  open _≈_ public

  bisim : ∀ {x y} → x ≈ y → x ≡ y
  bisim′ : ∀ {x y} → x ≈′ y → x ≡ y

  force (bisim p i) = bisim′ (p .force) i

  bisim′ {zero}  {zero}  [ p′ ] = refl
  bisim′ {suc x} {suc y} [ p′ ] = cong′ suc (bisim p′)

  misib : ∀ {x y} → x ≡ y → x ≈ y
  misib′ : ∀ {x y} → x ≡ y → x ≈′ y

  force (misib p) = misib′ (cong′ force p)
  misib′ {zero}  {zero}  p′ = [ _ ]
  misib′ {suc x} {suc y} p′ = [ misib (cong′ pred′′ p′) ]
  misib′ {zero}  {suc y} p′ = ⊥-elim (zero-suc-inv p′)
  misib′ {suc x} {zero}  p′ = ⊥-elim (zero-suc-inv (sym p′))

  iso : ∀ {x y} → (p : x ≈ y) → misib (bisim p) ≡ p
  iso′ : ∀ {x y} → (p : x ≈′ y) → misib′ (bisim′ p) ≡ p

  force (iso p i) = iso′ (p .force) i
  iso′ {zero}  {zero}  [ p ] = refl
  iso′ {suc x} {suc y} [ p ] = cong [_] (iso p)

  p0≡pi : ∀ {a} {A : Set a} {x y : A} (p : x ≡ y) → ∀ i → x ≡ p i
  p0≡pi {a} {A} {x} {y} p i j =
    hcomp (λ k → λ { (j = i0) → x
                   ; (j = i1) → p i
                   })
          (p (i ∧ j))

  _∋_ : ∀ {a} (A : Set a) → A → A
  A ∋ a = a

  ≡-stable  : {x y : Conat} → Stable (x ≡ y)
  ≡′-stable : {x y : Conat′} → Stable (x ≡ y)

  force (≡-stable ¬¬p i) = ≡′-stable (λ ¬p → ¬¬p (λ p → ¬p (cong′ force p))) i
  ≡′-stable {zero}  {zero}  ¬¬p′ = refl
  ≡′-stable {suc x} {suc y} ¬¬p′ =
    cong′ suc (≡-stable λ ¬p → ¬¬p′ (λ p → ¬p (cong′ pred′′ p )))
  ≡′-stable {zero}  {suc y} ¬¬p′ = ⊥-elim (¬¬p′ zero-suc-inv)
  ≡′-stable {suc x} {zero}  ¬¬p′ = ⊥-elim (¬¬p′ λ p → zero-suc-inv (sym p))

  ≡-unique : {m n : Conat} (p q : m ≡ n) → p ≡ q
  ≡-unique = Stable≡→isSet (λ a b → ≡-stable) _ _

  ≡′-unique : {m n : Conat′} (p q : m ≡ n) → p ≡ q
  ≡′-unique {m′} {n′} p′ q′ i = cong′ force (≡-unique {m} {n} p q i)
    where m = (λ where   .force → m′)  ; n = (λ where .force → n′)
          p = (λ where i .force → p′ i); q = (λ where i .force → q′ i)

  osi : ∀ {x y} → (p : x ≡ y) → bisim (misib p) ≡ p
  osi′ : ∀ {x y} → (p : x ≡ y) → bisim′ (misib′ p) ≡ p

  force (osi p i j) = osi′ (cong′ force p) i j
  osi′ {zero}  {zero}  p′ = ≡′-unique _ _
  osi′ {suc x} {suc y} p′ =
    (λ i → suc (bisim (misib (λ j → pred′′ (p′ j))) i))
      ≡⟨ ≡′-unique _ _ ⟩
    cong′ suc (bisim (misib (λ j → pred′′ (p′ j))))
      ≡⟨ cong′ (cong′ suc) (osi (cong′ pred′′ p′)) ⟩
    (λ i → suc (pred′′ (p′ i)))
      ≡⟨ ≡′-unique _ _ ⟩
    p′
      ∎
  osi′ {zero}  {suc y} p′ = ⊥-elim (zero-suc-inv p′)
  osi′ {suc x} {zero}  p′ = ⊥-elim (zero-suc-inv (sym p′))

  path≃bisim : ∀ {x y} → (x ≡ y) ≃ (x ≈ y)
  path≃bisim = isoToEquiv misib bisim iso osi

  path≡bisim : ∀ {x y} → (x ≡ y) ≡ (x ≈ y)
  path≡bisim = ua path≃bisim
```

gallais:
Ok, I'm daft: `osi` is a direct corrolary of `≡-unique`...

```agda
  osi : ∀ {x y} (p : x ≡ y) → bisim (misib p) ≡ p
  osi p = ≡-unique _ _
```

ice1000:
Ok, I've pushed @gallais@'s proof on my branch. Now it's ready to go.

gallais:
`zsuc-inv`, `≡-stable` and `≡-unique` should probably be outside of the `Bisimulation`
submodule as they are not related to bisimulation.

`zsuc-inv` irks me because `zero` is shortened but not `suc`. May be worth calling it
`no-confusion` after the usual lemma stating that an equality between two distinct
constructors is absurd.

ice1000:
Addressed @gallais@'s comment. Merge?

Cubical/Core/Prelude.agda
```diff
@@ -45,6 +45,10 @@ cong : ∀ {B : A → Set ℓ'} (f : (a : A) → B a) (p : x ≡ y)
        → PathP (λ i → B (p i)) (f x) (f y)
 cong f p = λ i → f (p i)
 
+cong′ : ∀ {B : Set ℓ'} (f : A → B) {x y : A} (p : x ≡ y)
+      → Path B (f x) (f y)
+cong′ f p = λ i → f (p i)
```
mortberg: Why is this here? Can't you just use `cong`?

mortberg:
Could you please add a header to the Base/Properties files saying what they contain? (we should probably do start doing this for all files...)

Cubical/Core/Prelude.agda
gallais: The fact that `B` is guaranteed to not be dependent makes constraint solving easier.
Sometimes if you use `cong` instead you get unsolved metas.

That's why we have a lot of primed combinators in [`Function`](https://agda.github.io/agda-stdlib/Function.html) in the stdlib.

mortberg: Oh I see! I guess we should have this as well then. Thanks for clarifying.

mortberg: Actually, @ice1000 can you please add a comment explaining this? Otherwise I will surely forget and delete this one day...

Cubical/Codata/Conat/Properties.agda
```diff
@@ -0,0 +1,127 @@
...
+
+  force (≡-stable ¬¬p i) = ≡′-stable (λ ¬p → ¬¬p (λ p → ¬p (cong force p))) i
+  ≡′-stable {zero}  {zero}  ¬¬p′ = refl
+  ≡′-stable {suc x} {suc y} ¬¬p′ =
+     cong′ suc (≡-stable λ ¬p → ¬¬p′ λ p → ¬p (cong pred′′ p))
+  ≡′-stable {zero}  {suc y} ¬¬p′ = ⊥-elim (¬¬p′ conat-absurd)
+  ≡′-stable {suc x} {zero}  ¬¬p′ = ⊥-elim (¬¬p′ λ p → conat-absurd (sym p))
+
+  ≡-unique : {m n : Conat} (p q : m ≡ n) → p ≡ q
+  ≡-unique = Stable≡→isSet (λ _ _ → ≡-stable) _ _
```
potato4444: Should this be named some variation on `isSetConat`? Also isn't that type signature exactly `isSet Conat`

Cubical/Core/Prelude.agda
ice1000: Yes I can.

Cubical/Codata/Conat/Properties.agda
ice1000: Fixed

ice1000:
Made all comments following the style in `BinNat` by @mortberg@, added comments on `cong'`, renamed `=-unique` into `isSetConat`.

Cubical/Codata/Conat/Properties.agda
```diff
@@ -1,3 +1,19 @@
...
+
+Coinduction makes infinity definable, and we can easily prove that ∞ + 1 is
+equivalent to ∞ by coinduction.
+
+A proof that conatrual is a hSet is included.
+
+Bisimulation on conatural and a proof that bisimulation is equivalent to
+equivalence (Coinductive Proof Principle) are also included. The standard
+library also defines bisimulation on conatruals:
+
```
gallais: Typo: conatruals -> conaturals

ice1000: Fixed

Cubical/Codata/Conat/Base.agda
```diff
@@ -0,0 +1,54 @@
...
+
+This file defines:
+
+- A coinductive natural number representation which is dual to
+  the inductive version (zero | suc Nat → Nat) of natural numbers.
+
+- Trivial operations (succ, pred) and the pattern synonyms on conaturals.
+
+While this definition can be seen as a coinductive wrapper of an inductive
+family, another way of definition is to define an inductive family that wraps
```
potato4444: The use of "inductive family" makes it seem like we are using indices, but we are not. Maybe just "inductive type", or possibly something else to that effect.

potato4444:
It is a fairly simple corollary of the fact the `conat` is a set and the equivalence between bisimilution and identity, but do we want an explicit proof of bisumlation being a proposition?

ice1000:
>It is a fairly simple corollary of the fact the `conat` is a set and the equivalence between bisimilution and identity, but do we want an explicit proof of bisumlation being a proposition?

I don't think it's possible to extract a general definition of bisimulation as a proposition, see our discussion above (we've come up with more than three definitions, each of them sounds amazingly reasonable but they all fail due to some mysteries until Mr. @gallais@'s final version based on `isSet`).

ice1000:
IIUC defining bisimulation on a coinductive structure requires you to "look into" the structure, which means you'll need to define a specific bisimulation for every coinductive structures.

If you have any idea about defining such general bisimulation proposition (except creating the definition by reflection) please explain.

Cubical/Codata/Conat/Base.agda
ice1000: Fixed.

gallais:
I think @potato4444 means that bisimulation satisfies [`isProp`](https://github.com/agda/cubical/blob/master/Cubical/Core/Id.agda#L159).

ice1000:
Proved that bisimulation is a prop.

Cubical/Codata/Conat/Properties.agda
```diff
@@ -145,3 +147,11 @@ module Bisimulation where
   path≡bisim : ∀ {x y} → (x ≡ y) ≡ (x ≈ y)
   path≡bisim = ua path≃bisim
 
+  isProp≈ : ∀ {x y} → isProp (x ≈ y)
+  prove (isProp≈ p q i) = isProp≈′ (prove p) (prove q) i
```
gallais: Can't you prove it instead by combining `isSetConat` and `path≡bisim` using `transport`?

gallais: Answering my own question:

```agda
  isProp≈ : ∀ {x y} → isProp (x ≈ y)
  isProp≈ p q =
    p               ≡⟨ sym (iso p) ⟩
    misib (bisim p) ≡⟨ cong misib (isSetConat _ _ _ _) ⟩
    misib (bisim q) ≡⟨ iso q ⟩
    q               ∎
```

ice1000: Is there some synonyms that can do all these (`p < sym (bla p) > orz < something isSetSth > zro < bla q > q`)?

gallais: More generally:

```agda

  SectionToProp-isProp :
    ∀ {a b} {A : Set a} {B : Set b} (f : A → B) g →
    section f g → isProp A → isProp B
  SectionToProp-isProp f g fg propA x y =
    x       ≡⟨ sym (fg x) ⟩
    f (g x) ≡⟨ cong′ f (propA _ _) ⟩
    f (g y) ≡⟨ fg y ⟩
    y ∎

  isProp≈ : ∀ {x y} → isProp (x ≈ y)
  isProp≈ {x} {y} = SectionToProp-isProp misib bisim iso (isSetConat x y)
```

ice1000: I like this! I'll put it into the Equiv module.

ice1000: We should have a function that proves "equivalence on hSet is hProp".. I'm trying to find an existing one in the library.

ice1000: Finally get a simplest proof: ` subst isProp path≡bisim (isProp[Set≡] isSetConat)`

ice1000:
Please take a look @mortberg@

Cubical/Codata/Conat/Properties.agda
```diff
@@ -148,10 +148,4 @@ module Bisimulation where
...
 
   isProp≈ : ∀ {x y} → isProp (x ≈ y)
-  prove (isProp≈ p q i) = isProp≈′ (prove p) (prove q) i
-    where
-    isProp≈′ : ∀ {x y} → isProp (x ≈′ y)
-    isProp≈′ {zero} {zero} (con tt) (con tt) = refl
-    isProp≈′ {zero} {suc x} (con ()) (con ())
-    isProp≈′ {suc x} {zero} (con ()) (con ())
-    isProp≈′ {suc x} {suc y} (con p) (con q) = cong con (isProp≈ p q)
+  isProp≈ = subst isProp path≡bisim (isProp[Set≡] isSetConat)
```
gallais: Ah! So the solution was to use `subst` and not `transport`.
I need to get a better intuition for one vs. the other.

Cubical/Codata/Conat/Properties.agda
mortberg: > We should have a function that proves "equivalence on hSet is hProp".. I'm trying to find an existing one in the library.

I don't understand what you're looking for... Regardless, the library is missing a lot of theory about HLevels/NTypes and equivalences. I was going to make an issue about this, but I forgot.

mortberg: > More generally:
> 
> ```agda
>   SectionToProp-isProp :
>     ∀ {a b} {A : Set a} {B : Set b} (f : A → B) g →
>     section f g → isProp A → isProp B
>   SectionToProp-isProp f g fg propA x y =
>     x       ≡⟨ sym (fg x) ⟩
>     f (g x) ≡⟨ cong′ f (propA _ _) ⟩
>     f (g y) ≡⟨ fg y ⟩
>     y       ∎
> 
>   isProp≈ : ∀ {x y} → isProp (x ≈ y)
>   isProp≈ = SectionToProp-isProp misib bisim iso (isSetConat _ _)
> ```

Nice, this should be included in `Foundations.HLevels`. I would restate it using retraction instead though. The general form of the lemma says that any retract `A` of an n-type `B` is an n-type (i.e. is of h-level n+2). This can of course be stated using a section instead, but at least in my head it is clearer what is going on if I think of it as a retraction as it says that `A` is a nice subtype of `B`.

Cubical/Core/Prelude.agda
```diff
@@ -132,3 +137,9 @@ isProp A = (x y : A) → x ≡ y
 
 isSet : Set ℓ → Set ℓ
 isSet A = (x y : A) → isProp (x ≡ y)
+
+-- Equivalence on hSets are hProps
+
+isProp[Set≡] : ∀ {A : Set ℓ} → isSet A →
```
mortberg: This is just changing the implicits in the definition of `isSet`, so I don't understand why it should be necessary. Also it should not be added to the Core. Lemmas of this kind should go into `Foundations.HLevels`. However I really don't see why this should be necessary at all.

Cubical/Core/Prelude.agda
```diff
@@ -45,6 +45,11 @@ cong : ∀ {B : A → Set ℓ'} (f : (a : A) → B a) (p : x ≡ y)
        → PathP (λ i → B (p i)) (f x) (f y)
 cong f p = λ i → f (p i)
 
+-- Less polymorphic version of `cong`, to avoid some unresolved metas
+cong′ : ∀ {B : Set ℓ'} (f : A → B) {x y : A} (p : x ≡ y)
```
mortberg: This should go into a suitable file in `Foundations` and not the core. I'm not sure which file though... Where does `cong` go in the standard library?

Cubical/Codata/Conat/Properties.agda
mortberg: What is `subst` and `transport` called in the standard library?

It might be more intuitive for computer scientists if we call `transport` coercion (so `coe` or `coerce`).

Cubical/Core/Prelude.agda
MatthewDaggitt: Relation.Binary.PropositionalEquality

Cubical/Codata/Conat/Properties.agda
```diff
@@ -0,0 +1,151 @@
...
+- Proof that ∞ + 1 is equivalent to ∞.
+
+- Proof that conatural is an hSet.
+
+- Bisimulation on conatural
+
+- Proof that bisimulation is equivalent to equivalence (Coinductive Proof
+  Principle).
+
+- Proof that this bisimulation is an hProp
```
mortberg: I would state this as "bisimulation is prop valued" instead

Cubical/Codata/Conat/Properties.agda
```diff
@@ -0,0 +1,151 @@
...
+  iso′ {suc x} {suc y} (con p) = cong con (iso p)
+  prove (iso p i) = iso′ (prove p) i
+
+  osi : ∀ {x y} → (p : x ≡ y) → bisim (misib p) ≡ p
+  osi p = isSetConat _ _ _ p
+
+  path≃bisim : ∀ {x y} → (x ≡ y) ≃ (x ≈ y)
+  path≃bisim = isoToEquiv misib bisim iso osi
+
+  path≡bisim : ∀ {x y} → (x ≡ y) ≡ (x ≈ y)
```
mortberg: Shouldn't this imply an induction principle for conat? (i.e. its universal property stated for bisimulation)

Cubical/Codata/Conat/Properties.agda
MatthewDaggitt: `subst` is called the same thing (Relation.Binary.PropositionalEquality). Interestingly I don't think we have `transport`... @gallais might correct me on this though?

Cubical/Core/Prelude.agda
mortberg: Ok, so maybe we should add a `Relation.Binary.Path` file.

Saizan: I would rather go for `Cubical.Equality` or `Cubical.Foundations.Equality`.

It's always been odd to go into `Relation.Binary.*` for equality.

mortberg: That is also fine with me. But can we please put it in `Cubical.Foundations.Paths` (or `.Path`)?

I think a lot of confusion that I've encountered among mathematicians when trying to explain what we are doing comes from the fact that we are using the word "equality" for paths. The problem with this is that "equality" in standard mathematical practice means (definitionally) proof-irrelevant equality. So the idea that univalence says "equivalence is equality" sounds really quite absurd. However if we make it clear that we are working with **proof-relevant** equality then it suddenly makes a lot more sense as equality is structure and not property. I like the idea of calling this notion "paths" following what Voevodsky did in his Foundations library.

Saizan: That's fine for me

Cubical/Codata/Conat/Properties.agda
ice1000: Is there any existing proofs that i can transport?

Cubical/Core/Prelude.agda
ice1000: Do we move the more polymorphic cong as well? What about sym, compPath?

Cubical/Codata/Conat/Properties.agda
ice1000: transport is dependent pattern matching on heterogeneous equality, afaik.

Cubical/Codata/Conat/Properties.agda
ice1000: I mean, are you expecting me to add some properties for conat?

I feel like it's already too much for one single PR. Maybe we can split the task.

Cubical/Codata/Conat/Properties.agda
ice1000: Fixed.

Cubical/Core/Prelude.agda
ice1000: Removed.

Cubical/Codata/Conat/Properties.agda
ice1000: > I don't understand what you're looking for...

It's just a simple `\p -> p _ _` 😉

Cubical/Core/Prelude.agda
ice1000: Created `C.F.Path` and moved `cong'`.

ice1000:
Addressed all comments.

mortberg:
It seems like you forgot to include `Foundations.Paths`

Cubical/Codata/Conat/Properties.agda
mortberg: I'm not sure what I had in mind... So let's mark this as resolved

Cubical/Core/Prelude.agda
mortberg: Let's not do it in this PR, but it might make sense to shrink the Core a bit as long as the HoTT-UF file still compiles

ice1000:
Added the file.

mortberg:
Thanks, I'm fine with merging as soon as Travis is happy

mortberg:
Thanks to everyone who contributed to this PR!

ice1000:
CI is very happy now.

mortberg:
@Saizan Can you merge if you have no more comments?

PR #58, opened 2019-02-05
Add +-assoc and +-comm

3abc:
It's much less painful to work with now.

### DIFF
--- a/Cubical/Data/Int/Properties.agda
+++ b/Cubical/Data/Int/Properties.agda
@@ -32,39 +32,6 @@ suc-equiv = (sucInt , isoToIsEquiv sucInt predInt sucPred predSuc)
 sucPathInt : Int ≡ Int
 sucPathInt = ua suc-equiv
 
--- Compose sucPathInt with itself n times. Transporting along this
--- will be addition, transporting with it backwards will be
--- subtraction.
-addEq : ℕ → Int ≡ Int
-addEq zero = refl
-addEq (suc n) = compPath sucPathInt (addEq n)
-
-subEq : ℕ → Int ≡ Int
-subEq n i = addEq n (~ i)
-
-_+_ : Int → Int → Int
-m + pos n    = transport (addEq n) m
-m + negsuc n = transport (subEq (suc n)) m
-
-_-_ : Int → Int → Int
-m - pos zero    = m
-m - pos (suc n) = m + negsuc n
-m - negsuc n    = m + pos (suc n)
-
--- We directly get that addition by a fixed number is an equivalence
--- without having to do any induction!
-isEquivAddInt : (m : Int) → isEquiv (λ n → n + m)
-isEquivAddInt (pos n)    = isEquivTransport (addEq n)
-isEquivAddInt (negsuc n) = isEquivTransport (subEq (suc n))
-
--- TODO: prove properties of + like:
---
--- +-comm : ∀ m n → m + n ≡ n + m
--- +-assoc : ∀ m n o → m + (n + o) ≡ (m + n) + o
---
--- Note: This might be a lot easier with the Cubical.Data.Int.Rewrite
-
-
 -- TODO: define multiplication by composing the addition equivalence
 -- with itself.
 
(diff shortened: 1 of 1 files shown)


Cubical/Data/Int/Properties.agda
```diff
@@ -133,3 +100,156 @@ discreteInt (negsuc n) (negsuc m) with discreteℕ n m
...
+_-_ : Int → Int → Int
+m - pos zero    = m
+m - pos (suc n) = m + negsuc n
+m - negsuc n    = m + pos (suc n)
+
+-- We directly get that addition by a fixed number is an equivalence
+-- without having to do any induction!
+isEquivAddInt : (m : Int) → isEquiv (λ n → n + m)
+isEquivAddInt (pos n)    = isEquivTransport (addEq n)
+isEquivAddInt (negsuc n) = isEquivTransport (subEq (suc n))
```
ice1000: Why moving these definitions? It's quite inconvenient for reviewing.

Cubical/Data/Int/Properties.agda
```diff
@@ -133,3 +100,156 @@ discreteInt (negsuc n) (negsuc m) with discreteℕ n m
...
+
+  L3- : ∀ m n → m + (sucInt (negsuc n)) ≡ sucInt (m + (negsuc n))
+  L3- m 0 = sym (sucPred m)
+  L3- m (suc n) = compPath (sym (sucPred _)) (cong sucInt (sym (L' m n)))
+  
+  Lsuc : ∀ m n → m + (sucInt n) ≡ sucInt (m + n)
+  Lsuc m (pos n)    = refl
+  Lsuc m (negsuc n) = L3- m n
+
+  Lpred : ∀ m n → m + (predInt n) ≡ predInt (m + n)
```
mortberg: I think this and some of the above lemmas should be exported, e.g. Lnegsuc, L', L3- and Lsuc. (with good names inspired by the standard library names)

Cubical/Data/Int/Properties.agda
```diff
@@ -133,3 +100,156 @@ discreteInt (negsuc n) (negsuc m) with discreteℕ n m
...
+
+  L0+ : ∀ n z → +pos (suc n) z ≡ +pos n (sucInt z)
+  L0+ zero z = refl
+  L0+ (suc n) z = cong sucInt (L0+ n z)
+
+  Lpos : ∀ n z → z + pos n ≡ +pos n z
+  Lpos zero z = refl
+  Lpos (suc n) z = cong sucInt (Lpos n z)
+
+  LHS RHS : {A : Set} {x y : A} → (x ≡ y) → A
```
mortberg: These are not used for anything so please delete them

Cubical/Data/Int/Properties.agda
```diff
@@ -133,3 +100,156 @@ discreteInt (negsuc n) (negsuc m) with discreteℕ n m
...
+  L-+base : ∀ {n} → negsuc 0 + pos n ≡ pos n + negsuc 0
+  L-+base {zero} = refl
+  L-+base {suc n} = 
+    negsuc 0 + pos (suc n)  ≡⟨ Lpos (suc n) (negsuc 0)⟩
+    +pos (suc n) (negsuc 0) ≡⟨ L0+ n _ ⟩
+    +pos n (pos 0)          ≡⟨ sym (Lpos n (pos 0)) ⟩
+    pos 0 + pos n           ≡⟨ L++base ⟩ 
+    pos (suc n) + negsuc 0 ∎
+    
+  L--base : ∀ {n} → negsuc 0 + negsuc n ≡ negsuc n + negsuc 0
```
mortberg: Some of these lemmas are also useful

Cubical/Data/Int/Properties.agda
3abc: Because it was a different definition under different name (i.e. `_+'_` instead of `_+_`), then it was changed to replace the original one.

mortberg:
By the way: I was expecting some of these proofs to rely on general lemmas about transp commuting in various ways. But it seems like you're not proving any general lemmas at all. Maybe one can generalize things a bit to make the proof slicker?

Cubical/Data/Int/Properties.agda
3abc: OK. Will remove these. 

I just found `LHS` `RHS` very useful when proving identities, e.g. to prove `complicatedIdentity` I want to know the normal form of it's left hand side, instead of entering it manually I can do `LHS complicatedIdentity` in `C-c C-n`

Cubical/Data/Int/Properties.agda
3abc: What are some example of good names? I have a hard time come up with these. 

How about renaming `Lpred` to `+predInt≡predInt+`? This still have risk of misunderstanding. 
Maybe `m+[predIntn]≡predInt[m+n]`?

mortberg:
> > By the way: I was expecting some of these proofs to rely on general lemmas about transp commuting in various ways. But it seems like you're not proving any general lemmas at all. Maybe one can generalize things a bit to make the proof slicker?
> 
> Oh with those lemmas it will definitely be slicker. Is it available or do I need to come up with those?

You will need to come up with them. Let me know if you need some help. I'm also not sure in which file to put it, but probably in some new file in Foundations.

Cubical/Data/Int/Properties.agda
ice1000: Thanks.

3abc:
> You will need to come up with them. Let me know if you need some help. I'm also not sure in which file to put it, but probably in some new file in Foundations.

I tried some general lemma like

```
Ltransport : ∀ {A B C : Set} (p : A ≡ B) (q : B ≡ C) → ∀ a
  → transport q (transport p a) ≡ transport (compPath p q) a
Ltransport p q a = cong (λ a → transport q (transport p a)) (sym (transportRefl a))
```

It turns out it's not easy to prove something like `compPath (addEq n) (addEq 1) ≡ addEq (suc n)` (probably due to use of `ua` and thus `Glue` in the definition.)

However it's much easier to prove `transport (addEq n)` and `transport (subEq n)` both commute with `sucInt` and `predInt`. (Although the definition of `addEq 1` used `ua`, the normal form of `transport (addEq 1) n` is automatically `sucInt n`, no need to use `uabeta` at all!)

Everything is rewritten with these lemmas.

Also the proof of `+-comm` is simplified.

ice1000:
Why CI is not running for this PR...?

3abc:
> Why CI is not running for this PR...?

I guess it's because it's PR to `addInt` branch, not `master`.

mortberg:
I merged the addInt branch. Is there some way to make this PR to be against master instead?

ice1000:
@mortberg Yes. Add his fork as your local git remote (`git remote add 3abc https://github.com/3abc/cubical`), fetch his branches (`git fetch 3abc`) and `git merge 3abc/patch-1`.

mortberg:
> @mortberg Yes. Add his fork as your local git remote (`git remote add 3abc https://github.com/3abc/cubical`), fetch his branches (`git fetch 3abc`) and `git merge 3abc/patch-1`.

Oops, by doing this and pushing I merged this into master... It seems like there is some name clash which wasn't caught by Travis as it wasn't run. I will fix and push directly to master.

mortberg:
I will also close this PR as all the code has been merged. 

@3abc Can you please add the other definitions of + in another PR and prove that they are equal to the one you did here?

ice1000:
> Oops, by doing this and pushing I merged this into master...

This should not happen, but maybe I forgot to say that you should be on the dev branch instead of master (because `git merge` merges the specified branch into current branch)?

ice1000:
Ooops, I've just realized that you're saying "make this PR against master" instead of "merge this PR into your dev branch"... 😅 
Apologize for all these.

Issue #19, opened 2019-01-17
Naming convention?


ice1000:
I see the definitions' names are in a mess, names like `binnat`, `HitInt` are used together .
Should we have a convention for naming? Like following the stdlib directly?

mortberg:
Yes, there is currently no naming convention. It might be a good idea to follow the HoTT-Agda convention (https://github.com/HoTT/HoTT-Agda) as we wouldn't have to think about changing names when porting developments from there.

Can you send me a link to the conventions for the standard library so that I can compare?

MatthewDaggitt:
> Can you send me a link to the conventions for the standard library so that I can compare?

Haha, I wish we were this organised. There is a very bare-bones skeleton of a style guide [here](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md) but mostly the convention is unspoken and is enforced by either @gallais or I coming down hard on names we don't like :laughing: I'll try and flesh it out a bit more in the next couple of days.

mortberg:
@MatthewDaggitt Thanks! I like the stylistic choices. I think I'll use the HoTT-Agda naming conventions and the stdlib style guide for things like indentation.

riaqn:
I think having a function named `elimPropTrunc` is kind of superflous when it's already in the file `PropTrunc/Properties.agda`. It's better to just call it `elim`. Then when other files import it, they can change the qualified name.
```
import Data.HITs.PropTrunc as PropTrunc

foo = ... PropTrunc.elim ...
```

This also kinds of gives a structure to the names, which is nice.

riaqn:
Speaking of qualified import, I think we should avoid unqualifed import as much as possible, except for maybe things from `Core`. This practice from Haskell is nice:
```
open import Data.HITs.PropTrunc (||_||)
import Data.HITs.PropTrunc as PropTrunc
```

gallais:
In Agda, the example

```
open import Data.HITs.PropTrunc (||_||)
import Data.HITs.PropTrunc as PropTrunc
```

can be written

```agda
open import Data.HITs.PropTrunc as PropTrunc using (||_||)
```

mortberg:
Closing this now that we finally merged https://github.com/agda/cubical/pull/451

PR #62, opened 2019-02-07
Define a faster _+_ and compare it with _+'_

3abc:


Cubical/Data/Int/Properties.agda
```diff
@@ -101,98 +101,74 @@ discreteInt (negsuc n) (negsuc m) with discreteℕ n m
 isSetInt : isSet Int
 isSetInt = Discrete→isSet discreteInt
 
--- Compose sucPathInt with itself n times. Transporting along this
--- will be addition, transporting with it backwards will be
--- subtraction.
++pos : ℕ → Int → Int
```
mortberg: Maybe make this infix? And also maybe call it `pos+` as the positive number is the first argument.

Cubical/Data/Int/Properties.agda
```diff
@@ -101,98 +101,74 @@ discreteInt (negsuc n) (negsuc m) with discreteℕ n m
...
+  +negsuc n m                    ≡⟨ cong (+negsuc n) (sym (predSuc m)) ⟩
+  +negsuc n (predInt (sucInt m)) ≡⟨ sym (predInt+negsuc n (sucInt m)) ⟩
+  predInt (+negsuc n (sucInt m)) ∎
+
+predInt+pos : ∀ n m → predInt (+pos n m) ≡ +pos n (predInt m)
+predInt+pos zero m = refl
+predInt+pos (suc n) m =     _ ≡⟨ predSuc _ ⟩
+  +pos n m                    ≡⟨ cong (+pos n) (sym (sucPred _)) ⟩
+  +pos n (sucInt (predInt m)) ≡⟨ sym (sucInt+pos n (predInt m))⟩
   _ ∎
```
mortberg: How about including the conclusion here for readability?

Cubical/Data/Int/Properties.agda
```diff
@@ -233,6 +205,55 @@ ind-assoc _·_ f g p q base m n (suc o) =
...
+addEq (suc n) = compPath (addEq n) sucPathInt
+
+subEq n i = addEq n (~ i)
+
+_+'_ : Int → Int → Int
+m +' pos n    = transport (addEq n) m
+m +' negsuc n = transport (subEq (suc n)) m
+
+private
+  FamilyOfEquiv : (f : Int → Int → Int) → Set
```
mortberg: This definition is never used again? Why keep it?

Cubical/Data/Int/Properties.agda
```diff
@@ -233,6 +205,55 @@ ind-assoc _·_ f g p q base m n (suc o) =
...
+
+  Lpos : ∀ n z → z +' pos n ≡ +pos n z
+  Lpos zero z = refl
+  Lpos (suc n) z = cong sucInt (Lpos n z)
+  
+  Lemma : ∀ m n → m +' n ≡ m + n
+  Lemma m (pos n) = Lpos n m
+  Lemma m (negsuc n) = Lnegsuc n m
+
++'≡+ : _+'_ ≡ _+_
```
mortberg: Very nice that this proof was so easy!

Cubical/Data/Int/Properties.agda
mortberg: Oh wait, it was used again later... How about moving it into a `where`-clause in that definition instead?

Cubical/Data/Int/Properties.agda
```diff
@@ -233,6 +205,55 @@ ind-assoc _·_ f g p q base m n (suc o) =
...
++'≡+ i m n = Lemma m n i
+
+-- We directly get that addition by a fixed number is an equivalence
+-- without having to do any induction!
+isEquivAddInt' : (m : Int) → isEquiv (λ n → n +' m)
+isEquivAddInt' (pos n)    = isEquivTransport (addEq n)
+isEquivAddInt' (negsuc n) = isEquivTransport (subEq (suc n))
+
+isEquivAddInt : (m : Int) → isEquiv (λ n → n + m)
+isEquivAddInt = subst FamilyOfEquiv +'≡+ isEquivAddInt'
```
mortberg: Very nice! I think this is an excellent example of how using univalence can make programming and proving more convenient. For comparison it would be nice to have the "naive" proof that `_ + m` is an equivalence (by defining `_ - m` and proving that this is an inverse to `_ + m` and then using `isoToEquiv`). Could you do this as well?

Cubical/Data/Int/Properties.agda
3abc: The reason I used `+pos` was 

```
m + pos n = +pos n m
```

so if you think `+pos n` as an operator, it's action on m will be `m + pos n`.

I think infix and swap the order of it's argument will make it work better. So we have

```
m + pos n = m +pos n
```

mortberg: Yeah, that's clearer!

3abc:
Updated.

By the way is there any simple examples of high dimensional `hcomp` and `hfill` available somewhere? I was trying to fill some cubes. (By "high dimensional" I just mean Kan operation on cubes or 4-cubes instead of squares.)

mortberg:
> Updated.
> 
> By the way is there any simple examples of high dimensional `hcomp` and `hfill` available somewhere? I was trying to fill some cubes. (By "high dimensional" I just mean Kan operation on cubes or 4-cubes instead of squares.)

Not exactly simple, because working with 4D (or higher) is intrinsically hard. But here's a not too complicated example: https://github.com/agda/cubical/blob/master/Cubical/HITs/Join/Base.agda#L57

3abc:
Do you think `-+m` and `+-m` used in `alternateProof` worth moving out as public function? Could be useful. Maybe I shall name them `minusPlus` and `plusMinus`.

Cubical/Data/Int/Properties.agda
```diff
@@ -269,8 +269,8 @@ minusPlus (negsuc (suc m)) n =
   
 plusMinus : ∀ m n → (n + m) - m ≡ n
 plusMinus (pos zero) n = refl
-plusMinus (pos (suc m)) n = minusPlus (negsuc m) n
-plusMinus (negsuc m) n = minusPlus (pos (suc m)) n
+plusMinus (pos (suc m)) = minusPlus (negsuc m)
+plusMinus (negsuc m) = minusPlus (pos (suc m))
```
3abc: I want to do `plusMinus z = minusPlus (neg z)`, but then I need to tell it `neg (neg m) = m` so I leave it like this.

Cubical/Data/Int/Properties.agda
```diff
@@ -233,6 +205,73 @@ ind-assoc _·_ f g p q base m n (suc o) =
...
     IH = ind-assoc _·_ f g p q base m n o
 
++-comm : ∀ m n → m + n ≡ n + m
++-comm m (pos n) = ind-comm _+_ pos sucInt refl sucInt+ +sucInt pos0+ m n
++-comm m (negsuc n) = ind-comm _+_ negsuc predInt refl predInt+ +predInt negsuc0+ m n
+
 +-assoc : ∀ m n o → m + (n + o) ≡ (m + n) + o
 +-assoc m n (pos o) = ind-assoc _+_ pos sucInt +sucInt refl (λ _ _ → refl) m n o
 +-assoc m n (negsuc o) = ind-assoc _+_ negsuc predInt +predInt refl +predInt m n o
+
```
mortberg: Can you add a comment here saying that we now define another version of addition by transporting along iterated successor paths (constructed using univalence) in order to get a direct proof that addition is an equivalence?

Cubical/Data/Int/Properties.agda
```diff
@@ -233,6 +205,73 @@ ind-assoc _·_ f g p q base m n (suc o) =
...
+  Lpos : ∀ n z → z +' pos n ≡ z +pos n
+  Lpos zero z = refl
+  Lpos (suc n) z = cong sucInt (Lpos n z)
+  
+  Lemma : ∀ m n → m +' n ≡ m + n
+  Lemma m (pos n) = Lpos n m
+  Lemma m (negsuc n) = Lnegsuc n m
+
++'≡+ : _+'_ ≡ _+_
++'≡+ i m n = Lemma m n i
```
mortberg: Why not inline the proof of Lemma here? It seems a bit unnecessarily verbose to make a private module for a proof that is this short.

Cubical/Data/Int/Properties.agda
```diff
@@ -233,6 +205,73 @@ ind-assoc _·_ f g p q base m n (suc o) =
...
+-- without having to do any induction!
+isEquivAddInt' : (m : Int) → isEquiv (λ n → n +' m)
+isEquivAddInt' (pos n)    = isEquivTransport (addEq n)
+isEquivAddInt' (negsuc n) = isEquivTransport (subEq (suc n))
+
+isEquivAddInt : (m : Int) → isEquiv (λ n → n + m)
+isEquivAddInt = subst FamilyOfEquiv +'≡+ isEquivAddInt'
+  where
+  FamilyOfEquiv : (f : Int → Int → Int) → Set
+  FamilyOfEquiv f = (m : Int) → isEquiv (λ n → f n m)
```
mortberg: Why not inline FamilyOfEquiv? I think it is even clearer if the proof is:

```agda
isEquivAddInt : (m : Int) → isEquiv (λ n → n + m)
isEquivAddInt = subst (λ add → ((m : Int) → isEquiv (λ n → add n m))) +'≡+ isEquivAddInt'
```

Cubical/Data/Int/Properties.agda
```diff
@@ -233,6 +205,73 @@ ind-assoc _·_ f g p q base m n (suc o) =
...
+
+minusPlus : ∀ m n → (n - m) + m ≡ n
```
mortberg: @MatthewDaggitt What is this lemma called in the standard library? I looked in https://github.com/agda/agda-stdlib/blob/master/src/Data/Nat/Properties.agda but I couldn't find it...

mortberg:
I forgot one thing in my review. Could you please add a header to the file saying what it contain? (make sure you include a pointer to the proof that add is an equivalence using univalence)

Cubical/Data/Int/Properties.agda
MatthewDaggitt: https://github.com/agda/agda-stdlib/blob/5e8b6aa91adc6d60606e17db7b579be356f72aec/src/Data/Nat/Properties.agda#L1335

It's called `m∸n+n≡m`.

mortberg: Oops, I of course wanted to find it for Int and of course I cannot find it in the Properties file for Int... Is there some way to search for lemmas in Agda using patterns? In `Coq` I would do "Search ((_ - _) + _)" and it would give me all the lemmas of this shape.

MatthewDaggitt: I think you've actually pulled ahead of us already here! I don't think we have that proof in `Data.Integer.Properties` yet. Not sure about the search functionality. I remember someone did mention that there exists something like that but I've never used it.

mortberg: Ok, thanks! I'm not sure if I like the name "m-n+n=n" (first of all because it doesn't match the statement of the lemma) and second of all because it is not so clear where the parantheses go...

Cubical/Data/Int/Properties.agda
3abc: To inline this proof elegantly I changed the definition of `subEq` a little bit. Otherwise I will have to do

```
+'≡+ i m (negsuc (suc n)) = compPath (λ i → (+'≡+ i (predInt m) (negsuc n))) (sym (predInt+negsuc n m)) i
```

The new definition make it as clean as the proof for the `pos` case:

```
+'≡+ i m (negsuc (suc n)) = predInt (+'≡+ i m (negsuc n))
```

Cubical/Data/Int/Properties.agda
```diff
@@ -233,6 +205,73 @@ ind-assoc _·_ f g p q base m n (suc o) =
...
 +-assoc m n (negsuc o) = ind-assoc _+_ negsuc predInt +predInt refl +predInt m n o
+
+-- Compose sucPathInt with itself n times. Transporting along this
+-- will be addition, transporting with it backwards will be subtraction.
+-- Use this to define _+'_ for which is easier to prove
+-- isEquiv (λ n → n +' m) since _+'_ is defined by transport
+
+-- recall
+-- sucPathInt : Int ≡ Int
+-- sucPathInt = ua (sucInt , isoToIsEquiv sucInt predInt sucPred predSuc)
```
3abc: Shall we move definition of `sucPathInt` down here?

That means removing some of the tests above.

Cubical/Data/Int/Properties.agda
```diff
@@ -1,4 +1,28 @@
 {-# OPTIONS --cubical --safe #-}
+
+{-
+
+This file defines
```
3abc: I leave `{-# OPTIONS --cubical --safe #-}` in line one. Let me know if you prefer it come after the header description. For me OPTIONS in line 1 seems cleaner. (Do we have a particular style yet?)

ice1000: No we don't, see #19 (it's about naming but I believe we don't have other kind of convensions right now)

mortberg: I don't have any preferences. Any of them are easy to read

Cubical/Data/Int/Properties.agda
```diff
@@ -1,4 +1,28 @@
...
+
+sucPred : ∀ (i : Int) → sucInt (predInt i) ≡ i
+predSuc : ∀ (i : Int) → predInt (sucInt i) ≡ i
+
+discreteInt : discrete Int
+isSetInt : isSet Int
+
+addition of Int is defined -- _+_ : Int → Int → Int
```
mortberg: There is addition `--`

mortberg:
I'll merge once Travis is done

PR #65, opened 2019-02-08
properties of equivalences

felixwellen:
'cong f' is an equivalence, if f is an equivalence
path split equivalences are logically the same as equivalences


Cubical/Foundations/EquivCharacterization.agda
```diff
@@ -0,0 +1,61 @@
...
+  Everything about path split equivalences is from https://arxiv.org/abs/1706.07526
+-}
+record isPathSplitEquiv {ℓ} {A B : Set  ℓ} (f : A → B) : Set ℓ where
+  field
+    s : B → A 
+    isSection : section f s
+    isSectionCong : (x y : A) → Σ[ s' ∈ (f(x) ≡ f(y) → x ≡ y) ] section (cong f) s'
+
+idIsPathSplitEquiv : ∀ {ℓ} {A : Set ℓ} → isPathSplitEquiv (λ (x : A) → x)
+idIsPathSplitEquiv = record {
```
mortberg: This would look nicer if it was defined using copatterns: https://agda.readthedocs.io/en/v2.5.2/language/copatterns.html

This should also have better computational behavior (as Agda doesn't eagerly unfold things defined using copatterns).

For a concrete example see:  https://github.com/agda/cubical/blob/master/Cubical/Data/BinNat/BinNat.agda#L282

Cubical/Foundations/EquivCharacterization.agda
```diff
@@ -0,0 +1,61 @@
...
+                       s = λ x → x ;
+                       isSection = λ x → refl ;
+                       isSectionCong = λ x y → (λ p → p) , λ p _ → p
+                     }
+
+module _ {ℓ} {A B : Set ℓ} where
+  toIsEquiv : (f : A → B) → isPathSplitEquiv f → isEquiv f 
+  toIsEquiv f record { s = r ; isSection = isSection ; isSectionCong = isSectionCong } =
```
mortberg: Please use copatterns

Cubical/Foundations/EquivCharacterization.agda
```diff
@@ -0,0 +1,61 @@
...
+    (isoToEquiv f r isSection λ (x : A) → (isSectionCong (r (f x)) x).fst (isSection (f x))) .snd
+
+  sectionOfEquiv : {A' B' : Set ℓ} → (f : A' → B') → isEquiv f → B' → A'
+  sectionOfEquiv f record { equiv-proof = all-fibers-contractible } x =
```
mortberg: Copatterns :)

Cubical/Foundations/EquivCharacterization.agda
```diff
@@ -0,0 +1,61 @@
...
+    all-fibers-contractible x .fst .fst
+
+  isSection : {A' B' : Set ℓ} → (f : A' → B') → (pf : isEquiv f) → section f (sectionOfEquiv f pf)
+  isSection f record { equiv-proof = all-fibers-contractible } =
```
mortberg: Copatterns

Cubical/Foundations/EquivCharacterization.agda
```diff
@@ -0,0 +1,61 @@
...
+    λ x → all-fibers-contractible x .fst .snd 
+
+  fromIsEquiv : (f : A → B) → isEquiv f → isPathSplitEquiv f
+  fromIsEquiv f pf = 
+    record {
```
mortberg: And guess what... please use copatterns :)

Cubical/Foundations/EquivCharacterization.agda
```diff
@@ -0,0 +1,61 @@
...
+open import Cubical.Foundations.Univalence
+open import Cubical.Foundations.Equiv
+
+isEquivCong : ∀ {ℓ} {A B : Set ℓ} {x y : A} (e : A ≃ B) → isEquiv (λ (p : x ≡ y) → (cong (e .fst) p))
+isEquivCong e = EquivJ (λ (B' A' : Set _) (e' : A' ≃ B') →
+                         (x' y' : A') → isEquiv (λ (p : x' ≡ y') → cong (e' .fst) p))
+                       (λ _ x' y' → idIsEquiv (x' ≡ y')) _ _ e _ _
+
+congEquiv : ∀ {ℓ} {A B : Set ℓ} {x y : A} (e : A ≃ B) → (x ≡ y) ≃ (e .fst x ≡ e .fst y)
+congEquiv e = ((λ (p : _ ≡ _) → cong (e .fst) p) , isEquivCong e)
```
mortberg: Could you please move this to the Equiv file? I think it makes more sense to have it there instead of in this file.

Cubical/Foundations/EquivCharacterization.agda
mortberg: Here's another example for `isEquiv` using copatterns: https://github.com/agda/cubical/blob/master/Cubical/Foundations/Equiv.agda#L19

felixwellen: Hi Anders! This one is different: the record is an argument now. I don't see how it is a copattern.

mortberg: Oh I see, I read it a bit too fast when I was tired...

Cubical/Foundations/EquivCharacterization.agda
mortberg: But this could be done with copatterns?

felixwellen: Yes. I forgot to commit. I was trying to figure out, what makes it so slow to type check.

Cubical/Foundations/EquivCharacterization.agda
felixwellen: It uses univalences, so that would introduce cyclic dependencies.

Cubical/Foundations/EquivCharacterization.agda
felixwellen: Ah. No. It is in my last commit.

Cubical/Foundations/EquivCharacterization.agda
mortberg: I see... That's unfortunate. I wonder how we could reorganize things to avoid this

Cubical/Foundations/EquivCharacterization.agda
mortberg: What is slow to typecheck?

Cubical/Foundations/EquivCharacterization.agda
mortberg: Maybe introducing a file on isomorphisms could help this. It would contain the proof that iso implies equiv which is used in Univalence. This would make the definition not be cyclic as the definition of equivalence is exported by the Core.

felixwellen: Did it. But I had to move things like 'compEquiv' to the new 'Isomorphism...'

Cubical/Foundations/EquivCharacterization.agda
felixwellen: The 'fromIsEquiv' takes something like 5 seconds. I made everything explicit now, but it is still slow.

Saizan: What's slow is checking that the inferred type of `isSec cong-f eq-cong` is the same as the expected type. And it's slow because Agda ends up unraveling `sectionOfEquiv (λ p i → f (p i)) eq-cong` to its WHNF on both sides when comparing the two types.

Because of `eq-cong` the terms being compared get quite large, hence the slowness.

Ironically, the two types being compared are pretty much the same, but Agda still does not recognize them as "syntactically equal", so they get reduced.

mortberg: How can one fix this Saizan? Put something in a record with a single constructor (this seems to always be the answer to this question...)?

Cubical/Foundations/Isomorphism.agda
```diff
@@ -0,0 +1,121 @@
+{-
+
+Theory about equivalences (definitions are in Core/Glue.agda)
```
mortberg: Comment is outdated

Cubical/Foundations/Isomorphism.agda
```diff
@@ -0,0 +1,121 @@
...
+
+-}
+{-# OPTIONS --cubical --safe #-}
+module Cubical.Foundations.Isomorphism where
+
+open import Cubical.Core.Everything
+open import Cubical.Foundations.HLevels
+
+-- Proof using isPropIsContr. This is slow and the direct proof below is better
+isPropIsEquiv' : ∀ {ℓ ℓ'} {A : Set ℓ} {B : Set ℓ'} (f : A → B) → isProp (isEquiv f)
```
mortberg: Can you please move this and the two following results back to Equiv?

Cubical/Foundations/Isomorphism.agda
```diff
@@ -0,0 +1,121 @@
...
+-- Section and retract
+module _ {ℓ ℓ'} {A : Set ℓ} {B : Set ℓ'} where
+  section : (f : A → B) → (g : B → A) → Set ℓ'
+  section f g = ∀ b → f (g b) ≡ b
+
+  -- NB: `g` is the retraction!
+  retract : (f : A → B) → (g : B → A) → Set ℓ
+  retract f g = ∀ a → g (f a) ≡ a
+
+-- Any iso is an equivalence
```
mortberg: Can you please package the data of an isomorphism into a Record?  (Maybe with the name `Iso`)

Cubical/Foundations/Isomorphism.agda
```diff
@@ -0,0 +1,121 @@
...
+
+  isoToIsEquiv : isEquiv f
+  isoToIsEquiv .equiv-proof y .fst .fst = g y
+  isoToIsEquiv .equiv-proof y .fst .snd = s y
+  isoToIsEquiv .equiv-proof y .snd z = lemIso y (g y) (fst z) (s y) (snd z)
+
+  isoToEquiv : A ≃ B
+  isoToEquiv = _ , isoToIsEquiv
+
+module _ {ℓ ℓ'} {A : Set ℓ} {B : Set ℓ'} (w : A ≃ B) where
```
mortberg: Let's move this back to Equiv as well. The secEq and retEq statements will have to be unfolded, but that's fine

Cubical/Foundations/Isomorphism.agda
```diff
@@ -0,0 +1,121 @@
...
+  secEq x = λ i → fst (snd (snd w .equiv-proof (fst w x)) (x , (λ j → fst w x)) i)
+
+  retEq : retract invEq (w .fst)
+  retEq y = λ i → snd (fst (snd w .equiv-proof y)) i
+
+isoToPath : ∀ {ℓ} {A B : Set ℓ} (f : A → B) (g : B → A)
+              (s : section f g) (t : retract f g) → A ≡ B
+isoToPath f g s t = ua (isoToEquiv f g s t)
+
+invEquiv : ∀ {ℓ ℓ'} {A : Set ℓ} {B : Set ℓ'} → A ≃ B → B ≃ A
```
mortberg: I guess this and the following result could also be moved to Equiv?

Cubical/Foundations/EquivCharacterization.agda
mortberg: I don't see why... In my head the dependency should be as follows:

1. Foundations.Isomorphism: section/retract, Iso, isoToEquiv

2. Foundations.Equiv: depends on 1 and contains all the theory about equivalences

3. Foundations.Univalence: depends on 1 and 2

4. Foundations.PathSplitEquiv: depends on 1-3 and contains in particular congEquiv.

This should not lead to any circularities? The fact that congEquiv ends up in 4 is a bit unfortunate, but we can add a comment to 2 with a pointer to where it can be found.

Cubical/Foundations/EquivCharacterization.agda
Saizan: In this case we can just make sure eq-cong doesn't show up in types that get compared:

```agda
  sectionOfEquiv' : {A' B' : Set ℓ} → (f : A' → B') → isEquiv f → B' → A'
  sectionOfEquiv' f record { equiv-proof = all-fibers-contractible } x =
    all-fibers-contractible x .fst .fst

  isSec : {A' B' : Set ℓ} → (f : A' → B') → (pf : isEquiv f) → section f (sectionOfEquiv' f pf)
  isSec f record { equiv-proof = all-fibers-contractible } x =
    all-fibers-contractible x .fst .snd

  sectionOfEquiv : {A' B' : Set ℓ} → (f : A' → B') → isEquiv f → Σ (B' → A') (section f)
  sectionOfEquiv f e = sectionOfEquiv' f e , isSec f e

  fromIsEquiv : (f : A → B) → isEquiv f → isPathSplitEquiv f
  s (fromIsEquiv f pf) = sectionOfEquiv' f pf
  sec (fromIsEquiv f pf) = isSec f pf
  secCong (fromIsEquiv f pf) x y = sectionOfEquiv cong-f eq-cong
	where
          cong-f : x ≡ y → f x ≡ f y
          cong-f = λ (p : x ≡ y) → cong f p
          eq-cong : isEquiv cong-f
          eq-cong = isEquivCong (f , pf)
```

Cubical/Foundations/Isomorphism.agda
felixwellen: Maybe "isIso f" and "Iso A B" analogous to "isEquiv f" and "A \simeq B"?

mortberg: I don't like the term `isIso` as it is not a proposition... I really like the convention that everything that starts with `is` is a proposition (which goes back to Voevodsky). 

How about `isoStruct` or something to emphasize that it is a structure?

felixwellen: Yes. Are you ok with calling the witnesses for "section f inverse" "rightInv" and similar for "rectract..."?

felixwellen: Should I keep the types of "isoToEquiv" and "isoToPath" or should I use the record everywhere?

mortberg: Use the record everywhere

mortberg: > Yes. Are you ok with calling the witnesses for "section f inverse" "rightInv" and similar for "rectract..."?

Sounds good to me

Cubical/Foundations/EquivCharacterization.agda
felixwellen: Thanks for fixing it and explaining:)

Cubical/Foundations/Isomorphism.agda
```diff
@@ -0,0 +1,94 @@
...
+
+- Definitions of [section] and [retract]
+- Definition of isomorphisms ([Iso])
+- Any isomorphism is an equivalence ([isoToEquiv])
+
+-}
+{-# OPTIONS --cubical --safe #-}
+module Cubical.Foundations.Isomorphism where
+
+open import Agda.Primitive
```
mortberg: Why is this needed?

Cubical/Foundations/Isomorphism.agda
```diff
@@ -0,0 +1,94 @@
...
+module _ {ℓ ℓ'} {A : Set ℓ} {B : Set ℓ'} where
+  section : (f : A → B) → (g : B → A) → Set ℓ'
+  section f g = ∀ b → f (g b) ≡ b
+
+  -- NB: `g` is the retraction!
+  retract : (f : A → B) → (g : B → A) → Set ℓ
+  retract f g = ∀ a → g (f a) ≡ a
+
+
+record isoStruct {ℓ ℓ'} {A : Set ℓ} {B : Set ℓ'} (f : A → B) : Set (ℓ ⊔ ℓ') where
```
mortberg: Why not put the f into the structure? 

The code would also look nicer if we had a constructor for this. For example `iso` or `_,_,_,_`.

Cubical/Foundations/Isomorphism.agda
```diff
@@ -0,0 +1,94 @@
...
+  field
+    inverse : B → A
```
mortberg: Why not call this simply `inv`? Then it would match the other projections better

Cubical/Foundations/PathSplitEquiv.agda
```diff
@@ -0,0 +1,176 @@
...
+
+- if f is an equivalence then (cong f) is an equivalence ([equivCong])
+- if f is an equivalence then pre- and postcomposition with f are equivalences ([preCompEquiv], [postCompEquiv])
+
+(those are not in 'Equiv.agda' because they need Univalence.agda (which imports Equiv.agda))
+-}
+{-# OPTIONS --cubical --safe #-}
+module Cubical.Foundations.PathSplitEquiv where
+
+open import Agda.Primitive
```
mortberg: Why is this needed?

felixwellen: For the max of universe levels.

Cubical/Foundations/Isomorphism.agda
felixwellen: For max of universe levels.

Cubical/Foundations/Isomorphism.agda
felixwellen: Yes, I can put one in.

Cubical/Foundations/PathSplitEquiv.agda
mortberg: This is already exported by Cubical.Core.Everything and it's called `ℓ-max`. Maybe this is a bad name, but let's use it for now.

Cubical/Foundations/Isomorphism.agda
felixwellen: Ok.

felixwellen: You say we call the maps "f" and "inv"?

Cubical/Foundations/Isomorphism.agda
felixwellen: I guess I can call the whole structure "Iso" then.

Cubical/Foundations/Isomorphism.agda
mortberg: What about `fun` and `inv`? (that makes it all a bit funnier :-) )

Saizan:
Btw, I decided to dig deeper about the slowness of the old `fromIsEquiv` and figured out that there was a bug in the syntactic equality heuristic: two path applications were never considered syntactically equal.

It's fixed in master now, so hopefully more things typecheck without reducing types too much

https://github.com/agda/agda/commit/e7bf512b58eb1af92a7878343d10dacf59c8ce5a

Cubical/Foundations/Isomorphism.agda
felixwellen: Not sure about a good name for the map. Maybe "map", since it is the map?

felixwellen: Ok;) Let's go about it the funny way.

felixwellen: Agda doesn't know how to parse with the _,_,_,_. I switch to iso.

mortberg:
I did a few cosmetic changes (remove parentheses and white space). I'm now very happy with this PR, so as soon as the CI is done checking that everything compiles I think we should merge. Thanks a lot for contributing!

PR #68, opened 2019-02-15
Define isOfHLevel 1 to be isProp, add hlevelsuc

3abc:
If we define `isOfHLevel 1` to be `isProp` and only use induction afterward then we don't need things like `isPropIsOfHLevel1` and proofs are simpler.

### DIFF
--- a/Cubical/Foundations/HLevels.agda
+++ b/Cubical/Foundations/HLevels.agda
@@ -109,7 +109,16 @@ hLevelPi 0 h = (λ x → fst (h x)) , λ f i y → snd (h y) (f y) i
 hLevelPi {B = B} 1 h f g i x = (h x) (f x) (g x) i
 hLevelPi (suc (suc n)) h f g = subst (isOfHLevel (suc n)) funExtPath (hLevelPi (suc n) λ x → h x (f x) (g x))
 
-hlevelsuc : (n : ℕ) (A : Set) → isOfHLevel n A → isOfHLevel (suc n) A
+private
+  variable
+    ℓ : Level
+
+hlevelsuc : (n : ℕ) (A : Set ℓ) → isOfHLevel n A → isOfHLevel (suc n) A
 hlevelsuc 0 A = isContr→isProp
 hlevelsuc 1 A = isProp→isSet
 hlevelsuc (suc (suc n)) A h a b =  hlevelsuc (suc n) (a ≡ b) (h a b)
+
+isPropOfHLevel : (n : ℕ) (A : Set ℓ) → isProp (isOfHLevel n A)
+isPropOfHLevel 0 A = isPropIsContr
+isPropOfHLevel 1 A = isPropIsProp
+isPropOfHLevel (suc (suc n)) A f g i a b = isPropOfHLevel (suc n) (a ≡ b) (f a b) (g a b) i
--- a/Cubical/Foundations/HLevels.agda
+++ b/Cubical/Foundations/HLevels.agda
@@ -58,9 +58,6 @@ isPropIsContr z0 z1 j =
 isPropIsProp : ∀ {ℓ} {A : Set ℓ} → isProp (isProp A)
 isPropIsProp f g i a b = isProp→isSet f a b (f a b) (g a b) i
 
-isPropIsSet : ∀ {ℓ} {A : Set ℓ} → isProp (isSet A)
-isPropIsSet f g i a b = isPropIsProp (f a b) (g a b) i
-
 -- A retract of a contractible type is contractible
 retractIsContr : ∀ {ℓ ℓ'} {A : Set ℓ} {B : Set ℓ'} (f : A → B) (g : B → A)
                  (h : (x : A) → g (f x) ≡ x) (v : isContr B) → isContr A
@@ -122,3 +119,6 @@ isPropOfHLevel : (n : ℕ) (A : Set ℓ) → isProp (isOfHLevel n A)
 isPropOfHLevel 0 A = isPropIsContr
 isPropOfHLevel 1 A = isPropIsProp
 isPropOfHLevel (suc (suc n)) A f g i a b = isPropOfHLevel (suc n) (a ≡ b) (f a b) (g a b) i
+
+isPropIsSet : ∀ {ℓ} {A : Set ℓ} → isProp (isSet A)
+isPropIsSet {A = A} = isPropOfHLevel 2 A
--- a/Cubical/Foundations/HLevels.agda
+++ b/Cubical/Foundations/HLevels.agda
@@ -24,7 +24,8 @@ hProp : {ℓ : Level} → Set (ℓ-suc ℓ)
 hProp {ℓ} = Σ (Set ℓ) isProp
 
 isOfHLevel : ∀ {ℓ} → ℕ → Set ℓ → Set ℓ
-isOfHLevel zero A = isContr A
+isOfHLevel 0 A = isContr A
+isOfHLevel 1 A = isProp A
 isOfHLevel (suc n) A = (x y : A) → isOfHLevel n (x ≡ y)
 
 HLevel : ∀ {ℓ} → ℕ → Set _
@@ -44,10 +45,6 @@ isProp→isSet h a b p q j i =
 
 inhProp→isContr : ∀ {ℓ} {A : Set ℓ} → A → isProp A → isContr A
 inhProp→isContr x h = x , h x
-
--- TODO: prove other direction
-isPropIsOfHLevel1 : ∀ {ℓ} {A : Set ℓ} → isProp A → isOfHLevel 1 A
-isPropIsOfHLevel1 h x y = inhProp→isContr (h x y) (isProp→isSet h x y)
 
 isPropIsContr : ∀ {ℓ} {A : Set ℓ} → isProp (isContr A)
 isPropIsContr z0 z1 j =
@@ -80,7 +77,10 @@ isContrSigma {A = A} {B = B} (a , p) q =
        , h (p (x .fst) i) (transp (λ j → B (p (x .fst) (i ∨ ~ j))) i (x .snd)) i))
 
 isContrPath : ∀ {ℓ} {A : Set ℓ} → isContr A → (x y : A) → isContr (x ≡ y)
-isContrPath cA = isPropIsOfHLevel1 (isContr→isProp cA)
+isContrPath cA x y = inhProp→isContr (pA x y) (sA x y)
+  where
+  pA = isContr→isProp cA
+  sA = isProp→isSet pA
 
 lemProp : ∀ {ℓ} {A : Set ℓ} → (A → isProp A) → isProp A
 lemProp h a = h a a
@@ -106,7 +106,10 @@ hLevelPi : ∀ {ℓ ℓ'} {A : Set ℓ} {B : A → Set ℓ'} n
          → ((x : A) → isOfHLevel n (B x))
          → isOfHLevel n ((x : A) → B x)
 hLevelPi 0 h = (λ x → fst (h x)) , λ f i y → snd (h y) (f y) i
-hLevelPi (suc n) h f g = subst (isOfHLevel n) funExtPath sub-lemma
-  where
-  sub-lemma : isOfHLevel n (∀ x → f x ≡ g x)
-  sub-lemma = hLevelPi n λ x → h x (f x) (g x)
+hLevelPi {B = B} 1 h f g i x = (h x) (f x) (g x) i
+hLevelPi (suc (suc n)) h f g = subst (isOfHLevel (suc n)) funExtPath (hLevelPi (suc n) λ x → h x (f x) (g x))
+
+hlevelsuc : (n : ℕ) (A : Set) → isOfHLevel n A → isOfHLevel (suc n) A
+hlevelsuc 0 A = isContr→isProp
+hlevelsuc 1 A = isProp→isSet
+hlevelsuc (suc (suc n)) A h a b =  hlevelsuc (suc n) (a ≡ b) (h a b)


mortberg:
Hmm, does this imply that `isSet A` is definitionally the same as `isOfHLevel 2 A`?

mortberg:
I'm feeling a bit uncertain whether we should change the definition of such a core notion as h-levels. This is one of the things that Voevodsky invented in his Foundation's library and it is very elegant how the specific definitions of isProp and isSet come out of this simple inductive definition. The fact that the definitions are not definitionally the same for hlevels 1 and 2 is a bit unfortunate, but in the presence of computing univalence this should not be much of an issue (all of these are hProp's (to be proved) so we can transport theorems back and forth using propositional extensionality (to be proved using Glue types) which is why I only started defining the maps going back and forth).

3abc:
> Hmm, does this imply that `isSet A` is definitionally the same as `isOfHLevel 2 A`?

Yes.

3abc:
> Hmm, does this imply that `isSet A` is definitionally the same as `isOfHLevel 2 A`?

If you want we can add this to it. But it's trivial.

```
isOfHLevel2≡IsSet : ∀ {ℓ} {A : Set ℓ} → isOfHLevel 2 A ≡ isSet A
isOfHLevel2≡IsSet = refl
```

3abc:
The thing is `isContr A` is a little tricky (some proofs of it used `isProp` and `isSet` for example.)

If we start the induction at `isProp A` then it's all about path spaces.

3abc:
To further convince you that the old definition is bad, consider the definition of 2-groupoid truncation. With my definition in this commit we can do

```
isTwoGroupoid : Set → Set
isTwoGroupoid = isOfHLevel 4

data g2Trunc (A : Set) : Set where
  g2inc : A → g2Trunc A
  g2TwoGroupoid : isTwoGroupoid (g2Trunc A)
```

The proof of `g2Trunc A` is a 2-groupoid is just a higher constructor!

With the old definition one have to struggle a little bit, since it will make Agda complain 
`The target of a constructor must be the datatype applied to its parameters`

mortberg:
Yeah, I see that this definition is quite nice. Maybe we could have the other version as well for historical purposes? If we prove that the two are equivalent then it should be possible to transport properties between the two.

@riaqn I think this PR could simplify some of your work about generalizing `isSet`?

mortberg:
Btw, about 2-groupoid truncation: Do you think you can find a proof of its eliminator that is not using `J`? 

The proof I did in cubicaltt some years ago uses `J` repeatedly and looks a bit like brainfuck: https://github.com/mortberg/cubicaltt/blob/pi4s3_dimclosures/examples/brunerie2.ctt#L1972

3abc:
> Btw, about 2-groupoid truncation: Do you think you can find a proof of its eliminator that is not using `J`?
> 
> The proof I did in cubicaltt some years ago uses `J` repeatedly and looks a bit like brainfuck: https://github.com/mortberg/cubicaltt/blob/pi4s3_dimclosures/examples/brunerie2.ctt#L1972

I can give it a shot. I just started with `propTruncElim` and `setTruncElim`, it's getting more and more complicated, but no need for `J`.

3abc:
> Btw, about 2-groupoid truncation: Do you think you can find a proof of its eliminator that is not using `J`?
> 
> The proof I did in cubicaltt some years ago uses `J` repeatedly and looks a bit like brainfuck: https://github.com/mortberg/cubicaltt/blob/pi4s3_dimclosures/examples/brunerie2.ctt#L1972

There is one thing I have to complain, namely splitting is horrible for HITs. Right now after splitting it complains (about not able to check end points of paths match) and to write down the type of the `L` below take me really long.

Do you think this can and should be automatic? 
I will submit an issue to Agda if no one have done this before.

```
g2TruncElim : (A : Set) (B : (g2Trunc A) → Set) (bG : (x : g2Trunc A) → isTwoGroupoid (B x))
  (f : (x : A) → B (g2inc x)) (x : g2Trunc A) → B x
g2TruncElim A B bG f = proof
  where
  proof : (x : g2Trunc A) → B x
  proof (g2inc x) = f x
  proof (g2TwoGroupoid x y p q r s u v i j k l) = L i j k l
    where
    L : PathP (λ i →
          PathP (λ j →
            PathP (λ k →
              PathP (λ l →
                B (g2TwoGroupoid x y p q r s u v i j k l)
                )
                (proof x)
                (proof y)
              )
              (λ i → proof (p i))
              (λ i → proof (q i))
            )
            (λ i j → proof (r i j))
            (λ i j → proof (s i j))
          )
          (λ i j k → proof (u i j k))
          (λ i j k → proof (v i j k))
    L = { }2
```

mortberg:
> > Btw, about 2-groupoid truncation: Do you think you can find a proof of its eliminator that is not using `J`?
> > The proof I did in cubicaltt some years ago uses `J` repeatedly and looks a bit like brainfuck: https://github.com/mortberg/cubicaltt/blob/pi4s3_dimclosures/examples/brunerie2.ctt#L1972
> 
> There is one thing I have to complain, namely splitting is horrible for HITs. Right now after splitting it complains (about not able to check end points of paths match) and to write down the type of the `L` below take me really long.
> 
> Do you think this can and should be automatic?
> I will submit an issue to Agda if no one have done this before.
> 
> ```
> g2TruncElim : (A : Set) (B : (g2Trunc A) → Set) (bG : (x : g2Trunc A) → isTwoGroupoid (B x))
>   (f : (x : A) → B (g2inc x)) (x : g2Trunc A) → B x
> g2TruncElim A B bG f = proof
>   where
>   proof : (x : g2Trunc A) → B x
>   proof (g2inc x) = f x
>   proof (g2TwoGroupoid x y p q r s u v i j k l) = L i j k l
>     where
>     L : PathP (λ i →
>           PathP (λ j →
>             PathP (λ k →
>               PathP (λ l →
>                 B (g2TwoGroupoid x y p q r s u v i j k l)
>                 )
>                 (proof x)
>                 (proof y)
>               )
>               (λ i → proof (p i))
>               (λ i → proof (q i))
>             )
>             (λ i j → proof (r i j))
>             (λ i j → proof (s i j))
>           )
>           (λ i j k → proof (u i j k))
>           (λ i j k → proof (v i j k))
>     L = { }2
> ```

Yeah, this is quite annoying! It's also very annoying that Agda does not display the boundary that you need for the holes. It is not sufficient just to see that you need something of type `A` when filling a hole if its boundary has to be `u` on `i=0`, `v` on `j=1` and so on...

mortberg:
> > Btw, about 2-groupoid truncation: Do you think you can find a proof of its eliminator that is not using `J`?
> > The proof I did in cubicaltt some years ago uses `J` repeatedly and looks a bit like brainfuck: https://github.com/mortberg/cubicaltt/blob/pi4s3_dimclosures/examples/brunerie2.ctt#L1972
> 
> I can give it a shot. I just started with `propTruncElim` and `setTruncElim`, it's getting more and more complicated, but no need for `J`.

Nice! Can you push your definitions? I did the `propTruncElim` here https://github.com/agda/cubical/blob/master/Cubical/Core/PropositionalTruncation.agda#L35 and I didn't think too much about it and just used `J`. This is not so bad as `J` is only used once, but it gets worse the higher the dimensions get so a direct cubical proof would be great.

Saizan:
> > Btw, about 2-groupoid truncation: Do you think you can find a proof of its eliminator that is not using `J`?
> > The proof I did in cubicaltt some years ago uses `J` repeatedly and looks a bit like brainfuck: https://github.com/mortberg/cubicaltt/blob/pi4s3_dimclosures/examples/brunerie2.ctt#L1972
> 
> There is one thing I have to complain, namely splitting is horrible for HITs. Right now after splitting it complains (about not able to check end points of paths match) and to write down the type of the `L` below take me really long.

By it complains you mean there are errors displayed in the *Agda information* buffer? Those are normal during interactive development, and I use them as a guide for what the terms should be.

At some point the `goal-and-context` command for a specific hole will display those too, so that you do not get just the type but also the constraints it has to satisfy.

Saizan:
@3abc in particular even if you have errors in that buffer you can keep interacting with the hole.

mortberg:
> By it complains you mean there are errors displayed in the _Agda information_ buffer? Those are normal during interactive development, and I use them as a guide for what the terms should be.
> 
> At some point the `goal-and-context` command for a specific hole will display those too, so that you do not get just the type but also the constraints it has to satisfy.

Thanks for the explanation! I've also noticed that Agda tells me what things has to be (slightly cryptically) if I load the file with the cursor outside of a hole. I'm really looking forward to the day the conditions will be printed in the goal and context buffer. I should come to some AIM to learn how I can contribute to these things.

Saizan:
Another way to get agda to show the currently unsolved constraints is the command `agda2-show-constraints`, i.e. `C-c C-=`.

Yeah, you should come to an AIM! (though I don't think I will be to the one in Japan in March..)

3abc:
> By it complains you mean there are errors displayed in the _Agda information_ buffer? Those are normal during interactive development, and I use them as a guide for what the terms should be.
> 
> At some point the `goal-and-context` command for a specific hole will display those too, so that you do not get just the type but also the constraints it has to satisfy.

Let me give a simple example of current behavior of Agda and a potential better version:

Say someone wants to prove that `f : (x : S1) → base ≡ x` (which is not provable since it will make `S1` contractible), and it goes

```
f : (x : S1) → base ≡ x
f x = { }0
```

where `?0 : base ≡ x`. Now split on `x`, current version becomes

```
f : (x : S1) → base ≡ x
f base = { }0
f (loop x) = { }1
```

where it shows

```
?0 : base ≡ base
?1 : base ≡ loop x

———— Errors —————————————————————————————————

Failed to solve the following constraints:
  _309 := λ x → ?1 (x = i1) x [blocked on problem 684]
  [684] base = ?1 (x = i1) i1 : S¹
  [684] base = ?1 (x = i1) i0 : S¹
```

I am kinda used to these error messages, since it always appear when one split with HITs. (I think you agree as you said "Those are normal during interactive development, and I use them as a guide for what the terms should be".) However the type of `?1` is actually misleading, as one may want to use `λ i → loop (i ∧ x)` to fill this hole. (After all, this IS a path that start at `base` and end at `loop x`, what's the problem?) However this attempt will fail because the value of `?1` when `x = i0` or `x = i1` needs to be the same as `f base`.

It will be much nicer if the splitting on HITs become something like this (which is where I always do by hand these days):

```
f : (x : S1) → base ≡ x
f base = { }0
f (loop i) = L i
  where
  L : PathP (λ i → base ≡ loop i) (f base) (f base)
  L = { }1
```

the restriction on `L` (namely its type) is much nicer and the information buffer no longer shows any error. So for a higher constructor, when we split, instead of `f (loop i) = { }1` we should have something like `f (loop i) = { }1 i`, where the type of `?1` is a dependent path indicating the correct endpoints. (Right now even if one start with `f (loop i) = ? i` Agda will end up with `Failed to solve the following constraints`...)

The example for `S1` is simple, so write down the type for `L` is not too hard. But for things like `g2TruncElim` above, even the type of `L` is hard to write down. It will be much nicer if Agda can describe the correct type of `L`.

jonsterling:
It is worth noting that we have implemented behavior similar to what was requested here in the `redtt` proof assistant for Cartesian cubical type theory (https://github.com/redprl/redtt). Specifically, rather than showing a confusing unification error, the goal will just show what type you need to construct and a list of face constraints.

This is made possible because typechecking in `redtt` is always with respect to a partial boundary, and holes store this information; holes are just global constants of extension [pre]type (meaning that they can include the boundary in their type), so we avoid unification errors like this by means of the eta rules for the extension types.

There are still some cases where it is not possible to exhibit this nice behavior in the maximal way, but it works pretty well already. Hopefully an idea like this can be adopted in Cubical Agda!

Saizan:
@3abc Fair enough, here the constraints you get are not helpful because of details of the unification algorithm, in other cases I've found them describing the same information as the path type you'd construct.

Eta expanding will get rid of that "blocked on problem .." and show the whole deal, which admittedly is a bit much:
```
f : (x : S¹) → base ≡ x
f base j = { }0
f (loop i) j = { }1

———— Errors ———————————————————————
Failed to solve the following constraints:
  base = ?1 (i = x) (j = i0) : S¹
  loop x = ?1 (i = x) (j = i1) : S¹
  ?1 (i = i1) (j = x) = ?1 (i = i0) (j = x) : S¹
  base = ?1 (i = i1) (j = i1) : S¹
  base = ?1 (i = i1) (j = i0) : S¹
```

Trying `? x` is never a good idea in Agda, so indeed that won't work. 

Clauses of the following form could be nice so that the new goal is actually a path between `f base` and `f base`. (not (yet?) supported)
```
cong f loop = ?1
```

@jonsterling Right, propagating constraints down should work nicely especially for pattern matching, but also for introductions. It's also something that came up for pattern-matching lambdas: if you don't have the boundary you can't infer the right type for them.

3abc:
> It is worth noting that we have implemented behavior similar to what was requested here in the `redtt` proof assistant for Cartesian cubical type theory (https://github.com/redprl/redtt). Specifically, rather than showing a confusing unification error, the goal will just show what type you need to construct and a list of face constraints.

Wow great. I should definitely check out redtt more! I only wish a documentation exists. 😂

3abc:
> Eta expanding will get rid of that "blocked on problem .." and show the whole deal, which admittedly is a bit much:

Is this something I can do in Agda? If so, how?

3abc:
> cong f loop = ?1

Oh that will be awesome. Even if it's not supported yet it will be awesome to have that in mind when using Agda. At least one can write this on paper and start thinking there 😄

3abc:
> Nice! Can you push your definitions? I did the `propTruncElim` here https://github.com/agda/cubical/blob/master/Cubical/Core/PropositionalTruncation.agda#L35 and I didn't think too much about it and just used `J`. This is not so bad as `J` is only used once, but it gets worse the higher the dimensions get so a direct cubical proof would be great.

Is the point of not using `J` just to avoid too much `transport`? I'm not ready to push yet, but you can see my definition [here](https://gist.github.com/3abc/08c0923a199b62c8dd03bf5cc641f324), I think I used the same amount of `transport` or even more.

Saizan:
> > Eta expanding will get rid of that "blocked on problem .." and show the whole deal, which admittedly is a bit much:
> 
> Is this something I can do in Agda? If so, how?

There's no command for doing it for every clause at once, but I meant adding the `j` argument like in my version of `f` above:
```
f : (x : S¹) → base ≡ x
f base j = { }0
f (loop i) j = { }1
```

mortberg:
This looks good to me. Shall I merge?

3abc:
> This looks good to me. Shall I merge?

Sure. :D

riaqn:
A downside of this definition is that, now when we do an induction on isOfHLevel, we will have do three cases: 0, 1 and (n + 1). I'd rather stick with the old definition and have conversion functions from isOfHLevel and isProp, isSet.

mortberg:
But isn't having things be definitionally equal more useful? Even if we have to do more 1 step deeper case analysis? Do you have a concrete example in mind?

3abc:
> But isn't having things be definitionally equal more useful? Even if we have to do more 1 step deeper case analysis? Do you have a concrete example in mind?

I agree. Every time that induction is actually used on `HLevels`, `n = 0` is always a special case that needs extra care. And the real base step is `n = 1`.

riaqn:
Yes, n=0 is always a special case. But what do you mean the real base step is n=1? In the original definition, two cases suffice.

Even from a practical view, this new definition means we need an extra case for anything induction of isOfHLevel. While the old definition means we sometimes need to call the conversion functions. I think in the long run the later is easier to manage.

3abc:
> Yes, n=0 is always a special case. But what do you mean the real base step is n=1? In the original definition, most of the time two cases suffice.
> 
> Even from a practical view, this new definition means we need an extra case for anything induction of isOfHLevel which will require some extra thinking. While the old definition means we sometimes need to call the conversion functions which is mechanical. I think in the long run the later is easier to manage.

I see your point. We will need two things 

1. ease of proving things (to just write down a term that type checks without caring about how complicated its normal form is.) 
2. being able to compute (closed term reduce to normal form quickly enough). 

What you pointed out is about 1. For 2. here things being definitionally equal is crucial. Maybe a good compromise is to put the old definition as `isOfHLevel'` and prove `isOfHLevel' ≡ isOfHLevel`.

riaqn:
I suggest we keep the old definition, but prove that:
isProp ≡ isOfHLevel 1
isSet ≡ isOfHLevel 2

This way we still have the shorter definition of isOfHLevel, but we get to transport results among isProp, isSet and isOfHLevel.

riaqn:
@3abc That's a good idea. If no one is opposing I will do this.

The only issue I might have is that the old/canonical definition should be `isOfHLevel`, and the new/sometime more convenient definition should be 'isOfHLevel'`

3abc:
> @3abc That's a good idea. If no one is opposing I will do this.
> @mortberg
> The only issue I might have is that the old/canonical definition should be `isOfHLevel`, and the new/sometime more convenient definition should be `isOfHLevel'`

I'm OK with that.

PR #69, opened 2019-02-18
Add some Nat properties and ordering on Nat

dolio:
I've been trying to prove that the ordering is a proposition, but haven't figured out how yet.

I've also been fooling around with using the ordering in a couple different ways to define finite sets, since indexed inductive definitions aren't available yet.

Thoughts?

### DIFF
--- a/Cubical/Data/Int/Properties.agda
+++ b/Cubical/Data/Int/Properties.agda
@@ -31,7 +31,7 @@ open import Cubical.Foundations.Equiv
 open import Cubical.Foundations.HLevels
 
 open import Cubical.Data.Empty
-open import Cubical.Data.Nat hiding (_+_ ; +-assoc)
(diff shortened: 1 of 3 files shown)


mortberg:
If you want to prove that <= is prop valued then you need to note that the k in <= is unique and then use the characterization of equality in Sigma types where the second component is a proposition: https://github.com/agda/cubical/blob/master/Cubical/Foundations/HLevels.agda#L98

mortberg:
PS: You will also need to use that nat is a set.

dolio:
Okay, that was the piece I needed. It's added now.

3abc:
CI failed. You need to rename the `+-comm` for integers.

dolio:
Okay. I tried running the check locally, but it failed on other stuff (constraint solving cubical stuff, I think). I'll try to get things fixed up.

Saizan:
@dolio things built fine here, so I renamed `+-comm` and merged.

PR #70, opened 2019-02-18
Add cabal v2-install instructions

phadej:


3abc:
`cabal update` is still the legacy version. Does `cabal v2-update` do the same?

phadej:
@3abc for these purposes yes. Can be either one.

phadej:
Changed `install` to `v2-install`

INSTALL.md
```diff
@@ -22,6 +22,84 @@ Agda see:
 
 https://agda.readthedocs.io/
 
+cabal v2-install instructions
+=============================
+
+cabal `v2-build` is a new operating mode, which makes projects
+not interfere with each other. To download and compile the development
+version Agda with a `v2-build`, you need to download recent
+version of [`cabal-install`](https://www.haskell.org/cabal/download.html).
```
mortberg: How recent does this have to be?

INSTALL.md
```diff
@@ -22,6 +22,84 @@ Agda see:
...
+```
+> cabal v2-update
+> git clone https://github.com/agda/agda
+> cd agda
+> touch doc/user-manual.pdf
+> cabal v2-install agda agda-mode
+```
+
+This should put the agda and agda-mode executables in the folder
+`~/.cabal/bin` (the location can be configured with `--symlink-bindir` flag).
```
mortberg: Does this work with sandboxes as well?

INSTALL.md
phadej: cabal-install-2.4

INSTALL.md
phadej: No, stuff in sandboxes stays in sandboxes.

phadej: Note: next `cabal-install` version won't have sandboxes at all, AFAIK

INSTALL.md
mortberg: Can you please add this so one knows whether an update is necessary or not

INSTALL.md
mortberg: Hmm, so how can I install things locally? Without sandboxes cabal v1 is more or less unusable for multiple projects.

phadej: Using `v2-install`.

phadej: Let me try to clarify: if you have `cabal-install-2.4` or later, there are hardly any reason to use old commands. You should use `v2-` prefixed ones.

I'm not sure what you mean by "install things locally", there are various things which that could mean. For example if you want a `ghci` session with some libraries, you can simply:

```
cabal new-repl -b `EdisonCore`
```

The documentation is at https://cabal.readthedocs.io/en/latest/nix-local-build-overview.html, it's not complete but quite comprehensive.

mortberg: I'm just wondering if I can install all of the dependencies for a project locally in a directory like I can do with sandboxes in v1? I never install anything globally in my `~` directory as it is easy to run into problems with various projects requiring different versions of the same package.

phadej: @mortberg the *nix-local-build* is "magic" which takes care of that. From [the documentation page](https://cabal.readthedocs.io/en/latest/nix-local-build-overview.html) I linked previously:

> Nix-style local builds combine the best of non-sandboxed and sandboxed Cabal:
>
> 1. Like sandboxed Cabal today, we build sets of independent local packages deterministically and independent of any global state. new-build will never tell you that it can’t build your package because it would result in a “dangerous reinstall.” Given a particular state of the Hackage index, your build is completely reproducible. For example, you no longer need to compile packages with profiling ahead of time; just request profiling and new-build will rebuild all its dependencies with profiling automatically.
> 2.  Like non-sandboxed Cabal today, builds of external packages are cached in ~/.cabal/store, so that a package can be built once, and then reused anywhere else it is also used. No need to continually rebuild dependencies whenever you make a new sandbox: dependencies which can be shared, are shared.

INSTALL.md
phadej: Added

INSTALL.md
mortberg: Wow, that sounds great!

mortberg:
Can you also update the list in the beginning of the file so that it says that there are 3 recommended ways to install the library:

- cabal-v2
- cabal-v1 with sandboxes
- stack

phadej:
@mortberg done.

mortberg:
Great, thanks!

PR #74, opened 2019-02-21
Proof for eliminators of truncations

3abc:
Up to h-level 4 i.e. 2-groupoid.

### DIFF
--- a/Cubical/Experiments/truncations.agda
+++ b/Cubical/Experiments/truncations.agda
@@ -1,5 +1,5 @@
 {-# OPTIONS --cubical --safe #-}
-module Cubical.Experiments.truncations where
+module truncations where
 
 open import Cubical.Data.Nat.Base
 open import Cubical.Core.Prelude
@@ -60,12 +60,12 @@ setTruncElim A B bS f = proof
     C = tIsSet x y p q
     M1 : {x y : sA} (p : x ≡ y) → B x → B y → Set
     M1 p a b = PathP (λ i → B (p i)) a b
-    M2 : {x y : sA} {p q : x ≡ y} (r : p ≡ q) (a : B x) (b : B y)
+    M2 : {x y : sA} {p q : x ≡ y} (r : p ≡ q) {a : B x} {b : B y}
        → (M1 p a b) → (M1 q a b) → Set
-    M2 r a b c d = PathP (λ i → M1 (r i) a b) c d
+    M2 r {a = a} {b = b} c d = PathP (λ i → M1 (r i) a b) c d
     B' : {x y : sA} {p q : x ≡ y} (r : p ≡ q) → Set
     B' {x} {y} {p} {q} r = (a : B x) (b : B y) (c : M1 p a b) (d : M1 q a b)
-                           → M2 r a b c d
+                           → M2 r c d
     L0 = bS x -- : B' (λ i j → C i0 i0) 
     B0 B1 : I → Set
     B0 k = B' (λ i j → C i0 (j ∧ k))
@@ -91,17 +91,17 @@ groupoidTruncElim A B bG f = proof
     C = tIsGroupoid x y p q r s
     M1 : {x y : gA} (p : x ≡ y) → B x → B y → Set
     M1 p a b = PathP (λ i → B (p i)) a b
-    M2 : {x y : gA} {p q : x ≡ y} (r : p ≡ q) (a : B x) (b : B y)
+    M2 : {x y : gA} {p q : x ≡ y} (r : p ≡ q) {a : B x} {b : B y}
        → (M1 p a b) → (M1 q a b) → Set
-    M2 r a b c d = PathP (λ i → M1 (r i) a b) c d
+    M2 r {a = a} {b = b} c d = PathP (λ i → M1 (r i) a b) c d
     M3 : {x y : gA} {p q : x ≡ y} {r s : p ≡ q} (u : r ≡ s)
-         (a : B x) (b : B y) (c : M1 p a b) (d : M1 q a b)
-         → (M2 r a b c d) → (M2 s a b c d) → Set
-    M3 u a b c d e f = PathP (λ i → M2 (u i) a b c d) e f
+         {a : B x} {b : B y} {c : M1 p a b} {d : M1 q a b}
+         → (M2 r c d) → (M2 s c d) → Set
+    M3 u {c = c} {d = d} e f = PathP (λ i → M2 (u i) c d) e f
     B' : {x y : gA} {p q : x ≡ y} {r s : p ≡ q} → (r ≡ s) → Set
     B' {x} {y} {p} {q} {r} {s} u =
       (a : B x) (b : B y) (c : M1 p a b) (d : M1 q a b)
-      (e : M2 r a b c d) (f : M2 s a b c d) → M3 u a b c d e f
+      (e : M2 r c d) (f : M2 s c d) → M3 u e f
     L0 = bG x -- : B' (λ i j k → C i0 i0 i0)
     B0 B1 B2 : I → Set
     B0 i = B' (λ j0 j1 j2 → C i0 i0 (i ∧ j2))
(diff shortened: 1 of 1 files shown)


3abc:
Here is the proof of `g2TruncElim`. To warm up I defined eliminators for Prop, Set and Groupoid as well. Hopefully it's less like brainfuck and more structured.

The idea is simple: we have a (higher) cube in a fiber and we want to expand it to one in the total space. We use a few transport to achieve this. The `M1`, `M2`, `M3`, `M4` are just auxiliary types.

I think this proof is more for the conceptual understanding instead of for computational, it's more or less the same as the existing one for propositional truncation at [PropositionalTruncation.agda#L35](https://github.com/agda/cubical/blob/master/Cubical/Core/PropositionalTruncation.agda#L35). Given the current performance using this to perform computation could be painful.

Cubical/Experiments/truncations.agda
```diff
@@ -0,0 +1,165 @@
...
+isTwoGroupoid = isOfHLevel 4
+
+transport' : (p : I → Set) → (p i0) → (p i1)
+transport' p a = transp (λ i → p i) i0 a
+
+data propTrunc (A : Set) : Set where
+  pinc : A → propTrunc A
+  pIsProp : isProp (propTrunc A)
+
+propTruncElim : (A : Set) (B : (propTrunc A) → Set)
```
mortberg: Isn't this proof definitionally equal to https://github.com/agda/cubical/blob/master/Cubical/Core/PropositionalTruncation.agda#L35 ?

Cubical/Experiments/truncations.agda
```diff
@@ -0,0 +1,165 @@
...
+    L1 : B' C
+    L1 = transport B0 L0
+    L : PathP (λ i → B (pIsProp x y i)) (proof x) (proof y)
+    L = L1 (proof x) (proof y)
+
+data setTrunc (A : Set) : Set where
+  tinc : A → setTrunc A
+  tIsSet : isSet (setTrunc A)
+
+setTruncElim : (A : Set) (B : (setTrunc A) → Set)
```
mortberg: Could you compress this proof a bit? It is quite hard to read because of all the local definitions with very short names.

I would like to understand how it compares to https://github.com/agda/cubical/blob/master/Cubical/HITs/SetTruncation/Properties.agda#L37 . Are the proofs the same or not? Note that J is just a transport with a `/\` connection https://github.com/agda/cubical/blob/master/Cubical/Core/Prelude.agda#L88

Cubical/Experiments/truncations.agda
```diff
@@ -0,0 +1,165 @@
...
+    B1 k = B' (λ i j → C (i ∧ k) j)
+    L1 = transp B0 i0 L0
+    L2 = transp B1 i0 L1
+    L = L2 (λ i → proof (p i)) (λ i → proof (q i))
+
+data groupoidTrunc (A : Set) : Set where
+  tinc : A → groupoidTrunc A
+  tIsGroupoid : isGroupoid (groupoidTrunc A)
+
+groupoidTruncElim : (A : Set) (B : (groupoidTrunc A) → Set)
```
mortberg: Can you please compress the proof of this as well?

Cubical/Experiments/truncations.agda
```diff
@@ -0,0 +1,165 @@
...
+    L1 = transp B0 i0 L0
+    L2 = transp B1 i0 L1
+    L3 = transp B2 i0 L2
+    L = L3 (λ i j → proof (r i j)) (λ i j → proof (s i j))
+
+data g2Trunc (A : Set) : Set where
+  g2inc : A → g2Trunc A
+  g2IsTwoGroupoid : isTwoGroupoid (g2Trunc A)
+
+g2TruncElim : (A : Set) (B : (g2Trunc A) → Set)
```
mortberg: Can you compress this proof as well? I want to understand how it compares to:

https://github.com/mortberg/cubicaltt/blob/pi4s3_dimclosures/examples/brunerie2.ctt#L2087

Cubical/Experiments/truncations.agda
```diff
@@ -0,0 +1,165 @@
...
+    B1 i = B' (λ j0 j1 j2 → C i0 (i ∧ j1) j2)
+    B2 i = B' (λ j0 j1 j2 → C (i ∧ j0) j1 j2)
+    L1 = transp B0 i0 L0
+    L2 = transp B1 i0 L1
+    L3 = transp B2 i0 L2
+    L = L3 (λ i j → proof (r i j)) (λ i j → proof (s i j))
+
+data g2Trunc (A : Set) : Set where
+  g2inc : A → g2Trunc A
+  g2IsTwoGroupoid : isTwoGroupoid (g2Trunc A)
```
mortberg: I wonder if having extension types would simplify these higher truncations as the higher constructor would just specify an n dimensional cube and we don't have to explicitly quantify over the points, lines, square... What do yinz think @ecavallo @jonsterling @cangiuli ? We cannot experiment with this in redtt yet right?

Cubical/Experiments/truncations.agda
3abc: > Can you please compress the proof of this as well?

By compress you mean inline everything? Sure I can do it. 

The current way of writing was just make sure I get how it generalizes. Before it wasn't entirely clear to me how to move from `propTruncElim` to `g2TruncElim`.

Let me do it for `groupoidTruncElim` first to see whether the compressed one is enlightening or not.

Cubical/Experiments/truncations.agda
3abc: > Isn't this proof definitionally equal to https://github.com/agda/cubical/blob/master/Cubical/Core/PropositionalTruncation.agda#L35 ?

Yeah, more or less yes. Although `refl` won't fill a path between those two in Agda for some reason. 

However I do think `propTruncElim` can have a simpler proof, namely

```Agda
propTruncElim' : {A : Set} {B : (propTrunc A) → Set} (bP : (x : propTrunc A) → isProp (B x))
  (f : (x : A) → B (pinc x)) (x : propTrunc A) → B x
propTruncElim' bP f (pinc x) = f x
propTruncElim' {B = B} bP f (pIsProp x y i) = L i
  where
  L' : PathP (λ i → B (pIsProp x y i)) (propTruncElim' bP f x) _
  L' i = transp (λ j → B (pIsProp x y (i ∧ j))) (~ i) (propTruncElim' bP f x)
  L : (i : I) → B (pIsProp x y i)
  L i = hcomp (u i) (L' i)
    where
    u : (i j : I) → Partial (i ∨ ~ i) (B (pIsProp x y i))
    u i j (i = i0) = propTruncElim' bP f x
    u i j (i = i1) = bP y (L' i1) (propTruncElim' bP f y) j
```

Note that `transp ... (~ i) ...` which simplifies a lot. The proof is basically a `compPath` except one of the paths is a dependent one. I will try to generalize this to other eliminators.

Cubical/Experiments/truncations.agda
3abc: > I wonder if having extension types would simplify these higher truncations as the higher constructor would just specify an n dimensional cube and we don't have to explicitly quantify over the points, lines, square... 

Even without extension types we can still do a version which quantifies over less things, I thought about the following alternate definition of `isGroupoid`.

```Agda
data S2 : Set where
  base2 : S2
  loop2 : refl {x = base2} ≡ refl

isGroupoid : Set → Set
isGroupoid A = (p : S2 → A) (x : S2) → p x ≡ p (base2)

data groupoidTrunc (A : Set) : Set where
  tinc : A → groupoidTrunc A
  tIsGroupoid : isGroupoid (groupoidTrunc A)
```

In this version we don't have to quantify over a pair of points, a pair of paths and a pair of squares. However it's still hard to prove the eliminators. And in this definition `isGroupoid A p base2` is not necessarily `refl`, which makes things tricky.

Cubical/Experiments/truncations.agda
mortberg: Hmm, so this is simpler than the proof of `propTruncElim` below as the transport in B0 computes to 1 hcomp with 2 transp as sides? It's really hard to for me to beta reduce the proof below in my head though...

Cubical/Experiments/truncations.agda
mortberg: Yes, inline and if something is definitionally the same as `J` then use `J`

Cubical/Experiments/truncations.agda
mortberg: Good point! Finding a definition for which we can directly prove the eliminators is a nice problem

Cubical/Experiments/truncations.agda
3abc: It's a `hcomp` with a `transp` as base, constant on one side, and a path given by `bP` on the other side. Does this help?

I want to tell how simple a thing is by asking Agda to compute it's normal form and check the length, but often it doesn't expand things in `where` clause. Isn't it supposed to expand everything? Is this a known issue? Sometimes normal forms end up containing things like `truncations.u` (where `truncations` is the name of the module).

mortberg: I would also like this. It would be useful if it would print the number of `transp` and `hcomp` in the term as that's a good measure for how complicated something is to compute with.

mortberg:
It would be nice if these elimination principles could be added to the suitable files @3abc . It would be great if you could also compress the proofs a bit as I suggested above (fewer temporary definitions and use `J` wherever the proof is just an unfolded version of it).

3abc:
> It would be nice if these elimination principles could be added to the suitable files @3abc . It would be great if you could also compress the proofs a bit as I suggested above (fewer temporary definitions and use `J` wherever the proof is just an unfolded version of it).

OK. I have proofs without using J written in Agda. What file do you suggest me to put it in?

mortberg:
For example https://github.com/agda/cubical/blob/master/Cubical/HITs/2GroupoidTruncation/Properties.agda and https://github.com/agda/cubical/blob/master/Cubical/HITs/GroupoidTruncation/Properties.agda, etc.

3abc:
> For example https://github.com/agda/cubical/blob/master/Cubical/HITs/2GroupoidTruncation/Properties.agda and https://github.com/agda/cubical/blob/master/Cubical/HITs/GroupoidTruncation/Properties.agda, etc.

I will close this one and I just added another PR.

PR #75, opened 2019-02-21
higher groupoid laws and a filler function for compPath

bbentzen:
- Homogeneous higher groupoid laws of types
- Filler function for compPath
- Group-like notations for compPath and sym

### DIFF
--- a/Cubical/Core/Prelude.agda
+++ b/Cubical/Core/Prelude.agda
@@ -54,8 +54,7 @@ compPath-filler {x = x} p q j i =
                   ; (i = i1) → q j }) (inc (p i)) j
 
 compPath : ∀ {x y z : A} → x ≡ y → y ≡ z → x ≡ z
-compPath {x = x} p q j =
-  (compPath-filler p q) i1 j
+compPath p q j = compPath-filler p q i1 j
 
 infix  3 _∎
 infixr 2 _≡⟨_⟩_
--- a/Cubical/Core/Prelude.agda
+++ b/Cubical/Core/Prelude.agda
@@ -47,10 +47,15 @@ cong f p = λ i → f (p i)
 
 -- This is called compPath and not trans in order to eliminate
 -- confusion with transp
-compPath : x ≡ y → y ≡ z → x ≡ z
-compPath {x = x} p q i =
-  hcomp (λ j → \ { (i = i0) → x
-                 ; (i = i1) → q j }) (p i)
+
+compPath-filler : ∀ {x y z : A} → x ≡ y → y ≡ z → I → I → A
+compPath-filler {x = x} p q j i =
+  hfill (λ j → λ { (i = i0) → x
+                  ; (i = i1) → q j }) (inc (p i)) j
+
+compPath : ∀ {x y z : A} → x ≡ y → y ≡ z → x ≡ z
+compPath {x = x} p q j =
+  (compPath-filler p q) i1 j
 
 infix  3 _∎
 infixr 2 _≡⟨_⟩_
--- a/Cubical/Foundations/GroupoidLaws.agda
+++ b/Cubical/Foundations/GroupoidLaws.agda
@@ -1,0 +1,95 @@
+{-
+
+This file proves the higher groupoid structure of types
+
+-}
+{-# OPTIONS --cubical --safe #-}
+module Cubical.Foundations.GroupoidLaws where
+
+open import Cubical.Core.Primitives
+open import Cubical.Core.Prelude
+
+private
+  variable
+    ℓ : Level
+    A : Set ℓ
+
+-- some useful notation (move to prelude?)
+
+_·_ :  {x y z : A} → (x ≡ y) → (y ≡ z) → (x ≡ z)
+x≡y · y≡z = compPath x≡y y≡z 
+
+infixr 30 _·_
+
+_⁻¹ : {x y : A} → (x ≡ y) → (y ≡ x)
+x≡y ⁻¹ = sym x≡y
+
+-- homogeneous groupoid laws
+
+symRefl : (x : A) →
+  Path (Path A x x) refl (refl ⁻¹)
+symRefl x i = refl 
+
+compPathRefl : (x : A) →
+  Path (Path A x x) refl (refl · refl)
+compPathRefl x j i =
+  hfill (λ j → λ { (i = i0) → x
+                  ; (i = i1) → x }) (inc x) j
+
+symInvo : (x y : A) (p : x ≡ y) →
+     Path (Path A x y) p ((p ⁻¹)⁻¹)
+symInvo x y p = refl
+
+rUnit : (x y : A) (p : x ≡ y) →
+  Path (Path A x y) p (p · refl)
+rUnit x y p j i =
+  hfill (λ j → λ { (i = i0) → x
+                 ; (i = i1) → y }) (inc (p i)) j
+
+lUnit : (x y : A) (p : x ≡ y) →
+  Path (Path A x y) p (refl · p)
+lUnit x y p k i =
+  hcomp (λ j → λ { (i = i0) → x
+                  ; (i = i1) → p (~ k ∨ j )
+                  ; (k = i0) → p i
+               -- ; (k = i1) → compPath-filler refl p j i
+                  }) (p (~ k ∧ i ))
+
+rCancel : (x y : A) (p : x ≡ y) →
+   Path (Path A x x) (p · (p ⁻¹)) refl
+rCancel x y p j i =
+  hcomp (λ k → λ { (i = i0) → x
+                  ; (i = i1) → p (~ k ∧ ~ j)
+               -- ; (j = i0) → hfill (λ w → λ { (i = i0) → x; (i = i1) → p (~ w) }) (inc (p i)) k
+                  ; (j = i1) → x }) (p (i ∧ ~ j))
+
+lCancel : (x y : A) (p : x ≡ y) →
+   Path (Path A y y) refl ((p ⁻¹) · p)
+lCancel x y p k i =
+  hcomp (λ j → λ { (i = i0) → p (j ∨ k)
+                 ; (i = i1) → p j
+                 ; (k = i0) → p j
+                 ; (k = i1) → (compPath-filler (p ⁻¹) p j i)
+                 }) (p (~ i ∧ k))
+
+3outof4 : (α : I → I → A) → (p : α i1 i0 ≡ α i1 i1) → (β : PathP (λ j → Path A (α j i0) (α j i1)) (λ i → α i0 i) p) →
+  Path (Path A (α i1 i0) (α i1 i1)) (λ i → α i1 i) p
+3outof4 α p β j i =
+  hcomp (λ k → λ { (i = i0) → α k i0
+                  ; (i = i1) → α k i1
+                  ; (j = i0) → α k i
+                  ; (j = i1) → β k i }) (α i0 i)
+
+preassoc : (x y z w : A) (p : x ≡ y) (q : y ≡ z) (r : z ≡ w) →
+  PathP (λ j → Path A x ((q · r) j)) p ((p · q) · r)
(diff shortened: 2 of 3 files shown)


Cubical/Foundations/GroupoidLaws.agda
```diff
@@ -0,0 +1,95 @@
...
+  Path (Path A x x) refl (refl · refl)
+compPathRefl x j i =
+  hfill (λ j → λ { (i = i0) → x
+                  ; (i = i1) → x }) (inc x) j
+
+symInvo : (x y : A) (p : x ≡ y) →
+     Path (Path A x y) p ((p ⁻¹)⁻¹)
+symInvo x y p = refl
+
+rUnit : (x y : A) (p : x ≡ y) →
```
mortberg: Why no prove compPathRefl using this?

Cubical/Foundations/GroupoidLaws.agda
```diff
@@ -0,0 +1,95 @@
...
+x≡y · y≡z = compPath x≡y y≡z 
+
+infixr 30 _·_
+
+_⁻¹ : {x y : A} → (x ≡ y) → (y ≡ x)
+x≡y ⁻¹ = sym x≡y
+
+-- homogeneous groupoid laws
+
+symRefl : (x : A) →
```
mortberg: How about renaming refl to 1?

Cubical/Core/Prelude.agda
```diff
@@ -47,10 +47,15 @@ cong f p = λ i → f (p i)
...
 -- confusion with transp
-compPath : x ≡ y → y ≡ z → x ≡ z
-compPath {x = x} p q i =
-  hcomp (λ j → \ { (i = i0) → x
-                 ; (i = i1) → q j }) (p i)
+
+compPath-filler : ∀ {x y z : A} → x ≡ y → y ≡ z → I → I → A
+compPath-filler {x = x} p q j i =
+  hfill (λ j → λ { (i = i0) → x
+                  ; (i = i1) → q j }) (inc (p i)) j
```
ice1000: Sounds like you need to reduce one space of indentation.

Cubical/Foundations/GroupoidLaws.agda
```diff
@@ -0,0 +1,95 @@
...
+  Path (Path A x x) refl (refl ⁻¹)
+symRefl x i = refl 
+
+compPathRefl : (x : A) →
+  Path (Path A x x) refl (refl · refl)
+compPathRefl x j i =
+  hfill (λ j → λ { (i = i0) → x
+                  ; (i = i1) → x }) (inc x) j
```
ice1000: Ditto, applies to all the same case below.

Cubical/Foundations/GroupoidLaws.agda
```diff
@@ -0,0 +1,95 @@
...
+   Path (Path A x x) (p · (p ⁻¹)) refl
+rCancel x y p j i =
+  hcomp (λ k → λ { (i = i0) → x
+                  ; (i = i1) → p (~ k ∧ ~ j)
+               -- ; (j = i0) → hfill (λ w → λ { (i = i0) → x; (i = i1) → p (~ w) }) (inc (p i)) k
+                  ; (j = i1) → x }) (p (i ∧ ~ j))
+
+lCancel : (x y : A) (p : x ≡ y) →
+   Path (Path A y y) refl ((p ⁻¹) · p)
+lCancel x y p k i =
```
3abc: I suggest making `(x y : A)` implicit. Also you can do

```
lCancel x y p i = rCancel y x (p ⁻¹) (~ i)
```

instead of a complicated `hcomp`.

Cubical/Core/Prelude.agda
```diff
@@ -47,10 +47,15 @@ cong f p = λ i → f (p i)
...
+
+compPath : ∀ {x y z : A} → x ≡ y → y ≡ z → x ≡ z
+compPath {x = x} p q j =
```
3abc: Just write `compPath p q = compPath-filler p q i1`

Cubical/Foundations/GroupoidLaws.agda
bbentzen: Oh, I was being silly. Actually compPathRefl was just an exercise, I forgot to go back to it after I proved rUnit!

Cubical/Foundations/GroupoidLaws.agda
bbentzen: That's really nice!

Cubical/Foundations/GroupoidLaws.agda
3abc: > I found a really cool unicode symbol instead, 𝟭.

It's a library, a weird unicode like 𝟭 will make people unable to use it.

I think what @mortberg meant was rename `symRefl` as `sym1`?

bbentzen: yeah, let's wait for his response :)

Cubical/Core/Prelude.agda
potato4444: @3abc that doesn't work because `compPath-filler p q i1` has a function type, but we are making a path.

Cubical/Foundations/GroupoidLaws.agda
mortberg: Ah I see, let's not do this then.

Cubical/Core/Prelude.agda
3abc: Oh OK.

3abc: I suggest giving `compPath-filler` a more specific type. 

Maybe bring back the definition of `compPath` and write

```
compPath-filler : ∀ {x y z : A} → (p : x ≡ y) → (q : y ≡ z)
                → PathP (λ i → x ≡ q i) p (compPath p q)
compPath-filler {x = x} p q j i =
  hfill (λ j → λ { (i = i0) → x
                  ; (i = i1) → q j }) (inc (p i)) j
```

This helps to reduce user's mental overhead when they want to use `compPath-filler`

bbentzen: > I suggest giving `compPath-filler` a more specific type.
> 
> Maybe bring back the definition of `compPath` and write
> 
> ```
> compPath-filler : ∀ {x y z : A} → (p : x ≡ y) → (q : y ≡ z)
>                 → PathP (λ i → x ≡ q i) p (compPath p q)
> compPath-filler {x = x} p q j i =
>   hfill (λ j → λ { (i = i0) → x
>                  ; (i = i1) → q j }) (inc (p i)) j
> ```
> This helps to reduce user's mental overhead when they want to use `compPath-filler`

I'd rather not have redundancy --- shouldn't the faces of the filler be explicitly clear from the hfill? It is natural to define the filler first and then extract the Kan composite from it, I think. That's one of the good practices of Redtt (see [this](https://github.com/RedPRL/redtt/blob/master/library/prelude/path.red) file). But that's just my opinion. :)

3abc: > I'd rather not have redundancy --- shouldn't the faces of the filler be explicitly clear from the hfill?

One of the points of a library is that user can have an interface to use it without having to look at details of the implementation. Yes it is kinda clear from `hfill`. But for a user who want to use `compPath-filler` when doing something else, the type of `compPath-filler` is the user interface, and `hfill` is just how you implement `compPath-filler`, which is irrelevant to the user.

So writing it as `I → I → A` it's rather unenlightening.

Also, `hfill` will reduce to `hcomp` anyway so define `compPath` first is not redundant at all.

3abc: Maybe at least put the more precise type `PathP (λ i → x ≡ q i) p (compPath p q)` in comment?

bbentzen: > Maybe at least put the more precise type `PathP (λ i → x ≡ q i) p (compPath p q)` in comment?

That's a great idea! :+1:

Cubical/Core/Prelude.agda
mortberg: Indeed. @bbentzen can you please fix?

Cubical/Foundations/GroupoidLaws.agda
```diff
@@ -0,0 +1,93 @@
...
+
+private
+  variable
+    ℓ : Level
+    A : Set ℓ
+
+-- some useful notation
+
+_·_ :  {x y z : A} → (x ≡ y) → (y ≡ z) → (x ≡ z)
+x≡y · y≡z = compPath x≡y y≡z
```
mortberg: This should be upstreamed and be the definition of compPath. How about doing it in this PR? (The font in my browser is a bit weird, but to input this symbol you use `\.` right?)

Cubical/Foundations/GroupoidLaws.agda
```diff
@@ -0,0 +1,93 @@
...
+
+_·_ :  {x y z : A} → (x ≡ y) → (y ≡ z) → (x ≡ z)
+x≡y · y≡z = compPath x≡y y≡z 
+
+infixr 30 _·_
+
+_⁻¹ : {x y : A} → (x ≡ y) → (y ≡ x)
+x≡y ⁻¹ = sym x≡y
+
+𝟭 : {x : A} → x ≡ x
```
mortberg: I think I suggested adding this, maybe that was a mistake. How about changing it back to refl?

Cubical/Foundations/GroupoidLaws.agda
```diff
@@ -0,0 +1,93 @@
...
+𝟭 = refl
+
+-- homogeneous groupoid laws
+
+symInvo : {x y : A} (p : x ≡ y) →
+     Path (Path A x y) p ((p ⁻¹)⁻¹)
```
mortberg: Why not use `≡` instead of the explicit `Path`?

Cubical/Foundations/GroupoidLaws.agda
```diff
@@ -0,0 +1,93 @@
...
+symInvo p = 𝟭
+
+rUnit : {x y : A} (p : x ≡ y) →
+  Path (Path A x y) p (p · 𝟭)
```
mortberg: Use `≡`

Cubical/Foundations/GroupoidLaws.agda
3abc: > This should be upstreamed and be the definition of compPath. How about doing it in this PR? (The font in my browser is a bit weird, but to input this symbol you use `\.` right?)

I think he used `\cdot`. I prefer `\.` too.

bbentzen: > This should be upstreamed and be the definition of compPath. How about doing it in this PR?

Do you mean it should be definition or notation of compPath?

>(The font in my browser is a bit weird, but to input this symbol you use `\.` right?)

I used `\cdot`, but I don't have any particular attachment to it. Let me change it to `\.`.

Cubical/Foundations/GroupoidLaws.agda
bbentzen: Yeah, that's true --- I will do that in my next commit.

Cubical/Foundations/GroupoidLaws.agda
bbentzen: Oh, how come I didn't think of it?

Cubical/Core/Prelude.agda
bbentzen: Sure! :)

Cubical/Core/Prelude.agda
bbentzen: @mortberg What do you think? Should fillers come before their composite or the other way around? (I need to define more fillers for the heterogeneous groupoid laws.)

mortberg: Let's do fillers first, but with the type in a comment

Cubical/Foundations/GroupoidLaws.agda
mortberg: It should be the definition of compPath

PR #83, opened 2019-03-11
Equality of Sigma-types, truncatedness of Sigma-types and products

loic-p:
Add an alternative equality for Sigma-types, following the HoTT Book. Use it to prove truncatedness results for Sigma-types, and specialize them to nondependent products.

PR #84, opened 2019-03-11
some stuff

riaqn:
proving that set quotient from
a discrete type modulo a decidable equivalance relation is discrete.
Along the way I proved lots of other useful results too, including:

1. defined half adjoint equivalence, and convert isomorphism to it
2. (cong f) where f is an equivalence, is also an equivalence
3. A ≡ B is a N-type, if both A and B are N-types.
4. the collection of all N-types is itself a N+1 type
5. some other results for compPath, including right unit, right
inverse, and associtivity; they are done purely cubically.


riaqn:
As some of my result overlaps with Loic's (https://github.com/agda/cubical/pull/83), I suggest we merge Loic's PR first, then I will do the conflict resolution and code re-organization. After that we can consider merge my PR.

Cubical/Core/Prelude.agda
```diff
@@ -78,12 +78,58 @@ compPathP : {A : I → Set ℓ} → {x : A i0} → {y : A i1} → {B_i1 : Set 
   (p : PathP A x y) → (q : PathP (λ i → B i) y z) → PathP (λ j → ((λ i → A i) ∙ B) j) x z
 compPathP p q j = compPathP-filler p q j i1
 
+∙-rInv : (p : x ≡ y) → p ∙ (sym p) ≡ refl
```
mortberg: This proof and the next are in https://github.com/agda/cubical/blob/master/Cubical/Foundations/GroupoidLaws.agda

Cubical/Core/Prelude.agda
```diff
@@ -78,12 +78,58 @@ compPathP : {A : I → Set ℓ} → {x : A i0} → {y : A i1} → {B_i1 : Set 
...
+                 ; (j = i0) → x
+                 ; (j = i1) → y }) (p j)
+
 _≡⟨_⟩_ : (x : A) → x ≡ y → y ≡ z → x ≡ z
 _ ≡⟨ x≡y ⟩ y≡z = x≡y ∙ y≡z
 
 _∎ : (x : A) → x ≡ x
 _ ∎ = refl
 
+-- another definition of composition, useful for some proofs
```
mortberg: Interesting, can you point me to some proofs where this is useful?

I think this should be defined using the triple composition of @loic-p :  https://github.com/agda/cubical/blob/master/Cubical/Foundations/GroupoidLaws.agda#L179

Maybe the same holds for the normal compPath (now called \.)? In fact, maybe we can develop some lemmas for the triple composition and then recover the lemmas for compPath and compPath' from the general one?

Cubical/Core/Prelude.agda
```diff
@@ -78,12 +78,58 @@ compPathP : {A : I → Set ℓ} → {x : A i0} → {y : A i1} → {B_i1 : Set 
...
+                       y
+
+∙-assoc' : {w : A} (p : x ≡ y) (q : y ≡ z) (r : z ≡ w) → (p ∙ q) ∙ r ≡ p ∙ (q □ r)
+∙-assoc' {x = x} p q r i j = hcomp (λ k → \ { (i = i0) → compPath-filler (p ∙ q) r k j
+                                            ; (i = i1) → compPath-filler p (q □ r) k j
+                                            ; (j = i0) → x
+                                            ; (j = i1) → compPath'-filler q r i k})
+                                   (compPath-filler p q (~ i) j)
+
+∙-assoc : {w : A} (p : x ≡ y) (q : y ≡ z) (r : z ≡ w) → (p ∙ q) ∙ r ≡ p ∙ (q ∙ r)
```
mortberg: This is already proved in the GroupoidLaws file, but maybe this proof is slicker

Cubical/Foundations/Equiv.agda
```diff
@@ -71,3 +87,13 @@ compEquiv f g = isoToEquiv
 --   invEquivInvol f i .fst = fst f
 --   invEquivInvol f i .snd = propIsEquiv (fst f) (snd (invEquiv (invEquiv f))) (snd f) i
 
+-- Transport is an equivalence
+isEquivTransport : ∀ {ℓ} {A B : Set ℓ} (p : A ≡ B) → isEquiv (transport p)
```
mortberg: This is and the next result is in: https://github.com/agda/cubical/blob/master/Cubical/Foundations/Transport.agda#L28

(with better proofs)

Cubical/Foundations/HAEquiv.agda
```diff
@@ -0,0 +1,120 @@
...
+HAEquiv : ∀ {ℓ ℓ'} (A : Set ℓ) (B : Set ℓ') → Set (ℓ ⊔ ℓ')
+HAEquiv A B = Σ (A → B) λ f → isHAEquiv f
+
+private
+  variable
+    ℓ ℓ' : Level
+    A : Set ℓ
+    B : Set ℓ'
+
+homotopyNatural' : {f g : A → B} (H : ∀ a → f a ≡ g a) {x y : A} (p : x ≡ y) → (H x) ∙ (cong g p) ≡ (cong f p) □ (H y)
```
mortberg: Maybe this and the next result should be put in `private` ?

Cubical/Foundations/HAEquiv.agda
```diff
@@ -0,0 +1,120 @@
...
+                 ; (i = i1) → compPath'-filler (cong f p) (H y) k j
+                 ; (j = i0) → cong f p (i ∧ (~ k))
+                 ; (j = i1) → cong g p (i ∨ k) })
+        (H (p i) j)
+
+homotopyNatural : {f g : A → B} (H : ∀ a → f a ≡ g a) {x y : A} (p : x ≡ y) → (H x) ∙ (cong g p) ≡ (cong f p) ∙ (H y)
+homotopyNatural H p = (homotopyNatural' H p) ∙ (□≡∙ _ _)
+
+Hfa≡fHa : ∀ {ℓ} {A : Set ℓ} (f : A → A) (H : ∀ a → f a ≡ a) → ∀ a → H (f a) ≡ cong f (H a)
+Hfa≡fHa {A = A} f H a = H (f a) ≡⟨ sym (∙-rUnit (H (f a))) ⟩
```
mortberg: Indentation is weird here, can you fix it so that it's aligned properly?

Cubical/Foundations/HAEquiv.agda
```diff
@@ -0,0 +1,120 @@
...
+    com a i j = hcomp (λ k → \ { (i = i0) → ε (f (η a j)) k
+                                ; (i = i1) → hfill (sides (f a) j) (inc (bot (f a) j)) k
+                                ; (j = i0) → ε (f (g (f a))) k
+                                ; (j = i1) → ε (f a) k})
+                       ( cong (cong f) (sym (Hfa≡fHa (λ x → g (f x)) η a)) i j)
+
+equiv→HAEquiv : A ≃ B → HAEquiv A B
+equiv→HAEquiv e = iso→HAEquiv (equivToIso e)
+
+congEquiv : ∀ {ℓ ℓ'} {A : Set ℓ} {B : Set ℓ'} {x y : A} (e : A ≃ B) → (x ≡ y) ≃ (e .fst x ≡ e .fst y)
```
mortberg: This is proved somewhere else as discussed, but that proof was not as general? (A and B had to be in the same universe)

Maybe this proof is the one that should be used everywhere?

Cubical/Foundations/HLevels.agda
```diff
@@ -25,7 +25,6 @@ hProp {ℓ} = Σ (Set ℓ) isProp
 
 isOfHLevel : ∀ {ℓ} → ℕ → Set ℓ → Set ℓ
 isOfHLevel 0 A = isContr A
-isOfHLevel 1 A = isProp A
```
mortberg: As discussed in some other thread I'm not sure if we should do this change or not... 

@3abc : what do you think?

Cubical/Foundations/HLevels.agda
```diff
@@ -103,22 +111,39 @@ hLevelPi : ∀ {ℓ ℓ'} {A : Set ℓ} {B : A → Set ℓ'} n
...
+  sub-lemma : isOfHLevel n (∀ x → f x ≡ g x)
+  sub-lemma = hLevelPi n λ x → h x (f x) (g x)
+
+isSetPi : ∀ {ℓ ℓ'} {A : Set ℓ} {B : A → Set ℓ'}
+  → ((x : A) → isSet (B x))
+  → isSet ((x : A) → B x)
+isSetPi Bset = isOfHLevel2→isSet (hLevelPi 2 (λ a → isSet→isOfHLevel2 (Bset a)))
+
+
+isSet→isSet' : ∀ {ℓ} {A : Set ℓ} → isSet A → isSet' A
```
mortberg: Long lines in this proof

Cubical/HITs/SetQuotients/Properties.agda
```diff
@@ -17,26 +17,147 @@ open import Cubical.HITs.SetTruncation
...
 open import Cubical.Foundations.HLevels
+open import Cubical.Relation.Nullary
+open import Cubical.Relation.Binary.Base
+open import Cubical.Data.Nat.Base
+open import Cubical.Foundations.Univalence
+open import Cubical.Data.Nat.Properties
+open import Cubical.Foundations.HAEquiv
+
+
+Σ≡ : ∀ {ℓ ℓ'} {A : Set ℓ} {B : A → Set ℓ'}
```
mortberg: Some of the results here are already proved in https://github.com/agda/cubical/pull/83, so let's remove/move them to the proper file once that is merged

Cubical/HITs/SetQuotients/Properties.agda
```diff
@@ -82,3 +203,53 @@ setQuotUniversal Bset = isoToEquiv (iso intro elim elimRightInv elimLeftInv)
     (λ sur → Bset (elim (intro g) x) (g x))
     (λ sur → cong (elim (intro g)) (sym (snd sur)) ∙ (cong g (snd sur))) ([]surjective x)
     )
+
+reflect : (Rprop : BinaryRelation.isProp R) (Requiv : BinaryRelation.isEquiv R) (a b : A) → [ a ] ≡ [ b ] → R a b
```
mortberg: Why is BinaryRelation qualified?

mortberg: Why is this called `reflect`? I would call it `effective`

Cubical/HITs/SetQuotients/Properties.agda
```diff
@@ -82,3 +203,53 @@ setQuotUniversal Bset = isoToEquiv (iso intro elim elimRightInv elimLeftInv)
...
+    elim : R a b → [ a ] ≡ [ b ]
+    elim = eq/ a b
+
+    intro-elim : ∀ x → intro (elim x) ≡ x
+    intro-elim ab = Rprop a b _ _
+
+    elim-intro : ∀ x → elim (intro x) ≡ x
+    elim-intro eq = squash/ _ _ _ _
+
+discreteSetQuotients : Discrete A → BinaryRelation.isProp R → BinaryRelation.isEquiv R → (∀ a₀ a₁ → Dec (R a₀ a₁)) → Discrete (A / R)
```
mortberg: Long lines

Cubical/Core/Prelude.agda
riaqn: yes, that sounds a good idea. In fact, even better, we can define compPath x y as tripleComp refl x y, and compPath' x y as tripleComp x y refl. I will work on this.

Cubical/Core/Prelude.agda
riaqn: I actually quite like the proof in GroupoidLaws, will use that.

Cubical/Foundations/HLevels.agda
3abc: > As discussed in some other thread I'm not sure if we should do this change or not...
> 
> @3abc : what do you think?

How about just keeping two `isOfHLevel`s, one is defined using three cases and one is defined using two? And prove their equivalence. The problem of the current version is not only we will need 

```Agda
isProp→IsOfHLevel1 : ∀ {ℓ} {A : Set ℓ} → isProp A → isOfHLevel 1 A
isOfHLevel1→isProp : ∀ {ℓ} {A : Set ℓ} → isOfHLevel 1 A → isProp A
isSet→isOfHLevel2 : ∀ {ℓ} {A : Set ℓ} → isSet A → isOfHLevel 2 A
isOfHLevel2→isSet : ∀ {ℓ} {A : Set ℓ} → isOfHLevel 2 A → isSet A
```

We will also need things like `isGroupoid→isOfHLevel3` and `isTwoGroupoid→isOfHLevel4` when working with higher truncations. For truncations definitional equality is good because then `isOfHLevel 4` can be used as higher constructors.

3abc: And if you really want truncation eliminators to compute fast, too much going back and forth will be bad. (This is ironical, what's good about HoTT is one can transport back and forth, in theory it's doable, in practice some of us try to avoid transporting too much so that things really compute.)

Cubical/HITs/SetQuotients/Properties.agda
loic-p: Note that the equality for sigma-types we used is different : in #83 my definition is basically the one from the HoTT book, while this one makes use of the native dependent paths in CuTT.

Even though the HoTT-style equality can be useful at times, this one is basically a primitive so I think it is a better way of proving properties for sigma-types, and therefore that the proof of hLevelSigma below should be the one kept when merging.

mortberg: Interesting! I didn't realize that @loic-p used the HoTT book version. I would expect the one using native dependent paths to be a bit easier to use, so let's indeed use this one.

PR #85, opened 2019-03-12
Renamed inc/ouc to inS/outS.

Saizan:


jonsterling:
How about `inS/auS`? :sunglasses:

Saizan:
Ok, unless any other PRs are ready today and could go first.

3abc:
How about Agda internals? Are they going to change? I assume now when one normalizes things one still gets `inc` `ouc`'s in the information window?

Saizan:
@3abc that shouldn't happen, because it's been done with the `renaming` feature.

3abc:
> @3abc that shouldn't happen, because it's been done with the `renaming` feature.

What is that? Any reference?

Saizan:
https://agda.readthedocs.io/en/v2.5.4.2/language/module-system.html#name-modifiers

PR #86, opened 2019-03-14
Prove Hedberg without J

3abc:


mortberg:
Wow, that's really nice!

PR #89, opened 2019-03-18
Privatise generalised declared variables

L-TChen:
I guess the generalised declared variables in this module are by accident not private.


mortberg:
Thanks!

PR #92, opened 2019-03-20
Bool properties

L-TChen:
Some simple properties are added to Cubical.Data.Bool: 

- true is not false
- Bool is discrete 
- Bool with _or_ is a bounded semi-lattice  

I need these facts to show that the free semi-lattice (Kuratowski finite sets) is not trivial. It might be helpful for them to stay in Cubical?


mortberg:
Thanks!

PR #93, opened 2019-03-20
Make Zesen's code happy

mortberg:


mortberg:
I'll merge this now even though it's not perfect. 

@riaqn will fix the duplication in Prelude and GroupoidLaws

I will try to figure out how to organize all of the results about equivalences and hlevels better

PR #95, opened 2019-03-22
Add cong₂ to Cubical.Core.Prelude 

L-TChen:
Add 

* cong₂ : (p : x ≡ y) (q : PathP (λ i → B (p i)) u v) → PathP (λ i → C (p i) (q i)) (f x u) (f y v)

It is handy in particular when proving properties about binary operations. 

I can also simplify other proofs using cong₂ later if time permits.

### DIFF
--- a/Cubical/Core/Prelude.agda
+++ b/Cubical/Core/Prelude.agda
@@ -35,7 +35,7 @@ infixr 2 _≡⟨_⟩_
 
 private
   variable
-    ℓ ℓ' : Level
+    ℓ ℓ' ℓ'' : Level
     A : Set ℓ
     x y z : A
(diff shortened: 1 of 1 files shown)


mortberg:
I almost always inline this kind of proofs, but I guess it could be instructive to see the types and sometimes handy to have when doing equational reasoning.

L-TChen:
> sometimes handy to have when doing equational reasoning.

Yes, that is exactly why I come to this.

mortberg:
> > sometimes handy to have when doing equational reasoning.
> 
> Yes, that is exactly why I come to this.

Yeah, thanks for adding it!

PR #96, opened 2019-03-22
add DiffInt

riaqn:
Some warning (constraint unsolved) remaining and I can't fix.

### DIFF
--- a/Cubical/Data/DiffInt.agda
+++ b/Cubical/Data/DiffInt.agda
@@ -1,0 +1,5 @@
+{-# OPTIONS --cubical --safe #-}
+module Cubical.Data.DiffInt where
+
+open import Cubical.Data.DiffInt.Base public
+open import Cubical.Data.DiffInt.Properties public
--- a/Cubical/Data/DiffInt/Base.agda
+++ b/Cubical/Data/DiffInt/Base.agda
@@ -1,0 +1,26 @@
+{-# OPTIONS --cubical --safe #-}
+module Cubical.Data.DiffInt.Base where
+
+open import Cubical.HITs.SetQuotients.Base
+open import Cubical.Data.Nat
+
+open import Cubical.Foundations.HLevels
+
+open import Cubical.Core.Prelude
+open import Cubical.Relation.Nullary
+open import Cubical.Relation.Nullary.DecidableEq
+open import Cubical.Relation.Binary.Base
+
+open import Cubical.Core.Glue
+open import Cubical.Foundations.Equiv
+open import Cubical.Data.Prod
+open import Cubical.Data.Sigma
+
+rel : (ℕ ×Σ ℕ) → (ℕ ×Σ ℕ) → Set
+rel (a₀ , b₀) (a₁ , b₁) = x ≡ y
+  where
+    x = a₀ + b₁
+    y = a₁ + b₀
+
+ℤ = (ℕ ×Σ ℕ) / rel
+
--- a/Cubical/Data/DiffInt/Properties.agda
+++ b/Cubical/Data/DiffInt/Properties.agda
@@ -1,0 +1,50 @@
+{-# OPTIONS --cubical --safe #-}
+module Cubical.Data.DiffInt.Properties where
+
+open import Cubical.Relation.Binary.Base
+open import Cubical.Relation.Nullary
+open import Cubical.Data.DiffInt.Base
+open import Cubical.Core.Prelude
+open import Cubical.Data.Nat
+open import Cubical.Core.Glue
+open import Cubical.Data.Sigma
+open import Cubical.HITs.SetQuotients
+open import Cubical.Data.Prod
+
+open BinaryRelation
+
+relIsEquiv : isEquivRel rel
+relIsEquiv = EquivRel relIsRefl relIsSym relIsTrans
+  where
+    relIsRefl : isRefl rel
+    relIsRefl {a = (a0 , a1)} = refl
+
+    relIsSym : isSym rel
+    relIsSym {a = (a0 , a1)} {b = (b0 , b1)} p = sym p
+
+    relIsTrans : isTrans rel
+    relIsTrans {a = (a0 , a1)} {b = (b0 , b1 )} {c = (c0 , c1)} p0 p1 =
+      inj-m+ {m = (b0 + b1)} ((b0 + b1) + (a0 + c1) ≡⟨ +-assoc (b0 + b1) a0 c1  ⟩
+            ((b0 + b1) + a0) + c1 ≡⟨ cong (λ x → x + a0 + c1) (+-comm b0 b1)⟩
+            ((b1 + b0) + a0) + c1 ≡⟨ cong (λ x → x + c1) (+-comm (b1 + b0) a0) ⟩
+            (a0 + (b1 + b0)) + c1 ≡⟨ cong (λ x → x + c1) (+-assoc a0 b1 b0) ⟩
+            (a0 + b1) + b0 + c1 ≡⟨ sym (+-assoc (a0 + b1) b0 c1) ⟩
+            (a0 + b1) + (b0 + c1) ≡⟨ cong (λ p → p . fst + p .snd) (transport (ua Σ≡) (p0 , p1))⟩
+            (b0 + a1) + (c0 + b1) ≡⟨ sym (+-assoc b0 a1 (c0 + b1))⟩
+            b0 + (a1 + (c0 + b1)) ≡⟨ cong (λ x → b0 + (a1 + x)) (+-comm c0 b1)⟩
+            b0 + (a1 + (b1 + c0)) ≡⟨ cong (λ x → b0 + x) (+-comm a1 (b1 + c0)) ⟩
+            b0 + ((b1 + c0) + a1) ≡⟨ cong (λ x → b0 + x) (sym (+-assoc b1 c0 a1))⟩
+            b0 + (b1 + (c0 + a1)) ≡⟨ +-assoc b0 b1 (c0 + a1)⟩
+            (b0 + b1) + (c0 + a1) ∎ )
+
+relIsProp : BinaryRelation.isPropValued rel
+relIsProp a b x y = isSetℕ _ _ _ _
+
+discreteℤ : Discrete ℤ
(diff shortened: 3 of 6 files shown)


mortberg:
I made a PR to your local branch fixing the yellow. Merge it in and then this should be ready

riaqn:
I merged it.

mortberg:
Can you please also add DiffInt to the Everything file in Data?

riaqn:
There seems to be some glitch in my code - when I try to normalize `discreteℤ [(2 , 4)] [(3 , 5)]`, it gives me back `yes` of something, which is expected.

But when I normalize `discreteℤ [(2 , 4)] [(3 , 4)]`, agda just freezes and takes 100% CPU. It should give me `no` of something.

3abc:
> There seems to be some glitch in my code (or Agda) - when I try to normalize `discreteℤ [(2 , 4)] [(3 , 5)]`, it gives me back `yes` of something, which is expected.
> 
> But when I normalize `discreteℤ [(2 , 4)] [(3 , 4)]`, agda just freezes and takes 100% CPU. It should give me `no` of something.

Likely just an Agda's performance problem... Maybe `discreteSetQuotients` is too complicated for Agda to normalize.

3abc:
> Likely just an Agda's performance problem...

My suspicion is, even if you make some relatively simple equivalence relation on `Nat` like `m ~ n` iff `m - n` is even. It's going to hang if you prove `discrete Nat/~` and evaluate it on `[0]` `[1]`

riaqn:
@3abc you are right. I tried an even more trivial one (that 0-0 = 0-1) and it still hangs.

However, I'm not convinced it's a performance problem than it's a BUG. I don't think the normalization involved is that complex.

mortberg:
> @3abc you are right. I tried an even more trivial one (that 0-0 = 0-1) and it still hangs.
> 
> However, I'm not convinced it's a performance problem than it's a BUG. I don't think the normalization involved is that complex.

Are you tests the same as:

https://github.com/mortberg/cubicaltt/blob/master/examples/hz.ctt#L63

?

Computing `0=1` and `1=1` is instant:
```
> test01
EVAL: false
Time: 0m0.089s
> test11
EVAL: true
Time: 0m0.093s
```
So this sounds like either a bug or that your proofs are too complicated.

Cubical/Relation/Binary/Base.agda
```diff
@@ -9,13 +9,13 @@ open import Cubical.HITs.SetQuotients.Base
 
 module BinaryRelation {ℓ ℓ' : Level} {A : Set ℓ} (R : A → A → Set ℓ') where
   isRefl : Set (ℓ-max ℓ ℓ')
-  isRefl = {a : A} → R a a
```
Alizter: What is the reason for changing these?

mortberg: Agda couldn't infer the argument in DiffInt.Properties so things got yellow there. I believe that we often want these arguments explicitly when doing these proofs, so I'd rather have them explicit and insert underscores sometimes.

3abc:
> However, I'm not convinced it's a performance problem than it's a BUG. I don't think the normalization involved is that complex.

Could be a bug too. But given that used `J` twice and `relIsTrans` contains too much `hcomp` the normal form will just explode for Agda to normalize...

Here is an example, I defined a really simple relation `rel2` below and asked Agda to normalize `discreteN/2 [ 0 ] [ 1 ]` it took 35.8 seconds of 100% CPU to produce a normal form of `42148` lines. In your case the normal form could be easily more than 4 million lines. (Now you might have some idea why in the other thread we want things to be definitionally equal badly, and try to avoid using `J`.)

```Agda
open import Cubical.Data.Empty public

data ⊤ : Set where
  t : ⊤

rel2 : ℕ → ℕ → Set
rel2 0 0 = ⊤
rel2 0 (suc n) = ⊥
rel2 (suc n) 0 = ⊥
rel2 (suc n) (suc m) = ⊤

N/2 = ℕ / rel2

rel2IsProp : BinaryRelation.isPropValued rel2
rel2IsProp zero zero t t = refl
rel2IsProp zero (suc b) x y = ⊥-elim x
rel2IsProp (suc a) zero x y = ⊥-elim x
rel2IsProp (suc a) (suc b) t t  = refl

rel2IsEquiv : isEquivRel rel2
rel2IsEquiv = EquivRel {A = ℕ} rel2IsRefl rel2IsSym rel2IsTrans
  where
    rel2IsRefl : isRefl rel2
    rel2IsRefl zero = t
    rel2IsRefl (suc a) = t

    rel2IsSym : isSym rel2
    rel2IsSym zero zero t = t
    rel2IsSym zero (suc b) x = ⊥-elim x
    rel2IsSym (suc a) zero x = ⊥-elim x
    rel2IsSym (suc a) (suc b) t = t
    
    rel2IsTrans : isTrans rel2
    rel2IsTrans zero zero c x y = y
    rel2IsTrans zero (suc b) c x _ = ⊥-elim x
    rel2IsTrans (suc a) zero c x _ = ⊥-elim x
    rel2IsTrans (suc a) (suc b) 0 x y = ⊥-elim y
    rel2IsTrans (suc a) (suc b) (suc c) t t = t

N/2IsDiscrete : (a b : ℕ) → Dec (rel2 a b)
N/2IsDiscrete zero zero = yes t
N/2IsDiscrete zero (suc b₁) = no λ x → x
N/2IsDiscrete (suc a₁) zero = no λ x → x
N/2IsDiscrete (suc a₁) (suc b₁) = yes t

discreteN/2 : Discrete N/2
discreteN/2 = discreteSetQuotients discreteℕ rel2IsProp rel2IsEquiv N/2IsDiscrete
```

riaqn:
@3abc thanks for the example. That's indeed suprising to me; I didn't expect such a simple proposition would need such a big proof.

Could you post the normalized form somewhere so I can look at it?

3abc:
> Could you post the normalized form somewhere so I can look at it?

Won't be too hard for you to just normalize it LoL (40 seconds!). It's so long that when I use `M-e` to jump to the end emacs hangs for a few seconds... And I've no idea how to export.

riaqn:
ahh, ok I got it.

yeah, this looks really bad. But I assume in general, if the type of the term is some inductive type, then by canonicity, the normalized form should be very short? Also I guess the process shouldn't take long because the term would be kept short?

riaqn:
> > @3abc you are right. I tried an even more trivial one (that 0-0 = 0-1) and it still hangs.
> > However, I'm not convinced it's a performance problem than it's a BUG. I don't think the normalization involved is that complex.
> 
> Are you tests the same as:
> 
> https://github.com/mortberg/cubicaltt/blob/master/examples/hz.ctt#L63
> 
> ?
> 
> Computing `0=1` and `1=1` is instant:
> 
> ```
> > test01
> EVAL: false
> Time: 0m0.089s
> > test11
> EVAL: true
> Time: 0m0.093s
> ```
> So this sounds like either a bug or that your proofs are too complicated.

I believe there is difference between just saying "yes/no" and saying "yes/no" with a reason. The reason is what costs us.

mortberg:
@Saizan I can think of two reasons for why this is so bad in Cubical Agda:

1. We are always doing "endpoint correction" in all parametrized HITs right? 

2. We should implement the transp for hcomp in U as in cubicaltt using the optimizations I mentioned a while ago. This should be more efficient than going through transp in Glue.

mortberg:
> > > @3abc you are right. I tried an even more trivial one (that 0-0 = 0-1) and it still hangs.
> > > However, I'm not convinced it's a performance problem than it's a BUG. I don't think the normalization involved is that complex.
> > 
> > 
> > Are you tests the same as:
> > https://github.com/mortberg/cubicaltt/blob/master/examples/hz.ctt#L63
> > ?
> > Computing `0=1` and `1=1` is instant:
> > ```
> > > test01
> > EVAL: false
> > Time: 0m0.089s
> > > test11
> > EVAL: true
> > Time: 0m0.093s
> > ```
> > So this sounds like either a bug or that your proofs are too complicated.
> 
> I believe there is difference between just saying "yes/no" and saying "yes/no" with a reason. The reason is what costs us.

Oh right, can you write a function returning a boolean without expecting the reason? That seems like what we really want anyway.

3abc:
> I assume in general, if the type of the term is some inductive type, then by canonicity, the normalized form should be very short? Also I guess the process shouldn't take long because the term would be kept short?

Could be. If I define
```Agda
r : {A : Set} → Dec A → ℕ
r (yes p) = 1
r (no ¬p) = 0
```
then it's easy for Agda to compute `r` of those terms. Returns instantly.

riaqn:
@mortberg  https://github.com/agda/cubical/pull/98

mortberg:
> > I assume in general, if the type of the term is some inductive type, then by canonicity, the normalized form should be very short? Also I guess the process shouldn't take long because the term would be kept short?
> 
> Could be. If I define
> 
> ```agda
> r : {A : Set} → Dec A → ℕ
> r (yes p) = 1
> r (no ¬p) = 0
> ```
> 
> then it's easy for Agda to compute `r` of those terms. Returns instantly.

Ok, very good! What might one want to have the normal form of the proof witnesses for? I don't find it too surprising that they are gigantic and hence unusable...

riaqn:
I think we are good to merge now. 

PS. we will soon need some tests to make sure the code works as we expected.

mortberg:
> I think we are good to merge now.
> 
> PS. we will soon need some tests to make sure the code works as we expected.

Sounds good. I restarted the Travis build and once it's green I'll merge this. 

Please make a PR with some tests checking that it all computes as expected

riaqn:
How about we put the tests in the same file with `private`?
```
private
  test0 : Dec→Bool (discreteℤ [ (3 , 5) ] [ (4 , 6) ]) ≡ true
  test0 = refl

  test1 : Dec→Bool (discreteℤ [ (3 , 5) ] [ (4 , 7) ]) ≡ false
  test1 = refl
```

mortberg:
> How about we put the tests in the same file with `private`?
> 
> ```
> private
>   test0 : Dec→Bool (discreteℤ [ (3 , 5) ] [ (4 , 6) ]) ≡ true
>   test0 = refl
> 
>   test1 : Dec→Bool (discreteℤ [ (3 , 5) ] [ (4 , 7) ]) ≡ false
>   test1 = refl
> ```

Exactly what I had in mind. You can even name them `_`

riaqn:
Great. Do you think it should be in this PR, or  should be a new PR?

mortberg:
Just do it in this one

PR #98, opened 2019-03-23
add Dec→Bool

riaqn:


mortberg:
Oops, I merged this too fast. It turns out that it doesn't typecheck (unsolved metas). I pushed a fix straight to master just now

PR #99, opened 2019-03-23
Logic properties

L-TChen:
Add few simple staff:

* `isPropUnit` in Data.Unit.Properties
* `elim-⊎` the elimination rule for coproduct type in Data.Sum.Base
* `⊔-identity` in Foundations.Logic 

Also fix the universe polymorphism in `Data.Sum.Base`. Hope these useful.

### DIFF
--- a/Cubical/Data/Sum/Base.agda
+++ b/Cubical/Data/Sum/Base.agda
@@ -3,22 +3,19 @@ module Cubical.Data.Sum.Base where
 
 open import Cubical.Core.Everything
 
-data _⊎_ {ℓ ℓ'} (P : Set ℓ) (Q : Set ℓ') : Set (ℓ-max ℓ ℓ') where
-  inl : P → P ⊎ Q
-  inr : Q → P ⊎ Q
-
 private
   variable
-    ℓ ℓ' ℓ'' ℓ''' : Level
-    A : Set ℓ
-    B : Set ℓ'
-    C : Set ℓ''
-    D : Set ℓ'''
+    ℓ ℓ' : Level
+    A B C D : Set ℓ
+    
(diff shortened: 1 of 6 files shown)


L-TChen:
And few more proofs 
* Add the proof that `isProp A ≡ (∀ (x y : A) → isContr (x ≡ y))`
* Remove the unnecessary dependency on HLevel in Isomoprhism
* Fix universe polymorphism of subtypeEquality in HLevel
* Simplify isProp→PathP (?)

I need the fact that the type of mere proposition is a set. Probably it is better to prove a more general statement that the type of n-types is (n+1)-type.

Saizan:
Great! By the way, I wouldn't mind universe polymorphism here, I just wanted to start simple.

L-TChen:
Thanks for clarification. I think I will close this PR and make it a draft later.

Cubical/Data/Sum/Base.agda
```diff
@@ -9,8 +9,16 @@ data _⊎_ {ℓ ℓ'} (P : Set ℓ) (Q : Set ℓ') : Set (ℓ-max ℓ ℓ') wher
...
 private
   variable
-    ℓ : Level
-    A B C D : Set ℓ
+    ℓ ℓ' ℓ'' ℓ''' : Level
+    A : Set ℓ
+    B : Set ℓ'
+    C : Set ℓ''
+    D : Set ℓ'''
+
```
Saizan: You don't need all the ```ℓ' ℓ'' ℓ'''``` just using the single ```ℓ``` in the signatures of ```A```, ```B```, ```C```, and ```D``` would obtain the same result.

L-TChen: Thanks. I have them removed. I should be more careful when using new features ...

L-TChen:
Okay, I am closing this PR now and return to it later after finish #100.

PR #100, opened 2019-03-24
A direct proof of ΣProp≡ without Σ≡

L-TChen:
In order to prove that the type of n-types is an (n+1) type, I come up with a direct proof of  `ΣProp≡`.

### DIFF
--- a/Cubical/Core/Id.agda
+++ b/Cubical/Core/Id.agda
@@ -36,7 +36,7 @@ open import Agda.Builtin.Cubical.Id public
   hiding ( primIdJ ) -- this should not be used as it is using compCCHM
 open import Cubical.Core.Primitives public  hiding ( _≡_ )
 open import Cubical.Core.Prelude public
-  hiding ( _≡_ ; _≡⟨_⟩_ ; _∎ )
+  hiding ( _≡_ ; _≡⟨_⟩_ ; _∎ ; isPropIsContr)
   renaming ( refl      to reflPath
            ; transport to transportPath
            ; J         to JPath
@@ -250,7 +250,6 @@ isPropIsContr (a0 , p0) (a1 , p1) j =
                                                 ; (j = i0) → idToPath (p0 x) (i ∧ k)
                                                 ; (j = i1) → idToPath (p1 x) i })
                                        (idToPath (p0 (idToPath (p1 x) i)) j))))
-
 
 -- We now prove that isEquiv is a proposition
 isPropIsEquiv : ∀ {A : Set ℓ} {B : Set ℓ} → {f : A → B} → (h1 h2 : isEquiv f) → Path _ h1 h2
--- a/Cubical/Core/Prelude.agda
+++ b/Cubical/Core/Prelude.agda
@@ -37,6 +37,7 @@ private
   variable
     ℓ ℓ' ℓ'' : Level
     A : Set ℓ
+    B : A → Set ℓ
     x y z : A
 
 refl : x ≡ x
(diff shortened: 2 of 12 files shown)


mortberg:
Great! I think the reason I wanted another proof of this was so that we could then reorganize some other things. But I don't remember what anymore...

L-TChen:
By the way, where shall I put the proof that the type of n-types is an (n+1) type? 

`Data.Sigma.Properties` depends on `Foundations.HLevels`, but naturally `ΣProp≡` used in the proof should be in `Data.Sigma.Properties`...

mortberg:
> By the way, where shall I put the proof that the type of n-types is an (n+1) type?
> 
> `Data.Sigma.Properties` depends on `Foundations.HLevels`, but naturally `ΣProp≡` used in the proof should be in `Data.Sigma.Properties`...Perhaps it is the reason why you need direct proof?

Yes! That's it! With your new proof we should be able to put both results in HLevels.

mortberg:
So please move `ΣProp≡` and `HLevel≡` to `Cubical.Foundations.HLevels`.

Please also take a look in `Cubical.Foundations.HAEquiv``and move whatever can be moved upstream after "TODO: all of these should be moved". I think those results ended up there only because of the dependence on Data.Sigma in that file.

L-TChen:
Done. But the following properties are moved to Prelude as I don't find a better place for them

* isPropIsContr
* isContr→isProp
* isProp→isSet

How about having a Prelude.Properties for h-level properties lower than or equal to `Set`?

L-TChen:
> By the way, where shall I put the proof that the type of n-types is an (n+1) type?

The proof was in HAEquiv ... Anyway, they are already moved to HLevel, so I don't have to prove again it myself. Hooray.

L-TChen:
Proofs in HLevels, I think, can be simplified or changed so that more dependencies can be removed. For example,  `congEquiv` in HAEquiv is used only once to prove that h-level is closed under equivalence, but the proof (Thm 7.1.4) in HoTT book uses that a retract of a n-type is n-type which is missing in HLevel. 

For now, I'd like to leave it as it is. Maybe make it an issue so we can come back later.

mortberg:
> Proofs in HLevels, I think, can be simplified or changed so that more dependencies can be removed. For example, `congEquiv` in HAEquiv is used only once to prove that h-level is closed under equivalence, but the proof (Thm 7.1.4) in HoTT book uses that a retract of a n-type is n-type which is missing in HLevel.
> 
> For now, I'd like to leave it as it is. Maybe make it an issue so we can come back later.

Indeed, the best way to prove this kind of lemma is typically to use the fact that if X is a retract of Y and Y is an n-type then X is. I've been very busy lately with other things so I haven't had time to sort out the best way to do these proofs for this library, so I very much appreciate that you are helping out with untangling these dependencies.

L-TChen:
> > Proofs in HLevels, I think, can be simplified or changed so that more dependencies can be removed. For example, `congEquiv` in HAEquiv is used only once to prove that h-level is closed under equivalence, but the proof (Thm 7.1.4) in HoTT book uses that a retract of a n-type is n-type which is missing in HLevel.
> > For now, I'd like to leave it as it is. Maybe make it an issue so we can come back later.
> 
> Indeed, the best way to prove this kind of lemma is typically to use the fact that if X is a retract of Y and Y is an n-type then X is. I've been very busy lately with other things so I haven't had time to sort out the best way to do these proofs for this library, so I very much appreciate that you are helping out with untangling these dependencies.

My pleasure!

PR #103, opened 2019-03-24
add Group

riaqn:


Cubical/Data/Group/Base.agda
```diff
@@ -0,0 +1,22 @@
...
+
+open import Cubical.Core.Prelude
+
+record isGroup {ℓ} (A : Set ℓ) : Set ℓ where
+  constructor group-struct
+  field
+    e  : A
+    _⁻¹  : A → A
+    _⊙_   : A → A → A
+    lUnit : ∀ a → e ⊙ a ≡ a
```
dolio: Shouldn't there be an `rUnit` as well?

riaqn: the right ones can be derived from the left ones:
https://github.com/HoTT/HoTT-Agda/blob/1037d82edcf29b620677a311dcfd4fc2ade2faa6/core/lib/types/Group.agda#L33

But I agree this might not worth the unsymmetry.

dolio: Oh, I see. I guess I'm more used to monoids. If it's derivable I'd skip it, I guess. If Agda had a facility for letting someone pick either one, that might be nice, but it doesn't really.

riaqn: I feel something like that can be done via type classes, but I always think that's often abused in projects to the point of ambiguity and should be used with caution.

Alizter: Its similar to how you can always adjointify equivalences but sometimes you want to give it explicitly. I think we should have the group with more axioms but add helper build functions.

mortberg: I agree with @Alizter

mortberg:
I think that this should go into a new package called `Cubical.Algebra`. It might be good to base the hierarchy of algebraic structures on https://github.com/agda/agda-stdlib/tree/master/src/Algebra.

However, I don't want this library to get cluttered with random algebraic structures unless we have interesting instances. Abstract algebra is only as fun as its instances. :-)

mortberg:
> I think that this should go into a new package called `Cubical.Algebra`. It might be good to base the hierarchy of algebraic structures on https://github.com/agda/agda-stdlib/tree/master/src/Algebra.
> 
> However, I don't want this library to get cluttered with random algebraic structures unless we have interesting instances. Abstract algebra is only as fun as its instances. :-)

PS: This is not a critique of the standard library! I just meant that in general it's good not to introduce too fine-grained algebraic structures.

Cubical/Data/Group/Base.agda
```diff
@@ -0,0 +1,22 @@
...
+    assoc  : ∀ a b c → (a ⊙ b) ⊙ c ≡ a ⊙ (b ⊙ c)
+    lCancel  : ∀ a → (a ⁻¹ ) ⊙ a ≡ e
```
mortberg: Please add the right laws to the record and a smart constructor showing that you only need one of them.  Also, please put this in a new folder called `Cubical.Algebra`.

riaqn:
Yeah I will do that. But actually i opened this as placeholder for the final version so others know someone is already doing this. I'm still working on it.

mortberg:
> Yeah I will do that. But actually i opened this as placeholder for the final version so others know someone is already doing this. I'm still working on it.

I see, good! There is a lot of people contributing to the library at the same time now so this is a good idea

riaqn:
I messed up the commit history, so I had to start-over the commit.

dolio:
By the way, I don't know what all your plans are for this, but an idea I was playing with a while back was constructing the free group over a set using the path space of a HIT with one base point. So, something like:

    data Gen (A : Set l) : Set l where
      base : Gen A
      loop : A -> Path (Gen A) base base

Then if `isSet A` `FreeGroup A = Path (Gen A) base base`.

I got stuck at some point because I think I didn't understand filling stuff well enough at the time, but it seems like this should be possible.

Alizter:
@dolio Showing that this is the free group is currently an open problem. See [this article on the HoTT wiki](https://ncatlab.org/homotopytypetheory/show/loop+space+of+a+wedge+of+circles). However if A has decidable equality then it can be shown to be "a free group".

One reason it might not be true in the general case it that by Hedbergs theorem we have that A is a set since A is decidable, so we know that the wedge of circles HIT is a 1-type. I don't think this will be the case for a more general A.

It may be worth investigating what happens if we add a truncation constructor to make the HIT a 1-type we may get something useful. I do not know.

Alizter:
I still think isSet should be part of isGroup, truncation is definitely necessary. Otherwise what we have is some "wile H-group".

PR #104, opened 2019-03-24
Finite sets two ways.

dolio:
This implements two sorts of finite sets.

The first is `Fin k` in the `Data` hierarchy. This implements finite sets using a pair of a natural and a proof that it is less than `k`.

The second is `Modulo k` in the `HITs` hierarchy, which has a base constructor including arbitrary naturals, with added paths between the image of `n` and `k + n`. This differs from `Fin` in the case where `k = 0`, as `Modulo 0` is isomorphic to `ℕ` (an extra truncation constructor is added for this).

A function for reducing naturals to their residue mod `suc k` is provided, and this is used to define a function for turning `Modulo (suc k)` into `Fin (suc k)`.

Some properties about the naturals that were needed were also added, and a module for embeddings is included, because I was originally planning to use that for the reduction functions (but ended up not doing so).


dolio:
I should note, the function that extracts from `Modulo (suc k)` is pretty slow to type check, and I'm not sure why. Toward the end of this I was bumping more into areas where cubical agda seemed to blow up in that regard. The `step` case of `mreduce` seems to be the slow part, but also, at some point when I did a `split` for that function, the `squash` case seemed to get stuck permanently, although it works all right when the absurd match is added, or the line is deleted.

Anyhow, if you want maybe `Modulo` should be removed from the `Everything` module to avoid the long checking for the time being.

Cubical/Data/Fin/Base.agda
```diff
@@ -0,0 +1,57 @@
...
+
+open import Cubical.Data.Empty
+open import Cubical.Data.Nat
+open import Cubical.Data.Nat.Order
+open import Cubical.Data.Sum
+
+open import Cubical.Relation.Nullary
+
+Fin : ℕ → Set
+Fin n = Σ[ k ∈ ℕ ] k < n
```
riaqn: Is there some particular reason you are using this definition? There is the other inductive family definition that I usually find behaves better.

riaqn: Namely this definition:
```
data Fin : (n : N) -> Set where
   zero : (n : N) -> Fin (suc n)
   suc : (n : N) -> Fin n -> Fin (suc n)
```

potato4444: @riaqn that definitionis an inductive family, which Agda doesn't support fully in cubical mode yet. For a different definition of Fin that would work you could define it recursively as a coproduct of unit types. I don't know the trade offs between that definition and the definition used here though.

riaqn: @potato4444 Hmm, I didn't know that. The definition itself passes the type check. Do you have some example results of this Fin that couldn't be accomplished in the current cubical agda?

potato4444: @riaqn this is what doesn't work https://github.com/agda/cubical/pull/57#issuecomment-461174095

dolio: Yeah, the lack of support is why I avoided it. I think there may be other things that don't work, although possibly they have been fixed since I encountered them. For instance, a while ago I was playing around with the usual `Fin` definition, and cubical agda wouldn't infer the value of implicit arguments `k` from explicit `Fin k` arguments.

This is also why I chose the `k + m = n` definition of natural ordering instead of the usual indexed  inductive definition.

L-TChen: > @riaqn that definition is an inductive family, which Agda doesn't support fully in cubical mode yet. For a different definition of Fin that would work you could define it recursively as a coproduct of unit types. I don't know the trade offs between that definition and the definition used here though.

But I assume this issue should only be temporary? If I understand it correctly, it only needs the computation rule for the least fixpoint, since an inductive family is the initial algebra for a dependent polynomial functor/indexed container.

Saizan: >For instance, a while ago I was playing around with the usual Fin definition, and cubical agda wouldn't infer the value of implicit arguments k from explicit Fin k arguments.

@dolio That's weird, if it happens again please report it. It could just be that the unification constraint wasn't in the pattern fragment (maybe because it only held for some (i = i0)), but otherwise type constructors should behave the same wrt type inference.

@L-TChen The problem is constructors that specialize the index argument of their result type, they need special care for `transp`.

mortberg: As @Saizan and others said already: inductive families are currently not supported by Cubical Agda in the sense that `transp` does not reduce correctly (or at all?) for them. There is no deep theoretical reason for this; it's "only" a matter of extending the internal definition of `transp` in Agda following http://www.cs.cmu.edu/~ecavallo/works/popl19.pdf. This change will require us to add "formal" `transp` values (just like `hcomp` is a constructor/value for HITs), for details see the special case of Id types in section 3.3 of the linked paper (note that `transp` is called `coe` as they are working in Cartesian cubical type theory, however the same idea works for De Morgan cubical type theory with `transp`/`hcomp`).

Cubical/Foundations/Embedding.agda
```diff
@@ -0,0 +1,81 @@
...
+private
+  variable
+    ℓ : Level
+    A B : Set ℓ
+    f : A → B
+    w x : A
+    y z : B
+
+isEmbedding : (A → B) → Set _
+isEmbedding f = ∀ w x → isEquiv {A = w ≡ x} (cong f)
```
mortberg: Can you please add the easy lemma that `isEmbedding` is a proposition?

Cubical/Foundations/Embedding.agda
```diff
@@ -0,0 +1,81 @@
...
+
+injEmbedding
```
mortberg: Hmm, maybe I'm wrong. But doesn't the hypotheses imply that A is a retract of B ? If that's the case then you don't need `isSet A` as any retract of a set is a set.

Cubical/Foundations/Embedding.agda
```diff
@@ -0,0 +1,81 @@
+{-# OPTIONS --cubical --safe #-}
```
mortberg: What is the relationship between the content of this file and https://github.com/agda/cubical/blob/master/Cubical/Foundations/PathSplitEquiv.agda ?

Cubical/HITs/Modulo/Properties.agda
```diff
@@ -0,0 +1,71 @@
...
+fembed : Fin k → Modulo k
+fembed (n , n<k) = embed n
+
+MReduction : (k : ℕ) → Modulo (suc k) → Set
+MReduction k (embed n) = Reduction (suc k) n
+MReduction k (step n i) = Reduction≡ (suc k) n (suc k + n) refl i
+
+mreduce : (m : Modulo (suc k)) → MReduction k m
+mreduce (embed n) = reduce _ n
+mreduce {k} (step n i) = reduce≡ k n (suc k + n) refl i
```
mortberg: Can you prove that `Fin k` is equivalent to `Modulo k` ?

Cubical/Foundations/Embedding.agda
dolio: Ah, okay. I'll try and generalize. I think the HoTT book said it was true "for sets" which was kind of ambiguous.

Cubical/Foundations/Embedding.agda
dolio: I'm not really sure. It looks like a `PathSplitEquiv` is similar to an equivalence. But embeddings are a notion similar to injective maps that are properly behaved for higher hlevel types.

Cubical/HITs/Modulo/Properties.agda
dolio: I guess I didn't get as far as I wanted with just this, because the reduction function started consuming all my effort (and then I forgot about the other stuff). However, there are some interesting differences between `Fin` and `Modulo`.

- `Fin 0` is empty and `Modulo 0` is the natural numbers
- It's easy to have a successor function `Fin k -> Fin (suc k)`, and harder to have one typed `Fin k -> Fin k`. However, the case is the opposite for `Modulo`.
- I think this probably generalizes somewhat, and it's easy to write arithmetic internal to `Modulo k`, and the difficulty related to writing it in `Fin k` is the reduction function (although I suppose I've factored that out in this).

I can add that `Fin (suc k)` is equivalent to `Modulo (suc k)`.

Cubical/Foundations/Embedding.agda
mortberg: Ok, it seemed to me that the results you proved were similar... But I was also very tired when I looked at it so I might have been mistaken

Cubical/Foundations/Embedding.agda
mortberg: I see, I might very well be wrong as I didn't try to prove it myself! Don't spend too much time on it if you get stuck.

Cubical/HITs/Modulo/Properties.agda
mortberg: This is interesting! Maybe you could include some of this comment in the top of the Modulo file to explain why it's a nice alternative to Fin? I'm very interested in programming examples that get simpler with univalence and HITs.

mortberg: And yes, please prove that the two types are equivalent. That would be nice to have as well as we can transport results between the two representations. I wonder if there is anything that is really painful with `Fin` that gets a lot simpler with `Modulo`....

Alizter:
Travis isn't happy, subTypeEquality isn't in scope?

Cubical/Data/Nat/Order.agda
```diff
@@ -127,3 +168,66 @@ private
...
+      { (inl x) → cong inl (m≤n-isProp x n<b)
+      ; (inr (m , p)) → ⊥-elim (<-asym n<b (m , sym (p ∙ +-comm b m)))
+      }
+
+    dichotomy+≡ : ∀ b m n → (p : n ≡ b + m) → dichotomy b n ≡ inr (m , p)
+    dichotomy+≡ b m n p
+      = case dichotomy b n return (λ d → d ≡ inr (m , p)) of λ
+      { (inl n<b) → ⊥-elim (<-asym n<b (m , +-comm m b ∙ sym p))
+      ; (inr (m' , q))
+      → cong inr (subtypeEquality
```
Alizter: I think this is where travis is tripping up

Cubical/Foundations/Embedding.agda
dolio: This has been proved.

Cubical/Foundations/Embedding.agda
dolio: I tried to do this, but it seemed impossible. Whichever `isSet` hypothesis I removed seemed to be the one I ended up needing to pull back the other `isSet`. So this might not actually be possible.

Cubical/HITs/Modulo/Properties.agda
dolio: `Modulo (suc k)` and `Fin (suc k)` have been shown equivalent.

Cubical/Data/Nat/Order.agda
dolio: Has `subTypeEquality` moved or been renamed?

dolio: Oh, it has. Now it's `ΣProp≡`.

Cubical/Foundations/Embedding.agda
mortberg: Ok, good to know!

mortberg: Maybe add a comment saying that the isSet assumptions seem necessary with a pointer to where this is defined in the HoTT book?

mortberg:
It looks like this PR is soon ready to be merged? 

Can you please add some headers to the files explaining their content. I would be good with a comment about what the pros and cons are with the different definitions so that someone who needs finite sets find the best definition for them (with pointers to the other file in each of the files)

mortberg:
I'm going to merge https://github.com/agda/cubical/pull/131 now. Let me know if you need some help resolving conflicts in this PR

dolio:
I think I'll be all right. And I'll write up some actual comments this weekend probably.

Cubical/HITs/Modulo/Properties.agda
dolio: By the way, I don't know if I mentioned this elsewhere, but I changed the definition of `Modulo` along the way. Instead of trying to truncate the 0 case, I avoid adding paths to it. This made things rather simpler, and a pattern synonym made it so that the successor cases weren't really any different.

Cubical/Foundations/Embedding.agda
```diff
@@ -0,0 +1,101 @@
...
+
+-- If `f` is an embedding, then the fibers of `f` are propositions,
+-- which is what we would expect from an injective function on sets.
+isEmbedding→isPropFiber : isEmbedding f → ∀ y → isProp (fiber f y)
+isEmbedding→isPropFiber iE y (x , p)
+  = subst (λ f → isProp f) (lemma₀ p) (isContr→isProp (lemma₁ iE x)) (x , p)
+
+private
+  lemma₂
+    : (p : f w ≡ f x)
```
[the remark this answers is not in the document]
mortberg: It seems to me like the CI is working? I restarted it just to be sure.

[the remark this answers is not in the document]
Alizter: It's all good.

Cubical/Data/Fin/Base.agda
Alizter: @mortberg Shall we open the support of inductive families as an issue?

Alizter:
Let's merge this before #132 and then make necessary adjustments in that pr.

Cubical/Data/Fin/Base.agda
Saizan: Oh, don't worry, I'm aware we need them

mortberg:
I'm merging this now so that I can merge the other PR. If you want to add more comments or add content please do it in a new PR. Thanks a lot for contributing!

PR #105, opened 2019-03-25
brunerie

ecavallo:
SLOW

### DIFF
--- a/Cubical/Experiments/Brunerie.agda
+++ b/Cubical/Experiments/Brunerie.agda
@@ -1,0 +1,305 @@
+{-# OPTIONS --cubical #-}
+module Cubical.Experiments.Brunerie where
+
+open import Cubical.Core.Everything
+open import Cubical.Foundations.Equiv
+open import Cubical.Data.Bool
+open import Cubical.Data.Nat
+open import Cubical.Data.Int
+open import Cubical.HITs.S1
+open import Cubical.HITs.S2
+open import Cubical.HITs.S3
+open import Cubical.HITs.Join
+open import Cubical.HITs.Hopf
+open import Cubical.HITs.SetTruncation
+
+ptType : Set₁
+ptType = Σ[ A ∈ Set ] A
+
+pt : ∀ (A : ptType) → A .fst
+pt A = A .snd
+
+ptBool ptS¹ ptS² ptS³ : ptType
+ptBool = (Bool , true)
+ptS¹ = (S¹ , base)
+ptS² = (S² , base)
+ptS³ = (S³ , base)
+
+Ω Ω² Ω³ : ptType → ptType
+Ω (A , a) = (Path A a a) , refl
+Ω² A = Ω (Ω A)
+Ω³ A = Ω (Ω² A)
+
+mapΩrefl : {A : ptType} {B : Set} (f : A .fst → B) → Ω A .fst → Ω (B , f (pt A)) .fst
+mapΩrefl f p i = f (p i)
+
+mapΩ²refl : {A : ptType} {B : Set} (f : A .fst → B) → Ω² A .fst → Ω² (B , f (pt A)) .fst
+mapΩ²refl f p i j = f (p i j)
+
+mapΩ³refl : {A : ptType} {B : Set} (f : A .fst → B) → Ω³ A .fst → Ω³ (B , f (pt A)) .fst
+mapΩ³refl f p i j k = f (p i j k)
+
+ptjoin : ptType → Set → ptType
+ptjoin (A , a) B = join A B , inl a
+
+meridS² : S¹ → Path S² base base
+meridS² base _ = base
+meridS² (loop i) j = surf i j
+
+alpha : join S¹ S¹ → S²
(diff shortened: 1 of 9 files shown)


Alizter:
This is awesome. Did you manage to run it? How long did it take?

mortberg:
> This is awesome. Did you manage to run it? How long did it take?

We tried a very simple special case of the computation and it was extremely slow compared to cubicaltt. The example we tried was: https://github.com/mortberg/cubicaltt/blob/pi4s3_nobug/examples/brunerie3.ctt#L1364

I don't have any hope for Cubical Agda to be able compute this any time soon without some major optimizations. This is quite disappointing as I was hoping that the ghcomp optimization would make a big difference.

mortberg:
@ecavallo Can you please add GroupoidTruncation and 2GroupoidTruncation to the HITs.Everything file?

Can you also please add test630, test631... ?

Cubical/Experiments/Brunerie.agda
```diff
@@ -0,0 +1,315 @@
...
+g10 : ∥ Int ∥₀ → Int
+g10 = elimSetTrunc (λ _ → isSetInt) (idfun Int)
+
+-- don't run me
+brunerie : Int
+brunerie = g10 (g9 (g8 (f7 (f6 (f5 (f4 (f3 (λ i j k → surf i j k))))))))
+
+-- simpler tests
+
+test63 : ℕ → Int
```
mortberg: @Saizan  This function is very fast in cubicaltt, but in Cubical Agda `test63 0` terminates fast but then it gets extremely slow directly. Any ideas?

I think that being able to disable "endpoint corrections" in trans for some HITs would help a lot. These corrections are in particular not necessary for truncations, but I believe that Cubical Agda still does them?

mortberg:
I'll merge this as soon as Travis gives me the green light!

Cubical/Experiments/Brunerie.agda
Saizan: Yeah, Cubical Agda does the corrections for any parametrized type. I suppose we could either make Agda detect when they can be skipped, or figure out a way to provide custom implementations of transp.

Saizan: By the way, what do you mean by

> ```test63 0``` terminates fast but then it gets extremely slow directly

?

What gets slow?

3abc: > By the way, what do you mean by
> 
> > `test63 0` terminates fast but then it gets extremely slow directly
> 
> ?
> 
> What gets slow?

He's talking about `test63 1` `test63 2` `test63 3`... etc. those computes fast in cubicaltt, but not in Cubical Agda.

mortberg: In cubicaltt the timings are as follows:
```
> test631
EVAL: pos (suc zero)
#hcomps: 0
Time: 0m0.005s
> test632
EVAL: pos (suc zero)
#hcomps: 0
Time: 0m0.037s
> test633
EVAL: pos (suc zero)
#hcomps: 0
Time: 0m0.095s
> test634
EVAL: pos (suc zero)
#hcomps: 0
Time: 0m0.209s
> test635
EVAL: pos (suc zero)
#hcomps: 0
Time: 0m0.504s
> test636
EVAL: pos (suc zero)
#hcomps: 0
Time: 0m1.125s
> test637
EVAL: pos (suc zero)
#hcomps: 0
Time: 0m2.103s
> test638
EVAL: pos (suc zero)
#hcomps: 0
Time: 0m6.096s
> test639
EVAL: pos (suc zero)
#hcomps: 0
Time: 0m9.633s
```

How can I get similar timings when running `C-c C-n test63 N` in Agda again?

mortberg: I don't know what @ecavallo was doing when we were testing this before on his laptop, but it seems to be instant to run for example `C-c C-n test63 100` on my machine?!

mortberg: Oh wait, `test63` is not testing the right thing! The base case should be `surf i j k`, not base!

mortberg: I pushed a fix now. Running `C-c C-n test63 1` does not terminate on my computer before using up all of my memory. In cubicaltt it's instant.

Saizan: @mortberg do you have some general schema on how to skip the corrections? e.g. for truncations it's not enough to remove the other ```hcomp``` you also need to skip those in the path arguments

mortberg: I'm not sure if you're asking for a syntactic criteria for when the endpoint corrections could be skipped or just how to skip them? 

For the first I don't know, maybe the best is to add a pragma to let experts flag the HIT definition themselves for now? This would let us test if this is the problem here. Alternatively we could have a way to disable them globally just so that we can test this example.

For how to implement not doing any corrections see: https://github.com/mortberg/cubicaltt/blob/pi4s3_nobug/Eval.hs#L664

PR #106, opened 2019-03-26
Function properties

m-yac:
It felt strange that I had to define `id` and `const` myself when using this library. I added these functions and some basic properties about composition to `Cubical.Foundations.Function`.

### DIFF
--- a/Cubical/Foundations/Function.agda
+++ b/Cubical/Foundations/Function.agda
@@ -12,8 +12,28 @@ _∘_ : ∀ {ℓ ℓ′ ℓ″} {A : Set ℓ} {B : A → Set ℓ′} {C : (a : A
         (g : {a : A} → (b : B a) → C a b) → (f : (a : A) → B a) → (a : A) → C a (f a)
 g ∘ f = λ x → g (f x)
 
-case_of_ : ∀{ℓ ℓ'} {A : Set ℓ} {B : A → Set ℓ'} → (x : A) → (∀ x → B x) → B x
+∘-assoc : ∀ {ℓ ℓ′ ℓ″ ℓ‴} {A : Set ℓ} {B : A → Set ℓ′} {C : (a : A) → B a → Set ℓ″} {D : (a : A) (b : B a) → C a b → Set ℓ‴}
+            (h : {a : A} {b : B a} → (c : C a b) → D a b c) (g : {a : A} → (b : B a) → C a b) (f : (a : A) → B a)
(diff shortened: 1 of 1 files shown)


3abc:
> It felt strange that I had to define `id` and `const` myself when using this library. I added these functions and some basic properties about composition to `Cubical.Foundations.Function`.

There is `idfun` defined in `Glue.agda`. Not sure about the constant function.

```
Cubical/Core/Glue.agda
75:idfun : ∀ {ℓ} → (A : Set ℓ) → A → A
76:idfun _ x = x
```

m-yac:
> There is `idfun` defined in `Glue.agda`.

Ah! Though looks like `idFun` has an explicit type argument, is it worth having a duplicate with this argument implicit?

3abc:
> Ah! Though looks like `idFun` has an explicit type argument, is it worth having a duplicate with this argument implicit?

I don't know, ask @mortberg .

mortberg:
> > Ah! Though looks like `idFun` has an explicit type argument, is it worth having a duplicate with this argument implicit?
> 
> I don't know, ask @mortberg .

It doesn't sound reasonable to have a duplicate of this. Can you make the type implicit in `idfun` without things breaking (after adapting the places where it is used)?

3abc:
> > > Ah! Though looks like `idFun` has an explicit type argument, is it worth having a duplicate with this argument implicit?
> > 
> > 
> > I don't know, ask @mortberg .
> 
> It doesn't sound reasonable to have a duplicate of this. Can you make the type implicit in `idfun` without things breaking (after adapting the places where it is used)?

I just took a look, a lot of time if we make thing explicit we will need to write things like `idfun {A = B}`. Not sure if it is worth it.

mortberg:
> > > > Ah! Though looks like `idFun` has an explicit type argument, is it worth having a duplicate with this argument implicit?
> > > 
> > > 
> > > I don't know, ask @mortberg .
> > 
> > 
> > It doesn't sound reasonable to have a duplicate of this. Can you make the type implicit in `idfun` without things breaking (after adapting the places where it is used)?
> 
> I just took a look, a lot of time if we make thing explicit we will need to write things like `idfun {A = B}`. Not sure if it is worth it.

Right, that's what I seem to remember. So let's keep only idfun the way it is.

m-yac:
> So let's keep only idfun the way it is.

2d91e62 Is this what you had in mind?

Looking at `Cubical.Core.Glue` I agree that the explicit argument is needed.

mortberg:
Yeah, this looks good now! Thanks for contributing

PR #107, opened 2019-03-26
Reasoning with paths

WorldSEnder:
Introduces a new syntax
```
syntax ≡⟨⟩-syntax x (λ i → B) y = x ≡[ i ]⟨ B ⟩ y
```
that let's one explicitly reason along one dimension of a path. This emulates the term one is reasoning about, without having to use `cong`, `sym` etc. For example:
```diff
-  m +negsuc n                    ≡⟨ cong (λ z → z +negsuc n) (sym (predSuc m)) ⟩
+  m +negsuc n                    ≡[ i ]⟨ predSuc m (~ i) +negsuc n ⟩
  (predInt (sucInt m)) +negsuc n
```

### Downsides:

Currently, agda does not unify terms at the end of paths, which - in some places - forces one to insert arguments, that could be deduced, explicitly. This can lead to more verbose (but imo still easier to follow) code. See for example

https://github.com/agda/cubical/commit/0021690ed0a5b9d48024e1efb8716dc8a662811e#diff-c8d357f7cb4918b983a46285d4d349dcL266
```diff
-   sucInt (n +negsuc (suc m)) +pos (suc m)   ≡⟨ cong (λ z → z +pos (suc m)) (sucPred _) ⟩ 
+   sucInt (n +negsuc (suc m)) +pos (suc m)   ≡[ i ]⟨ sucPred (n +negsuc m) i +pos (suc m) ⟩ 
```

### DIFF
--- a/Cubical/Core/Prelude.agda
+++ b/Cubical/Core/Prelude.agda
@@ -88,6 +88,12 @@ compPathP p q j = compPathP-filler p q j i1
 _≡⟨_⟩_ : (x : A) → x ≡ y → y ≡ z → x ≡ z
 _ ≡⟨ x≡y ⟩ y≡z = x≡y ∙ y≡z
 
+
+≡⟨⟩-syntax : (x : A) → x ≡ y → y ≡ z → x ≡ z
+≡⟨⟩-syntax = _≡⟨_⟩_
+infixr 2 ≡⟨⟩-syntax
+syntax ≡⟨⟩-syntax x (λ i → B) y = x ≡[ i ]⟨ B ⟩ y
+
 _∎ : (x : A) → x ≡ x
 _ ∎ = refl
 
--- a/Cubical/Data/Bool/Properties.agda
+++ b/Cubical/Data/Bool/Properties.agda
@@ -72,12 +72,12 @@ or-comm true y =
 or-assoc     : ∀ x y z → x or (y or z) ≡ (x or y) or z
 or-assoc false y z =
   false or (y or z)   ≡⟨ or-identityˡ _ ⟩
-  y or z              ≡⟨ cong (_or z) (sym (or-identityˡ _)) ⟩
+  y or z              ≡[ i ]⟨ or-identityˡ y (~ i) or z ⟩
   ((false or y) or z) ∎ 
 or-assoc true y z  =
  true or (y or z)  ≡⟨ zeroˡ _ ⟩
   true             ≡⟨ sym (zeroˡ _) ⟩
-  true or z        ≡⟨ cong (_or z) (sym (zeroˡ _)) ⟩
+  true or z        ≡[ i ]⟨ zeroˡ y (~ i) or z ⟩
   (true or y) or z ∎
   
 or-idem      : ∀ x → x or x ≡ x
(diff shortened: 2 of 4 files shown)


mortberg:
Very cool! Is it possible to handle more than one abstracted dimension?

WorldSEnder:
I don't think more than one dimension is currently possible with the restrictions on syntax declarations. Maybe if someone is more craftier than me? Do you have a specific example in mind?

Something I find really intuitive is to reason about multiple parts of one equality (although care is always appropriate not to overdo it).

```
silly-example : ∀ a b c d → a + c ≡ 0 → b + d ≡ 0 → (a + c) + (b + d) ≡ 0
silly-example a b c d ac≡0 bd≡0 =
  (a + c) + (b + d) ≡[ i ]⟨ ac≡0 i + bd≡0 i ⟩
     0    +    0    ≡⟨ refl ⟩
          0         ∎
```

mortberg:
> I don't think more than one dimension is currently possible with the restrictions on syntax declarations. Maybe if someone is more craftier than me? Do you have a specific example in mind?
> 
> Something I find really intuitive is to reason about multiple parts of one equality (although care is always appropriate not to overdo it).
> 
> ```
> silly-example : ∀ a b c d → a + c ≡ 0 → b + d ≡ 0 → (a + c) + (b + d) ≡ 0
> silly-example a b c d ac≡0 bd≡0 =
>   (a + c) + (b + d) ≡[ i ]⟨ ac≡0 i + bd≡0 i ⟩
>      0    +    0    ≡⟨ refl ⟩
>           0         ∎
> ```

We often do proofs with more than one dimension variable in context, however I'm not sure if this ever happens in equational reasoning. 

That example is very cute!

3abc:
> ```diff
> -  m +negsuc n                    ≡⟨ cong (λ z → z +negsuc n) (sym (predSuc m)) ⟩
> +  m +negsuc n                    ≡[ i ]⟨ predSuc m (~ i) +negsuc n ⟩
>   (predInt (sucInt m)) +negsuc n
> ```

So this is the same as

> ```Agda
>  m +negsuc n                    ≡⟨ λ i → predSuc m (~ i) +negsuc n ⟩
>  (predInt (sucInt m)) +negsuc n
> ```

only that one doesn't have to write `λ` and `→`? Doesn't seem like it saved much here. But I think it could be very useful if it can handle things like `[ i j k ]` instead of just `[ i ]`.

WorldSEnder:
> So this is the same as
> 
> > ```agda
> >  m +negsuc n                    ≡⟨ λ i → predSuc m (~ i) +negsuc n ⟩
> >  (predInt (sucInt m)) +negsuc n
> > ```

@3abc When I try that, agda complains that it can not parse the application here, I needed an extra pair of brackets like so:

```
 m +negsuc n                    ≡⟨( λ i → predSuc m (~ i) +negsuc n )⟩
 (predInt (sucInt m)) +negsuc n
```
I would guess because the closing bracket `⟩` is parsed as part of the lambda. That was enough of an idiom for me to warrant a bit of extra "fairy dust".

3abc:
> @3abc When I try that, agda complains that it can not parse the application here, I needed an extra pair of brackets like so:

Oh OK... this syntax is cool then :D

mortberg:
Yeah, I agree! This is really neat. Merging

PR #108, opened 2019-03-26
Some logic properties

L-TChen:
1. Add more logic operations and their properties in Logic: 

* pseudo-complement
* conjunction
* implication
* bi-implication
* bi-implication gives a path
* (Ω, ⊓, ⊤) is a bounded ⊓-semilattice
* left and right ⊔-identities

2. Replace proofs with `Ω≡ (pequivToPath …)` by `⇔toPath`
3. Tidy up Cubical.Data.Prod.Base

* Move projections to Base from Properties
* Add elimination rules and the bifunctorial map
* Rename `swapInv` to swap-invol`

4. Add the elimination for sum type `elim-⊎`
5. Add simple facts `isPropUnit`, `isContrUnit`

This PR includes a patch from #106, so it should be merged after #106 (I guess?). Making it universe polymorphic is currently not needed to me, so I just leave it for now.

### DIFF
--- a/Cubical/Core/Prelude.agda
+++ b/Cubical/Core/Prelude.agda
@@ -35,7 +35,7 @@ infixr 2 _≡⟨_⟩_
 
 private
   variable
-    ℓ ℓ' ℓ'' : Level
+    ℓ ℓ' : Level
     A : Set ℓ
     B : A → Set ℓ
     x y z : A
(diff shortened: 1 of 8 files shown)


L-TChen:
wait ..

L-TChen:
`⇔toPath` is a high-order function instead of taking a pair of proofs.

PR #109, opened 2019-03-27
Datatype of Rationals

Saizan:
I didn't define any operation, but it's a start if someone wants.

### DIFF
--- a/Cubical/HITs/Everything.agda
+++ b/Cubical/HITs/Everything.agda
@@ -10,6 +10,7 @@ open import Cubical.HITs.Pushout public
 open import Cubical.HITs.S1 public
 open import Cubical.HITs.S2 public
 open import Cubical.HITs.S3 public
+open import Cubical.HITs.Rational public
 open import Cubical.HITs.Susp public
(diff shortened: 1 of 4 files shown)


mortberg:
Nice! We could also define this as a set quotient and prove that the two representations are equivalent. I don't have any suggestions for improvements, but let's leave it open for a bit if someone else wants to comments.

@riaqn You might want to look at this.

Alizter:
The quotient version is defined as Z x N / ~. Where (a, b) ~ (c, d) iff a(d + 1) = c(b + 1).

Alizter:
I think you forgot to add Cubical.HITs.Rational you've only added Base.

Saizan:
@Alizter do you think that would work better? it seems a bit of an artificial representation trick that would need more messing around with the signs of these interegers, though we won't really know until we implement things.

Alizter:
@Saizan that is the definition of the rationals in the HoTT book, just showing they are equivalent is enough. I like your definition of rationals better however.

Saizan:
See issue #116

PR #110, opened 2019-03-27
Cubical.HITs.ListedFiniteSet

Saizan:
Finite sets represented as a list, with membership and append.


Cubical/Foundations/Logic.agda
```diff
@@ -35,6 +35,14 @@ inr x = ∣ D.inr x ∣
                                ; (D.inr x) → g x
                                })
 
+-- Mere equality
+
+infix 30 _≡m_
+
+_≡m_ : {A : Set} (x y : A) → Ω
```
mortberg: Hmm, isn't Ω just hProp? I would not have two names for it... Also I don't like to call it Ω, it's not really a subobject classifier (because of predicativity) right?

Cubical/HITs/ListedFiniteSets.agda
```diff
@@ -0,0 +1,79 @@
...
+  = comm x y (assoc-++ xs ys zs j) i
+assoc-++ (trunc xs xs' p q i k) ys zs j
+  = trunc (assoc-++ xs ys zs j) (assoc-++ xs' ys zs j)
+          (cong (\ xs -> assoc-++ xs ys zs j) p) (cong (\ xs -> assoc-++ xs ys zs j) q) i k
+
+
+
+infix 30 _≡m_  _∈_
+
+_≡m_ : ∀ (x y : A) → Ω
```
mortberg: This is already defined in Cubical/Foundations/Logic.agda (by this PR)...

Saizan: Uh, I didn't mean to commit this file, removed.

Cubical/Foundations/Logic.agda
Saizan: it's hProp at zero, yeah.

We should make most of this polymorphic, but not in this PR. When we do that we can remove Ω?

mortberg: Sounds good, I'll make an issue for it

L-TChen:
Incidentally, formalising finite sets to be used in formalising some structural type systems is exactly the reason I start working with cubical agda. 

I have the equivalence between the type of listed finite sets and the free semilattice representation Kuratowski already. 

https://github.com/L-TChen/FiniteSets/tree/master/FiniteSets

Saizan:
@L-TChen Nice, I wanted to use this for a talk but I didn't have further plans for it.

Alizter:
Did travis not get to finish?

Issue #111, opened 2019-03-27
Clean up and make Cubical.Foundations.Logic universe polymorphic


mortberg:
The file https://github.com/agda/cubical/blob/master/Cubical/Foundations/Logic.agda should be made universe polymorphic and Ω should be removed in favour of hProp

L-TChen:
I almost have finished this issue. Since I need this to have a recursion from free semilattices to types, so I proved a few more than I need. I will create a PR when it is ready. Hope you don't mind. 

 https://github.com/L-TChen/cubical/blob/logic-properties/Cubical/Foundations/Logic.agda

Saizan:
Fixed by #113

PR #113, opened 2019-03-27
Logic

L-TChen:
I improve Foundations.Logic with more operations, properties, and some notations. Hope this is useful and good enough. #111 is also resolved by this PR.

I add a mixfix notation for proving bi-implication, e.g., 

```agda
⊔-comm : (P : hProp {ℓ}) (Q : hProp {ℓ'}) → P ⊔ Q ≡ Q ⊔ P
⊔-comm a b =
  ⇒∶ (⊔-elim a b (\ _ → (b ⊔ a)) inr inl)
  ⇐∶ (⊔-elim b a (\ _ → (a ⊔ b)) inr inl)
```
Both order `⇒∶_⇐∶_`  and `⇐∶_⇒∶_` are defined, since people tend to prove easy parts first or the other round. 

In the end of Logic, I put some syntax experiments, somehow inspired by #107. (I almost forgot this feature.) To show that `P ⊔ (Q ⊓ R) ⇒ (P ⊔ Q) ⊓ (P ⊔ R)`, we can use a case analysis syntax like

```agda
⊔-dist-⊓′ : (P Q R : hProp {ℓ}) → P ⊔ (Q ⊓ R) ≡ (P ⊔ Q) ⊓ (P ⊔ R)
⊔-dist-⊓′ P Q R =
  ⇒∶ case P     ∶ D.intro-× inl inl
     case Q ⊓ R ∶ D.map-× inr inr
     --------------------------------
     ⊢ (P ⊔ Q) ⊓ (P ⊔ R) ⊔∎
...
```
but I still feel unsatisfied. If you have any suggestion, please feel free to change or remove it.

Please have a look. Thanks. 

1. Add more logic operations and their properties in Logic:

* quantifiers
* pseudo-complement
* conjunction
* implication
* bi-implication
* (Ω, ⊓, ⊤) is a bounded ⊓-semilattice
* left and right ⊔-identities
* distributivity laws

2. Replace `Ω≡ (pequivToPath …)` by `⇔toPath`
3. Tidy up Cubical.Data.Prod.Base

* Move projections to Base from Properties
* Add elimination rules and the bifunctorial map
* Rename `swapInv` to `swap-invol`

4. Add the elimination for sum type `elim-⊎`
5. Add simple facts `isPropUnit`, `isContrUnit`
6. Logic is universe polymorphic
7. Rename \Omega to hProp
8. Some syntax experiments
9. Fix whitespaces

### DIFF
--- a/Cubical/Core/Prelude.agda
+++ b/Cubical/Core/Prelude.agda
@@ -107,7 +107,7 @@ compPath'-filler {z = z} p q j i =
 _□_ : x ≡ y → y ≡ z → x ≡ z
 (p □ q) j = compPath'-filler p q i1 j
 
-□≡∙ : (p : x ≡ y) (q : y ≡ z) → p □ q ≡ p ∙ q 
+□≡∙ : (p : x ≡ y) (q : y ≡ z) → p □ q ≡ p ∙ q
 □≡∙ {x = x} {y = y} {z = z} p q i j = hcomp (λ k → \ { (i = i0) → compPath'-filler p q k j
                                              ; (i = i1) → compPath-filler p q k j
                                              ; (j = i0) → p ( ~ i ∧ ~ k)
@@ -226,4 +226,4 @@ isProp→isSet h a b p q j i =
   hcomp (λ k → λ { (i = i0) → h a a k
                  ; (i = i1) → h a b k
                  ; (j = i0) → h a (p i) k
-                 ; (j = i1) → h a (q i) k }) a 
+                 ; (j = i1) → h a (q i) k }) a
--- a/Cubical/Core/Prelude.agda
+++ b/Cubical/Core/Prelude.agda
@@ -35,7 +35,7 @@ infixr 2 _≡⟨_⟩_
 
 private
   variable
-    ℓ ℓ' ℓ'' : Level
+    ℓ ℓ' : Level
     A : Set ℓ
     B : A → Set ℓ
     x y z : A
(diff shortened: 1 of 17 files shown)


dolio:
Have you considered not having this import `Cubical.Data.Everything`? I ran into a module loop at one point doing the finite set stuff because of that (although I ended up not using `Logic`, so the problem went away).

L-TChen:
> Have you considered not having this import `Cubical.Data.Everything`? I ran into a module loop at one point doing the finite set stuff because of that (although I ended up not using `Logic`, so the problem went away).

It is possible. `Logic` only needs `Empty`, `Prod`, `Sum`, and `Unit`. However, my gut feeling is that these basic types and properties shall go to `Core.Somewhere` as well as properties about `isContr`, `isProp`, and `isSet`. To be precise, the results in Chapter 1-3 of the HoTT book, except natural number and `Bool`, shall be part of `Core`. That's why I did not try to minimise its dependency on `Cubical.Data.Everything`.

Saizan:
@L-TChen I also feel we should have such an expanded prelude, but it still it won't import all of ```Cubical.Data.Everything```. Maybe we can have ```Cubical.Foundations.Prelude``` ?

Would probably be best to discuss this in another PR, so it'd be great if you minimized the imports here.

After that, do you think the PR is in a good state to be merged?

L-TChen:
> @L-TChen I also feel we should have such an expanded prelude, but it still it won't import all of `Cubical.Data.Everything`. Maybe we can have `Cubical.Foundations.Prelude` ? 
> Would probably be best to discuss this in another PR, 

Yeah. Maybe an issue first.

>so it'd be great if you minimized the imports here.
Done. 

I also removed the syntax experiment. It doesn't seem convenient or instructive enough to me. 

> After that, do you think the PR is in a good state to be merged?

If you are all happy with those changes I made, then I think it is good to go.

mortberg:
Yes, let's discuss this in an issue. I want to move as much as possible from the Core to Foundations one day. I've come to the conclusion that the Core should only get Cubical Agda up and running and not have any real library code.

PR #119, opened 2019-03-29
Add a definition of finite multisets

vikraman:
Following the definition of finite sets, I defined a HIT for finite multisets and proved a universal property showing it's the free commutative monoid. Suggestions welcome!

### DIFF
--- a/Cubical/HITs/FiniteMultiset.agda
+++ b/Cubical/HITs/FiniteMultiset.agda
@@ -1,0 +1,6 @@
+{-# OPTIONS --cubical --safe #-}
+module Cubical.HITs.FiniteMultiset where
+
+
+open import Cubical.HITs.FiniteMultiset.Base public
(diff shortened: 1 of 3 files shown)


Alizter:
It would be nice to show this is equivalent to the type of functions `f : A -> N` where `f(a)` is non-zero for finitely many `a`. Something like
```
(f : A -> N) x (n : N) x (a : A) x (¬(f(a) = 0) <~> Fin n)
```

Saizan:
Looks great, thanks for the contrubution!

I just wonder if ```f#``` and ```f#-uniq``` are good names for the universal morphism and its uniqueness.

Have you checked if we already have a naming convention in other files? Otherwise we might want to think of one.

vikraman:
I renamed it to `f-extend` and `f-extend-unique`, following the convention in HoTT-Agda.

PR #121, opened 2019-04-02
Few more logic

L-TChen:
This PR is about some helper functions for conversion between hProp and Set and to fix the universe issue of the definition of bi-implication as identity. 

Also  the naming convention issue: shall we use `m` for mere proposition or `p` for proposition? The mere equality uses `m`. However it is probably more mnemonic and consistent to use `p` instead, since `p` is naturally used for some other helper functions, e.g., `∥_∥ₚ: Set ℓ → hProp` for truncating types into hProp.

### DIFF
--- a/Cubical/Foundations/Logic.agda
+++ b/Cubical/Foundations/Logic.agda
@@ -1,4 +1,5 @@
 {-# OPTIONS --cubical --safe #-}
+
 module Cubical.Foundations.Logic where
 
 import Cubical.Data.Empty as D
@@ -9,16 +10,30 @@ import Cubical.Data.Unit  as D
 open import Cubical.Core.Prelude
 open import Cubical.Core.PropositionalTruncation
 
-open import Cubical.Foundations.HLevels     using (hProp; ΣProp≡; isPropIsProp; propPi; isSetHProp) public
+open import Cubical.Foundations.HLevels  using (hProp; ΣProp≡; isPropIsProp; propPi) public
 open import Cubical.Foundations.Isomorphism
 open import Cubical.Foundations.Function
 
-infixr 7 _⊔_
-infixr 7 _⊔′_
-infixr 7 _⊓_
-infixr 7 _⊓′_
-infixr 5 _⇒_
+open import Cubical.Relation.Nullary hiding (¬_)
+
+infix 10 ¬_
+infixr 8 _⊔_
+infixr 8 _⊔′_
+infixr 8 _⊓_
+infixr 8 _⊓′_
+infixr 6 _⇒_
 infixr 4 _⇔_
+infix 30 _≡ₚ_
+infix 30 _≢ₚ_
+
+infix 2 ∃[]-syntax
+infix 2 ∃[∶]-syntax
+
+infix 2 ∀[∶]-syntax
+infix 2 ∀[]-syntax
+
+infix 2 ⇒∶_⇐∶_
+infix 2 ⇐∶_⇒∶_
 
 --------------------------------------------------------------------------------
 -- The type hProp of mere propositions
@@ -34,14 +49,11 @@ private
 [_] : hProp → Set ℓ
 [_] = fst
 
-infix 30 _≡ₘ_
-infix 30 ∥_∥ₚ
-
 ∥_∥ₚ : Set ℓ → hProp
 ∥ A ∥ₚ = ∥ A ∥ , propTruncIsProp
 
-_≡ₘ_ : (x y : A) → hProp
-x ≡ₘ y = ∥ x ≡ y ∥ₚ
+_≡ₚ_ : (x y : A) → hProp
+x ≡ₚ y = ∥ x ≡ y ∥ₚ
 
 hProp≡ : [ P ] ≡ [ Q ] → P ≡ Q
 hProp≡ p = ΣProp≡ (\ _ → isPropIsProp) p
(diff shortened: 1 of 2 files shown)


Saizan:
We do say "mere equality" (also because "propositional equality" means something else in some contexts), so it seems appropriate to use ```m``` for that.

"mere propositional truncation" doesn't sound as good, and I suppose ```∥_∥ₚ``` can be read as "propositional truncation as an HProp".

It does seem appealing to standardize on ```ₚ``` for "as a HProp" and change mere equality, which builds an ```HProp```, to ```_≡ₚ_``` to be consistent.

@mortberg do you have any experience/opinion on this?

mortberg:
I'm personally not using the "mere" terminology very often. The notion of propositions in univalent mathematics is hProp, so I don't use a special qualifier for it. So I think `p` is better

L-TChen:
If the following changes make sense to you, this PR should be ready to merge. 

* use the subscript p instead m for the h-prop version of operations/types (_≡ₚ_, substₚ, Decₚ)
* `pathTo⇒ : P ≡ Q → [ P ⇒ Q ]` and the other way around
* a hprop version of `subst`: `substₚ : {x y : A} (B : A → hProp {ℓ}) → [ x ≡ₚ y ⇒ B x ⇒ B y ]`
* bi-implication is now conjunction of implications in two directions, avoiding increasing universe
* another syntax for quantifiers without type annotation
* decidable mere proposition
* two more simple proofs: negation commutes with truncation, conjunction commutes with the universal quantifier

The type of decidable mere equality can now be given in the following way 

`[ ∀[ x ∶ A ] ∀[ y ∶ A ] Decₚ (x ≡ₚ y) ]`

which makes it clear that it is a proposition.

L-TChen:
> I thought it was fairly standard to have no subscript for (-1)-truncation?

True, but that is achieved by coercive subtyping. For example, in UniMath, hProp can always be converted to Type implicitly because of the following coercion

```coq
Coercion hProptoType : hProp >-> UU.
```

So in Coq there is no need to define propositional truncation twice (one for `Set`; one for the sigma type `Σ Set isProp`). Agda does not support coercion subtyping so this kind of translation has to be explicit.

mortberg:
> > I thought it was fairly standard to have no subscript for (-1)-truncation?
> 
> True, but that is achieved by coercive subtyping. For example, in UniMath (written in Coq), hProp can always be converted to Type implicitly because of the following coercion
> 
> ```coq
> Coercion hProptoType : hProp >-> UU.
> ```
> 
> So in UniMath there is no need to define propositional truncation twice (one for `Set`; one for the sigma type `Σ Set isProp`). Agda does not support coercion subtyping so this kind of translation has to be explicit.

@martinescardo Yet another argument for why Agda should have coercions! Another of my favorite arguments is that I want to be able to write `f a` when applying an equivalence and not `f .fst x`. In general, whenever the second argument is a proposition I was the first project to be a coercion.

Alizter:
What is the correct way to handle coercions in a type checker? Is it a preprocessing step? What references do we have for it? The only one I can think of is calculus of inductive constructions on which Coq is based. It would seem to me that coercions add potential complexity if not handled correctly.

martinescardo:
@mortberg We should try to persuade the Agda developers to implement coercions in the way mathematicians use it, to make Agda code reflect mathematical practice and intuition.

mortberg:
> @mortberg We should try to persuade the Agda developers to implement coercions in the way mathematicians use it, to make Agda code reflect mathematical practice and intuition.

Indeed, I'm quite happy with how they work in Coq. I'm not sure how they are implemented, but I'm sure there is a lot of literature on it and that some of the Agda developers know much more about it than I do. What was the thread where you proposed it and got a lot of backlash? I would like to weigh in on your side. If people don't like coercions then they can simply not use them, but if you want to formalize mathematics that is close to the way mathematicians do it then you need coercions.

Alizter:
It could be argued that coercions are widely used already in object oriented programming, under the name of inheritance. I guess this is related to subtyping and so on. However a key difference is that we don't have universes there.

martinescardo:
@mortberg https://lists.chalmers.se/pipermail/agda/2018/010605.html

L-TChen:
Personally, I think there should be a refined notion of coercive subtyping. From a mathematical viewpoint, one can formally distinguish [staff, structure, and properties](https://ncatlab.org/nlab/show/stuff,+structure,+property) and something in-between, e.g. [property-like structure](http://www.tac.mta.ca/tac/volumes/1997/n9/3-09abs.html) or structure-like stuff. In homotopy type theory, we also have different h-levels. Since what we want to forget is nothing but properties in the aforementioned examples, i.e. the proof of `isProp A`, or `isEquiv f`, it shouldn't be a big problem.

However general coercion allows you to apply arbitrary function (but uniquely declared?) which often causes ambiguity and difficulty to understand. 

Huh, it sounds something interesting for me to investigate in the future. :-)

Alizter:
I think coercion has the ability to be used and abused. Mortberg outlines some very real advantages of coercion, but the agda mailing list discussion also notes many disadvantages to their use.

I can see the fear of people writing really unmaintainable code if coercion were to be implemented. But I am also of the opinion that if used sparingly it can make code more maintainable. The problem then boils down to "what does sparingly mean".

martinescardo:
The fear of people writing bad code is already here without the need of facilities for coercion.

martinescardo:
A language is not something to prevent people from saying things in a certain way, but to allowi them to say it in a variety of ways. A good language will not make all its users Pulitzer Prize winners, but will allow some of them to be. You will not make the world better by preventing modes of expression.

martinescardo:
(Even more if these modes of expression are precisely the ones that helped make mathematics to be a success story.)

Alizter:
@martinescardo Preventing modes of expression is exactly what typing sets out to do. So I am not fully convinced lax programming specifications can be a force for good.

For what we want to do with coercion I fully agree that it is a good idea. What I disagree with is adding it to agda without further thought. It will probably have unintended consequences.

But who knows until we try right? :)

L-TChen:
One has at least be aware of the diamond problem or, more formally, the coherence property in coercive subtyping. Coq disregards every but the oldest path of [coercions](https://coq.inria.fr/refman/addendum/implicit-coercions.html), so the order of declaration in Coq does matter. There are different implementations in other languages. We might restrict to a smaller set of functions for coercions, say, only those forget properties, so that multiple paths of coercions are essentially the same.

Alizter:
Some coercions, though they ought to exist, are completely uselss. Take for example coercions from Groups to Magmas or Semigroups. There are all sorts of diamond issues here and at the end of the day declaring these coercions is kind of useless even though we implicitly do it in mathematics without much thought. So perhaps there should be a guidline of coercing when needed rather than when we can.

mortberg:
Indeed, the content of this PR looked good and this discussion had really derailed quite a bit.

PR #123, opened 2019-04-05
Cubical proof of the 3x3 lemma for pushouts

loic-p:
The proof is quite short, which is nice !

### DIFF
--- a/Cubical/HITs/Pushout/Properties.agda
+++ b/Cubical/HITs/Pushout/Properties.agda
@@ -54,156 +54,156 @@ module 3x3 (A00 A02 A04 A20 A22 A24 A40 A42 A44 : Set)
            (H33 : ∀ x → f43 (f32 x) ≡ f34 (f23 x))
   where
 
-  A●0 : Set
-  A●0 = Pushout f10 f30
+  A□0 : Set
+  A□0 = Pushout f10 f30
 
-  A●2 : Set
-  A●2 = Pushout f12 f32
+  A□2 : Set
+  A□2 = Pushout f12 f32
 
-  A●4 : Set
-  A●4 = Pushout f14 f34
+  A□4 : Set
+  A□4 = Pushout f14 f34
 
-  f●1 : A●2 → A●0
-  f●1 (inl x) = inl (f01 x)
-  f●1 (inr x) = inr (f41 x)
-  f●1 (push a j) = ((λ i → inl (H11 a i))
+  f□1 : A□2 → A□0
+  f□1 (inl x) = inl (f01 x)
+  f□1 (inr x) = inr (f41 x)
+  f□1 (push a j) = ((λ i → inl (H11 a i))
                    ∙∙ push (f21 a)
                    ∙∙ (λ i → inr (H31 a (~ i)))) j
                    
-  f●3 : A●2 → A●4
-  f●3 (inl x) = inl (f03 x)
-  f●3 (inr x) = inr (f43 x)
-  f●3 (push a j) = ((λ i → inl (H13 a i))
+  f□3 : A□2 → A□4
+  f□3 (inl x) = inl (f03 x)
+  f□3 (inr x) = inr (f43 x)
+  f□3 (push a j) = ((λ i → inl (H13 a i))
                    ∙∙ push (f23 a)
                    ∙∙ (λ i → inr (H33 a (~ i)))) j
   
-  A●○ : Set
-  A●○ = Pushout f●1 f●3
+  A□○ : Set
+  A□○ = Pushout f□1 f□3
 
-  A0● : Set
-  A0● = Pushout f01 f03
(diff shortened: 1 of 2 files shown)


loic-p:
Some style questions :
- Not sure about the 4 blobs of 150-characters lines for fillers, should I try to split them to a more respectable length ?
- I am not sure about how the module system works in Agda, should I make some parts private or is it sufficient to put them inside the module ?

Cubical/HITs/Pushout/Properties.agda
```diff
@@ -0,0 +1,209 @@
...
+A00 ←—f01—— A02 ——f03—→ A04
+ ↑           ↑           ↑
+f10   H11   f12   H13   f14
+ |           |           |
+A20 ←—f21—— A22 ——f23—→ A24
+ |           |           |
+f30   H31   f32   H33   f34
+ ↓           ↓           ↓
+A40 ←—f41—— A42 ——f43—→ A44
+
```
mortberg: Can you add the statement using the names in the diagram?

Cubical/HITs/Pushout/Properties.agda
```diff
@@ -0,0 +1,209 @@
+{-# OPTIONS --cubical --safe #-}
+
```
mortberg: You can add your name in a comment up here if you want so that the authorship is preserved

Cubical/HITs/Pushout/Properties.agda
```diff
@@ -0,0 +1,209 @@
...
+  backward-l (inl x) = inl (inl x)
+  backward-l (inr x) = inr (inl x)
+  backward-l (push a i) = push (inl a) i
+
+  backward-r : A4□ → A□○
+  backward-r (inl x) = inl (inr x)
+  backward-r (inr x) = inr (inr x)
+  backward-r (push a i) = push (inr a) i
+
+  backward-filler : A22 → I → I → I → A□○
```
mortberg: The forward and backward fillers are very similar... But it seems kind of hard to refactor them into a lemma as you would have to abstract over so many things?

Cubical/HITs/Pushout/Properties.agda
```diff
@@ -0,0 +1,209 @@
...
+                   ; (i = i1) → backward-r (doubleCompPath-filler (λ k → inl (H31 a (~ k))) (push (f32 a)) (λ k → inr (H33 a k)) j (~ t))
+                   ; (j = i0) → homotopy2-l (doubleCompPath-filler (λ k → inl (H11 a k)) (push (f21 a)) (λ k → inr (H31 a (~ k))) i t) k
+                   ; (j = i1) → homotopy2-r (doubleCompPath-filler (λ k → inl (H13 a k)) (push (f23 a)) (λ k → inr (H33 a (~ k))) i t) k
+                   ; (k = i0) → A○□→A□○ (forward-filler a i j t)
+                   ; (k = i1) → backward-filler a j i (~ t) })
+          (backward-filler a j i i1)
+
+  
+
+  3x3 : A□○ ≡ A○□
```
mortberg: Maybe call this `Pushout3x3` ?

mortberg: Could you also put this statement outside of the module?

mortberg:
One more thing, it might be nice to add some more abstractions to make the statement a bit nicer. Like spans, square of spans, transpose of square of spans, etc. This way we could state the theorem like:

https://github.com/HoTT/HoTT-Agda/blob/master/theorems/homotopy/3x3/Commutes.agda#L21

But I don't think this has to be done for this PR.

Cubical/HITs/Pushout/Properties.agda
loic-p: I am not sure about that…
Since everything comes in two versions (inl/inr, f12/f21, etc) I could indeed add a 0/1 index to every function to refactor things. However, I do not think the saved space justifies the loss in readability.

Cubical/HITs/Pushout/Properties.agda
loic-p: > Could you also put this statement outside of the module?
Do you mean something like what I added at the end of the file in commit 1109cc5 ? It does not look very pretty, but probably using abstractions like spans and transpose would make it better (as you already suggested)

Cubical/HITs/Pushout/Properties.agda
mortberg: Yeah, makes sense to me

Cubical/HITs/Pushout/Properties.agda
mortberg: Yeah, packaging things up will make it look nicer. But let's keep it as it is for now

mortberg:
Thanks a lot for contributing this! I'm very impressed that it is so slick!

PR #124, opened 2019-04-05
test speed of univalence using it to compute funext

martinescardo:


Alizter:
How does it compare to the proof of funext in cubical type theory?

martinescardo:
> How does it compare to the proof of funext in cubical type theory?

Difficult to say. The type checking of the file is instant.

mortberg:
I'll merge this, but some results should be moved to Foundations

PR #128, opened 2019-04-10
Two relatively simple consequences of isProp

WorldSEnder:
We notice that we not only have `isProp→PathP` but that this `PathP` is actually contractible.

This PR also includes `isProp→isContr≡` which is the non-indexed variant of the above.

Prove this without J or is it fine as is?

### DIFF
--- a/Cubical/Foundations/HLevels.agda
+++ b/Cubical/Foundations/HLevels.agda
@@ -231,10 +231,10 @@ isContrPartial→isContr : ∀ {ℓ} {A : Set ℓ}
                        → (∀ u → u ≡ (extend i1 λ { _ → u}))
                        → isContr A
 isContrPartial→isContr {A = A} extend law
-  = x , λ y → law x ∙ (λ i → Aux.v y i) ∙ sym (law y)
-    where x = extend i0 empty
+  = ex , λ y → law ex ∙ (λ i → Aux.v y i) ∙ sym (law y)
+    where ex = extend i0 empty
           module Aux (y : A) (i : I) where
             φ = ~ i ∨ i
             u : Partial φ A
-            u = λ { (i = i0) → x ; (i = i1) → y }
+            u = λ { (i = i0) → ex ; (i = i1) → y }
             v = extend φ u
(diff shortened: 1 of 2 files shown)


L-TChen:
You may find the same kind of proofs in Foundations.HLevels. A while ago I moved some of them to Prelude to move dependency on HLevels since those are used heavily in other modules. 

@mortberg once said he wanted to reorganise the library so that Prelude only includes essential stuff to run Cubical Agda.

WorldSEnder:
I see `isContrPath` which is not quite the same as `isProp→isContr≡` (in this commit) but actually a composition of `isContr→isProp` and `isProp→isContr≡`. It would make sense to move the definitions there. Currently `isProp→PathP` is used in Foundations.Equiv which would be my guess as to why it remains in the Prelude?

L-TChen:
> Currently `isProp→PathP` is used in Foundations.Equiv which would be my guess as to why it remains in the Prelude?

Yes. And there are more instances like this or something confusing, see #100 for details.

Cubical/Foundations/HLevels.agda
```diff
@@ -64,11 +65,26 @@ isContrSigma {A = A} {B = B} (a , p) q =
      , ( λ x i → p (x .fst) i
        , h (p (x .fst) i) (transp (λ j → B (p (x .fst) (i ∨ ~ j))) i (x .snd)) i))
 
+isProp→isPropPathP : (∀ a → isProp (B a))
```
mortberg: Wouldn't it be more natural to prove this using https://github.com/agda/cubical/blob/master/Cubical/Core/Prelude.agda#L172 ?

You get two PathP's, convert them to Path with transp using that lemma and then use isPropB ?

WorldSEnder: I'm not sure what you mean. Wouldn't one need to first show that `toPathP` is an equivalence since applying isPropB to the transported path otherwise can't be pulled back?

mortberg: Ah right, nevermind.

PR #131, opened 2019-04-10
Refactor Core

mortberg:
I took the opportunity with few open PRs to refactor the library. I mainly moved all of the mathematical content out of the Core and to Foundations so that the Core now only exports the basic cubical primitives. This way the Core is really the Core of the library and all of the mathematical content start in Foundations. 

This PR doesn't add any new mathematics, it's just a lot of moving of files and fixing imports.


L-TChen:
Wow, great. 

I see some modules import Cubical.Core.Primitives but it has been exported publicly by Prelude already. Given that Prelude is the main module one need to start using Cubical Agda, perhaps it makes sense simply call it Cubical.Prelude?

Alizter:
Would it be a good time to rename Set to Type?

mortberg:
> Would it be a good time to rename Set to Type?

Not in this PR. I don't like to mix large scale refactoring with any other changes. But once this PR is merged we should be able to do this.

mortberg:
> Wow, great.
> 
> I see some modules import Cubical.Core.Primitives but it has been exported publicly by Prelude already. Given that Prelude is the main module one need to start using Cubical Agda, perhaps it makes sense simply call it Cubical.Prelude?

Good point, I will remove the extraneous imports. I might merge the contents of the Prelude into more suitable places in other files in Foundations, so let's keep the name as it is for now.

Issue #118, opened 2019-03-29
Rename Set to Type


Alizter:
It is rather unfortunate that Agda calls its univereses `Set`. Renaming it to `Type` will be much more readable.

 1. We need a good way to rename `Set` into `Type`.
 2. We need to go about replacing all occurrences of `Set` with `Type`.

I have no idea how these will be achieved in Agda. Some people discussed possible solutions in #77 so let's collect them here and discuss.

L-TChen:
Simply adding the following code to `Core.Prelude` would do the trick as suggested in https://github.com/agda/cubical/issues/77#issuecomment-473402355

```
Type : (ℓ : Level) → Set (ℓ-suc ℓ)
Type ℓ = Set ℓ

Type0 : Type (ℓ-suc ℓ-zero)
Type0 = Type ℓ-zero
```

Since `Set` is a keyword, replacing every token `Set` with `Type` if followed by a token except `→` or `Set` with `Type0` if followed by `→` or nothing, should be okay. Alright, maybe I can do it later ...

martinescardo:
You may not like my solution, but it does look close to e.g. the HoTT book, and so I suggest in here just in case:

https://www.cs.bham.ac.uk/~mhe/HoTT-UF-in-Agda-Lecture-Notes/MLTT-Agda.html#universes

L-TChen:
> You may not like my solution, but it does look close to e.g. the HoTT book, and so I suggest it here just in case:
> 
> https://www.cs.bham.ac.uk/~mhe/HoTT-UF-in-Agda-Lecture-Notes/MLTT-Agda.html#universes

It looks nice to me. Just in case, Quail/Agda-input is customisable , so it can be typed by two keystrokes only, e.g. `\U`, fewer than `Type`.

Alizter:
@martinescardo The link is not working.

martinescardo:
Sorry, I was redesigning some things and the link changed. It should be stable now.
I edited it above, repeated here: https://www.cs.bham.ac.uk/~mhe/HoTT-UF-in-Agda-Lecture-Notes/HoTT-UF-Agda.html#universes

Apologies.

martinescardo:
(I had to change the links so that the automatically generated pdf file had the links pointing to itself rather than the html file. This involved redesigning the html link strategy.)

Alizter:
@martinescardo These notes are really well written!

martinescardo:
[...]

PR #132, opened 2019-04-11
Rename Set to Type

mortberg:
This resolves https://github.com/agda/cubical/issues/118 by adding:
```
Type : (ℓ : Level) → Set (ℓ-suc ℓ)
Type ℓ = Set ℓ

Type₀ : Type (ℓ-suc ℓ-zero)
Type₀ = Type ℓ-zero

Type₁ : Type (ℓ-suc (ℓ-suc ℓ-zero))
Type₁ = Type (ℓ-suc ℓ-zero)
```
to Cubical.Core.Primitives. It also renames `Setω`  to `Typeω`.

All uses of `Set ℓ` are then changed to `Type ℓ`. One drawback is that simply `Set` used to mean `Set 0`, but now it has to be `Type₀`. However this seems better to me as it is a bit clearer.

Question: can we somehow forbid the use of `Set` somehow?

### DIFF
--- a/Cubical/HITs/FiniteMultiset/Properties.agda
+++ b/Cubical/HITs/FiniteMultiset/Properties.agda
@@ -1,57 +1,57 @@
 {-# OPTIONS --cubical --safe #-}
 module Cubical.HITs.FiniteMultiset.Properties where
 
-open import Cubical.Core.Everything
+open import Cubical.Foundations.Prelude
 open import Cubical.Foundations.HLevels
 
 open import Cubical.HITs.FiniteMultiset.Base
 
 private
   variable
-    A : Type
+    A : Type₀
 
 infixr 30 _++_
 
-_++_ : ∀ (xs ys : FMType A) → FMType A
+_++_ : ∀ (xs ys : FMSet A) → FMSet A
 [] ++ ys = ys
 (x ∷ xs) ++ ys = x ∷ xs ++ ys
 comm x y xs i ++ ys = comm x y (xs ++ ys) i
 trunc xs zs p q i j ++ ys =
   trunc (xs ++ ys) (zs ++ ys) (cong (_++ ys) p) (cong (_++ ys) q) i j
 
-unitl-++ : ∀ (xs : FMType A) → [] ++ xs ≡ xs
+unitl-++ : ∀ (xs : FMSet A) → [] ++ xs ≡ xs
 unitl-++ xs = refl
 
-unitr-++ : ∀ (xs : FMType A) → xs ++ [] ≡ xs
-unitr-++ = FMTypeElimProp.f (trunc _ _)
+unitr-++ : ∀ (xs : FMSet A) → xs ++ [] ≡ xs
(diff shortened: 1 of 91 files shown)


L-TChen:
> Question: can we somehow forbid the use of `Set` somehow?

Perhaps one can use the Agda's parser to check if `Set` appear anywhere except `Cubical.Core.Primitives`? https://github.com/agda/agda/blob/master/src/full/Agda/Syntax/Parser.hs

Cubical/HITs/FiniteMultiset/Properties.agda
```diff
@@ -44,14 +44,14 @@ comm-++ = FMSetElimProp.f (propPi (λ _ → trunc _ _))
                  ∙ cong (_++ xs) (cons-++ x ys)
                  ∙ sym (assoc-++ ys [ x ] xs))
 
-module FMSetUniversal {ℓ} {M : Set ℓ} (MSet : isSet M)
+module FMSetUniversal {ℓ} {M : Type ℓ} (MType : isSet M)
```
potato4444: Find and replace error.

riaqn:
I this this is also good since now we can actually have `Set := HLevel 2`.

Cubical/HITs/FiniteMultiset/Properties.agda
```diff
@@ -44,14 +44,14 @@ comm-++ = FMSetElimProp.f (propPi (λ _ → trunc _ _))
...
   (e : M) (_⊗_ : M → M → M)
   (comm-⊗ : ∀ x y → x ⊗ y ≡ y ⊗ x) (assoc-⊗ : ∀ x y z → x ⊗ (y ⊗ z) ≡ (x ⊗ y) ⊗ z)
   (unit-⊗ : ∀ x → e ⊗ x ≡ x)
   (f : A → M) where
 
   f-extend : FMSet A → M
-  f-extend = FMSetRec.f MSet e (λ x m → f x ⊗ m)
+  f-extend = FMSetRec.f MType e (λ x m → f x ⊗ m)
```
potato4444: Same error a few times it this section.

Cubical/HITs/FiniteMultiset/Properties.agda
mortberg: Fixed

mortberg:
> I this this is also good since now we can actually have `Set := HLevel 2`.

Agda doesn't seem to allow this as `Set` has a special status.

Alizter:
> Agda doesn't seem to allow this as `Set` has a special status.

We should open this as an issue for agda then.

>But I think it is better to do it when improving Makefile as well as #26.

Let's add this as an issue so we don't forget.

Alizter:
So we need to now modify the new files from #104

mortberg:
> So we need to now modify the new files from #104

Yes. Can you do it? I think you can make a PR to this PR.

Alizter:
@mortberg Sure, I'll send a pr to your branch in a bit.

PR #134, opened 2019-04-24
Import basics on modalities from HoTT-Agda

felixwellen:
Define modalities and prove the following basic facts:

* Modalities preserve equivalences

* Pi types of modal dependent types are modal

* Sigmas of modal dependent types on a modal type are modal

### DIFF
--- a/Cubical/Modalities/Everything.agda
+++ b/Cubical/Modalities/Everything.agda
@@ -1,0 +1,4 @@
+{-# OPTIONS --cubical --safe #-}
+module Cubical.Modalities.Everything where
+
+open import Cubical.Modalities.Modality public
--- a/Cubical/Modalities/Modality.agda
+++ b/Cubical/Modalities/Modality.agda
@@ -1,0 +1,165 @@
+{-# OPTIONS --cubical --safe #-}
+module Cubical.Modalities.Modality where
+
+{-
+  translated from
+  https://github.com/HoTT/HoTT-Agda/blob/master/core/lib/types/Modality.agda
+-}
+
+open import Cubical.Core.Everything
+open import Cubical.Foundations.Everything
+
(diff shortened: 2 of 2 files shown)


jonsterling:
Excellent! Funnily enough, I just spent the day getting owned by being confused about how modalities work :laughing:

felixwellen:
I'm happy to discuss modalities next week, when I'm back in Pittsburgh. There is also a lot to formalize, I just did some basics. I guess I should make a list and open an issue.

felixwellen:
I should have had coffee first;)

Alizter:
LGTM I think the next thing to do would be to redefine truncations in terms of modalities. That will act as a test case if you will.

felixwellen:
I see three things that could be done (concerning truncations):
1) Show that the (already defined) truncations are modalities.
2) Define truncations as nullification modalities.
3) Show that these two are the same.
The first one could be based directly on this PR. It might be a good idea to prove a suitable characterization of modalities first. Everyone is welcome to do this - I will work on other things first.
I'm still working on the localization/nullification needed for 2) (and indirectly for 3). I certainly don't insist on doing this alone...

mortberg:
> I see three things that could be done (concerning truncations):
> 
>     1. Show that the (already defined) truncations are modalities.
> 
>     2. Define truncations as nullification modalities.
> 
>     3. Show that these two are the same.
>        The first one could be based directly on this PR. It might be a good idea to prove a suitable characterization of modalities first. Everyone is welcome to do this - I will work on other things first.
>        I'm still working on the localization/nullification needed for 2) (and indirectly for 3). I certainly don't insist on doing this alone...

Great! How about making issues for each of these? That way you or someone else can assign themselves and the specific project can be discussed in the issue.

PR #135, opened 2019-04-24
Change comp to the builtin one

mortberg:
This has the benefit that we need fewer `inS`. However the builtin `comp` takes the phi explicitly.


mortberg:
I made a PR to agda making the phi implicit in the primitive comp now: https://github.com/agda/agda/pull/3713

This PR will need to be updated if that is merged.

PR #136, opened 2019-04-25
Elimination of propositional truncation to higher types

dolio:
The propositional truncation can be eliminated into non-propositional types via 'coherently constant' functions. These patches formulate the notion of coherently constant functions for low dimensions and adds the eliminators for them. This allows, for instance, definition of families of `hProp`s and `hSet`s by cases.

### DIFF
--- a/Cubical/Foundations/HLevels.agda
+++ b/Cubical/Foundations/HLevels.agda
@@ -20,7 +20,7 @@ open import Cubical.Foundations.Isomorphism
 open import Cubical.Foundations.Transport
 open import Cubical.Foundations.HAEquiv      using (congEquiv)
 open import Cubical.Foundations.Equiv        using (isoToEquiv; isPropIsEquiv; retEq; invEquiv)
-open import Cubical.Foundations.Univalence   using (univalence)
+open import Cubical.Foundations.Univalence   using (ua; univalence)
 
 open import Cubical.Data.Sigma  using (ΣPathP; sigmaPath→pathSigma; pathSigma≡sigmaPath; _Σ≡T_)
 open import Cubical.Data.Nat    using (ℕ; zero; suc; _+_; +-comm)
@@ -114,33 +114,12 @@ hLevelPi (suc (suc n)) h f g =
 isSetPi : ((x : A) → isSet (B x)) → isSet ((x : A) → B x)
 isSetPi Bset = hLevelPi 2 (λ a → Bset a)
 
-squeezeSq≡
-  : ∀{w x y z : A}
-  → (p : w ≡ y) (q : w ≡ x) (r : y ≡ z) (s : x ≡ z)
-  → (q ≡ p ∙∙ r ∙∙ sym s) ≡ (Square p q r s)
-squeezeSq≡ p q r s k
-  = Square
-      (λ j → p (j ∧ k))
-      q
-      (λ j → doubleCompPath-filler p r (sym s) j (~ k))
-      (λ j → s (j ∧ k))
-
 isSet→isSet' : isSet A → isSet' A
 isSet→isSet' {A = A} Aset {x} {y} {z} {w} p q r s
   = transport (squeezeSq≡ r p q s) (Aset _ _ p (r ∙∙ q ∙∙ sym s))
 
 isSet'→isSet : isSet' A → isSet A
 isSet'→isSet {A = A} Aset' x y p q = Aset' p q refl refl
-
-squeezeFace≡
-  : ∀{w x y z w' x' y' z' : A}
-  → {p : w ≡ y} {q : w ≡ x} {r : y ≡ z} {s : x ≡ z}
-  → {p' : w' ≡ y'} {q' : w' ≡ x'} {r' : y' ≡ z'} {s' : x' ≡ z'}
-  → {a : w ≡ w'} {b : x ≡ x'} {c : y ≡ y'} {d : z ≡ z'}
-  → (ps : Square a p p' c) (qs : Square a q q' b)
-  → (rs : Square c r r' d) (ss : Square b s s' d)
-  → Square p q r s ≡ Square p' q' r' s'
-squeezeFace≡ ps qs rs ss k = Square (ps k) (qs k) (rs k) (ss k)
 
 squeezeCu≡
   : ∀{w x y z w' x' y' z' : A}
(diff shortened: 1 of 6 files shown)


dolio:
This is a work in progress. Currently it allows elimination of propositional truncation into sets. I have the elimination into groupoids as well, but it uses a different notion of groupoid than you get from `HLevel` similar to `isSet'` (the version where all heterogeneous cubes are filled). I'm still working on the proof that `HLevel 3` implies the alternate version (although I think I'm relatively close on that as well).

One question I have is: I have some helper definitions `Square` and `Cube` for these heterogeneous shapes. Is putting them in `Prelude` appropriate? Or somewhere else?

Alizter:
Typically those shapes are put in a folder called Cubical in the other HoTT libraries, but that has some naming issues here.

In the original Licata-Brunerie paper I believe "dependent squares" were the heterogeneous shapes and squares simply referred to the homogeneous case.

Perhaps a folder called `Shapes` might be appropriate?

Saizan:
I would define both ```Square``` and ```Cube``` in the same file (and also the the heterogeneous variants).
Something like ```Cubical.Foundations.Path``` might be a good place.

mortberg:
Putting them in a separate folder seems a bit silly as the whole project is cubical and about shapes. I would put them in the Prelude.

Cubical/HITs/PropositionalTruncation/Properties.agda
```diff
@@ -39,3 +44,59 @@ elimPropTrunc' : ∀ {P : ∥ A ∥ → Type ℓ} → ((a : ∥ A ∥) → isPro
...
+  mkKmap f = f ∘ ∣_∣ , kcomp f
+
+  fib : (g : Σ (A → B) 2-Constant) → fiber mkKmap g
+  fib (g , kg) = recPropTrunc→Set g kg , refl
+
+  eqv : (g : Σ (A → B) 2-Constant)
+      → ∀ fi → fib g ≡ fi
+  eqv g (f , p) =
+    ΣProp≡ (λ f → isOfHLevelΣ 2 Fset Kset _ _)
+      (cong (uncurry recPropTrunc→Set) (sym p) ∙ setRecLemma Bset f)
```
Alizter: uncurry is causing problems for travis

dolio:
Here's another more general question. `hProp` is currently defined with an implicit level argument. However, this isn't very readily inferred in a lot of cases. Is there a reason not to make it explicit, similar to `Type`?

dolio:
So, this is mostly done. I was working on the proof that the groupoid elimination is an equivalence, but my desktop machine (which I use for Agda) turned off over night and won't turn back on. So I'm out of commission for a bit.

Alizter:
@dolio we can keep this open until you are able to finish it. Thank you for the additions!

dolio:
Finally got through the groupoid equivalence by following the ideas of the paper more closely (I found out my computer wasn't really broken, but it still took me a while to figure out the proof). This was all the work I'd planned to do, so just let me know if any modifications are desired.

Cubical/Foundations/HLevels.agda
```diff
@@ -111,16 +114,92 @@ hLevelPi (suc (suc n)) h f g =
 isSetPi : ((x : A) → isSet (B x)) → isSet ((x : A) → B x)
 isSetPi Bset = hLevelPi 2 (λ a → Bset a)
 
+squeezeSq≡
```
mortberg: I think this should go into GroupoidLaws

Cubical/Foundations/HLevels.agda
```diff
@@ -111,16 +114,92 @@ hLevelPi (suc (suc n)) h f g =
...
+      (λ j → qs (j ∧ τ))
+      (λ j → rs (j ∧ τ))
+      (λ j → ss (j ∧ τ))
+      f0
+      (toPathP {A = F} (transportTransport⁻ (squeezeFace≡ ps qs rs ss) f1) τ)
+  where
+  F : I → Set _
+  F k = Square (ps k) (qs k) (rs k) (ss k)
+
+transposeSq
```
mortberg: Do you really need this as a separate lemma? It's just swap.

Cubical/Foundations/HLevels.agda
```diff
@@ -111,16 +114,92 @@ hLevelPi (suc (suc n)) h f g =
...
+  : ∀{w x y z : A} {p : w ≡ y} {q : w ≡ x} {r : y ≡ z} {s : x ≡ z}
+  → Square p q r s
+  → Square q p s r
+transposeSq sq i j = sq j i
+
+transposeCu
```
mortberg: I don't think we need this as a separate lemma

Cubical/Foundations/HLevels.agda
mortberg: By "swap" I mean flip :-)

Cubical/Foundations/HLevels.agda
dolio: I can move it. I've been wondering though, GroupoidLaws often takes an extremely long time to type check for me. I actually don't know how long, because I've never waited for it to complete, just walked away and it was done an hour or two later. Does this not happen for other people?

It also only seems to happen when not loading the file directly, but as a dependency of another module.

mortberg: I think was related to some Agda bug. If you update your Agda installation this should go away, I've at least never experienced this for the last month or something.

Alizter: The whole library takes `	1m44.752s` to compile for me, on a Chromebook. Like mortberg says, reinstalling agda should fix this problem.

Saizan: eta equality for ```isEquiv``` were making it slow if you had the builtins for ```Glue``` loaded.

dolio: Should `squeezeCu≡` (and `squeezeFace≡`) go in GroupoidLaws as well?

Cubical/Foundations/HLevels.agda
dolio: You're right, it was easy enough to inline these. I'm not sure I could have discovered the inlining without the separate definitions, though. :)

Cubical/Foundations/HLevels.agda
mortberg: I guess so. But I'm a bit confused why you need both isGroupoid1 and 2? Shouldn't it be sufficient to just have the most general one?

dolio: I couldn't understand how to do it all in one step. I guess now that I figured it out I could collapse it into one definition.

dolio: Sorry for the delay. I got rid of the middle `isGroupoid` type. Took me a while to figure out, since inlining seems to make it more difficult for Agda to infer all the metas (and it was difficult for me to infer them, as well :)).

I didn't move `squeezeCu≡` because its use of stuff from `Transport` causes a cycle. I could make it private if you'd prefer.

mortberg:
I think this looks good now. Is it ready to be merged @dolio ?

dolio:
Yep, it's ready.

mortberg:
Great, thanks for contributing!

PR #142, opened 2019-05-06
Make everything compatible with the new arguments to comp

mortberg:
The Agda patch https://github.com/agda/agda/pull/3713 changed the implicit arguments to `comp` so that the formula is implicit. This patch makes the agda/cubical library typecheck with this change. 

**NOTE**: Once this PR has been merged everyone will have to update their Agda installation in order to typecheck the library.

### DIFF
--- a/Cubical/Codata/M.agda
+++ b/Cubical/Codata/M.agda
@@ -26,7 +26,7 @@ module Helpers where
   compPathD F {A = A} P Q {x} p q i =
      comp (\ j → F (hfill (λ j → \ { (i = i0) → A ; (i = i1) → Q j })
                           (inS (P i))
-                          j)) _
+                          j))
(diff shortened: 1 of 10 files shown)


Alizter:
What is the most painless way to update agda?

I had installed it in a sandbox, and cabal was giving me trouble when I pulled and tried to make again. My solution has been to delete and reinstall. Is there a better way?

mortberg:
> What is the most painless way to update agda?
> 
> I had installed it in a sandbox, and cabal was giving me trouble when I pulled and tried to make again. My solution has been to delete and reinstall. Is there a better way?

I have Agda installed in a sandbox. I just do "rm .cabal-sandbox/bin/agda && git pull && make" and it always works.

Alizter:
@mortberg Let's add that to the INSTALL file so people who are clueless about agda's installation, like me, know what to do :)

PR #152, opened 2019-06-05
Add Cubical.Data.Maybe

phadej:
isOfHLevelMaybe is essentially the same proof as for Sum.

---

Motivation: Maybe is useful in programming, and more ergonomic than `Unit ⊎ A`

### DIFF
--- a/Cubical/Data/Maybe/Properties.agda
+++ b/Cubical/Data/Maybe/Properties.agda
@@ -1,0 +1,111 @@
+{-# OPTIONS --cubical --safe #-}
+module Cubical.Data.Maybe.Properties where
+
+open import Cubical.Core.Everything
+open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.HLevels
+open import Cubical.Foundations.Isomorphism
+open import Cubical.Data.Empty
+open import Cubical.Data.Unit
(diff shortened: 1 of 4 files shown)


Cubical/Data/Maybe/Properties.agda
```diff
@@ -0,0 +1,105 @@
...
+fromJust-def a nothing = a
+fromJust-def _ (just a) = a
+
+just-inj : (x y : A) → just x ≡ just y → x ≡ y
+just-inj x _ eq = cong (fromJust-def x) eq 
+
+discreteMaybe : Discrete A → Discrete (Maybe A)
+discreteMaybe eqA nothing nothing    = yes refl
+discreteMaybe eqA nothing (just a')  = no (λ p → lower (subst (caseMaybe (Maybe _) (Lift ⊥)) p (just a')))
+discreteMaybe eqA (just a) nothing   = no (λ p → lower (subst (caseMaybe (Lift ⊥) (Maybe _)) p (just a)))
```
potato4444: I think this should be factored out into its own definition stating that just and nothing are never equal

phadej: You're right, that lemma is useful. Done.

PR #153, opened 2019-06-05
Add discreteList

phadej:


phadej:
In this PR and #152 I chose some somewhat arbitrary names for lemmas, especially injectivity.

Seems that proper names should be `injJust`, `inj-∷₁` and `inj-∷₂`; I'll be happy to rename these (and other) lemmas if there is some convention.

Saizan:
I would say we only need one of ```¬cons≡nil``` and its symmetric, because we have ```sym``` already.

phadej:
@Saizan I followed the 

https://github.com/agda/cubical/blob/8ad53109f579d63d5caa08f9322429cd5673d29e/Cubical/Data/Nat/Properties.agda#L36-L40

approach. (Also in `Maybe` case there are symmetric ones).

mortberg:
I think the code looks fine and I'm fine with the names. If someone has a proposal for a naming convention that work for some other Agda library feel free to propose it. I'm very bad at thinking about these things...

Issue #150, opened 2019-06-04
One more HIT Int


phadej:
I was playing with Cubical Agda and got https://gist.github.com/phadej/527ad34e3bed28e5ea3e8cf9d9a2cb79 to work.

```agda
data ℤ : Set where
  zero : ℤ
  succ : ℤ → ℤ
  pred : ℤ → ℤ

  succpred : ∀ z → succ (pred z) ≡ z
  predsucc : ∀ z → pred (succ z) ≡ z

  trunc : isSet ℤ

...

Int≡ℤ : Int ≡ ℤ
Int≡ℤ = isoToPath (iso Int→ℤ ℤ→Int ℤ→Int→ℤ Int→ℤ→Int)
```

Should it be included in this library: it demonstrates the truncation problem as mentioned in http://www.cs.nott.ac.uk/~psztxa/talks/bonn18.pdf

m-yac:
This is inspiring me to finish a port of redprl's [biinv-int](https://github.com/RedPRL/redtt/blob/master/library/cool/biinv-int.red) that I started last year (https://github.com/m-yac/cubical/tree/biinv-int). It adds the following HIT equivalent to Int:
```agda
data BiInvInt : Type₀ where
  zero : BiInvInt
  suc : BiInvInt -> BiInvInt
  -- suc is a bi-invertible equivalence:
  predr : BiInvInt -> BiInvInt
  suc-predr : ∀ z -> suc (predr z) ≡ z
  predl : BiInvInt -> BiInvInt
  predl-suc : ∀ z -> predl (suc z) ≡ z
```
A close cousin of this type is mentioned* in the linked slides as an example of the other way to solve the 'truncation problem' -- adding coherences. I think these twos HITs would complement one another well.

\* What Thorsten actually states is hard to work with, I tried and failed to prove it equivalent to Int and it seems others have too: https://github.com/agda/cubical/blob/master/Cubical/Experiments/HInt.agda

mortberg:
It would be great to have both the truncated and untruncated (based on biinv-int) versions in the library. So please make PRs.

The version from Thorsten's slides is indeed a nightmare to work with and I only pushed what worked to the HInt file. I tried many different approaches to filling those holes, but it always led straight to coherence hell.

phadej:
@mortberg Should I put my variant into `Cubical.HITs.IntTrunc` ?

mortberg:
That sounds good. Make sure you follow the same folder and file structure
as the other HITs.

PR #154, opened 2019-06-07
Port of biinv-int from redtt

m-yac:
As discussed in #150, contains a port of [the biinv-int type from redtt](https://github.com/RedPRL/redtt/blob/master/library/cool/biinv-int.red) in `HITS/BiInvInt`:

```agda
data BiInvInt : Type₀ where
  zero : BiInvInt
  suc : BiInvInt -> BiInvInt
  
  -- suc is a bi-invertible equivalence:
  predr : BiInvInt -> BiInvInt
  suc-predr : ∀ z -> suc (predr z) ≡ z
  predl : BiInvInt -> BiInvInt
  predl-suc : ∀ z -> predl (suc z) ≡ z

...

Int≡BiInvInt : Int ≡ BiInvInt
Int≡BiInvInt = isoToPath (iso fwd bwd fwd-bwd bwd-fwd)
```

I also included for contrast the following example of a 'wrong' way to define the integers as a HIT in `HITS/IsoInt`:

```agda
data IsoInt : Type₀ where
  zero : IsoInt
  suc : IsoInt -> IsoInt
  
  -- suc is an isomorphism:
  pred : IsoInt -> IsoInt
  suc-pred : ∀ z -> suc (pred z) ≡ z
  pred-suc : ∀ z -> pred (suc z) ≡ z

...

¬isSet-IsoInt : ¬ (isSet IsoInt)
¬isSet-IsoInt pf = NonTrivial.p₁≢p₂ (pf _ _ NonTrivial.p₁ NonTrivial.p₂)

¬Int≡IsoInt : ¬ (Int ≡ IsoInt)
¬Int≡IsoInt p = ¬isSet-IsoInt (subst isSet p isSetInt)
```
I tried to keep this PR's effects on everything else as minimal as possible, but I did change a few things outside of these files (see ff057da) and want to make two comments:

1. When filling cubes for the proof of  `Int≡BiInvInt` I found `Square′`, which just switches the order of the arguments to `Square` to match the convention for `hcomp`, very useful. (See `com-op` in `Foundations/HAEquiv.agda` for the simplest example of where I used it.) That being said, I don't know if we need two definitions of `Square` and would be fine removing it and refactoring.

2. I moved the general properties of equivalences which currently live at the top of `Foundations/PathSplitEquiv.agda` to `Foundations/Equiv/Properties.agda` and added a few more that I needed. Moving them wasn't technically necessary, but it seemed like an oversight that they were in that file to begin with. That being said I am happy to move them all back.

Also not sure if the little demo I left at the end of `HITS/IsoInt/Base.agad` is necessary, or if it should be moved elsewhere.

### DIFF
--- a/Cubical/Data/Int/Properties.agda
+++ b/Cubical/Data/Int/Properties.agda
@@ -43,14 +43,12 @@ open import Cubical.Relation.Nullary
 open import Cubical.Relation.Nullary.DecidableEq
 
 sucPred : ∀ i → sucInt (predInt i) ≡ i
-sucPred (pos zero)       = refl
-sucPred (pos (suc n))    = refl
-sucPred (negsuc zero)    = refl
-sucPred (negsuc (suc n)) = refl
+sucPred (pos zero)    = refl
+sucPred (pos (suc n)) = refl
+sucPred (negsuc n)    = refl
 
 predSuc : ∀ i → predInt (sucInt i) ≡ i
-predSuc (pos zero)       = refl
-predSuc (pos (suc n))    = refl
+predSuc (pos n)          = refl
 predSuc (negsuc zero)    = refl
 predSuc (negsuc (suc n)) = refl
 
--- a/Cubical/Data/Int/Properties.agda
+++ b/Cubical/Data/Int/Properties.agda
@@ -36,11 +36,20 @@ open import Cubical.Foundations.Univalence
 
 open import Cubical.Data.Empty
 open import Cubical.Data.Nat hiding (_+_ ; +-assoc ; +-comm)
+open import Cubical.Data.Bool
 open import Cubical.Data.Sum
 open import Cubical.Data.Int.Base
 
 open import Cubical.Relation.Nullary
 open import Cubical.Relation.Nullary.DecidableEq
+
+abs : Int → ℕ
+abs (pos n) = n
+abs (negsuc n) = suc n
+
+sgn : Int → Bool
+sgn (pos n) = true
+sgn (negsuc n) = false
 
 sucPred : ∀ i → sucInt (predInt i) ≡ i
 sucPred (pos zero)    = refl
--- a/Cubical/Foundations/BiInvEquiv.agda
+++ b/Cubical/Foundations/BiInvEquiv.agda
@@ -1,0 +1,73 @@
+{-
+
+Theory about Bi-Invertible Equivalences
+
+- BiInvEquiv to Iso
+- BiInvEquiv to Equiv
+- Iso to BiInvEquiv
+
+-}
+{-# OPTIONS --cubical --safe #-}
+module Cubical.Foundations.BiInvEquiv where
+
+open import Cubical.Core.Glue
+
+open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.Isomorphism
+
+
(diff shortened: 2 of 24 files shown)


mortberg:
Very nice! I'm currently traveling, but will look closely at this code as soon as I'm back.

m-yac:
Great, thanks! -- I think it's about time I learn how to spell-check in emacs.

Without using `Data.Int`, addition, negation, and subtraction as well as associativity and identity for addition are all very easy to define/prove, but proving commutativity and inverses for addition seems to be much more involved (the definition of multiplication needs associativity, identity, and inverses for addition). I'm going to try a bit more on these before adding to the PR.

The idea with `Square′` is that it when filling a cube it seems one typically writes the pattern-matching lambda in `hcomp` (or `hfill`) in the order `(i = i0)`, `(i = i1)`, `(j = i0)`, then `(j = i1)`. For example, in `HITs/BiInvInt/Base.agda`:
```agda
fwd-predSuc (pos n) i j
  = hcomp (λ k → λ { (i = i0) → rUnit (sym (pred-suc (fwd (pos n)))) k j
                   ; (i = i1) → fwd (pos n)
                   ; (j = i0) → fwd (pos n)
                   ; (j = i1) → pred-suc (fwd (pos n)) i
                   })
          (sym-filler (pred-suc (fwd (pos n))) i j)
```
I could have made the type of `fwd-predSuc` just `Int → I → I → BiInvInt`, but when both writing and using this function I found it helpful for me to specify the edge paths -- thus I defined `Square′` so I could write:
```agda
fwd-predSuc : ∀ (x : Int)
              → Square′ {- (i = i0) -} (fwd-predInt (sucInt x) ∙ (λ i → pred (fwd-sucInt x i)))
                        {- (i = i1) -} (λ _ → fwd x)
                        {- (j = i0) -} (λ i → fwd (predSuc x i))
                        {- (j = i1) -} (pred-suc (fwd x))
```
with the edge paths in the same order as before -- this way I could glance quickly between the definition, type, and usages of `fwd-predSuc` and not be confused. That's what I meant by 'more natural' before, sorry that was poor word choice.

I agree both versions of `Square` shouldn't be included, so one resolution is that I simply change the type of `fwd-predSuc` to `Int → I → I → BiInvInt`, since I realize there may be no reason the reader needs to see this type now that agda accepts it all. Another is that I reorder the cases in `hcomp` everywhere I use it with a `Square′` so that it matches `Square`.

mortberg:
Ok, let me know when you have added everything you want and I'll merge.

Regarding Square vs Square': I have no strong opinions which one is best. At this point I have a very symmetric mind, so I don't care which way the cubes are rotated. :-) If it's not too much trouble I suggest using the old Square as it is the one that is being used in the rest of the library.

m-yac:
Sorry I got distracted by travel -- I pushed what properties I was able to prove last week, and will try my hand the rest when I have more time (in a new PR).

Also I restructured things a bit, moving all the various HIT integers into their own folder.

mortberg:
> Sorry I got distracted by travel -- I pushed what properties I was able to prove last week, and will try my hand the rest when I have more time (in a new PR).
> 
> Also I restructured things a bit, moving all the various HIT integers into their own folder.

Great, so this is ready to be merged now?

m-yac:
Yep! Sorry for the flurry of commits there at the end.

mortberg:
Very good! Thanks for the contribution

PR #155, opened 2019-06-07
Truncation for any positive n

riaqn:


riaqn:
The check failed, but according to the log it seems more like a network failure?

Cubical/HITs/Truncation/Base.agda
```diff
@@ -0,0 +1,13 @@
+{-# OPTIONS --cubical --safe #-}
+module Cubical.HITs.Truncation.Base where
+
+open import Cubical.Data.Nat
+open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.HLevels
+open import Cubical.HITs.Sn
+
+-- note: only works when n is non-zero
```
mortberg: Maybe we should define the type of nonzero naturals?

Cubical/HITs/Sn/Base.agda
```diff
@@ -0,0 +1,11 @@
...
+module Cubical.HITs.Sn.Base where
+
+open import Cubical.HITs.Susp
+open import Cubical.Data.Nat
+open import Cubical.Data.Empty
+open import Cubical.Foundations.Prelude
+
+S : ℕ → Type₀
+S zero = ⊥
+S (suc n) = Susp (S n)
```
mortberg: Can you please add equivalences with the other spheres in the library? (I think everything  you need should be in https://github.com/agda/cubical/blob/master/Cubical/HITs/Susp/Base.agda)

Cubical/HITs/Truncation/Base.agda
riaqn: Or maybe in the definition, instead of `n` I say `suc n`. Therefore `∥A∥0` would actually be prop truncation.

riaqn:
Also, do you think we should make the existing truncations (prop, set) special cases of this general truncation? I imagine there are cases where definitional equality is needed.

Cubical/HITs/Truncation/Base.agda
mortberg: That sounds like a better solution than my suggestion! Please do it

mortberg:
> @mortberg Also, do you think we should make the existing truncations (prop, set) special cases of this general truncation? I imagine there are cases where definitional equality is needed.

I don't think we should replace the existing definitions (which as you say probably has some better definitional computation rules). But I think you should prove that the special cases of the general definition are equivalent to the direct definitions. First you'll need the elimination principle for the general definition.

riaqn:
> > @mortberg Also, do you think we should make the existing truncations (prop, set) special cases of this general truncation? I imagine there are cases where definitional equality is needed.
> 
> I don't think we should replace the existing definitions (which as you say probably has some better definitional computation rules). But I think you should prove that the special cases of the general definition are equivalent to the direct definitions. First you'll need the elimination principle for the general definition.

I already have the dependent elimination principle for the general truncation (will push later), which is why i think it makes sense to unitize the specific ones. Unless we can think of a scenario where we can only use the specific ones?

mortberg:
> > > @mortberg Also, do you think we should make the existing truncations (prop, set) special cases of this general truncation? I imagine there are cases where definitional equality is needed.
> > 
> > 
> > I don't think we should replace the existing definitions (which as you say probably has some better definitional computation rules). But I think you should prove that the special cases of the general definition are equivalent to the direct definitions. First you'll need the elimination principle for the general definition.
> 
> I already have the dependent elimination principle for the general truncation (will push later), which is why i think it makes sense to unitize the specific ones. Unless we can think of a scenario where we can only use the specific ones?

I think the specific ones might have better computational behavior. Plus it's easier to work with the direct definitions if one wants to write code directly by pattern-matching instead of using eliminators. So let's keep both, but prove that they are equivalent.

Cubical/HITs/Truncation/Base.agda
riaqn: actually I think it might be easier to keep it this way - otherwise I will be proving `isOfHLevel (suc n) (∥A∥n)` which is kind of awkward and hard to memorize. Much easier to remember `isOfHLevel n (∥A∥n)` for any positive `n`

riaqn: Or maybe even better, define `∥A∥0` to be unit type, and `∥A∥n+1` to be this HIT. Then from user's perspective we have a neat definition.

mortberg: Hmm, the most sensible thing would be to have `||A|| (-1)` be propositional truncation, `||A|| 0` be set truncation, etc. Then you would have `isOfHLevel (suc (suc n)) (||A|| n)` which what one would expect.

riaqn: Well, maybe we should have a `N-2` type from the beginging, then the def. of `isOfHLevel` would also be more sensible.

But anyway, if we do what I said above, we can still have `isOfHLevel n (∥A∥n)` for any natural number `n`

mortberg: The confusing thing is that Vladimir defined the H-levels to start at 0 while the truncation levels in the HoTT book start at -2. So there will always be a off-by-two mismatch between truncations and H-levels. In order to not have any more confusion I think the best is to stick to this convention and have `isOfHLevel (suc (suc n)) ||A||n`.

riaqn: OK. But if we already picked hlevels, doesn't it make sense to stick to it? Use hLevels throughout the project.

mortberg: But ||A||n is n-truncation... Which is of hlevel n+2. I think it would be more confusing to use ||A||n for n+2 truncation than defining ℕ₋₂ as in https://github.com/HoTT/HoTT-Agda/blob/e48a16bcd719e2fcb409d79e3f7df6c6b81223bb/old/Types.agda#L79

Alizter: Starting at 0 or -2 is a completely superficial choice whose only benefit is readability. Given that the HoTT book starts at -2 and very few people use h-levels, I would argue that starting at -2 is the correct choice.

ecavallo:
@riaqn Do you mind if I try to wrap this up? (And could you push the dependent elim that you mentioned above?)

riaqn:
Oh sorry for delaying - of course please go ahead. I will try to push later
today.

riaqn:
Hi @ecavallo the code for dependent elimination is here https://github.com/riaqn/cubical/blob/hgroup/Cubical/HITs/Truncation/Properties.agda

But I don't know whether it's compatible with current version of agda. I would like to see to it myself but compiling agda on chromebook is very slow. So maybe you can just copy the code you need and close this pull request?

Cubical/Data/NatMinusOne/Base.agda
```diff
@@ -0,0 +1,21 @@
...
+data ℕ₋₁ : Set where
+  neg1 : ℕ₋₁
+  suc  : (n : ℕ₋₁) → ℕ₋₁
+
+_+₋₁_ : ℕ → ℕ₋₁ → ℕ₋₁
+0 +₋₁ n = n
+suc m +₋₁ n = suc (m +₋₁ n)
+
+ℕ→ℕ₋₁ : ℕ → ℕ₋₁
+ℕ→ℕ₋₁ n = n +₋₁ suc neg1
```
m-yac: Changing line this to `ℕ→ℕ₋₁ n = suc (n +₋₁ neg1)` would make the definition of `S n` easier to use. As-is one has to pattern match on `n` in order to do anything with `S n`, but with this change we would have definitionally that `S n ≡ Susp (S₊ (n +₋₁ neg1))` and thus, for example, would have that `north : S n` for all `n`.

Cubical/Data/NatMinusOne/Base.agda
```diff
@@ -0,0 +1,21 @@
...
+
+1+_ : ℕ₋₁ → ℕ
+1+ neg1 = zero
+1+ suc n = suc (1+ n)
+
```
m-yac: For completeness, also add the following function? It's also a clearer way to write `n +₋₁ neg1`.
```agda
_-1 : ℕ → ℕ₋₁
zero -1 = neg1
(suc n) -1 = suc (n -1)
```

Cubical/HITs/Truncation/Base.agda
```diff
@@ -0,0 +1,13 @@
...
+
+open import Cubical.Data.NatMinusOne
+open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.HLevels
+open import Cubical.HITs.Sn
+
+data  ∥_∥_ {ℓ} (A : Type ℓ) (n : ℕ₋₁) : Type ℓ where
+  ∣_∣ : A  → ∥ A ∥ n
+  top : (f : S (1+ n) → ∥ A ∥ n) → ∥ A ∥ n
+  rays : (f : S (1+ n) → ∥ A ∥ n) (x : S (1+ n)) → top f ≡ f x
```
m-yac: To be consistent with the HoTT book's terminology, in #214 I named these constructors `hub` and `spoke` (as opposed to `top` and `rays`). Is that a change worth making here as well?

ecavallo: Sounds like a good idea to me.

ecavallo:
I'm ready to merge this now.

mortberg:
Thanks a lot to everyone that contributed to this PR!

PR #156, opened 2019-06-08
[ fix ] make `List` public

L-TChen:
Expose data type `List` and apply `fix-agda-whitespace` ( and some random fix to trigger CI test ...).


mortberg:
Thanks for the cleaning!

PR #158, opened 2019-06-26
[ fix #114 ] `make clean` removes all `.agda` & check whitespace issue

### LINKED ISSUE #114: Check whitespace in Makefile
There is a [tool](https://github.com/agda/agda/blob/master/src/fix-agda-whitespace/FixWhitespace.hs) in Agda to fix the whitespace issue.  The following is the explanation from the tool

> Background: Agda was reported to fail to compile on Windows because a file did not end with a newline character (Agda uses -Werror).

If this is still an issue, then this check should be included in the Makefile so that Travis can check it as well. 

Unfortunately, I will be unavailable for a few week, so I won't be able to make a PR on this.

L-TChen:
This PR is to improve `Makefile` (#114). 

* Travis will install `fix-agda-whitespace` and check if there is any violation. 

* `make` by default fix whitespace issue for you, so please remember to run `make` or `make test` before pushing any commit.    

* `make clean` remove any `.agdai` recursively starting from the root directory instead of using hardwired pattern. 

If I have some time to spare, I will also take a look of  #26 in another PR.


Saizan:
I feel a bit uneasy about ```make``` altering source files by default. Maybe ```make``` should just check and suggest ```make fix-whitespace``` when needed?

L-TChen:
> I feel a bit uneasy about `make` altering source files by default. Maybe `make` should just check and suggest `make fix-whitespace` when needed?

Done.

mortberg:
LGTM

Issue #144, opened 2019-05-11
Licence?


dylanede:
Unlike the main Agda repository, this one doesn't have a licence file in it. It would probably be a good idea to add one, perhaps similar to the one in the main Agda repository? That one appears to be based on the MIT licence.

Saizan:
Sounds good to me, by looking at a blank MIT license https://opensource.org/licenses/MIT

I guess we just need to put together a list of copyright holders.

PR #161, opened 2019-07-29
Re #144: Add a MIT License file.

Saizan:
The copyright "github.com/agda/cubical contributors" is maybe odd, but accurate.

I'm no expert of licensing though, what do you think?

Agda's own LICENSE file has a long list of names in the copyright section which ultimately links to the contributors page though.


mortberg:
I think this makes sense, but I'm also no expert

Alizter:
LGTM

PR #163, opened 2019-08-05
[ build ] make "make" only build the library by default

Saizan:
I think it's fairly conventional for ```make``` to just "build" the project, which for an agda library means typechecking.

So now ```test``` is not the default target anymore, but it's still the one travis runs.

The new ```all``` target is the default, which just runs ```check```.

Also, all of these are declared ```.PHONY```, as they are.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,4 +62,4 @@ install:
     cp -a $HOME/.cabal/lib $HOME/.cabal/share $HOME/.cabal/bin $HOME/installplan.txt $HOME/.cabsnap/;
 
 script:
-  - make
+  - make test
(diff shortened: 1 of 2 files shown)


Alizter:
Why did agda fail to install?

ice1000:
Why did agda faiI to install?

Saizan:
Good questions!

Saizan:
I will merge this even if travis is still failing on Bool.agdai

ice1000:
I don think it's the problem of Bool.agdai specifically, it's the problem of all buitin libraries. It's shown as `Bool.agdai` just because "Agda.Builtin.Bool" is alphabetically the first builtin file.

ice1000:
My guess: master Agda is now generating `_build` for `.agdai`s for libraries and (possibly) so does for the builtin modules. We, developers, succeeded in installing Agda because we're reinstalling, our `Bool.agdai`s are there because old Agda generates them and they're always there. For the cute innocent Travis CI, the problem is exposed.

Not sure though, haven't verified

PR #168, opened 2019-08-08
DeltaInt

ice1000:
Define Int as the diff of two Nats

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@ instructions see the
 [INSTALL](https://github.com/agda/cubical/blob/master/INSTALL.md)
 file.
 
+If you want to use Agda 2.6.0.1 instead of the latest development version, you
+can check out the tag `v0.1` of this library.
 
 The type theory that Cubical Agda implements is a variation of the
 cubical type theory of:


Alizter:
I don't think this is equivalent to the integers.

mortberg:
> I don't think this is equivalent to the integers. More specifically the missing coherence conditions in your attempt at an equivalence are probably precisely what you are missing in the definition.

Why? It's essentially just the integers defined as the ring of differences of nat...

It would be nice to complete the proof that they are equivalent to the non-HIT integers. Maybe it would make help to try to prove that equality is decidable so that we see that they form a set.

ice1000:
> Maybe it would make help to try to prove that equality is decidable so that we see that they form a set.

Well, I was thinking about proving `DeltaInt = Int` and use substitute `isSet Int` to `isSet DeltaInt`.

ice1000:
This is like defining the thing in #96 with a HIT.

ice1000:
The CI fails with

```
Creating package registration file: /tmp/pkgConf-Agda-2.6308846135608580149.1
src/data/lib/prim/Agda/Builtin/Bool.agdai: copyFile: does not exist (No such
file or directory)
cabal: Leaving directory '.'
Failed to install Agda-2.6.1
```

which is not my fault.

ice1000:
Well, I lost my work by jumping through branches and doing `git reset --hard`s... I'll try to take them back (or rewrite them 😭)

PR #169, opened 2019-08-09
Fix CI

ice1000:
It looks like the CI is broken somehow -- `cabal install` errors with some file does not exist error.
I'm changing it into `make install` to see if it can fix the issue.


ice1000:
This is very confusing

ice1000:
Does not fix anything, closing

PR #175, opened 2019-08-12
Homotopy colimits of graphs

m-yac:
This pull request adds a type of undirected multigraphs (in `Data.Graph`) and a HIT giving the unique homotopy colimit of a diagram over a graph (in `HITs.Colimit`). I also included some examples of graphs and proved that pushouts are indeed colimits.

Also included in this PR is yet another equivalent definition of `Fin`. When defining finite graphs I found the definition of `Fin` in `Data.Fin` very hard to work with, specifically in regards to pattern matching. The new definition (in `Data.SumFin`) behaves identically to the usual inductive version.

In future PRs I will add more examples of colimits (coequalizers, sums, and direct limits were what I had in mind), define homotopy *limits* of graphs, and prove some more categorical facts about cocones... but this pull request is already quite large as-is so I'm cutting it off!

I wrote most of this code a while ago, but I believe I drew inspiration from:
https://homotopytypetheory.org/2016/01/08/colimits-in-hott/
(Perhaps formalizing the constructions in this post is another good PR?)

### DIFF
--- a/Cubical/Data/Everything.agda
+++ b/Cubical/Data/Everything.agda
@@ -15,3 +15,4 @@ open import Cubical.Data.DiffInt public
 open import Cubical.Data.Group public hiding (_≃_)
 open import Cubical.Data.HomotopyGroup public
 open import Cubical.Data.List public
+open import Cubical.Data.Graph public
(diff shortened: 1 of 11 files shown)


Alizter:
Let's reset this now that #179 has been merged.

PR #183, opened 2019-09-06
[ delta-int ] Transport some proofs from Int, update commit message

ice1000:
As title, /cc @mortberg


ice1000:
I've just found a huge giant typo in the issue title and the commit message :sob:

Issue #166, opened 2019-08-08
3x3 via symmetric double pushout


Alizter:
I wonder if we can make the proof of the 3x3 lemma for pushouts shorter in the following way:

Given the 3x3 commutative diagram, we can encode this data as 4 types (the corners), 4 type families indexed by types on the edge (the edges) and finally a type family over the total spaces of the four type families (the middle).

It is easy to show that the 4 homotopies witnessing the commutativity of the diagram can also be recovered from this "type theoretric" encoding of a 3x3 diagram. You can get the original 3x3 by looking at the total spaces of the families and the original maps become the projection maps.

Now from this consider the HIT with 4 point constructors for each of the 4 corner types. 4 path constructors making sure that coming from the middle of an edge through a corner agrees and finally a square constructor for these 4 paths.

Now the idea is that it should be simpler to show that a double pushout is equivalent to this symmetric double pushout HIT. Which you can do both ways.

I tried to [formalise this in coq](https://github.com/Alizter/HoTT/blob/dep3x3/contrib/HITPushout.v) earlier this year, but the path algebra got a bit heavy for me. Seeing the success of the Torus, I expect this to work easily.

This was [originally Mike Shulmans idea](https://github.com/HoTT/HoTT/issues/1001#issuecomment-462858919).

Here is the following coq code you can use as pseudocode:
```coq
 Section Push.
    Context
      {A00 A10 A01 A11 : Type}
      {A0x : A00 -> A01 -> Type}
      {A1x : A10 -> A11 -> Type}
      {Ax0 : A00 -> A10 -> Type}
      {Ax1 : A01 -> A11 -> Type}
      (Axx : forall a00 a01 a10 a11,
        A0x a00 a01 ->
        A1x a10 a11 ->
        Ax0 a00 a10 ->
        Ax1 a01 a11 -> Type).

  Local Definition B00 := A00.
  Local Definition B10 := A10.
  Local Definition B01 := A01.
  Local Definition B11 := A11.

  Local Definition B0x := {a00 : A00 & {a01 : A01 & A0x a00 a01}}.
  Local Definition B1x := {a10 : A10 & {a11 : A11 & A1x a10 a11}}.
  Local Definition Bx0 := {a00 : A00 & {a10 : A10 & Ax0 a00 a10}}.
  Local Definition Bx1 := {a01 : A01 & {a11 : A11 & Ax1 a01 a11}}.

[...]

PR #184, opened 2019-09-30
3x3 reorganization, pushout span equivalences, join associativity

loic-p:
- Make better use of the module system to render the 3x3 lemma more practical.
- Define pushout spans, homotopy natural equivalences, and prove that two equivalent spans induce equivalent pushouts.
- Use 3x3-lemma and homotopy natural equivalences to prove associativity of the join.

### DIFF
--- a/Cubical/HITs/Join/Properties.agda
+++ b/Cubical/HITs/Join/Properties.agda
@@ -33,39 +33,39 @@ private
 
 
 -- Alternative definition of the join using a pushout
-joinAlt : (A : Type ℓ) → (B : Type ℓ') → Type (ℓ-max ℓ ℓ')
-joinAlt A B = Pushout {_} {_} {_} {A × B} {A} {B} proj₁ proj₂
+joinPushout : (A : Type ℓ) → (B : Type ℓ') → Type (ℓ-max ℓ ℓ')
+joinPushout A B = Pushout {_} {_} {_} {A × B} {A} {B} proj₁ proj₂
 
 -- Proof that it is equal
-joinAlt-iso-join : (A : Type ℓ) → (B : Type ℓ') → Iso (joinAlt A B) (join A B)
-joinAlt-iso-join A B = iso (joinAlt→join A B) (join→joinAlt A B) (join→joinAlt→join A B) (joinAlt→join→joinAlt A B)
+joinPushout-iso-join : (A : Type ℓ) → (B : Type ℓ') → Iso (joinPushout A B) (join A B)
+joinPushout-iso-join A B = iso joinPushout→join join→joinPushout join→joinPushout→join joinPushout→join→joinPushout
   where
-    joinAlt→join : (A : Type ℓ) → (B : Type ℓ') → joinAlt A B → join A B
-    joinAlt→join A B (inl x) = inl x
-    joinAlt→join A B (inr x) = inr x
-    joinAlt→join A B (push y i) = push (proj₁ y) (proj₂ y) i
+    joinPushout→join : joinPushout A B → join A B
+    joinPushout→join (inl x) = inl x
+    joinPushout→join (inr x) = inr x
+    joinPushout→join (push y i) = push (proj₁ y) (proj₂ y) i
 
-    join→joinAlt : (A : Type ℓ) → (B : Type ℓ') → join A B → joinAlt A B
-    join→joinAlt A B (inl x) = inl x
-    join→joinAlt A B (inr x) = inr x
-    join→joinAlt A B (push a b i) = push (a , b) i
+    join→joinPushout : join A B → joinPushout A B
+    join→joinPushout (inl x) = inl x
+    join→joinPushout (inr x) = inr x
+    join→joinPushout (push a b i) = push (a , b) i
 
-    joinAlt→join→joinAlt : (A : Type ℓ) → (B : Type ℓ') → ∀ x → join→joinAlt A B (joinAlt→join A B x) ≡ x
-    joinAlt→join→joinAlt A B (inl x) = refl
-    joinAlt→join→joinAlt A B (inr x) = refl
-    joinAlt→join→joinAlt A B (push (a , b) j) = refl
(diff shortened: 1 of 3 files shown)


mortberg:
I restarted the build. Not sure why it didn't terminate... Apparently it seemed to be stuck on compiling Agda.

Cubical/HITs/Join/Properties.agda
```diff
@@ -0,0 +1,274 @@
...
+open import Cubical.HITs.Pushout
+
+private
+  variable
+    ℓ : Level
+    ℓ' : Level
+
+
+-- Alternative definition of the join using a pushout
+joinAlt : (A : Type ℓ) → (B : Type ℓ') → Type (ℓ-max ℓ ℓ')
```
mortberg: Maybe call this `joinPushout`?

Cubical/HITs/Join/Properties.agda
```diff
@@ -0,0 +1,274 @@
...
+joinAlt A B = Pushout {_} {_} {_} {A × B} {A} {B} proj₁ proj₂
+
+-- Proof that it is equal
+joinAlt-iso-join : (A : Type ℓ) → (B : Type ℓ') → Iso (joinAlt A B) (join A B)
+joinAlt-iso-join A B = iso (joinAlt→join A B) (join→joinAlt A B) (join→joinAlt→join A B) (joinAlt→join→joinAlt A B)
+  where
+    joinAlt→join : (A : Type ℓ) → (B : Type ℓ') → joinAlt A B → join A B
```
mortberg: Can't you just remove A and B from this and the other helper functions? They are bound in the top-level function.

Cubical/HITs/Pushout/Properties.agda
```diff
@@ -16,13 +19,166 @@ open import Cubical.Core.Glue
...
+     e2 : 3-span.A2 s1 ≃ 3-span.A2 s2
+     e4 : 3-span.A4 s1 ≃ 3-span.A4 s2
+     H1 : ∀ x → 3-span.f1 s2 (e2 .fst x) ≡ e0 .fst (3-span.f1 s1 x)
+     H3 : ∀ x → 3-span.f3 s2 (e2 .fst x) ≡ e4 .fst (3-span.f3 s1 x)
+
+{-
+  Proof that the pushouts of equivalent diagrams are equal
+-}
+
+spanEquivToPath : {s1 : 3-span} → {s2 : 3-span} → (e : 3-span-equiv s1 s2)
```
mortberg: Hmm, I have a feeling that there must be a more direct proof of this using Glue types... But maybe we can investigate that later.

Cubical/HITs/Join/Properties.agda
```diff
@@ -0,0 +1,274 @@
...
+join-assoc A B C = (joinAlt≡join (join A B) C) ⁻¹
+                   ∙ (spanEquivToPath sp3≃sp4) ⁻¹
+                   ∙ (3x3-span.3x3-lemma span) ⁻¹
+                   ∙ (spanEquivToPath sp1≃sp2)
+                   ∙ (joinAlt≡join A (join B C))
+  where
+    -- the meat of the proof is handled by the 3x3 lemma applied to this diagram
+    span : 3x3-span
+    span = record {
+      A00 = A;      A02 = A × B;      A04 = B;
```
mortberg: It would be nice to not have to give all of these fields explicitly. How about making a smarter constructor for 3x3-span that has most of these things implicit?

Cubical/HITs/Pushout/Properties.agda
loic-p: Yes, retrospectively this proof is a bit silly since it does not even use univalence (except in the very end). I am pretty sure I could do it with some simple transport.

loic-p: I could indeed do it in a much shorter manner, and it barely took 1 hour. Now I feel stupid for spending so much time on this. Oh well, a reminder to take univalence seriously is always good :)

Cubical/HITs/Join/Properties.agda
loic-p: Indeed !
Btw, I am becoming less and less of a fan of the A→B, B→A, A→B→A, B→A→B naming scheme for helper functions, because it is annoyingly long and already explicit in the type. 
Since they should probably only exist during the definition of an isomorphism, maybe a simpler naming scheme like f, inv, sec, ret should be adopted ?

Cubical/HITs/Pushout/Properties.agda
mortberg: Can you push the new proof?

Cubical/HITs/Join/Properties.agda
mortberg: I can never remember which of section and retraction is which... But sure, you can use other names for these things. They are local anyway.

Alizter:
There might be a simpler proof of the 3x3 lemma, as I mention in #166.

Cubical/HITs/Pushout/Properties.agda
loic-p: Done. Despite the old one not being totally devoid of interest (it does the job in a non-univalent world), I don't think there is much of a point to keep it around, so I just removed it from the PR.

mortberg:
> There might be a simpler proof of the 3x3 lemma, as I mention in #166.

Sure, feel free to experiment with it! The cubical proof is already pretty short now though.

Cubical/HITs/Join/Properties.agda
loic-p: I can do one that only asks for the four homotopies, but that makes it not very convenient — in this instance, Agda will not typecheck the proof anymore. Maybe with some more type annotations in other places it could work.
I can also do one that asks for both functions and homotopies, but then the improvement over the vanilla constructor is not very significant…

mortberg: I'm confused... Why won't Agda check the proof?

loic-p: Well, the functions involved are not so straightforward to extract from the type of the homotopies, since their types only feature compositions of two functions.

PR #189, opened 2019-10-26
Prove false

ice1000:
Prove false.

### DIFF
--- a/Cubical/Data/Equality.agda
+++ b/Cubical/Data/Equality.agda
@@ -15,7 +15,7 @@ These definitions and lemmas are proved without Axiom K.
 
 {-# OPTIONS --cubical --safe #-}
 
-module Cubical.WithK.WithoutK where
+module Cubical.Data.Equality where
 
 open import Cubical.Foundations.Prelude
   renaming ( _≡_ to _≡c_ ; refl to reflc )
@@ -24,7 +24,8 @@ open import Agda.Builtin.Equality
   renaming ( _≡_ to _≡p_ ; refl to reflp )
   public
 
-variable
+private
+ variable
   ℓ : Level
   A : Set ℓ
   x y z : A
--- a/Cubical/Data/Everything.agda
+++ b/Cubical/Data/Everything.agda
@@ -4,6 +4,7 @@ module Cubical.Data.Everything where
 open import Cubical.Data.BinNat public
 open import Cubical.Data.Bool public
 open import Cubical.Data.Empty public
+open import Cubical.Data.Equality public
 open import Cubical.Data.Fin public
 open import Cubical.Data.Nat public
 open import Cubical.Data.Int public renaming (_+_ to _+Int_ ; +-assoc to +Int-assoc; +-comm to +Int-comm)
--- a/Cubical/WithK/Base.agda
+++ b/Cubical/WithK/Base.agda
@@ -1,0 +1,62 @@
+{- Cubical Agda with K
+
+This file defines:
+
+ptoc : x ≡p y → x ≡c y
+ctop : x ≡c y → x ≡p y
+
+and proves:
+
+p-c : (x ≡c y) ≡c (x ≡p y)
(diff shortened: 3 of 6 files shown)


ice1000:
Note: some of the codes here do not belong to me. They come from dram, where dram was inspired by someone in the IRC.

dramforever:
The proof that `--with-K` and `--cubical` leads to inconsistency came from `tomjack` on IRC: https://www.irccloud.com/pastebin/NFolVTgR/

mortberg:
I'm not sure if we should merge this... Maybe it is interesting to newcomers? But the fact that UIP/K is incompatible with univalence is very well-known

Alizter:
If it is going to be useful for newcomers, then perhaps it will be more useful if we add some more comments talking about axiom K, what that means and how the proof works. I don't see any reason not to merge this.

mortberg:
The reason not to merge it is that it is not very useful to have a proof of false from two principles that are well known to be contradictory. But I do agree that if it is properly documented then it might be instructive for beginners.

Alizter:
Perhaps add a README to WithK in order to emphasize that importing any of these files is very contradictory.

@mortberg I would agree that for people interested in HoTT, univalence and K being incompatible is well known, however if I understand correctly, the point of cubical agda is not just to be another model of HoTT but rather as a programming languages with HITs. There are many people interested in programming languages who might not really care about univalence (yet).

Say for example somebody unknowingly used K everywhere in their agda code and were interested in porting it to cubical agda for whatever reason. I think this file can be your canonical example of what goes wrong.

Just some ideas.

ice1000:
> Perhaps add a README to WithK in order to emphasize that importing any of these files is very contradictory.

I got banner comments for this fact. Is a README really needed? I'm not sure what should I put into the README though.

Alizter:
@ice1000 I would just put a warning about importing these files. The readme would be seen when browsing on github at least.

As I understand @mortberg thinks this is not really useful since it is so well known.

I am trying to say that it would be instructional, especially for people unfamiliar with univalence. The other issue is if I was new to cubical agda, withK isn't really a folder that grabs my attention.

So perhaps this should be part of an intro or a course on cubical agda rather than part of the main library?

ice1000:
> I'm not sure if we should merge this... Maybe it is interesting to newcomers? But the fact that UIP/K is incompatible with univalence is very well known

Yes, it is interesting to newcomers! At least it takes you some time to come up with the proof of `ptoc-ctop`.

ice1000:
> @ice1000 I would just put a warning about importing these files. The readme would be seen when browsing on github at least.

These files cannot be imported with the `--safe` flag, while other files (with a `--safe` flag) can.

ice1000:
I have split the without-K definitions used in `WithK` into a separate safe module.

ice1000:
Travis fails with:

```
[336 of 373] Compiling Agda.TypeChecking.Coverage ( src/full/Agda/TypeChecking/Coverage.hs, dist/build/Agda/TypeChecking/Coverage.o )
cabal: Leaving directory '.'
Failed to install Agda-2.6.1
cabal: Error: some packages failed to install:
Agda-2.6.1 failed during the building phase. The exception was:
ExitFailure (-9)
This may be due to an out-of-memory condition.
The command "cabal install" failed and exited with 1 during .
```

Saizan:
I think the content of `Cubical.WithK.WithoutK` should actually be moved to `Cubical.Data.Equality`, as it is useful to have the equivalence between paths and inductive equality.

The other modules can also provide more context in their comments, like probably citing the HoTT book, and how `--safe` prevents mixing the two flags this way.

ice1000:
> I think the content of `Cubical.WithK.WithoutK` should actually be moved to `Cubical.Data.Equality`, as it is useful to have the equivalence between paths and inductive equality.
> 
> The other modules can also provide more context in their comments, like probably citing the HoTT book, and how `--safe` prevents mixing the two flags this way.

Done

PR #191, opened 2019-11-02
Added inf-nat type

oisdk:
Added inf-nat: a type for nat+infinity that keeps decidable equality. (from [here](https://gist.github.com/ice1000/d0a1aa5a17bfb3d66bfeec3045d11b60))


ice1000:
Question: can we prove that `inf-nat = Conat`?

potato4444:
@ice1000 No, I believe that is equivalent to WLPO (Weak Limited Princple of Omniscience). Here is Martin Escardo showing WLPO is equivalent to decreasing binary sequences (which are isomorphic to `Conat`) having decidable equality: https://www.cs.bham.ac.uk/~mhe/agda-new/WLPO.html

I believe that decidable equality for `Conat` is enough to show it isomorphic to the natural numbers and hence `inf-nat`.

PR #192, opened 2019-11-05
simplify ΩTorus≡Int×Int

Rotsor:
Signed-off-by: Arseniy Alekseyev <rotsor@gmail.com>


mortberg:
Wow, I wonder what I was thinking when I did that proof... At least I realized that it could probably be simplified :-)

Issue #77, opened 2019-03-04
Cubical discussion forum


mortberg:
I'll open this "issue" as a place to discuss various cubical things, like "how to prove X cubically", as suggested by @3abc . 

Let me open up with two questions that came up in https://github.com/agda/agda/issues/3599. 

1. Give a direct cubical proof that:
```
transport p ≡ transport q → p ≡ q
```

2. Is there a more direct proof of

https://github.com/agda/agda/blob/master/src/data/lib/prim/Agda/Builtin/Cubical/Glue.agda#L104

A slick proof that `transport` is an equivalence is the following:
```
isEquivTransport : ∀ {ℓ} {A B : Set ℓ} (p : A ≡ B) → isEquiv (transport p)
isEquivTransport {A = A} {B = B} p =
  transport (λ i → isEquiv (λ x → transp (λ j → p (i ∧ j)) (~ i) x)) (idIsEquiv A)

transportEquiv : ∀ {ℓ} {A B : Set ℓ} → A ≡ B → A ≃ B
transportEquiv p = (transport p , isEquivTransport p)
```

What's going on in `pathToEquiv` is very similar, but I can't figure out how to get an analogous proof to typecheck... Any ideas? @Saizan

mortberg:
Btw, who are you @3abc ? Your github profile is very mysterious.

3abc:
> Btw, who are you @3abc ? Your github profile is very mysterious.

Sent you an email (to your CMU mailbox).

3abc:
Anyone know how to generalize this proof?

```Agda
mapHLevel : ∀ {ℓ} {A B : Set ℓ} (n : ℕ) (p : isOfHLevel n B) → isOfHLevel n (A → B)
mapHLevel 0 p = (λ _ → p .fst) , λ y i a → snd p (y a) i
mapHLevel 1 p a b i x = p (a x) (b x) i
mapHLevel 2 p a b x y i j z = p (a z) (b z) (λ i → x i z) (λ i → y i z) i j 
mapHLevel 3 p a b x y r s i j k z = p (a z) (b z) (λ i → x i z) (λ i → y i z) (λ i j → r i j z) (λ i j → s i j z) i j k 
mapHLevel (suc (suc n)) p a b = {!!}
```

The pattern is clear but I don't know how to implement the induction step.

jonsterling:
[...]

PR #193, opened 2019-11-05
prove that transport is injective

ecavallo:
As requested in https://github.com/agda/cubical/issues/77#issuecomment-546688162.

Putting this somewhere reasonable in Cubical.Foundations was problematic because of circular dependencies, so I created a new folder in Data for facts about the universe. (This one is a fact about the identity type of the universe.) Suggestions are welcome for better places to put it, but my experience suggests that it's helpful to separate lemmas that are used to prove basic facts about primitive type formers and HITs (like univalence itself) from lemmas that are mainly useful in more complex results (like this one).


mortberg:
I think this location is reasonable enough for now.

PR #194, opened 2019-11-06
define klein bottle and prove it is a groupoid

ecavallo:
just for fun


mortberg:
Nice!

mortberg:
Weill merge once Travis is happy. For some reason it keeps timing out sometimes when compiling Agda...

PR #197, opened 2019-11-16
Explain the computational behavior of Glue, glue, unglue

Rotsor:
This PR adds versions of `Glue`, `glue`, `unglue` constrained by `Sub` to demonstrate how they compute because it can be hard to figure out on your own. (you can't just look up the definitions)
This is purely for documentation, so the definitions are currently private.

The change in `Cubical/Foundations/Id.agda` resolves a warning (`equivCtrPath` name clash) I'm getting with a recent version of Agda.


mortberg:
How about adding a reference to section 6 of https://arxiv.org/abs/1611.02108 where we originally introduced these types?

Rotsor:
Sounds good. Are you thinking instead or in addition?

Rotsor:
Added a paragraph. I could add the `b = glue _ (unglue b)` and `b = unglue (glue _ b)` proofs too for completeness if people think that's a good idea.

Saizan:
I recently added some equalities in the agda documentation [1] but it's nice to have these in the lib too.

[1] https://agda.readthedocs.io/en/latest/language/cubical.html#glue-types

PR #199, opened 2019-11-21
More efficient definition of compIso

marcinjangrzybowski:
As discussed on April 22 on cubical forum


Saizan:
Any clue on why travis is not running on this?

PR #200, opened 2019-11-22
Equivalence of strictly-descending lists and ListedFiniteSet

### LINKED ISSUE #117: ListedFiniteSets: prove equivalence with sorted duplication-free lists
See https://github.com/agda/cubical/pull/110#pullrequestreview-219453495

Also @L-TChen has developed more about finite sets

https://github.com/L-TChen/FiniteSets/blob/master/FiniteSets/List/Properties.agda#L47

Rotsor:
This PR contributes a proof of equivalence between strictly-descending lists and the ListedFiniteSet HIT.

Strictly-descending lists are defined by instantiating the `Cubical.Data.DescendingList` module with a strict relation `_>_`.

Assumptions include:

```
   (A-discrete : Discrete A)
   (>-isProp : ∀ {x y} → isProp (x > y))
   (tri : ∀ x y → Tri (y > x) (x ≡ y) (x > y))
   (>-trans : ∀ {x y z} → x > y → y > z → x > z)
   (>-irreflexive : ∀ {x} → Type¬ x > x)
```
(where `Tri` is a trichotomy type defined in this PR)

This is intended to close #117. (I picked this one because it was marked "good first issue")

### DIFF
--- a/Cubical/Data/DescendingList/Strict-Properties.agda
+++ b/Cubical/Data/DescendingList/Strict-Properties.agda
@@ -1,0 +1,221 @@
+{-# OPTIONS --cubical --safe #-}
+
+open import Cubical.Core.Everything
+open import Cubical.Foundations.Everything
+open import Agda.Primitive using (lzero)
+
+module Cubical.Data.DescendingList.Strict-Properties
+ (A : Type₀)
+ (_>_ : A → A → Type₀)
+ where
+
+open import Cubical.Data.DescendingList.Strict A _>_
+open import Cubical.HITs.ListedFiniteSet renaming (_∈_ to _∈ʰ_)
+
+import Cubical.Data.Empty as Empty
+open Empty using (⊥-elim)
+open import Cubical.Relation.Nullary.DecidableEq
+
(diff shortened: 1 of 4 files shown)


Saizan:
I think `Cubical.Data.DescendingList.Strict-Properties` should actually be `Cubical.Data.DescendingList.Strict.Properties` to fit the current naming scheme.

Also I would export the isomorphism as well, rather than just the path `SDL≡LFSet`, in case it's useful to use directly.

@mortberg do you have any comments? Since you proposed proving the equivalence originally.

mortberg:
> I think `Cubical.Data.DescendingList.Strict-Properties` should actually be `Cubical.Data.DescendingList.Strict.Properties` to fit the current naming scheme.
> 
> Also I would export the isomorphism as well, rather than just the path `SDL≡LFSet`, in case it's useful to use directly.
> 
> @mortberg do you have any comments? Since you proposed proving the equivalence originally.

I think this is a nice contribution. I agree with @Saizan's suggestions for changes. Once this is fixed I think we should merge.

Rotsor:
I pushed fixes to the issues mentioned so far and fixed compilation after merging with the latest master. (should I rebase?)

PR #203, opened 2019-11-28
Mutual definitions of Binℕ and Pos

Rotsor:
Currently matching on `Binℕ` asks you to consider four cases:

- `binℕ0`
- `binℕpos pos1`
- `binℕpos (x0 p)`
- `binℕpos (x1 p)`

It turns out that if you make the definitions mutual like so, you can subsume the `pos1` case into `x1`:

```
-- Positive binary numbers
data Binℕ : Type₀
data Pos : Type₀ where
  pos1 : Pos
  x0   : Pos → Pos
  x1-gen   : Binℕ → Pos

-- Binary natural numbers
data Binℕ where
  binℕ0   : Binℕ
  binℕpos : Pos → Binℕ
```

This change is compatible with all existing proofs by defining pattern synonyms as can be seen in commit 2bfdfb0767ba64f599b9c796097a280c98da3cc4:

```
pattern pos1 = x1-gen binℕ0
pattern x1 n = x1-gen (binℕpos n)
```

The follow-up commits use the new power to make some definitions compute better and simplify proofs, making use of the extra computation. (although it does go a little beyond that)

I also rename `[old]x1 -> x1-pos` and `x1-gen -> x1`

(This idea came up in the context of [dyadics](https://github.com/Rotsor/cubical/blob/dya/Cubical/DyaBin.agda) where this representation worked better for me)

### DIFF
--- a/Cubical/Data/BinNat/BinNat.agda
+++ b/Cubical/Data/BinNat/BinNat.agda
@@ -74,11 +74,20 @@ open import Cubical.Data.Empty
 
 open import Cubical.Relation.Nullary
 
+
 -- Positive binary numbers
+data Binℕ : Type₀
 data Pos : Type₀ where
-  pos1 : Pos
   x0   : Pos → Pos
-  x1   : Pos → Pos
+  x1-gen   : Binℕ → Pos
+
+-- Binary natural numbers
+data Binℕ where
+  binℕ0   : Binℕ
+  binℕpos : Pos → Binℕ
+
+pattern pos1 = x1-gen binℕ0
+pattern x1 n = x1-gen (binℕpos n)
 
 sucPos : Pos → Pos
 sucPos pos1    = x0 pos1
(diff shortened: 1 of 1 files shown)


mortberg:
It's really annoying that Travis keeps timing out because Agda is so slow to compile...

Saizan:
It really is annoying. The agda/agda-stdlib repo manages to build though, we should look at their .travis file.

Saizan:
@Rotsor can you run `make fix-whitespace` or otherwise make sure you do not have trailing whitespace?

Apparently there's some in `Cubical.Data.BinNat.BinNat`

Rotsor:
I think I removed the trailing space now. Thanks for doing it for me in the other PR.

Rotsor:
I now also noticed that the type `binnat` is isomorphic to `Pos` in a boring way: it's possible to redefine `binnat` with pattern synonyms and keep all existing proofs working (91c0292859fc9dac2d422d93b707e48a89ee855a):

```
binnat : Type₀
binnat = Pos

pattern zeroB = x1 binℕ0
pattern consOdd x = x0 x
pattern consEven x = x1-pos x
```

We can then remove proofs that become duplicated (d8286696e90a361191322cc3905f6f7e33b6b968). 

I found one proof that's different though.
It's a proof of `∀ n → ℕ→binnat (binnat→ℕ n) ≡ n`, also known as `ℕ⇒Pos (Pos⇒ℕ p) ≡ p`. For `Pos` we prove it by induction while for binnat the proof is more direct. We could get rid of one of the proofs or we could keep both.

What do you think? Should I make a PR out of this?

Saizan:
@Rotsor it's a bit strange that `zeroB` would be represented by `x1 binℕ0` as `x1 _` has the connotation of adding a "1" digit to me.
There is some value in retaining intuitive/easy to work with representations for every type, and just have isomorphisms to transport things between them.

Rotsor:
@Saizan, that's true. I wish it was possible to have the cake (abstraction) and eat it too (compute through it).

I'll try to see what can be gotten from an isomorphism proof here. In the worst case. At the minimum I'll make a PR that points out this relationship in comments.

Rotsor:
@Saizan, that's true. I wish it was possible to have the cake (abstraction) and eat it too (compute through it).

I'll try to see what can be gotten from an isomorphism proof here. In the worst case. At the minimum I'll make a PR that points out this relationship in comments.

PR #207, opened 2019-12-11
Allowing diferent levels of arguments for some definitions in BiInvEquiv

marcinjangrzybowski:
I am not sure how to create pull request without all those unnecesary 14 merge commits. Are those problematic? I do not want to create unnecesary mess in main branch commit history...


Saizan:
> I am not sure how to create pull request without all those unnecesary 14 merge commits. Are those problematic? I do not want to create unnecesary mess in main branch commit history...

It seems like you are locally working directly on your `master` branch where you pull from `cubical/master`. That usually doesn't work so well without taking a lot of care.

What's easiest is to do your changes in a different local branch you create for a specific purpose. It's referred to as a "feature branch".  Then you can push that branch to your fork and create a PR from it. 
Your local `master` is instead reserved just for pulling commits from the `agda/cubical` repo, and it's used as the base for new feature branches.

For this PR we can squash all the commits into one while we merge.

PR #208, opened 2019-12-16
direct proof of associativity for join

ecavallo:


Cubical/HITs/Join/Properties.agda
```diff
@@ -323,3 +323,99 @@ joinSwitch = isoToEquiv (iso map map invol invol)
...
+
+  forwardBack : ∀ u → forward (back u) ≡ u
+  forwardBack (inl a) = refl
+  forwardBack (inr (inl b)) = refl
+  forwardBack (inr (inr c)) = refl
+  forwardBack (inr (push b c j)) = refl
+  forwardBack (push a (inl b) i) = refl
+  forwardBack (push a (inr c) i) = refl
+  forwardBack (push a (push b c j) i) l =
+    comp
```
mortberg: Doesn't hcomp work here?

Cubical/HITs/Join/Properties.agda
```diff
@@ -323,3 +323,99 @@ joinSwitch = isoToEquiv (iso map map invol invol)
...
+
+  backForward : ∀ u → back (forward u) ≡ u
+  backForward (inl (inl a)) = refl
+  backForward (inl (inr b)) = refl
+  backForward (inl (push a b i)) = refl
+  backForward (inr c) = refl
+  backForward (push (inl a) c j) = refl
+  backForward (push (inr b) c j) = refl
+  backForward (push (push a b i) c j) l =
+    comp
```
mortberg: hcomp?

Cubical/HITs/Join/Properties.agda
ecavallo: Nope. These arise from applying forward/back to an hcomp, which produces a comp.

ecavallo: Specifically, the `l = i0` face of the clause would not be correct if we used an hcomp.

mortberg: I see. In this case I guess it could have been an hcomp though, as forward/back are non-dependent functions...

mortberg: I mean that Cubical Agda could in principle have reduced to an hcomp instead of a comp when applying forward/back to an hcomp

ecavallo: yep

mortberg:
I restarted the build. Hopefully Travis will succeed this time... Then I'll merge

PR #209, opened 2019-12-17
clean up hlevel code

### LINKED ISSUE #182: Definition of Square and isSet'
The definition of `Square` in the library order the paths by going around the perimeter, and then `isSet'` shuffles them around again.
https://github.com/agda/cubical/blob/c9fc29e4677002122f497de8618b886f0ba61898/Cubical/Foundations/Prelude.agda#L185-L195

I think the best order is to pair up opposite sides, since that matches up nicely with the nested `PathP`s:
```
Square p q r s = PathP (λ i → r i ≡ s i) p q
```

also we should be consistent about it, at least when there's only one square talked about.

ecavallo:
Organize the results in Foundations.HLevels and resolve #182

### DIFF
--- a/Cubical/Foundations/HLevels.agda
+++ b/Cubical/Foundations/HLevels.agda
@@ -69,8 +69,8 @@ isOfHLevelDep (suc (suc  n)) {A = A} B = {a0 a1 : A} (b0 : B a0) (b1 : B a1) →
 isOfHLevel→isOfHLevelDep : (n : ℕ)
   → {A : Type ℓ} {B : A → Type ℓ'} (h : (a : A) → isOfHLevel n (B a)) → isOfHLevelDep n {A = A} B
 isOfHLevel→isOfHLevelDep 0 h {a} =
-  (h a .fst , λ b' p → isProp→PathP (λ x → isContr→isProp (h x)) p (h a .fst) b')
-isOfHLevel→isOfHLevelDep 1 h = λ b0 b1 p → isProp→PathP h p b0 b1
+  (h a .fst , λ b' p → isProp→PathP (λ i → isContr→isProp (h (p i))) (h a .fst) b')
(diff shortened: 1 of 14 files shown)


mortberg:
The code looks fine.

Does anyone have any idea how to make Travis not time out after 10 minutes? It's pretty much useless to have Travis if it cannot compile Agda...

scott-fleischman:
> Does anyone have any idea how to make Travis not time out after 10 minutes? It's pretty much useless to have Travis if it cannot compile Agda...

https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

> If you have a command that doesn’t produce output for more than 10 minutes, you can prefix it with travis_wait, a function that’s exported by our build environment.

scott-fleischman:
Added in #209

Cubical/Foundations/HLevels.agda
```diff
@@ -262,57 +268,43 @@ hLevel≃ (suc n) hA hB =
...
 hLevel≡ n hA hB = hLevelRespectEquiv n (invEquiv univalence) (hLevel≃ n hA hB)
 
-hLevelHLevel1 : isProp (HLevel ℓ 0)
-hLevelHLevel1 x y = ΣProp≡ (λ _ → isPropIsContr) ((hLevel≡ 0 (x .snd) (y .snd) .fst))
-
-hLevelHLevelSuc : ∀ n → isOfHLevel (suc (suc n)) (HLevel ℓ (suc n))
-hLevelHLevelSuc n x y = subst (λ e → isOfHLevel (suc n) e) HLevel≡ (hLevel≡ (suc n) (snd x) (snd y))
+-- h-level of HLevel
 
-hProp≡HLevel1 : hProp ℓ ≡ HLevel ℓ 1
```
ecavallo: this is true by reflexivity

Issue #176, opened 2019-08-12
Why doesn't Foundations.Prelude export isEquiv?


m-yac:
It seems strange that every time one wants to use `isEquiv` they have to import `Cubical.Core.Glue` – almost every other primitive is exported by `Foundations.Prelude`. Is there some reason for excluding `isEquiv` like this?

PR #210, opened 2019-12-17
export equivalences and glue types in Foundations

ecavallo:
Proposed resolution to #176: 
- export the equivalence definitions from Cubical.Core.Glue in Foundations.Equiv,
- export the Glue type definitions in Foundations.Univalence.


mortberg:
Thanks!

PR #214, opened 2019-12-20
Mapping cones and dunce cap contraction

m-yac:
Mapping cones (AKA the homotopy cofiber/cokernel) as a HIT and proof that the dunce cap (defined as a mapping cone) is contractible.

I got stuck fairly quickly trying to prove that the following HIT is equivalent to the mapping cone definition of `Dunce` – I'll give it another go tomorrow but maybe someone else will have more luck.
```agda
data DunceHIT : Type₀ where
  base : DunceHIT
  loop : base ≡ base
  surf : PathP (λ i → loop i ≡ loop i) loop refl
```


m-yac:
There are two pairs of holes left that seem like they should be really easy to fill – but I just haven't been able to work it out yet.

For example, I the following two types *should* be equivalent:
```agda
data Dunce : Type₀ where
  base : Dunce
  loop : base ≡ base
  surf : PathP (λ i → loop i ≡ loop i) loop refl

data Dunce₂ : Type₀ where
  base₂ : Dunce₂
  loop₂ : base₂ ≡ base₂
  surf₂ : loop₂ ⁻¹ ∙∙ loop₂ ∙∙ loop₂ ≡ refl
```
since I proved:
```agda
PathP≡doubleCompPath : (PathP (λ i → p i ≡ s i) q r) ≡ (p ⁻¹ ∙∙ q ∙∙ s ≡ r)
PathP≡doubleCompPath k = PathP (λ i → p (i ∨ k) ≡ s (i ∨ k)) (λ j → doubleCompPath-filler (p ⁻¹) q s j k) r
```
but I wasn't able to work out how to state the `leftInv` and `rightInv` cases of the isomorphism.

ecavallo:
One way to do it is to prove that `Dunce` is also contractible, which can be done like so:
```
contract : (d : Dunce) → d ≡ base
contract base k = loop k
contract (loop i) k = surf k i
contract (surf i j) k =
  hcomp
    (λ l → λ
      { (i = i0) → surf k j
      ; (i = i1) → cube l
      ; (j = i0) → cube l
      ; (j = i1) → cube l
      ; (k = i0) → surf i j
      ; (k = i1) → base
      })
    (surf (k ∨ i) j)
  where
  cube : I → Dunce
  cube l =
    hfill
      (λ i → λ
        { (k = i0) → loop i
        ; (k = i1) → base
        ; (l = i0) → loop (k ∨ i)
        ; (l = i1) → surf k i
        })
      (inS (loop k))
      i
```
(I arrived at this definition by thinking about how to do the much simpler case where the 2-d constructor is instead `surf : loop = refl`.)

m-yac:
Wow, thanks! I couldn't work out this proof on my own, so I'm definitely going to study that hcomp.

Although I used this to prove `Dunce` and `DunceCone` equivalent, I would still be very interested in understanding how one might do it without appealing to contractibility – i.e. understanding general strategies for proving HITs with squares vs. HITs with an hcomp equivalent (or even more generally, HITs with path constructors of equivalent types equivalent). I noticed this is similar to a recent discussion on the cubical forum, so I'll continue the conversation there.

It's worth noting:
- Because of my interest in a direct proof that `Dunce ≡ DunceCone`, I moved my progress to `Experiments/DunceEquiv.agda`. Happy to remove this file, though, if people think it's not necessary.
- I moved `isoToEquiv` and `compIso` from `Foundations.Equiv` to `Foundations.Isomorphism` because they seemed to be in the wrong place.

Cubical/Experiments/DunceEquiv.agda
```diff
@@ -0,0 +1,146 @@
+{-
```
mortberg: I don't think we need this attempt as the other proof is so slick. But this file contains some generally useful lemmas that should be moved elsewhere.

mortberg:
There are some conflicts here now. Let me know when you've fixed them and I'll merge

m-yac:
Thanks! All the conflicts should be resolved now. I also added `ua-ungluePath` for completeness.

mortberg:
Excellent! Will merge once travis is done (I had a to fix another small conflict after merging the nullification PR)

PR #215, opened 2019-12-23
Yoneda lemma

jonsterling:
I'm putting this here so that I can get critiqued by people who know how to use cubical agda better than me :)

The code is currently not organized super well, but I thought I'd open a PR so that people can offer suggestions for improving it, matching the idiom of this library, etc.


jonsterling:
Travis seems to be kind of F'd :wink:

scott-fleischman:
It seems to be running out of memory. Not sure what the Agda build itself does to improve that.

Saizan:
Yeah, I'll have to seriously look into travis when I have a bit more time.

@jonsterling it might be nice to factor out the types and operations of `Precategory` into its own record so that we can have a lemma about two categories being equal as long as that data is equal.

The `no-eta-equality` is great, you might even want to add it to `isCategory`.

jonsterling:
@Saizan good idea! I'll do some refactoring when I have a moment.

Cubical/Categories/Presheaves.agda
```diff
@@ -0,0 +1,76 @@
...
+open import Cubical.HITs.PropositionalTruncation
+
+open import Cubical.Categories.Category
+open import Cubical.Categories.Functor
+open import Cubical.Categories.NaturalTransformation
+open import Cubical.Categories.Sets
+
+module _ (ℓ : Level) where
+  PSH : Precategory ℓ → Precategory (ℓ-suc ℓ)
+  PSH 𝒞 = FUNCTOR (𝒞 ^op) (SET ℓ)
```
phadej: this resulting category `isCategory`, as `SET` `isCategory`, isn't it so? (probably useful result to add to `NaturalTransformations`). If i'm not mistaken, that's HoTT theorem 9.2.5

mortberg: Yes it is, but I think it can be added to a future PR

mortberg:
I think the code looks good and will merge as soon as Travis is happy. Thanks for contributing and sorry for taking so long to look a this.

 @Saizan Didn't someone at Chalmers already develop some CT in some earlier version of Cubical Agda?

mortberg:
> Yeah, I'll have to seriously look into travis when I have a bit more time.
> 
> @jonsterling it might be nice to factor out the types and operations of `Precategory` into its own record so that we can have a lemma about two categories being equal as long as that data is equal.
> 
> The `no-eta-equality` is great, you might even want to add it to `isCategory`.

Oh, this is a good point that I didn't think about. 

@jonsterling Could you do the refactoring? Then I'll merge

Saizan:
@mortberg yeah, it's at https://github.com/Saizan/cat however it would need to be ported from `cubical-demo` and I guess reorganized a bit.

jonsterling:
I think I may be able to make a little time this afternoon to get this refactored. thanks for the feedback!

jonsterling:
I was thinking about it, and I realized that while I really like the idea of separating structure from property in theory, in the Agda that currently exists, this will most likely lead to some kind of hell.

1. I am not keen on writing `C .struct .seq (C.struct .idn x, f)` everywhere, etc. This can be solved by defining some helper functions, but then see the next issue

2. When you construct a category explicitly by copatterns, you will get wrecked under a lot of projections.

Both of these could be resolved if we had copattern synonyms (https://github.com/agda/agda/issues/2622), but I guess these aren't implemented yet. Does anyone have any ideas on how I should proceed?

Saizan:
@jonsterling Yeah, I guess constructing elements of these types would get quite tedious. No need to refactor this then.

Just FYI, the module system helps with deifning the helper projections:

```agda
module Nested where

postulate
  A B : Set

record Data : Set where
  field
    someA : A

open Data public
record Thing : Set where
  field
    someData : Data
    someB : B

  open Data someData public

open Thing public

foo : (t : Thing) → A
foo t = t .someA
```

But it still doesn't allow to use `.someA` as copattern.

jonsterling:
@Saizan Wow, thanks for the tip, that's great to know!

jonsterling:
BTW feel free to merge if you folks are ok with it, I won't be editing this PR any time soon. (But if there are more requests for changes, I can do it in about a month.)

fredefox:
> @Saizan Didn't someone at Chalmers already develop some CT in some earlier version of Cubical Agda?

@Saizan co-supervised my masters thesis where I implemented some things in Cubical Agda. You can find it here https://github.com/fredefox/cat

@bafain just updated it to work with the latest and greatest versions of Agda, The standard library and the cubical library. I just tested it myself today as well.

Saizan:
@fredefox that's great! Any plans for the future of the library?

fredefox:
Realistically *I* probably want be spending much time on it. Some ideas came up yesterday when I presented my thesis at ITC. Perhaps it would be interesting to define the notion of categories indexed by their homotopy level - so as to not just have small categories. And then define the category of categories.

It could also be interesting to compare the way the code is structured with how the code in https://github.com/agda/agda-categories is structured (which uses setoids, not cubical Agda).

There are some other partial implementations in the library itself. Like the notion of monoid, monoidal categories, various concrete categories like the functor category and CwFs (@anka-213 did some work on this).

PR #216, opened 2019-12-30
Add listings make rule

phadej:
Adding also the README module.

---

The HTML version is handy, it's nice to click around the source to explore what's in the library.
We can setup travis to upload html, so it would be available as it's done for `agda-stdlib` (https://agda.github.io/agda-stdlib/README.html is super handy)

---

It would be cleaner to move sources into `src` directory (like in `agda-stdlib`), then `README` module won't be "accidentally" part of the library, as we could say

```
name: cubical
include: src
```

in `cubical.agda-lib`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 *.agdai
 *~
+html
--- a/Cubical/Core/README.md
+++ b/Cubical/Core/README.md
@@ -6,21 +6,11 @@ the following things:
 
 * **Primitives**: exposes cubical agda primitives.
 
-* **Prelude**: basic cubical prelude.
-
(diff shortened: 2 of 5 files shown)


phadej:
The travis is happy!

Saizan:
> It would be cleaner to move sources into `src` directory (like in `agda-stdlib`), then `README` module won't be "accidentally" part of the library, as we could say
> 
> ```
> name: cubical
> include: src
> ```
> 
> in `cubical.agda-lib`

It might be simpler to have it as `Cubical.README`, it still won't clash with other namespaces.

phadej:
Changed to have `Cubical.README`, it's indeed less fuss.

PR #226, opened 2020-01-16
A "near inverse" to `∣_∣ : A → ∥ A ∥`

m-yac:
Adds a function `recover` which definitionally satisfies `recover ∣ x ∣ ≡ x`, meaning agda's normalizer can uncover the information supposedly hidden by `∣_∣` for any type. Based on Nicolai Kraus' blog post: [The Truncation Map |_| : ℕ -> ‖ℕ‖ is nearly Invertible](https://homotopytypetheory.org/2013/10/28/the-truncation-map-_-%E2%84%95-%E2%80%96%E2%84%95%E2%80%96-is-nearly-invertible/).

The definition of `recover` and a small demo can be found in `HITs.PropositionalTruncation.MagicTrick`. This PR also adds:
- some facts about pointed types in `Foundations.Pointed.Properties`
- a definition and some facts about homogenous types in `Foundations.Pointed.Homogeneous`
- HoTT book Lemma 3.11.8 and equivalences defined by `_∙_` in `Foundations.Path`
- a useful function (`ua-gluePath`) in `Foundations.Univalence`

### DIFF
--- a/Cubical/Foundations/Path.agda
+++ b/Cubical/Foundations/Path.agda
@@ -75,9 +75,31 @@ toPathP-isEquiv A {x} {y} = isoToIsEquiv (iso toPathP fromPathP to-from from-to)
                           })
                           (transp (\ j → A ((i ∨ h) ∧ j)) (~ (i ∨ h)) x)
 
+compPathl-cancel : ∀ {ℓ} {A : Type ℓ} {x y z : A} (p : x ≡ y) (q : x ≡ z) → p ∙ (sym p ∙ q) ≡ q
+compPathl-cancel p q = p ∙ (sym p ∙ q) ≡⟨ assoc p (sym p) q ⟩
+                       (p ∙ sym p) ∙ q ≡⟨ cong (_∙ q) (rCancel p) ⟩
+                              refl ∙ q ≡⟨ sym (lUnit q) ⟩
+                                     q ∎
+
+compPathr-cancel : ∀ {ℓ} {A : Type ℓ} {x y z : A} (p : z ≡ y) (q : x ≡ y) → (q ∙ sym p) ∙ p ≡ q
+compPathr-cancel p q = (q ∙ sym p) ∙ p ≡⟨ sym (assoc q (sym p) p) ⟩
+                       q ∙ (sym p ∙ p) ≡⟨ cong (q ∙_) (lCancel p) ⟩
+                              q ∙ refl ≡⟨ sym (rUnit q) ⟩
+                                     q ∎
+
+compPathl-isEquiv : ∀ {ℓ} {A : Type ℓ} {x y z : A} (p : x ≡ y) → isEquiv (λ (q : y ≡ z) → p ∙ q)
+compPathl-isEquiv p = isoToIsEquiv (iso (p ∙_) (sym p ∙_) (compPathl-cancel p) (compPathl-cancel (sym p)))
+
+compPathr-isEquiv : ∀ {ℓ} {A : Type ℓ} {x y z : A} (p : y ≡ z) → isEquiv (λ (q : x ≡ y) → q ∙ p)
+compPathr-isEquiv p = isoToIsEquiv (iso (_∙ p) (_∙ sym p) (compPathr-cancel p) (compPathr-cancel (sym p)))
+
 -- Variation of isProp→isSet for PathP
 isProp→isSet-PathP : ∀ {ℓ} {B : I → Type ℓ} → ((i : I) → isProp (B i))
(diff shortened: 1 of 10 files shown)


mortberg:
Cool! I'll take a closer look next week when I'm less busy, but I guess this is the same example as what we formalized in the (by now quite ancient) predecessor of cubicaltt called cubical: https://github.com/simhu/cubical/blob/master/examples/Kraus.cub

Cubical/Foundations/Path.agda
```diff
@@ -75,9 +75,31 @@ toPathP-isEquiv A {x} {y} = isoToIsEquiv (iso toPathP fromPathP to-from from-to)
...
+compPathr-isEquiv p = isoToIsEquiv (iso (_∙ p) (_∙ sym p) (compPathr-cancel p) (compPathr-cancel (sym p)))
+
 -- Variation of isProp→isSet for PathP
 isProp→isSet-PathP : ∀ {ℓ} {B : I → Type ℓ} → ((i : I) → isProp (B i))
                    → (b0 : B i0) (b1 : B i1)
                    → isProp (PathP (λ i → B i) b0 b1)
 isProp→isSet-PathP {B = B} hB b0 b1 =
   transport (λ i → isProp (PathP≡Path B b0 b1 (~ i))) (isProp→isSet (hB i1) _ _)
+
+-- Lemma 3.11.8 in the HoTT book
```
mortberg: This is essentially: https://github.com/agda/cubical/blob/master/Cubical/Foundations/Prelude.agda#L153

Cubical/Foundations/Pointed/Base.agda
```diff
@@ -0,0 +1,11 @@
+{-# OPTIONS --cubical --safe #-}
+module Cubical.Foundations.Pointed.Base where
+
+open import Cubical.Foundations.Prelude
+
+record Pointed ℓ : Type (ℓ-suc ℓ) where
```
mortberg: Pointed types are already defined in two places in the library:
https://github.com/agda/cubical/blob/master/Cubical/Data/HomotopyGroup/Base.agda#L13
https://github.com/agda/cubical/blob/master/Cubical/Foundations/SIP.agda#L413

It's much better to have them in a separate file, but is it really worth using a record and not just a Sigma? If we define them with a Sigma then the SIP proof will just work as it is. You can still define `typ` and `pt` instead of the normal Sigma projections.

Once this is done could you please remove the duplicate definition in HomotopyGroups? The one in SIP can stay for now as we are working on that file and will refactor later

Cubical/Foundations/Univalence.agda
```diff
@@ -37,9 +37,20 @@ uaIdEquiv : {A : Type ℓ} → ua (idEquiv A) ≡ refl
...
+            → B [ _ ↦ (λ { (i = i0) → e .fst x ; (i = i1) → x }) ]
+ua-unglue e i x = inS (unglue (i ∨ ~ i) x)
+
+-- Give detailed type to glue
+ua-glue : ∀ {A B : Type ℓ} (e : A ≃ B) (i : I) (x : A) (y : B)
+          → B [ _ ↦ (λ { (i = i0) → e .fst x ; (i = i1) → y }) ]
+          → (ua e i) [ _ ↦ (λ { (i = i0) → x ; (i = i1) → y }) ]
+ua-glue e i x y s = inS (glue (λ { (i = i0) → x ; (i = i1) → y }) (outS s))
+
+ua-gluePath : ∀ {A B : Type ℓ} (e : A ≃ B) {x : A} {y : B}
```
mortberg: This is really nice to have as a lemma!

Cubical/Foundations/Pointed/Base.agda
mortberg: By "SIP proof" I mean the fact that isomorphic pointed types are equal. It's not yet stated in the SIP file, but should be easy to add. Maybe this could be used in some places in your code here (like for `isHomogeneousPath`).

m-yac: Yes, I agree now that this should just be a sigma. My initial rationale was that defining it as a record would force one to use `typ` and `pt` instead of `fst` and `snd`, since using the latter makes code involving pointed types far less readable, from my experience. But having `Pointed ℓ` be definitionally a sigma is definitely more useful.

I had already removed the duplicate definition in HomotopyGroups, but I'm not sure what you have in mind when you suggest to use SIP to prove `isHomogenousPath`.

mortberg: Great! I'm not sure what I had in mind either, so marking this as resolved :-)

PR #230, opened 2020-01-22
Localization, Nullification, and Truncation from -2

### LINKED ISSUE #137: Show that the (already defined) truncations are modalities
as discussed in #134

### LINKED ISSUE #138: Define truncations as nullification modalities
as discussed in #134 

I suppose this issue also includes defining localization and finally showing that the two definitions of truncation are the same.

### LINKED ISSUE #219: Defining truncation from -2
Once #155 is merged I plan to make a PR adding n-truncated and n-connected maps (plus a few easy facts) and it would be nice to have a definition of truncation which is defined from -2.

I experimented with this here: https://github.com/m-yac/cubical/commit/309f3ed021be2ccf993f461ad9254a64039cb950

In that commit I defined `ℕ₋₂` in an analagous way to `ℕ₋₁` (along with literals for both) and used the following definition of truncation from negative two:
```agda
∥_∥_ : ∀ {ℓ} (A : Type ℓ) (n : ℕ₋₂) → Type ℓ
∥ A ∥ neg2  = Lift Unit
∥ A ∥ suc n = FromMinusOne.∥ A ∥ (1+ n) -- the old HIT definition
```

Since one needs to pattern match on `n` when doing anything with this definition, I don't think it is ideal (e.g. the definition of `|_|` is already clunky). I'm curious if anyone knows of a better one.

m-yac:
Based on discussion in #219, this PR defines truncation from -2 through nullification (`HITs.Nullification`) as well as defines localization (`HITs.Localization`). In a future PR I will clean up `Foundations.PathSplitEquiv` and prove some more facts about localizations from the ['Modalities in Homotopy Type Theory' paper](https://arxiv.org/pdf/1706.07526.pdf).

Most of the work on this PR went into proving the final two cases of `ind` in `HITs.Nullification.Properties`. I tried lots of different versions of these final two path constructors based on different notions of `const : Null S A → (S → Null S A)` being an equivalence, but the only one for which I could manage to define this function was the one below based on `const` being a path-split equivalence.

```agda
data Null {ℓ ℓ'} (S : Type ℓ) (A : Type ℓ') : Type (ℓ-max ℓ ℓ') where
  ∣_∣ : A → Null S A
  -- the image of every map (S → Null S A) is contractible in Null S A
  hub   : (f : S → Null S A) → Null S A
  spoke : (f : S → Null S A) (s : S) → hub f ≡ f s
  -- the image of every map (S → x ≡ y) for x y : A is contractible in x ≡ y
  ≡hub   : ∀ {x y} (p : S → x ≡ y) → x ≡ y
  ≡spoke : ∀ {x y} (p : S → x ≡ y) (s : S) → ≡hub p ≡ p s
```

This PR also includes:
- A type `ℕ₋₂` and integer literals for `ℕ₋₂` and `ℕ₋₁`
- A change to the the type of `isOfHLevelPath` to reflect how it is actually used
- The function `ua-gluePath` from #226 (It's really useful!)
- A name-swap of `S` and `S₊`

Check the commit messages for the complete list of other additions/changes.

Resolve #219 
Resolve #138
Resolve #137


Cubical/Data/Nat/Base.agda
```diff
@@ -7,6 +7,16 @@ open import Agda.Builtin.Nat public
   using (zero; suc; _+_; _*_)
   renaming (Nat to ℕ)
 
+open import Agda.Builtin.FromNat public
```
mortberg: What is this?

Cubical/Data/Nat/Base.agda
```diff
@@ -7,6 +7,16 @@ open import Agda.Builtin.Nat public
...
+  renaming (Number to HasFromNat)
+open import Agda.Builtin.FromNeg public
+  renaming (Negative to HasFromNeg)
+open import Cubical.Data.Unit.Base public
+
+instance
+  fromNatℕ : HasFromNat ℕ
+  fromNatℕ = record { Constraint = λ _ → Unit ; fromNat = λ n → n }
```
mortberg: Also, what is this?

Cubical/Data/NatMinusTwo.agda
```diff
@@ -0,0 +1,4 @@
+{-# OPTIONS --cubical --safe #-}
```
mortberg: Could you please add this file to Data.Everything?

Cubical/Foundations/Prelude.agda
```diff
@@ -141,6 +141,10 @@ substRefl px = transportRefl px
 funExt : {f g : (x : A) → B x} → ((x : A) → f x ≡ g x) → f ≡ g
 funExt p i x = p x i
 
+-- the inverse to funExt (see Foundations.FunExtEquiv)
+appl : ∀ {f g : (x : A) → B x} → f ≡ g → (x : A) → f x ≡ g x
```
mortberg: A better name might be `funExt⁻`?

Cubical/HITs/Truncation/Base.agda
```diff
@@ -1,13 +1,11 @@
...
+open import Cubical.Data.NatMinusTwo
+open import Cubical.HITs.Nullification
 open import Cubical.HITs.Sn
 
-data  ∥_∥_ {ℓ} (A : Type ℓ) (n : ℕ₋₁) : Type ℓ where
-  ∣_∣ : A  → ∥ A ∥ n
-  hub : (f : S (1+ n) → ∥ A ∥ n) → ∥ A ∥ n
-  spoke : (f : S (1+ n) → ∥ A ∥ n) (x : S (1+ n)) → hub f ≡ f x
-
+∥_∥_ : ∀ {ℓ} → Type ℓ → ℕ₋₂ → Type ℓ
```
mortberg: I think we should keep the old n-truncation code. It might have better/worse computational capabilities, so it would be nice to be able to easily switch definition and do benchmarks. Can you please put the old code in a module?

mortberg: This goes also for the proofs about it in the Properties file

Cubical/Modalities/Modality.agda
```diff
@@ -58,15 +58,14 @@ record Modality ℓ : Type (ℓ-suc ℓ) where
...
-    abstract
-      to-from : ∀ ◯b → ◯-map f (◯-map inv ◯b) ≡ ◯b
-      to-from = ◯-elim
-        (λ ◯b → ◯-=-isModal (◯-map f (◯-map inv ◯b)) ◯b)
-        (λ b → cong (◯-map f) (◯-map-β inv b) ∙ ◯-map-β f (inv b) ∙ cong η (rightInv b))
-      from-to : ∀ ◯a → ◯-map inv (◯-map f ◯a) ≡ ◯a
-      from-to = ◯-elim
-          (λ ◯a → ◯-=-isModal (◯-map inv (◯-map f ◯a)) ◯a)
-          (λ a → cong (◯-map inv) (◯-map-β f a) ∙ ◯-map-β inv (f a) ∙ cong η (leftInv a))
+    to-from : ∀ ◯b → ◯-map f (◯-map inv ◯b) ≡ ◯b
```
mortberg: I wonder why these things were abstract before? Efficiency?

Cubical/HITs/Nullification/Properties.agda
```diff
@@ -0,0 +1,129 @@
...
+            (fst (secCongDep' nB (≡hub p) (transport refl (ind nB g x)) (transport refl (ind nB g y)))
+                 (λ i s → transport (λ j → B (≡spoke p s (~ j) i)) (ind nB g (p s i))) i)
+
+  ind nB g (≡spoke {x} {y} p s j i)
+    = hcomp (λ k → λ { (i = i0) → toPathP⁻-sq (ind nB g x) k j
+                     ; (i = i1) → toPathP⁻-sq (ind nB g y) k j
+                     ; (j = i1) → ind nB g (p s i) })
+            (q₂ j i)
+
+    where q₁ : Path (PathP (λ i → B (≡hub p i)) (transport refl (ind nB g x)) (transport refl (ind nB g y)))
```
m-yac: Should these be inlined? Having the types of these explicit was super helpful in actually proving this case, but inlining them might make more apparent its similarities to the other three cases.

Cubical/Modalities/Modality.agda
m-yac: I'm not sure, with all of them removed it is slightly slower, but not to an extent that is unreasonable for files in this library (e.g. it's faster than `GroupoidLaws`). For some reason I was getting an error when compiling this file with `Σ-isModal` marked abstract, so first I removed just that instance of 'abstract', but then after testing just removed them all.

Cubical/HITs/Truncation/Base.agda
m-yac: I moved the old code to `Truncation/FromNeg1/` and added an explanation of the difference to `Truncation.Base`. Is something like this what you had in mind?

Cubical/Data/Nat/Base.agda
mortberg: Ok, maybe you can add a comment about this so that we don't forget what the instance arguments are there for?

Cubical/HITs/Nullification/Properties.agda
mortberg: This probably makes the proof a bit more readable... So let's keep them

Cubical/Data/Nat/Base.agda
mortberg: Marking this as resolved, but please add some comments about why these instance arguments are here

Cubical/HITs/Truncation/Base.agda
mortberg: Yes, exactly!

mortberg:
I think everything looks fine now, except for some comments about the instance arguments. Let me know when you have fix this and I'll merge

mortberg:
Oops, clicked the wrong button!

m-yac:
Thanks for the comments! If what I added to `Data.Nat`, `Data.NatMinusOne`, etc. were the comments you were looking for then this is ready to merge!

PR #233, opened 2020-02-04
Add list functions and properties

limemloh:
Hi there
Here are some definitions and properties from a case study I did together with @paldepind.

This will add the `length` function and `∷ʳ`(snoc) together with related properties and some properties regarding nil.

### DIFF
--- a/Cubical/Data/List/Properties.agda
+++ b/Cubical/Data/List/Properties.agda
@@ -122,6 +122,24 @@ cons-inj₂ = cong safe-tail
 ¬nil≡cons : ∀ {x : A} {xs} → ¬ ([] ≡ x ∷ xs)
 ¬nil≡cons {A = A} p = lower (subst (caseList (List A) (Lift ⊥)) p [])
 
+¬snoc≡nil : ∀ {x : A} {xs} → ¬ (xs ∷ʳ x ≡ [])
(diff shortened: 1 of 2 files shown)


Cubical/Data/List/Properties.agda
```diff
@@ -122,6 +122,25 @@ cons-inj₂ = cong safe-tail
 ¬nil≡cons : ∀ {x : A} {xs} → ¬ ([] ≡ x ∷ xs)
 ¬nil≡cons {A = A} p = lower (subst (caseList (List A) (Lift ⊥)) p [])
 
+¬snoc≡nil : ∀ {x : A} {xs} → ¬ (xs ∷ʳ x ≡ [])
+¬snoc≡nil {xs = []} contra = ¬cons≡nil contra
+¬snoc≡nil {xs = x ∷ xs} contra = ¬cons≡nil contra
+
+¬nil≡snoc : ∀ {x : A} {xs} → ¬ ([] ≡ xs ∷ʳ x)
```
mortberg: Can't you just prove this in one line using `sym` and `¬snoc≡nil`?

limemloh: Yes, you are right! This would be a better solution.

limemloh:
@mortberg I have adjusted the code to your comment

mortberg:
Thanks for the contribution!

PR #235, opened 2020-02-12
Cohomology + Identity on truncation of loop spaces

aljungstrom:
Hi! This PR contains:
- A definition of cohomology + the very beginning of a properties-file.
- A wee test computation of H_0 (S^1)
- A formalisation of the proof of Theorem 7.3.12. in the HoTT book (truncation of identity types). Probably there is a shorter proof but...
- Some very minor additions such as a proof that isSet A -> A = || A ||_0, a definition of pointed functions, etc.

Criticise away!


Cubical/Foundations/HLevels.agda
```diff
@@ -246,6 +246,16 @@ hLevelSuc 0 A = isContr→isProp
 hLevelSuc 1 A = isProp→isSet
 hLevelSuc (suc (suc n)) A h a b = hLevelSuc (suc n) (a ≡ b) (h a b)
 
+hLevelAdd : (n : ℕ) (m : ℕ) (A : Type ℓ) → isOfHLevel n A → isOfHLevel (n + m) A
```
mortberg: This is already proved in: https://github.com/agda/cubical/blob/master/Cubical/Foundations/HLevels.agda#L254

These lemmas should really be renamed to be of the form `isOfHLevelX`... I'm not suggesting that you do it in this PR, but if someone has time a PR fixing the naming in the HLevels file would be very much appreciated. (@m-yac maybe? :-) )

Cubical/Foundations/Prelude.agda
```diff
@@ -131,6 +131,7 @@ transport p a = transp (λ i → p i) i0 a
 transportRefl : (x : A) → transport refl x ≡ x
 transportRefl {A = A} x i = transp (λ _ → A) i x
 
+
```
mortberg: Why add a newline here?

Cubical/HITs/SetTruncation/Properties.agda
```diff
@@ -56,3 +56,23 @@ elimSetTrunc3 : {B : (x y z : ∥ A ∥₀) → Type ℓ}
                 (x y z : ∥ A ∥₀) → B x y z
 elimSetTrunc3 Bset g = elimSetTrunc2 (λ _ _ → hLevelPi 2 λ _ → Bset _ _ _) (λ a b →
                        elimSetTrunc (λ _ → Bset _ _ _) (λ c → g a b c))
+
+
+setTruncIsSet : isSet ∥ A ∥₀
+setTruncIsSet a b p q = squash₀ a b p q
+
+setId : isSet A → A ≡ ∥ A ∥₀
```
mortberg: Why is the conclusion flipped compared to `propId`? I'm not sure which version is the easiest to use, but it would be good if they were the same

Cubical/HITs/Truncation/Properties.agda
```diff
@@ -124,10 +124,20 @@ isSnNull→isOfHLevel {n = suc n} nA = isSphereFilled→isOfHLevelSuc (λ f →
 isOfHLevel∥∥ : (n : ℕ₋₂) → isOfHLevel (2+ n) (∥ A ∥ n)
 isOfHLevel∥∥ neg2    = hub ⊥-elim , λ _ → ≡hub ⊥-elim
 isOfHLevel∥∥ (suc n) = isSphereFilled→isOfHLevelSuc isSphereFilled∥∥
+
```
mortberg: Why add whitespace here?

Cubical/HITs/Truncation/Properties.agda
```diff
@@ -204,3 +217,174 @@ groupoidTrunc≃Trunc1 =
       (ind (λ _ → squash₂) ∣_∣₂)
       (ind (λ _ → hLevelPath 4 (isOfHLevel∥∥ 2) _ _) (λ _ → refl))
       (g2TruncElim _ _ (λ _ → hLevelPath 4 squash₂ _ _) (λ _ → refl)))
+
+
+
+---- ∥ Ω A ∥ ₙ ≡ Ω ∥ A ∥ₙ₊₁  ----
+
+  
+  {- Proofs of Theorem 7.3.12. and Corollary 7.3.13. in the HoTT book  -}
```
mortberg: I suspect that this has a more direct cubical proof. Did someone try to prove this at some point? (maybe @ecavallo ?)

Cubical/HITs/Truncation/Properties.agda
```diff
@@ -204,3 +217,174 @@ groupoidTrunc≃Trunc1 =
...
+                                                      {b})
+
+        {- The final Iso established -}
+        IsoFinal : ∀ {ℓ} {B : Type ℓ} {n : ℕ₋₂} (x y : ∥ B ∥ (suc n)) → Iso (x ≡ y) (P x y)
+        IsoFinal x y = iso (encode-fun x y ) (decode-fun x y) (P-linv x y) (P-rinv x y)
+
+IsoIdTrunc : {a b : A} (n : ℕ₋₂) → Iso (_≡_ {A = ∥ A ∥ (suc n)} ∣ a ∣ ∣ b ∣) (∥ a ≡ b ∥ n)
+IsoIdTrunc {a = a} {b = b} n = IsoFinal {n = n} ∣ a ∣ ∣ b ∣
+
+IsoΩ : {a : A} (n : ℕ₋₂) → Iso (_≡_ {A = ∥ A ∥ (suc n)} ∣ a ∣ ∣ a ∣) (∥ a ≡ a ∥ n)
```
mortberg: Why state these as Iso and not Equiv or Path? I think it would be useful to have the other formulations as well (especially Path)

Cubical/ZCohomology/Base.agda
```diff
@@ -0,0 +1,51 @@
...
+open import Cubical.HITs.Truncation.Base
+
+private
+  variable
+    ℓ : Level
+    A : Type ℓ
+
+{-
+Definition of cohomology with cocefficients in the integers. For now, this definition uses ℕ₋₂, for instance defining H⁰ as
+coHom neg2, H¹ as coHom (suc neg2), and so on. This has so far been easier to work with than a definition relying on ℕ→ℕ₋₂.
```
mortberg: This indexing is a bit strange and confusing... Let's look at it together next week

Cubical/Foundations/Pointed/Base.agda
```diff
@@ -11,3 +11,8 @@ typ = fst
 
 pt : ∀ {ℓ} (A∙ : Pointed ℓ) → typ A∙
 pt = snd
+
+{- Pointed functions -}
+_→*_ : ∀{ℓ ℓ'} → (A : Pointed ℓ) → (B : Pointed ℓ') → Type (ℓ-max ℓ ℓ')
```
guillaumebrunerie: Maybe it would be useful to also have the version that returns a pointed type ?

Cubical/HITs/Truncation/Properties.agda
```diff
@@ -145,6 +155,9 @@ ind2 : {n : ℕ₋₂}
 ind2 {n = n} hB g = ind (λ _ → hLevelPi (2+ n) (λ _ → hB _ _)) λ a →
                     ind (λ _ → hB _ _) (λ b → g a b)
 
+
+
+
```
guillaumebrunerie: Newlines?

Cubical/HITs/Truncation/Base.agda
```diff
@@ -22,3 +22,4 @@ open import Cubical.HITs.Sn
 
 ∥_∥_ : ∀ {ℓ} → Type ℓ → ℕ₋₂ → Type ℓ
 ∥ A ∥ n = Null (S (1+ n)) A
+
```
guillaumebrunerie: Why a newline?

Cubical/ZCohomology/Base.agda
guillaumebrunerie: I agree with Anders, it seems confusing to use ℕ₋₂ here. And it will get even more confusing when you want to do things like (H^n × H^m) → H^{n + m}, because then addition will need to be translated to n + m + 2 or something like that.

Cubical/ZCohomology/Base.agda
```diff
@@ -0,0 +1,51 @@
...
+
+--- Reduced cohomology ---
+
+{- Pointed version of Kₙ  -}
+coHomK-ptd : (n : ℕ₋₂) → Pointed (ℓ-zero)
+coHomK-ptd neg2 = coHomK neg2 , (pos 0)
+coHomK-ptd (suc n) = (coHomK (suc n) , ∣ north ∣)
+
+{- Reduced cohomology -}
+coHomRed : (n : ℕ₋₂) → (A : Pointed ℓ) → Type ℓ
```
guillaumebrunerie: Why do you pattern match on `n` here? Can’t it be defined simply in one line like `coHom`?

Cubical/ZCohomology/Properties.agda
```diff
@@ -0,0 +1,64 @@
...
+open import Cubical.Data.Sum.Base
+open import Cubical.Data.HomotopyGroup
+private
+  variable
+    ℓ : Level
+    A : Type ℓ
+
+
+
+{- Equivalence between cohomology and the reduced version -}
```
guillaumebrunerie: I would suggest something like "Equivalence between cohomology of A and reduced cohomology of (A + 1)" instead (and maybe a different name for the theorem), given that there is another "equivalence between cohomology and the reduced version", namely the fact that they are equivalent for n>0.

Cubical/HITs/Truncation/Properties.agda
ecavallo: I wouldn't think so, it's an encode-decode proof and those usually don't leave much room for cubical improvement.

mortberg:
Let me know when you have addressed the comments and consider the PR ready for merging. I'll then take another look and merge once Travis is happy unless I have more comments

aljungstrom:
@mortberg Ok, that's me done editing. I think it should be ready to merge. Although having seen @m-yac 's PR with a new definition of N-2, I'm thinking that might be more suitable for the definitions. Perhaps we should put my PR on hold? It's up to you.

mortberg:
Merging this now

PR #237, opened 2020-02-24
cleanup brunerie

ecavallo:
Some minor updates to use consistent notation and the pointed types / hlevel facts from the library.


mortberg:
Thanks, but no progress on computing anything?

ecavallo:
I didn't try any of the tests, but these are all cosmetic changes so I doubt it.

PR #238, opened 2020-02-25
Make ℕ₋₁ and ℕ₋₂ more usable

m-yac:
As mentioned in a (now outdated) comment by @aljungstrom, the current definitions of `ℕ₋₁` and `ℕ₋₂` are hard to work with. This is due to the fact that most of the key operations on `ℕ₋₁` and `ℕ₋₂` are defined by induction, and therefore many simple identities do not hold definitionally. I ran into this issue myself when trying to define `isSphereFilled→isOfHLevelSuc`, and my workaround at the time was to simply rig the definitions of `ℕ→ℕ₋₂`, `2+_`, `-2+_`, etc. in a way that ensured `isSphereFilled→isOfHLevelSuc` typechecked! – if I recall correctly basically any modification of those definitions, or a change of index, would've required `subst`'s in the type of `isSphereFilled→isOfHLevelSuc`.

This PR redefines `ℕ₋₁` and `ℕ₋₂` using records with eta-equality, which ensures lots of simple identities will now hold definitionally. There are now a number of different ways to give the type of `isSphereFilled→isOfHLevelSuc`, I chose one which indexes by `ℕ`. Finally, this PR also adds `ℕ₊₁` in this same style.


Saizan:
Looks good to me. I guess this would also make it less painful to switch our HLevel numbering to -2 based indexing?

m-yac:
> I guess this would also make it less painful to switch our HLevel numbering to -2 based indexing?

Maybe! I am worried though about the fact that there are a number of proofs in Foundations.HLevels that look like:
```agda
isOfHLevelX : (n : ℕ) → ...
isOfHLevelX zero = ...
isOfHLevelX (suc n) = ... isOfHLevelX n
```
which after refactoring to `ℕ₋₂`, due to it's non-inductive nature, would become harder to read:
```agda
isTruncX : (n : ℕ₋₂) → ...
isTruncX neg2 = ...
isTruncX (-1+ n) = ... isOfHLevelX (-2+ n)
```
(Note that `n` has type `ℕ` here, not `ℕ₋₂`.)

For someone new to the library, it might be more confusing to have to frequently deal with proofs that look like the latter than it would be to learn a new indexing scheme.

Saizan:
I making ```suc₋₂``` a pattern synonym could get around the awkwardness there:
```agda
pattern  suc₋₂ n = (-2+ (suc n))
```

Anyhow, we can discuss this in a separate issue.

@mortberg are there PRs we should merge before this?

mortberg:
> I making `suc₋₂` a pattern synonym could get around the awkwardness there:
> 
> ```agda
> pattern  suc₋₂ n = (-2+ (suc n))
> ```
> 
> Anyhow, we can discuss this in a separate issue.
> 
> @mortberg are there PRs we should merge before this?

Yeah, let's merge the cohomology PR first. I'll take a look at it now. Probably this PR will need to be updated, but that should be straightforward

m-yac:
> I making `suc₋₂` a pattern synonym could get around the awkwardness there:
> 
> ```agda
> pattern  suc₋₂ n = (-2+ (suc n))
> ```

Hmm I might be missing something obvious here, but to me this `suc₋₂` has the wrong type: `ℕ → ℕ₋₂` not `ℕ₋₂ → ℕ₋₂`. Regardless, this pattern is actually identical to the pattern `-1+_` I used above, so I'm not sure how this renaming would make the inductive step clearer:
```agda
isTruncX (suc₋₂ n) = ... isTruncX (-2+ n) -- (n : ℕ)
```

I agree this PR is not ready to be merged until some conclusions are reached about naming.

Saizan:
You are right, pattern synonyms are not expressive enough to define an actual `suc₋₂ : ℕ₋₂ → ℕ₋₂` pattern, we would have to use a "view" instead.

mortberg:
I find these issues with various naturals starting at different numbers really annoying. I think the best way to fix this is to just use one numbering and not mix hlevels and truncation levels. So either we use  ℕ₋₂ everywhere or ℕ everywhere. I'm in favor of using ℕ as it is kind of silly and annoying to redevelop operations and theory about ℕ for ℕ₋₂. To avoid confusion we can then call things "h-truncated" and "h-connected" to make it clear that the counting differs by 2 compared to what's in the HoTT book. What do you think?

Saizan:
I do agree that

- We should stick to just one indexing scheme everywhere
- It would be nice if it was 0-based so that we can just use ℕ.

However I think I've seen plenty of people use "homotopy truncated", "h-level", and so on.. while counting from -2. So it wouldn't really get around the name clash to use `h-truncated`.

Also, are we going to rename set truncation to `∥_∥₂` ? Otherwise we are inconsistent again.

@m-yac way of implementing `ℕ₋₂` kinda makes starting from -2 seem the less painful option to me.

Saizan:
I suppose set truncation could just be `∥_∥set` and the documentation for `isOfHLevel` (with that name or a new one) can explain the numbering scheme.

mortberg:
The h-levels are due to Voevodsky and start at 0 (see his old Foundations Coq code).

For the direct HIT definitions I think we can use ∥_∥ for propositional truncation, setTrunc for set truncation, gpdTrunc for groupoid trunc, etc. For the general n-truncation HIT we then use `hLevelTrunc` to avoid confusion.

Saizan:
Ok that works as long as we include `level` in the wording. 

The `h-` prefix in `h-truncated` and `h-connected` wouldn't achieve the disambiguation we'd like.

mortberg:
We don't need h-truncated as we have isOfHLevel. For connected we can use "hLevel-connected" or something just to indicate that we are using hLevels for the index.

m-yac:
I'm convinced! My only real concern with this plan is that we would be further distancing ourselves from the notation used in the literature, especially with `hLevelTrunc A : ℕ → Type` instead of `∥ A ∥_ : ℕ₋₂ → Type`, which will create some visual dissonance between between versions of results proved here and elsewhere. It will also add unexpected +2's in places like cohomology definitions. But, I think the hassle of `ℕ₋₂` is just not worth it.

I'm happy to make these changes (i.e. renaming/retyping the various truncations). We would presumably still keep `ℕ₋₁`, as it's needed for `S -1` (used in the definition `hLevelTrunc A 0`), and I don't think it makes sense to re-index spheres. Also, the way this type is used I don't think it causes nearly as many problems.

Should I update this PR and change the title, or close this and make a new one?

mortberg:
Can't we just keep `∥ A ∥_ : ℕ₋₂ → Type` and define it using hLevelTrunc? It might be nice to use hLevelTrunc under the hood just to not get bogged down by ℕ₋₂ when developing the basic theory (recursor, eliminator, etc.), but we can then also expose `∥ A ∥_ : ℕ₋₂ → Type` with its usual recursor and eliminator. That way the user can use either `hLevelTrunc` or `∥ _ ∥` depending on their needs (for instance in the definition of cohomology it might be less annoying to use `hLevelTrunc`).

This might be a generally good design for the library: work internally with `ℕ` but also expose standard interfaces with `ℕ₋₂`?

I don't think we should close this PR, but rather merge it before switching to ℕ in more places. Having ℕ₋₁ and ℕ₋₂ more usable is still worthwhile, especially if we do what I suggest above.

Saizan:
I wonder if still having `ℕ₋₂` around would cause unnecessary friction. 
We could have `∥ A ∥lvl_ : ℕ -> Type` and be close enough to the notation from the book.

However I'm fine with both.

m-yac:
@mortberg oh, that makes sense!

> This might be a generally good design for the library: work internally with ℕ but also expose standard interfaces with ℕ₋₂?

I think this is a great idea.

mortberg:
Should I go ahead and merge this now then? The other changes we discussed can be done in another PR

m-yac:
Yep, I think this is ready to merge.

PR #240, opened 2020-02-25
Clean up and make naming consistent in Foundations.HLevels

m-yac:
As requested by @mortberg, this PR refactors the naming of the lemmas in Foundations.HLevels to match the form `isOfHLevelX`. I also shuffled around the order of the lemmas in that file to (hopefully) make it more readable and made a few small style changes for consistency.

Along the way, I also:
- fixed the ad-hoc inlining of +-comm in Foundations.HLevels by removing the unused import of Data.Nat.Algebra from Data.Nat.Properties
- added Data.Nat.Algebra to Data.Everything
- moved `isPropIsProp` to Foundations.Prelude
- added `hLevelPi⁻` (been meaning to do this for a while)
- generalized `isProp→isPropPathP` to `isOfHLevelPathP'` and `isOfHLevelPathP`
- added `ΣProp≡-equiv` and used it in place of (the old) `isOfHLevel≡`
- replaced unnecessary use of of `isProp→isContrPathP` in `isInjectiveTransport`
- moved `isInjectiveTransport` to Foundations.Transport because there is no longer a cyclic dependency
- removed Data.Universe, as it is now empty
- added `isProp→isOfHLevelSuc` to better reflect how `isOfHLevelPlus` is actually used
- added `isContr→isOfHLevel` for completeness, and used it to prove `isOfHLevelUnit`


mortberg:
Merged this now. Thanks a lot @m-yac !

PR #243, opened 2020-02-27
Eliminator naming conventions

### LINKED ISSUE #241: naming conventions in HIT/Data properties
Some Data/HITs Properties files contain definitions like `elim`, `ind`, `comm`, etc. that are meant to  used in a qualified import or with `renaming`, while others have names like `elimSetProp`, etc. that incorporate the module name. Thoughts on which the library should prefer? I'm asking because I want to make the names in the various `Truncation` modules consistent with each other.

ecavallo:
In progress, to resolve #241.


mortberg:
This is really nice! After merging the HLevels file there are a bunch of conflicts... 

Please ping me when you consider this PR to be ready.

ecavallo:
@mortberg Done!

m-yac:
This is awesome!

What are your thoughts about using unicode names for qualified imports? e.g. importing `Data.Nat` as `ℕ` giving `ℕ.elim`, `Data.Empty` as `⊥` giving `⊥.elim`, etc. Right now it looks like you chose to always go for full names (e.g. giving `Nat.elim` and `Empty.elim`). For the two examples I gave I think unicode looks better, but that definitely doesn't always apply. I think we should probably decide on a convention for this and write it down somewhere.

More generally, I think the conclusions reached in #241 about rec/elim, omitting redundant module names, and qualifying imports should all be written down somewhere. Maybe in the main README under a 'conventions' header?

mortberg:
> This is awesome!
> 
> What are your thoughts about using unicode names for qualified imports? e.g. importing `Data.Nat` as `ℕ` giving `ℕ.elim`, `Data.Empty` as `⊥` giving `⊥.elim`, etc. Right now it looks like you chose to always go for full names (e.g. giving `Nat.elim` and `Empty.elim`). For the two examples I gave I think unicode looks better, but that definitely doesn't always apply. I think we should probably decide on a convention for this and write it down somewhere.
> 
> More generally, I think the conclusions reached in #241 about rec/elim, omitting redundant module names, and qualifying imports should all be written down somewhere. Maybe in the main README under a 'conventions' header?

Having unicode names for these does indeed seem nice. 

Yes, this might be the beginning of the long awaited styleguide for the library! Feel free to make a PR with it

ecavallo:
Is it a policy that `Cubical.HITs.X` should import all the modules in `Cubical/HITs/X/`? This is not always the case currently, and I think some files are being missed in `make` as a result.

mortberg:
> Is it a policy that `Cubical.HITs.X` should import all the modules in `Cubical/HITs/X/`? This is not always the case currently, and I think some files are being missed in `make` as a result.

Yes, can you fix this as well? Everything should be in the Everything files as well...

ecavallo:
Trying to fix it is creating some cyclic module dependencies... I'll make sure everything is in `Everything` and leave the rest as it is for this PR.

m-yac:
> Trying to fix it is creating some cyclic module dependencies... I'll make sure everything is in `Everything` and leave the rest as it is for this PR.

I've noticed this as well, maybe importing but not opening is a good solution?

ecavallo:
I'm ready to merge now.

I think there are some larger questions about the purpose of the modules in `Data` and `HITs`. These could work in different ways:
1. `Data.X` is intended to contain all possible theorems about `X`. In that case, importing the entirety of `Data.X` elsewhere (e.g. importing `Data.Nat` in `Foundations.HLevels`) is bound to lead to cyclic dependency problems, so files should only import the specific submodules they need. (As an egregious example, consider that `HITs.S2` could end up depending on all sorts of math.)
2. `Data.X` only contains basic theorems about `X`.
a. More advanced theorems are also in the `Data/X/` directory, but not imported by `Data.X`.
b. More advanced theorems get categorized in new top-level modules like `ZCohomology`, etc.

Right now the organization is a mix of these.

Saizan:
I would say that `2` is more reasonable, we still want `Data.X` to be useful to import.

mortberg:
Sigh, conflicts again. Can you please fix and then I'll merge this? 

Maybe the discussions should be moved into Issues instead so that they don't get lost when I merge the PR.

mortberg:
Thanks! Merge when Travis is done?

ecavallo:
Yep!

mortberg:
Thanks a lot for this contribution!

PR #247, opened 2020-03-11
Equivalence of `SumFin` and `Fin`

rei1024:
* Add computation rules for the eliminator of `Fin`

* Add the eliminator for `SumFin.Fin`

* Prove equivalence of `SumFin` and `Fin`


mortberg:
Thanks for the contribution! Everything looks good to me so I'll merge.

PR #248, opened 2020-03-12
Rm sns

mzeuner:
Simplifying the Foundatioins.SIP file by reducing the number of different versions of standard notion of structure (SNS).

### DIFF
--- a/Cubical/Foundations/SIP.agda
+++ b/Cubical/Foundations/SIP.agda
@@ -48,7 +48,7 @@ ua-pathToEquiv e = uaHAEquiv _ _ .snd .ret e
 -- as S-structures. This we call a standard notion of structure or SNS.
 -- We will use a different definition, but the two definitions are interchangeable.
 SNS-≡ : (S : Type ℓ₁ → Type ℓ₂) (ι : StrIso S ℓ₃) → Type (ℓ-max (ℓ-max (ℓ-suc ℓ₁) ℓ₂) ℓ₃)
-SNS-≡ {ℓ₁} S ι = ∀ {X : Type ℓ₁} (s t : S X) → ((s ≡ t) ≃ ι (X , s) (X , t) (idEquiv X))
+SNS-≡ {ℓ₁} S ι = ∀ {X : Type ℓ₁} (s t : S X) → (ι (X , s) (X , t) (idEquiv X) ≃ (s ≡ t))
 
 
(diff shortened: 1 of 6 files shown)


Cubical/Foundations/SIP.agda
```diff
@@ -24,115 +24,93 @@ open import Cubical.Foundations.Structure public
...
-SNS₃ {ℓ} S ι = (A B : TypeWithStr ℓ S) (e : typ A ≃ typ B)
-             → (transport (λ i → S (ua e i)) (str A) ≡ (str B)) ≃ (ι A B e)
-
-SNS₂≡SNS₃ : (S : Type ℓ → Type ℓ') (ι : StrIso S ℓ'') → SNS₂ S ι ≡ SNS₃ S ι
-SNS₂≡SNS₃ S ι = refl
+-- The following PathP version of SNS₁ is a bit easier to
+-- work with for the proof of the SIP
+SNS₂ : (S : Type ℓ₁ → Type ℓ₂) (ι : StrIso S ℓ₃) → Type (ℓ-max (ℓ-max (ℓ-suc ℓ₁) ℓ₂) ℓ₃)
+SNS₂ {ℓ₁} S ι = (A B : TypeWithStr ℓ₁ S) (e : typ A ≃ typ B)
+             → (PathP (λ i → S (ua e i)) (str A) (str B)) ≃ (ι A B e)
```
mortberg: Can you flip the equivalence in the conclusion? They way we don't need the `invEquiv` in `sip` and the formulation is more natural. It might make sense to do the same in `SNS₁`.

mzeuner: I changed the SNS-names, but I haven't changed the order of the equivalence, as this would break a few things in the proof of the SIP and files in Cubical.Structures. I agree that it would be more natural, but I'm not sure if that is worth the effort. Maybe I could do that in a separate PR?

mortberg: I think this is a good time before we write even more code that relies on this definition. So it would be great if you could do it in this PR.

Shouldn't it be straightforward to implement? You just drop some `invEq` and flip the order in some proofs using equivalence reasoning with `_≃⟨_⟩_`.

mzeuner: You're right, better do it now. I hope I can do it today.

PR #249, opened 2020-03-14
Simplify some proofs related to the SIP

mortberg:
- Get rid of `Σ≡` in the proof of SIP
- More direct proof of join-SNS
- Upstream code and add more implicit arguments here and there

### DIFF
--- a/Cubical/Foundations/FunExtEquiv.agda
+++ b/Cubical/Foundations/FunExtEquiv.agda
@@ -1,24 +1,24 @@
 {-# OPTIONS --cubical --safe #-}
-
 module Cubical.Foundations.FunExtEquiv where
 
 open import Cubical.Foundations.Prelude
 open import Cubical.Foundations.Equiv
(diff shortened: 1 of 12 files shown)


Cubical/Foundations/FunExtEquiv.agda
```diff
@@ -1,28 +1,91 @@
...
+  equiv-proof funExt₃-isEquiv p = (fib p , funExt₃-fiber-isContr p)
+
+  funExt₃Equiv : (∀ x y z → f x y z ≡ g x y z) ≃ (f ≡ g)
+  funExt₃Equiv = (funExt₃ , funExt₃-isEquiv)
+
+  funExt₃Path : (∀ x y z → f x y z ≡ g x y z) ≡ (f ≡ g)
+  funExt₃Path = ua funExt₃Equiv
+
+
+-- Puzzle: Can one generalize this to n-ary functions?
```
mortberg: If some wants a puzzle take a look at this!

PR #251, opened 2020-03-19
Association Lists

mzeuner:
We define Association Lists using HITs and show that they're equivalent  to Finite Multi Sets


mortberg:
Thanks for the contribution @mzeuner !

PR #252, opened 2020-03-21
Real projective spaces

m-yac:
This PR defines the real projective spaces following the paper:

- U. Buchholtz, E. Rijke, The real projective spaces in homotopy type theory. (2017) https://arxiv.org/abs/1704.05770

The core addition of this PR is the file `Cubical.HITs.RPn.Base`, which defines `RP n`, it's double cover, and proves that the total space of this double cover is `S n`. It also adds: 
- `HITs.Pushout.Flattening`: the flattening lemma for pushouts, used in `HITs.RPn`. This file ended up being a pretty good example of using `glue`/`unglue`, as well as how to deal with terms of the form `f (hcomp ...)` – the latter is something I've struggled with in the past.
- in `Foundations.Univalence`: since they keep coming up for me more and more (e.g. in `HITs.Pushout.Flattening`) I gave `ua-unglue` and `ua-glue` easier types to use and added a few other nice definitions about / related to these functions
- `Foundations.Bundle` and `Structures.TypeEqvTo`
- lots of small lemmas in lots of different files (sorry @mortberg !)

Eventually someone (possibly me) should add the long exact fiber sequence, from which computing the homotopy groups for `RP n` follows easily.

### DIFF
--- a/Cubical/Data/Prod/Properties.agda
+++ b/Cubical/Data/Prod/Properties.agda
@@ -48,11 +48,17 @@ private
   testrefl = refl
 
 -- equivalence between the sigma-based definition and the inductive one
-A×B≡A×ΣB : A × B ≡ A ×Σ B
-A×B≡A×ΣB = isoToPath (iso (λ { (a , b) → (a , b)})
+A×B≃A×ΣB : A × B ≃ A ×Σ B
+A×B≃A×ΣB = isoToEquiv (iso (λ { (a , b) → (a , b)})
                           (λ { (a , b) → (a , b)})
                           (λ _ → refl)
                           (λ { (a , b) → refl }))
+
+A×B≡A×ΣB : A × B ≡ A ×Σ B
+A×B≡A×ΣB = ua A×B≃A×ΣB
+
+swapΣEquiv : (A : Type ℓ) (B : Type ℓ') → A ×Σ B ≃ B ×Σ A
+swapΣEquiv A B = compEquiv (compEquiv (invEquiv A×B≃A×ΣB) (swapEquiv A B)) A×B≃A×ΣB
 
 -- truncation for products
 isOfHLevelProd : (n : ℕ) → isOfHLevel n A → isOfHLevel n B → isOfHLevel n (A × B)
--- a/Cubical/Foundations/Bundle.agda
+++ b/Cubical/Foundations/Bundle.agda
@@ -1,0 +1,26 @@
+{-# OPTIONS --cubical --safe #-}
+module Cubical.Foundations.Bundle where
+
+open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.Equiv
+open import Cubical.Foundations.Structure
+
+open import Cubical.Structures.TypeEqvTo
+
+module _ {ℓb ℓf} {B : Type ℓb} {F : Type ℓf} {ℓ} where
+
+{-
+    A fiber bundle with base space B and fibers F is a map `p⁻¹ : B → TypeEqvTo F`
+     taking points in the base space to their respective fibers.
+
+    e.g. a double cover is a map `B → TypeEqvTo Bool`
+-}
+
+  Total : (p⁻¹ : B → TypeEqvTo ℓ F) → Type (ℓ-max ℓb ℓ)
+  Total p⁻¹ = Σ[ x ∈ B ] p⁻¹ x .fst
+
+  pr : (p⁻¹ : B → TypeEqvTo ℓ F) → Total p⁻¹ → B
+  pr p⁻¹ = fst
+
+  inc : (p⁻¹ : B → TypeEqvTo ℓ F) (x : B) → p⁻¹ x .fst → Total p⁻¹
+  inc p⁻¹ x = (x ,_)
--- a/Cubical/Foundations/Bundle.agda
+++ b/Cubical/Foundations/Bundle.agda
@@ -2,7 +2,9 @@
 module Cubical.Foundations.Bundle where
 
 open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.Isomorphism
 open import Cubical.Foundations.Equiv
+open import Cubical.Foundations.Transport
 open import Cubical.Foundations.Structure
 
 open import Cubical.Structures.TypeEqvTo
@@ -24,3 +26,30 @@ module _ {ℓb ℓf} {B : Type ℓb} {F : Type ℓf} {ℓ} where
 
   inc : (p⁻¹ : B → TypeEqvTo ℓ F) (x : B) → p⁻¹ x .fst → Total p⁻¹
   inc p⁻¹ x = (x ,_)
+
+  module FibPrIso (p⁻¹ : B → TypeEqvTo ℓ F) (x : B) where
+
+    fwd : fiber (pr p⁻¹) x → typ (p⁻¹ x)
+    fwd ((x' , y) , p) = subst (λ z → typ (p⁻¹ z)) p y
+
+    bwd : typ (p⁻¹ x) → fiber (pr p⁻¹) x
+    bwd y = (x , y) , refl
+
+    fwd-bwd : ∀ x → fwd (bwd x) ≡ x
+    fwd-bwd y = transportRefl y
+
+    bwd-fwd : ∀ x → bwd (fwd x) ≡ x
+    bwd-fwd ((x' , y) , p) i = h (q i)
+      where h : Σ[ s ∈ singl x ] typ (p⁻¹ (s .fst)) → fiber (pr p⁻¹) x
+            h ((x , p) , y) = (x , y) , sym p
+            q : Path (Σ[ s ∈ singl x ] typ (p⁻¹ (s .fst)))
+                     ((x  , refl ) , subst (λ z → typ (p⁻¹ z)) p y)
+                     ((x' , sym p) , y                            )
+            q i = (contrSingl (sym p) i) , toPathP {A = λ i → typ (p⁻¹ (p (~ i)))}
+                                           (transport⁻Transport (λ i → typ (p⁻¹ (p i))) y) i
+
+    isom : Iso (fiber (pr p⁻¹) x) (typ (p⁻¹ x))
+    isom = iso fwd bwd fwd-bwd bwd-fwd
+
+  fibPrEquiv : (p⁻¹ : B → TypeEqvTo ℓ F) (x : B) → fiber (pr p⁻¹) x ≃ p⁻¹ x .fst
+  fibPrEquiv p⁻¹ x = isoToEquiv (FibPrIso.isom p⁻¹ x)
--- a/Cubical/Foundations/Univalence.agda
+++ b/Cubical/Foundations/Univalence.agda
@@ -36,18 +36,18 @@ ua {A = A} {B = B} e i = Glue B (λ { (i = i0) → (A , e)
 uaIdEquiv : {A : Type ℓ} → ua (idEquiv A) ≡ refl
 uaIdEquiv {A = A} i j = Glue A {φ = i ∨ ~ j ∨ j} (λ _ → A , idEquiv A)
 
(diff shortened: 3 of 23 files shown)


Cubical/Structures/Pointed.agda
```diff
@@ -24,6 +24,10 @@ pointed-iso : StrIso pointed-structure ℓ
 pointed-iso A B f = equivFun f (pt A) ≡ pt B
 
 pointed-is-SNS : SNS {ℓ} pointed-structure pointed-iso
-pointed-is-SNS {A = A} {B = B} f =
-  transportEquiv ( (λ j → transportRefl (equivFun f (pt A)) (~ j) ≡ pt B)
-                 ∙ sym (PathP≡Path (λ i → ua f i) (pt A) (pt B)))
+pointed-is-SNS f = invEquiv (ua-ungluePath-Equiv f)
```
m-yac: Simplified this slightly using glue types, hope that's okay.

In fact, I suspect that the proof of SIP could similarly be simplified (in the sense of removing `transport`'s) using glue/unglue. I would be happy to give it a try if it doesn't conflict with plans @mzeuner or @mortberg have.

Cubical/Structures/Pointed.agda
```diff
@@ -24,6 +24,10 @@ pointed-iso : StrIso pointed-structure ℓ
...
+
+pointed-SIP : (A B : Pointed ℓ) → A ≃[ pointed-iso ] B ≃ (A ≡ B)
+pointed-SIP = SIP pointed-structure pointed-iso pointed-is-SNS
+
+pointed-sip : (A B : Pointed ℓ) → A ≃[ pointed-iso ] B → (A ≡ B)
+pointed-sip A B = equivFun (pointed-SIP A B) -- ≡ λ (e , p) i → ua e i , ua-gluePath e p i
```
m-yac: This has a nice form now using `ua-ungluePath-Equiv` to prove `pointed-is-SNS`.

mortberg:
Wow, amazing! Will give detailed feedback on the code soon.

How far did you get in the paper? I think there were some parts where Lean was running out of memory when typechecking, did you get to these?

It would be good to have comments for each major result to where they can be found in the paper.

Cubical/Data/Prod/Properties.agda
```diff
@@ -48,12 +48,18 @@ private
...
 -- equivalence between the sigma-based definition and the inductive one
-A×B≡A×ΣB : A × B ≡ A ×Σ B
-A×B≡A×ΣB = isoToPath (iso (λ { (a , b) → (a , b)})
+A×B≃A×ΣB : A × B ≃ A ×Σ B
+A×B≃A×ΣB = isoToEquiv (iso (λ { (a , b) → (a , b)})
                           (λ { (a , b) → (a , b)})
                           (λ _ → refl)
                           (λ { (a , b) → refl }))
 
+A×B≡A×ΣB : A × B ≡ A ×Σ B
```
mortberg: Btw, what is the point of having both of these definitions of products around? Having products defined using Sigma seems like the better option as we can easier inherit a lot of theory.

Cubical/Foundations/Bundle.agda
```diff
@@ -0,0 +1,55 @@
+{-# OPTIONS --cubical --safe #-}
+module Cubical.Foundations.Bundle where
```
mortberg: I'm starting to wonder what really belongs in Foundations... Could this go somewhere else, maybe in a new directory? I would like to move a lot of the content from Foundations just to keep it minimal. I think files should only be in Foundations if they are used in almost all other files in the library

Cubical/Foundations/Univalence.agda
```diff
@@ -119,6 +166,11 @@ EquivJ : (P : (A B : Type ℓ) → (e : B ≃ A) → Type ℓ')
        → (A B : Type ℓ) → (e : B ≃ A) → P A B e
 EquivJ P r A B e = subst (λ x → P A (x .fst) (x .snd)) (contrSinglEquiv e) (r A)
 
+-- Based equivalence induction
+EquivJ' : {A B : Type ℓ} (P : (B : Type ℓ) → (e : B ≃ A) → Type ℓ')
+        → (r : P A (idEquiv A)) → (e : B ≃ A) → P B e
+EquivJ' P r e = subst (λ x → P (x .fst) (x .snd)) (contrSinglEquiv e) r
```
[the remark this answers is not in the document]
mortberg: I find the type of EquivJ very unintuitive... I think I proved it in cubical around 2014 and I have no clue why I did it that way. 

Maybe we should drop EquivJ in favor for EquivJ'? (in another PR as lots of files will have to change)

Cubical/HITs/Pushout/Flattening.agda
```diff
@@ -0,0 +1,91 @@
+{-
+
+   The flattening lemma for pushouts (Lemma 8.5.3 in the HoTT book) proved in a cubical style.
+
+   The proof in the HoTT book (the core lying in Lemma 6.12.2, the flattening lemma for coequalizers)
+    consists mostly of long strings of equalities about transport. This proof follows almost
+    entirely from definitional equalities involving glue/unglue.
```
mortberg: This is really cool!

Cubical/HITs/Pushout/Flattening.agda
```diff
@@ -0,0 +1,91 @@
...
+      comp (λ _ → Σ (Pushout f g) E)
+           (λ k → λ { (i = i0) → push a (~ k) , ua-gluePt (e a) (~ k) x
+                    ; (i = i1) → inr (g a) , x
+                    ; (j = i1) → push a (i ∨ ~ k) , sq (e a) i k x })
+            (inr (g a) , ua-unglue (e a) i x)
+
+    isom : Iso (Σ (Pushout f g) E) (Pushout Σf Σg)
+    isom = iso bwd fwd bwd-fwd fwd-bwd
+
+  flatten : Σ (Pushout f g) E ≃ Pushout Σf Σg
```
mortberg: I guess there are proofs of this in the various HoTT libraries. How do they compare in length?

Cubical/HITs/RPn/Base.agda
```diff
@@ -0,0 +1,308 @@
...
+  -- we've already shown that this map is an equivalence on the right
+
+  isEquivʳ : (y : typ X) → isEquiv (y ⊕*_)
+  isEquivʳ y = invEquiv (fst (fst (isContr-2-EltPointed-iso (fst X , y , snd X)))) .snd
+
+  Equivʳ : typ X → typ X ≃ Bool
+  Equivʳ y = (y ⊕*_) , isEquivʳ y
+
+  -- any mere proposition that holds for (Bool, _⊕_) holds for (typ X, _⊕*_)
+  -- this ammounts to just carefully unfolding the PropTrunc.elim and J in isContr-2-EltPointed-iso
```
mortberg: "ammounts"

Cubical/HITs/RPn/Base.agda
```diff
@@ -0,0 +1,308 @@
...
+          S (ℕ→ℕ₋₁ n)                      ■
+
+-- the usual covering map S n → RP n, with fibers exactly cov⁻¹
+
+cov : (n : ℕ₋₁) → S n → RP n
+cov n = pr (cov⁻¹ n) ∘ invEq (TotalCov≃Sn n)
+
+fibcov≡cov⁻¹ : ∀ n (x : RP n) → fiber (cov n) x ≡ cov⁻¹ n x .fst
+fibcov≡cov⁻¹ n x =
+  fiber (cov n)        x ≡[ i ]⟨ fiber {A = ua e i} (pr (cov⁻¹ n) ∘ ua-unglue e i) x ⟩
```
mortberg: I really like this equality reasoning over `i`

Cubical/Structures/Pointed.agda
mortberg: That would be cool! I don't think @mzeuner or I are planning to do anything about the SIP proofs at the moment. Please do it in a new PR if you succeed.

Cubical/Structures/Pointed.agda
mortberg: Yeah, it's cool how those lemmas simplify these proofs!

m-yac:
> I think there were some parts where Lean was running out of memory when typechecking, did you get to these?

I did! In the nomenclature of this PR, I believe they ran into trouble using `isContr-2-EltPointed-iso` to prove `⊕*.isEquivˡ`. This is the slowest part of the file to typecheck, but still only takes a second and a half on my laptop. The only parts of the paper I didn't implement are Corollary III.6 (needs the long exact fiber sequence) and Section IV (needs sequential colimits).

As for your question about the flattening lemma in other HoTT libraries, I couldn't find it in UniMath, but [HoTT-Agda's proof](https://github.com/HoTT/HoTT-Agda/blob/e48a16bcd719e2fcb409d79e3f7df6c6b81223bb/core/lib/types/Flattening.agda#L62) is quite long (there's actually even more pushout-specific code [here](https://github.com/HoTT/HoTT-Agda/blob/e48a16bcd719e2fcb409d79e3f7df6c6b81223bb/core/lib/types/PushoutFlattening.agda)). I also found [a proof in lean2](https://github.com/leanprover/lean2/blob/14c8fbfea30d217657f40c0cf91197cc4f1770b6/hott/hit/pushout.hlean#L169), but I don't know how to compare it's length since I'm not familar with lean.

Cubical/Data/Prod/Properties.agda
m-yac: I tend to agree, this is what `Data.Prod.Base` says:
```agda
-- If × is defined using Σ then transp/hcomp will be compute
-- "negatively", that is, they won't reduce unless we project out the
-- first of second component. This is not always what we want so the
-- default implementation is done using a datatype which computes
-- positively.
```
I have no experience with this so I can't speak to how important it is.

Cubical/Foundations/Bundle.agda
m-yac: I very much agree! I don't know what this other directory should be called, though. `Cubical.Theories`? If we want to make a less drastic change we could make a `Cubical.Maps` or `Cubical.Functions` directory and just move `Bundle`, `TotalFiber`, `Surjection`, and `Embedding`, but I think a directory like you're describing will have to be made eventually.

Cubical/HITs/Pushout/Flattening.agda
m-yac: Oh, I answered this in a top-level comment.

Cubical/Data/Prod/Properties.agda
mortberg: Oh, right... Sometimes I write some good comments! I just have to remember to read them...

Cubical/Foundations/Bundle.agda
mortberg: Cubical.Functions sounds pretty good for the things you mention. There might be some other things that could go there, e.g. FunExt.

Make an issue and fix in a future PR?

mortberg:
> > I think there were some parts where Lean was running out of memory when typechecking, did you get to these?
> 
> I did! In the nomenclature of this PR, I believe they ran into trouble using `isContr-2-EltPointed-iso` to prove `⊕*.isEquivˡ`. This is the slowest part of the file to typecheck, but still only takes a second and a half on my laptop. The only parts of the paper I didn't implement are Corollary III.6 (needs the long exact fiber sequence) and Section IV (needs sequential colimits).
> 

Very cool! This is an interesting observation. I'm surprised that Lean was so slow for this... Maybe @UlrikBuchholtz or @EgbertRijke have some interesting comments about this?

mortberg:
I'd be up for merging as soon as you've fixed the conflicts and made an issue about moving some files out of Foundations. Please ping me when this is done.

mortberg:
What does "(trying to make TotalCov≃Sn compute)" mean?

mortberg:
Thanks a lot for this contribution!

PR #253, opened 2020-03-23
Re-index n-truncation

### LINKED ISSUE #211: hlevel indexing
This was discussed a bit in #155. Currently the library indexes hlevels from contractible=0 but truncations (∥_∥,∥_∥₀,∥_∥₁) from contractible=-2, which can be confusing. Given that different parts of the community have different conventions for these things, I'm partial to the approach we tried in [redtt](https://github.com/RedPRL/redtt/blob/master/library/prelude/hlevel.red), which was to use names for hlevels and avoid committing to some mapping onto integers.
```
data hLevel : Type₀ where
  contr : hLevel
  suc : hLevel → hLevel

pattern prop = suc contr
pattern set = suc prop

_-gpd : ℕ → hLevel
zero -gpd = set
suc n -gpd = suc (n -gpd)

...
```
The disadvantages of introducing a new type (as opposed to using ℕ as we are now) are that (a) we'll have to introduce conversions between `ℕ` and `hLevel` in various places, (b) we may have to re-prove various arithmetical facts about adding hlevels together and so on. But I think both of these lead to more readable code in the long term, and (b) is mostly an upfront cost.

Thoughts?

m-yac:
As discussed in #238, use `hLevelTrunc : ℕ → Type ℓ → Type ℓ` for n-truncation, but still expose `∥ A ∥_ : ℕ₋₂ → Type`. This PR also fixes a weird importing issue with natural number and negative integer literals for `ℕ₋₁`, etc.

Resolves #211

### DIFF
--- a/Cubical/Foundations/HLevels.agda
+++ b/Cubical/Foundations/HLevels.agda
@@ -24,6 +24,7 @@ open import Cubical.Foundations.Univalence   using (ua; univalence)
 
 open import Cubical.Data.Sigma using (ΣPathP; sigmaPath→pathSigma; pathSigma≡sigmaPath; _Σ≡T_)
 open import Cubical.Data.Nat   using (ℕ; zero; suc; _+_; +-zero; +-comm)
+open import Cubical.Data.NatMinusTwo
 
 private
   variable
@@ -37,6 +38,11 @@ isOfHLevel : ℕ → Type ℓ → Type ℓ
 isOfHLevel 0 A = isContr A
 isOfHLevel 1 A = isProp A
 isOfHLevel (suc (suc n)) A = (x y : A) → isOfHLevel (suc n) (x ≡ y)
+
+-- a re-indexing of `isOfHLevel` which is consistent with standard notation,
+--  in the HoTT book this is `is-n-type(X)`
+isTrunc : ℕ₋₂ → Type ℓ → Type ℓ
+isTrunc n = isOfHLevel (2+ n)
 
 HLevel : ∀ ℓ → ℕ → Type (ℓ-suc ℓ)
 HLevel ℓ n = TypeWithStr ℓ (isOfHLevel n)
--- a/Cubical/Foundations/HLevels.agda
+++ b/Cubical/Foundations/HLevels.agda
@@ -24,7 +24,6 @@ open import Cubical.Foundations.Univalence   using (ua; univalence)
 
 open import Cubical.Data.Sigma using (ΣPathP; sigmaPath→pathSigma; pathSigma≡sigmaPath; _Σ≡T_)
 open import Cubical.Data.Nat   using (ℕ; zero; suc; _+_; +-zero; +-comm)
-open import Cubical.Data.NatMinusTwo
 
 private
   variable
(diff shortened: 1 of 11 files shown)


Cubical/Foundations/HLevels.agda
```diff
@@ -38,6 +39,11 @@ isOfHLevel 0 A = isContr A
 isOfHLevel 1 A = isProp A
 isOfHLevel (suc (suc n)) A = (x y : A) → isOfHLevel (suc n) (x ≡ y)
 
+-- a re-indexing of `isOfHLevel` which is consistent with standard notation,
+--  in the HoTT book this is `is-n-type(X)`
+isTrunc : ℕ₋₂ → Type ℓ → Type ℓ
+isTrunc n = isOfHLevel (2+ n)
```
m-yac: I'm also unsure about the inclusion of this definition. It is nice to have in conjunction with `∥_∥_` (and in the future, `isConn`, `isConnMap`, `isTruncMap`), but every fact about it is currently stated in terms of h-levels. Should we therefore have a file that consists solely of re-indexed versions of theorems in `HLevels.agda`? (e.g. `isTruncSuc`, `isTruncPath`, `isTruncPi`, etc.) This seems excessive, but it is in the spirit of @mortberg's suggestion in #238 to "work internally with ℕ but also expose standard interfaces with `ℕ₋₂`".

As-is, using it is clunky (e.g. in the definition of `decode-fun` in `Truncation.Properties` there are terms like: `isOfHLevelSuc (2+ n) (isTrunc∥∥ n A)`).

mortberg: I don't know... I'm personally fine with working with h-levels, but it's nice to have this definition as documentation to reduce confusion for people who read the HoTT book.

mortberg: And yeah, I don't think we should redo the HLevels file shifted. I was mainly thinking that we could provide the ℕ₋₂ interface for n-truncation. But maybe it will be hard to use because all theory is stated for h-levels?

Cubical/Foundations/HLevels.agda
```diff
@@ -38,6 +39,11 @@ isOfHLevel 0 A = isContr A
 isOfHLevel 1 A = isProp A
 isOfHLevel (suc (suc n)) A = (x y : A) → isOfHLevel (suc n) (x ≡ y)
 
+-- a re-indexing of `isOfHLevel` which is consistent with standard notation,
+--  in the HoTT book this is `is-n-type(X)`
```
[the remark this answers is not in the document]
mortberg: Why not just say "HoTT book terminology" instead of "standard notation"?

Cubical/Foundations/HLevels.agda
m-yac: For now I think I should remove `isTrunc` from this PR and make an issue asking the community about this. `∥_∥_` is nice to have in theory, but mostly unusable without similarly indexed lemmas about hlevels / truncation levels.

mortberg: Sure, makes sense to me

m-yac:
This should be ready to merge!

mortberg:
Thanks a lot @m-yac!

PR #255, opened 2020-03-25
Connected funs/types + Freudenthal Suspension Theorem + some lemmas

aljungstrom:
Hi!

This PR contains definitions of connected functions and types. It also contains some lemmas from chapter 7 and 8 in the HoTT book - in particular, those that are needed for the proof of the Freudenthal Suspension Theorem. They follow the book quite closely, apart from the last bit of path algebra in the proof of Freudenthal which has to be a bit different in Cubical agda due to some equalities not holding definitionally. 

I placed the lemmas leading up the the theorem (including the proof of the theorem) in a separate folder - a lot of it is incredibly long. Because there is no support of cyclic module dependencies, I had to restate some of them in Properties.agda. I felt like this was a better solution than putting ~2000 lines of code in the properties file. Beware of Prelims.agda. It consists entirely of trivial lemmas (many of which already in the library) defined by J to make the final (horrible) path algebra a bit easier.

As I said, my proof follows that in the HoTT book closely, so if anyone is up for a challange, there is definitely room for some cubical improvement in 8-6-4.agda and 8-6-4Code.agda.

Finally, I should say, the proof uses some abstract definitions because I have had some termination issues. I do not know what is causing this. Believe it or not, the problem is not the massive path algebra file 8-6-4.agda, but the definition of CODE in 8-6-4Code.agda. I would love to know if it runs better on someone else's computer (i.e. on a computer that is not from the middle ages).


ecavallo:
I was trying some of this myself, just up to wedge connectivity:
https://github.com/ecavallo/cubical/blob/connected/Cubical/Experiments/Freudenthal.agda
Dunno if any of it is useful to you, I also mostly followed the book.

aljungstrom:
> I was trying some of this myself, just up to wedge connectivity:
> https://github.com/ecavallo/cubical/blob/connected/Cubical/Experiments/Freudenthal.agda
> Dunno if any of it is useful to you, I also mostly followed the book.

This is definitely useful! My lemma 8.6.1 is a bit of a mess. Your proof is far more direct. I will try to incorporate your proofs later, if I may. Perhaps things will compute better then.

ecavallo:
I added an attempt at defining the Code family. Using Glue instead of ua helps to avoid some wrangling with PathP, though I didn't prove the 8.6.11-12 equalities so it might make those harder.

aljungstrom:
Oh wow, did you just finish it??

ecavallo:
Yeah, except for going from connectedness to equivalence of truncations. One change I made that seemed to simplify the argument was to instead first prove that ``merid : A -> north = south`` is connected, then transfer that fact to ``σ`` at the end. That way there's one less `merid` to deal with. And I used a kind of unpleasant trick of redefining `rCancel` in a particular way so that it's easier to compute the result of calling `encode` on `merid`.

aljungstrom:
Yeah, that seems to have worked wonders. Very nice.

One last question, before I surrender to your proof completely. Does it run ok? I noticed you have some stuff in abstract. Is there a reason for this?

ecavallo:
It takes about 2 seconds to check on my machine with abstract, about 15 without.

aljungstrom:
Great! If you make a PR, I will just go ahead and close mine.

ecavallo:
OK. I'm going to wait until #253 is settled.

aljungstrom:
Sounds reasonable:-) Closing now.

PR #245, opened 2020-03-06
Added construction of M-types from Signatures/Containers

cmester0:

PR #256, opened 2020-03-27
[ perfomance ] improved normal forms of isoToPath and compIso

Saizan:
Core functions like `isoToPath` and `compIso` should really use copatterns as much as possible to keep normal forms small.

### DIFF
--- a/Cubical/Foundations/Equiv.agda
+++ b/Cubical/Foundations/Equiv.agda
@@ -22,8 +22,7 @@ open import Cubical.Foundations.Prelude
 open import Cubical.Foundations.Isomorphism
 open import Cubical.Foundations.GroupoidLaws
 
-open import Cubical.Core.Glue public
-  using ( isEquiv ; equiv-proof ; _≃_ ; equivFun ; equivProof )
+open import Cubical.Foundations.Equiv.Base public
 
 private
   variable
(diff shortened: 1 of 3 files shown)


mortberg:
Great! Did you time it with and without these changes? I'm curious how much difference it makes.

Saizan:
We would have to setup specific benchmarks, but if you compare the effect on `shift` from #245 it's quite dramatic: without this change the normal form of `shift ?` is intractably large, while with this change it's 53 lines (still not small I guess, but some duplication is due to `primGlue`) and computed in 52 ms.

It doesn't improve typechecking time of `make check`, but it should make fewer things blow up.

guillaumebrunerie:
Does it matter how big normal forms are? Ideally Agda should only use weak-head reduction, not normalization.
I’m not saying it’s bad to use copatterns, I’m just wondering if the size of normal forms is the right metric.

Saizan:
Sometimes it matters just because you want to see what the normal form of your goal is.

Normalization is not the exact metric we should use, but conversion checking recurses on the subterms after computing the `whnf`, so it's not unrelated either.

Copatterns improve the whnf of `compIso ? ?` as well, btw.

To be fair, to get all the benefits of copatterns for conversion checking we should add `no-eta-equality` to it though, I guess I'll test that.

mortberg:
Should I merge or do you want to experiment with `no-eta-equality` first?

Saizan:
We should merge as-is.

We can revisit `no-eta-equality` once canonicity is fixed https://github.com/agda/agda/issues/4560

PR #257, opened 2020-03-28
Kraus-von Raumer induction for pushouts

ecavallo:
From https://arxiv.org/abs/1901.06022.

The lemma I called "``flippedHomotopyIsInverse``" here has shown up before (``HITs.S1.Base.rotLoopInv`` is an instance) but I don't know if it should be moved up somewhere or what it should be called.

### DIFF
--- a/Cubical/Foundations/Path.agda
+++ b/Cubical/Foundations/Path.agda
@@ -17,10 +17,6 @@ cong′ : ∀ {B : Type ℓ'} (f : A → B) {x y : A} (p : x ≡ y)
       → Path B (f x) (f y)
 cong′ f = cong f
 
-PathP→Path : ∀ {ℓ} {A B : Type ℓ} {C : A ≡ B} {a : A} {b : B}
-  → PathP (λ i → C i) a b → transport C a ≡ b
-PathP→Path {C = C} p i = transp (λ j → C (i ∨ j)) i (p i)
-
 PathP≡Path : ∀ (P : I → Type ℓ) (p : P i0) (q : P i1) →
              PathP P p q ≡ Path (P i1) (transport (λ i → P i) p) q
 PathP≡Path P p q i = PathP (λ j → P (i ∨ j)) (transp (λ j → P (i ∧ j)) (~ i) p) q
--- a/Cubical/Foundations/Path.agda
+++ b/Cubical/Foundations/Path.agda
@@ -17,6 +17,10 @@ cong′ : ∀ {B : Type ℓ'} (f : A → B) {x y : A} (p : x ≡ y)
       → Path B (f x) (f y)
 cong′ f = cong f
 
+PathP→Path : ∀ {ℓ} {A B : Type ℓ} {C : A ≡ B} {a : A} {b : B}
+  → PathP (λ i → C i) a b → transport C a ≡ b
+PathP→Path {C = C} p i = transp (λ j → C (i ∨ j)) i (p i)
+
 PathP≡Path : ∀ (P : I → Type ℓ) (p : P i0) (q : P i1) →
              PathP P p q ≡ Path (P i1) (transport (λ i → P i) p) q
 PathP≡Path P p q i = PathP (λ j → P (i ∨ j)) (transp (λ j → P (i ∧ j)) (~ i) p) q
--- a/Cubical/HITs/Pushout/KrausVonRaumer.agda
+++ b/Cubical/HITs/Pushout/KrausVonRaumer.agda
@@ -43,8 +43,8 @@ private
 
 module ElimL {ℓ ℓ' ℓ'' ℓ'''} {A : Type ℓ} {B : Type ℓ'} {C : Type ℓ''}
   {f : A → B} {g : A → C} {b₀ : B}
-  {P : ∀ b → Path (Pushout f g) (inl b₀) (inl b) → Type ℓ'''}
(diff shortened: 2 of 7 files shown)


ecavallo:
This would presumably also be useful for other types like suspension, mapping cone, join. If the lemmas were all moved upstream than the code duplication wouldn't be so bad.

But it doesn't look like transp is being optimized for these special case types, so is there any reason not to define them as instances of the pushout in the first place?

m-yac:
> But it doesn't look like transp is being optimized for these special case types, so is there any reason not to define them as instances of the pushout in the first place?

I noticed this as well, I think it would be nicer for `Susp`, `Cone`, and `Join` to be defined in terms of `Pushout`. We could still include the HIT versions in a comment, like in `Truncation.Base`.

We would probably also want pattern synonyms for each (e.g. `north` for `inl tt`).

Cubical/Foundations/Path.agda
```diff
@@ -17,6 +17,10 @@ cong′ : ∀ {B : Type ℓ'} (f : A → B) {x y : A} (p : x ≡ y)
       → Path B (f x) (f y)
 cong′ f = cong f
 
+PathP→Path : ∀ {ℓ} {A B : Type ℓ} {C : A ≡ B} {a : A} {b : B}
+  → PathP (λ i → C i) a b → transport C a ≡ b
+PathP→Path {C = C} p i = transp (λ j → C (i ∨ j)) i (p i)
```
mortberg: This is kind of the same as: https://github.com/agda/cubical/blob/master/Cubical/Foundations/Prelude.agda#L174

Cubical/HITs/Pushout/KrausVonRaumer.agda
```diff
@@ -0,0 +1,147 @@
...
+  (Q : ∀ c → Path (Pushout f g) (inl b₀) (inr c) → Type ℓ''')
+  (r : P b₀ refl)
+  (e : (a : A) (q : inl b₀ ≡ inl (f a)) → P (f a) q ≃ Q (g a) (q ∙ push a))
+  where
+
+  Codes : (d : Pushout f g) (q : inl b₀ ≡ d) → Type ℓ'''
+  Codes (inl b) q = P b q
+  Codes (inr c) q = Q c q
+  Codes (push a i) q =
+    Glue
```
mortberg: Can't you use `ua` here?

Cubical/HITs/Pushout/KrausVonRaumer.agda
```diff
@@ -0,0 +1,147 @@
...
+  (Q : ∀ c → Path (Pushout f g) (inr c₀) (inr c) → Type ℓ''')
+  (r : Q c₀ refl)
+  (e : (a : A) (q : inr c₀ ≡ inl (f a)) → P (f a) q ≃ Q (g a) (q ∙ push a))
+  where
+
+  Codes : (d : Pushout f g) (q : inr c₀ ≡ d) → Type ℓ'''
+  Codes (inl b) q = P b q
+  Codes (inr c) q = Q c q
+  Codes (push a i) q =
+    Glue
```
mortberg: Use `ua`?

Cubical/HITs/Pushout/KrausVonRaumer.agda
ecavallo: Not as it is, the line being extended is not constant.

Cubical/Foundations/Path.agda
ecavallo: Oops, I'll use that one.

ecavallo:
> But why can't we just keep the definitions as they are and have paths between them and the pushout definitions so that we can transport whatever results we want?

Well, any time we do this we negate the efficiency gains of special casing in the first place. And it can be annoying to set up if the theorem talks about the constructors or eliminator.

Cubical/HITs/Pushout/KrausVonRaumer.agda
```diff
@@ -0,0 +1,147 @@
...
+
+{-# OPTIONS --cubical --safe #-}
+module Cubical.HITs.Pushout.KrausVonRaumer where
+
+open import Cubical.Foundations.Everything
+open import Cubical.Data.Sigma
+open import Cubical.HITs.Pushout.Base
+
+private
+  flippedHomotopyIsInverse : ∀ {ℓ} {A : Type ℓ} {f : A → A} (p : ∀ a → f a ≡ a)
```
mortberg, line 17: How about moving this to Foundations.Function and calling it "homotopySymInv" or something similar?

mortberg:
> > But why can't we just keep the definitions as they are and have paths between them and the pushout definitions so that we can transport whatever results we want?
> 
> Well, any time we do this we negate the efficiency gains of special casing in the first place. And it can be annoying to set up if the theorem talks about the constructors or eliminator.

Right... So it might be better to use the Pushout definitions after all in order to avoid duplication of proofs. I wouldn't mind keeping the direct definitions as well, but encourage the use of the Pushout ones. No matter what this should be done in future PRs.

Cubical/HITs/Pushout/KrausVonRaumer.agda
ecavallo: Done! Sadly the other path lemmas are somewhat weird... I hoped that there was a way to build them out of lemmas already in the library but I didn't come up with anything.

mortberg:
Thanks for the contribution!

PR #258, opened 2020-03-29
Friendlier definitions of composition in Foundations.Prelude

m-yac:
This PR comes from the following ideas:
- As far as I can tell, the definition of `compPath-filler` doesn't need to come before the definition of `_∙_`, in fact I think it's more helpful to newcomers if it comes after:
   ```agda
   _∙_ : x ≡ y → y ≡ z → x ≡ z
   _∙_ {x = x} p q i =
     hcomp (λ j → λ { (i = i0) → x
                    ; (i = i1) → q j })
           (p i)
   
   compPath-filler : (p : x ≡ y) (q : y ≡ z) → PathP (λ j → x ≡ q j) p (p ∙ q)
   compPath-filler {x = x} p q j i =
     hfill (λ j → λ { (i = i0) → x
                    ; (i = i1) → q j })
           (inS (p i)) j
   ```
- The reason to use `_□_` over `_∙_` is because of `_□_`'s filler, but it's straightforward to define the analogous filler for `_∙_`:
   ```agda
   compPath-filler' : (p : x ≡ y) (q : y ≡ z) → PathP (λ j → p (~ j) ≡ z) q (p ∙ q)
   compPath-filler' {z = z} p q j i =
     hcomp (λ k → λ { (i = i0) → p (~ j)
                    ; (i = i1) → q k
                    ; (j = i0) → q (i ∧ k) })
           (p (i ∨ ~ j))
   ```
   Using this instead of `compPath'-filler` is much more convenient, and can simply things. See `homotopyNatural`, for example.

I ended up defining both versions of composition in terms of `_∙∙_∙∙_`, which makes for a nicer proof that they are equal and (in my opinion) makes for nicer exposition. In general, I focused on clarity of exposition, spelling out in comments a few things I wish I would've seen when I was first trying to grok cubical type theory.

My goal here is to make reading `Foundations.Prelude` and using `_∙_` a bit friendlier to new folks, so if this complicates things in other ways I'm very happy to greatly modify or close this PR.

### DIFF
--- a/Cubical/Foundations/Equiv.agda
+++ b/Cubical/Foundations/Equiv.agda
@@ -120,13 +120,9 @@ PropEquiv→Equiv Aprop Bprop f g = isoToEquiv (iso f g (λ b → Bprop (f (g b)
 
 homotopyNatural : {f g : A → B} (H : ∀ a → f a ≡ g a) {x y : A} (p : x ≡ y) →
                   H x ∙ cong g p ≡ cong f p ∙ H y
-homotopyNatural H p = homotopyNatural' H p ∙ □≡∙ _ _
-  where
-  homotopyNatural' : {f g : A → B} (H : ∀ a → f a ≡ g a) {x y : A} (p : x ≡ y) →
-                     H x ∙ cong g p ≡ cong f p □ H y
-  homotopyNatural' {f = f} {g = g} H {x} {y} p i j =
+homotopyNatural {f = f} {g = g} H {x} {y} p i j =
     hcomp (λ k → λ { (i = i0) → compPath-filler (H x) (cong g p) k j
-                   ; (i = i1) → compPath'-filler (cong f p) (H y) k j
+                   ; (i = i1) → compPath-filler' (cong f p) (H y) k j
                    ; (j = i0) → cong f p (i ∧ (~ k))
                    ; (j = i1) → cong g p (i ∨ k) })
           (H (p i) j)
--- a/Cubical/Foundations/Prelude.agda
+++ b/Cubical/Foundations/Prelude.agda
@@ -64,12 +64,12 @@ cong₂ f p q i = f (p i) (q i)
 {- The most natural notion of homogenous path composition
     in a cubical setting is double composition:
 
-     x ∙ ∙ ∙ > w
-     |         ^
-   p |         | r        ^
-     V         |        j |
-     y — — — > z          ∙ — >
-          q                 i
+       x ∙ ∙ ∙ > w
+       ^         ^
+   p⁻¹ |         | r        ^
+       |         |        j |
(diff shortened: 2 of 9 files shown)


Cubical/Foundations/Equiv.agda
```diff
@@ -120,13 +120,9 @@ PropEquiv→Equiv Aprop Bprop f g = isoToEquiv (iso f g (λ b → Bprop (f (g b)
 
 homotopyNatural : {f g : A → B} (H : ∀ a → f a ≡ g a) {x y : A} (p : x ≡ y) →
                   H x ∙ cong g p ≡ cong f p ∙ H y
-homotopyNatural H p = homotopyNatural' H p ∙ □≡∙ _ _
-  where
-  homotopyNatural' : {f g : A → B} (H : ∀ a → f a ≡ g a) {x y : A} (p : x ≡ y) →
-                     H x ∙ cong g p ≡ cong f p □ H y
-  homotopyNatural' {f = f} {g = g} H {x} {y} p i j =
+homotopyNatural {f = f} {g = g} H {x} {y} p i j =
```
mortberg: Wow, nice!

Cubical/Foundations/Prelude.agda
```diff
@@ -61,35 +61,112 @@ cong₂ : ∀ {C : (a : A) → (b : B a) → Type ℓ} →
...
--- The filler of homogeneous path composition:
--- compPath-filler p q = PathP (λ i → x ≡ q i) p (p ∙ q)
+{- The most natural notion of homogenous path composition
+    in a cubical setting is double composition:
 
-compPath-filler : ∀ {x y z : A} → x ≡ y → y ≡ z → I → I → A
-compPath-filler {x = x} p q j i =
-  hfill (λ j → λ { (i = i0) → x
-                  ; (i = i1) → q j }) (inS (p i)) j
+     x ∙ ∙ ∙ > w
```
mortberg: It's a bit weird to have p go in this direction... Why not have p^⁻1 and have it look like a normal composition diagram?

m-yac: I actually find it easier to visualize squares/cubes/etc. this way (drawing p^-1 as an arrow going the opposite direction) but happy to change it if it's nonstandard or less intuitive to others.

mortberg: I guess it's OK in 2D, but it doesn't scale very well to higher dimensions... At least I get very confused when the arrows all go in different direction when you draw a cube. Plus, this drawing doesn't really match the semantic meaning of hcomp as an operation that gives you the missing side of an open box.

m-yac: Ah, these are very good points. I should probably stop thinking about the edges of cubes this way.

PR #260, opened 2020-03-29
Make the type of `EquivJ` better match the type of `J`

m-yac:
As requested by @mortberg in #252.


mortberg:
This was surprisingly easy to fix. I thought we used EquivJ much more... Thanks a lot for fixing it!

PR #261, opened 2020-03-29
Explanation of "A is constant on φ" in Cubical.Core.Primitives

m-yac:
The terminology "A is constant on φ" can be counterintuitive and confusing when seeing it for the first time. This PR adds an explanation and some examples, mostly taken from the docs, to `Cubical.Core.Primitives`.

The third example I added I think is particularly helpful in making clear what this phrase means, let me know if folks think this example should be added to the docs as well.


mortberg:
Thanks for the clarification! Feel free to update the docs as well.

Btw, I don't think the docs are linked to anywhere on this github page?

PR #262, opened 2020-03-31
Different levels of objects and morphisms in categories

Schippmunk:
This PR adds another level argument to the record Precategory such that the ob and hom may take values in different universes.

This has the advantage that the lifting of morphisms, as seen for instance in Presheaves.agda, can be omitted.


mortberg:
Neat, thanks for contributing!

PR #264, opened 2020-04-03
Comments, Univalent Precategories and Pullbacks in Precategories

Schippmunk:
This PR adds:
- Comments regarding the terminology of categories to Categories.agda.
- The definition of a univalent precategory: a precategory in which for every two objects, pathToIso is an equivalence (of types)
- The definition of category-theoretical cospans, cones, cone-map and pullbacks to Categories/Pullback.agda


mortberg:
Thanks for the contribution! 

Before adding more category theory to the library I think we should think about whether we want to use the SIP more when developing CT. We could follow https://www.cs.bham.ac.uk/~mhe/HoTT-UF-in-Agda-Lecture-Notes/HoTT-UF-Agda.html#categories-sip. What do we gain by such an approach?

Schippmunk:
I'm not sure I fully understand your question, so please rephrase it if I'm missing the point.

To me it seems that above definition of pullbacks and isos in categories is orthogonal to whether an SIP for categories is commonly used or not.
And unless `isUnivalent` is not invariant under traditional equivalences of precategories, I see no harm in defining it like that either.

mortberg:
My point is that if we want to benefit from the SIP as currently stated when we should rewrite the definitions about categories to fit the pattern in https://github.com/agda/cubical/blob/master/Cubical/Structures/Monoid.agda. This is discussed in detail for Martíns formalization in the link in my previous message. It's not clear to me what is gained from this, but it might be possible to simplify some later developments (e.g. reducing duplication related to developing theory for limits and colimits, etc.). Maybe someone else has more thoughts about it? @mzeuner @m-yac @jonsterling 

One drawback with this approach is that we can't use records, but I personally have nothing against using nested Sigma-types. But it might have some consequences for efficiency? @Saizan 

So my point is just that we should figure out what we want to do before adding a lot more CT which we later might have to completely rewrite if we want to experiment with the cubical SIP. I think it could be a worthwhile and interesting project to develop CT using the cubical SIP, but it might also be that nothing is gained from it and a more traditional approach is better.

mortberg:
PS: note that the SIP as stated in the HoTT book (about categories) is, at least a priori, different from the cubical SIP that we have in the library.

Saizan:
Comparing nested sigma types for equality can be quite more expensive, e.g. it's proportional to how many fields you have and how big their types are, rather than just proportional to how many parameters a record type would have.

Also projecting from nested sigma types or defining an element by copatterns can be quite more inconvenient.

It's however very easy to prove an equivalence with the nested sigma type if you want to use a result for them. I would suggest we define stuff as standalone records and prove such equivalences if we have a use for them.

mortberg:
> Comparing nested sigma types for equality can be quite more expensive, e.g. it's proportional to how many fields you have and how big their types are, rather than just proportional to how many parameters a record type would have.
> 
> Also projecting from nested sigma types or defining an element by copatterns can be quite more inconvenient.
> 
> It's however very easy to prove an equivalence with the nested sigma type if you want to use a result for them. I would suggest we define stuff as standalone records and prove such equivalences if we have a use for them.

Sounds like a good strategy. I'll make an issue about it and merge this one

Cubical/Categories/Pullback.agda
```diff
@@ -0,0 +1,48 @@
...
+record Pullback {𝒞 : Precategory ℓ ℓ'} (cspn : Cospan 𝒞) : Type (ℓ-max ℓ ℓ') where
+  constructor pullback
+  field
+    c : ob 𝒞
+    cn : Cone cspn c
+    universal : {c' : ob 𝒞} (cn' : Cone cspn c') → isContr (Σ[ f ∈ 𝒞 .hom c' c ] Σ[ q ∈ Cone.p₁ cn' ≡ 𝒞 .seq f (Cone.p₁ cn) ] (Cone.p₂ cn' ≡ 𝒞 .seq f (Cone.p₂ cn)))
+
+-- whisker the parallel morphisms g and g' with f
+lPrecatWhisker : {𝒞 : Precategory ℓ ℓ'} {x y z : 𝒞 .ob} (f : 𝒞 .hom x y) (g g' : 𝒞 .hom y z) (p : g ≡ g') → (𝒞 .seq f g ≡ 𝒞 .seq f g')
+lPrecatWhisker {𝒞 = 𝒞} f g g' p = J (λ h q → 𝒞 .seq f g ≡ 𝒞 .seq f h) refl p
```
mortberg, line 38: Why prove this with J?  It's just `cong (𝒞 .seq f) p`. 

Also, why have parantheses around the conclusion?

Cubical/Categories/Pullback.agda
```diff
@@ -0,0 +1,48 @@
...
+
+-- extend a cone on c by a morphism c'→c using precomposition
+coneMap : {𝒞 : Precategory ℓ ℓ'} {cspn : Cospan 𝒞} {c c' : ob 𝒞} (cn : Cone cspn c) (f : hom 𝒞 c' c) → Cone cspn c'
+coneMap {𝒞 = 𝒞} {cspn = cspn} cn f = cone (𝒞 .seq f p1)  (𝒞 .seq f p2) ((𝒞 .seq-α f p1 s1 ) ∙ (lPrecatWhisker {𝒞 = 𝒞} f (𝒞 .seq p1 s1) (𝒞 .seq p2 s2) sq) ∙ (sym (𝒞 .seq-α f p2 s2)))
```
mortberg, line 42: Why not use triple composition https://github.com/agda/cubical/blob/master/Cubical/Foundations/Prelude.agda#L78 ? It's a bit more natural

Cubical/Categories/Pullback.agda
Schippmunk: That's true, cong is obviously a better choice here

Cubical/Categories/Pullback.agda
Schippmunk: That didn't exist at the time of the PR but I'll merge and then use it.

mortberg:
Thanks for the contribution!

PR #265, opened 2020-04-03
Prove `(B → Type ℓ) ≃ Σ[ E ∈ Type ℓ ] (E → B)`

m-yac:
I'm not totally sure about:
- the names I chose for the lemmas in `Foundations.Fibration`,
- my decision to move the result of `Foundations.TotalFiber` to `Foundations.Fibration`.


mortberg:
This looks good to me! Thanks for adding this result

PR #267, opened 2020-04-03
Reorganize Cubical.Foundations

m-yac:
This PR:
- Moves Bundle, Embedding, FunExtEquiv, Surjection, and TotalFiber into Cubical/Functions
- Moves the different notions of equivalence (PathSplit, BiInvertible, HalfAdjoint) into Cubical/Foundations/Equiv
- Fixes all the import errors that creates


mortberg:
Great, thanks a lot! 

I think we can do even more in this PR:

- CartesianKanOps, HoTT-UF > Experiments

- Pointed -> Structures    (not sure about this move, but it might make sense?)

- The Logic file isn't really used anywhere. Maybe it should be moved or maybe it should be populated with some more useful results... @mzeuner Maybe the stuff about Omega that we discussed yesterday can go into this file?

- Merge UnivalenceId and Id?

mortberg:
Thinking more about this, let's do the following in this PR
 
* Move CartesianKanOps, HoTT-UF -> Experiments
* Pointed -> Structures   
* Merge UnivalenceId and Id

Can you do it @m-yac ? 

Sorry for the conflicts from my truncation PR. Hopefully they will be easy to fix

m-yac:
> * Move CartesianKanOps, HoTT-UF -> Experiments
> * Merge UnivalenceId and Id

Done.

> * Pointed -> Structures

I'm not sure I agree with this move. The notion of a pointed type is pretty foundational, and the `Pointed.Properties` and `Pointed.Homogeneous` files don't really feel like they belong in `Structures` (based on the files there right now). This is based on me starting to make the move and realizing it didn't really fit.

I think that the contents of `Pointed.Base` should stay in `Foundations` and that the `Pointed.Properties` and `Pointed.Homogeneous` files could be moved somewhere else – but I don't think `Structures` is the right place. 

Of course, I don't really care that much and am happy to make the move if you think that's best.

m-yac:
Does anyone know why github sometimes doesn't pick up on travis builds? This PR does pass btw: https://travis-ci.org/github/agda/cubical/builds/673582807

mortberg:
Ok, let's keep Pointed in Foundations for now. 

I have no clue what went wrong with Travis on this PR... It doesn't seem like I can even restart the build.

mortberg:

PR #269, opened 2020-04-05
[ travis ] Only reinstall agda when necessary

m-yac:
In this PR travis now caches the tag of the commit used to build the (already cached) agda executable, and does not re-build this executable if the cached tag matches the current one.

The result is that travis now takes only **4-5m** if the master branch of agda has not been updated since the last build, and the usual **20-30m** otherwise.

See below for the results of some experimenting. All the builds there shared the same `travis.yml` file and were triggered with empty commits to my `travis` branch – my comments are just describing what happened automatically. (Also, the commit tags I reference are commits in [agda/agda](https://github.com/agda/agda/commits/master).)

empty-commit-testing

I'm still a noob with CI/travis, so @phadej / @Saizan / anyone else, is what I've done reasonable?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,6 +59,7 @@ install:
       echo "cabal build-cache MISS";
       rm -rf $HOME/.cabsnap;
       mkdir -p $HOME/.ghc $HOME/.cabal/lib $HOME/.cabal/share $HOME/.cabal/bin;
+      travis_wait 30 cabal install --only-dependencies;
     fi
   # only installing agda if we have to
   - touch $HOME/oldHEAD.txt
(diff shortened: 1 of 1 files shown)


m-yac:
Hmm... don't know why github isn't showing the latest travis build of this PR, but it passed (in 4m 30s): https://travis-ci.org/github/agda/cubical/builds/671318433

Saizan:
I'm not a travis expert either, but it does make sense.
I was kinda assuming we were doing this before too, but I guess not.

I wonder how this compares to what https://github.com/agda/agda-stdlib does.

PR #270, opened 2020-04-05
Lex modalities

jonsterling:
Axiomatizing and proving some properties of lex modalities. Unfortunately, I need the modality itself to be universe polymorphic, so I can't share the code with the existing modality module. This contains a proof that lex modalities commute with dependent sums.

Please point out any stylistic issues --- I'm still learning the style of this library.

### DIFF
--- a/Cubical/Modalities/Lex.agda
+++ b/Cubical/Modalities/Lex.agda
@@ -105,9 +105,9 @@ module LexModality
 
     unpush-sg-split : (x : ◯ A) (y : ⟨◯⟩ B x) → ◯Σ
     unpush-sg-split =
-      ◯-ind (λ _ → Π-modal (λ _ → idemp)) λ x →
+      ◯-ind (λ _ → Π-modal λ _ → idemp) λ x →
       abstract-along {C = λ _ → ◯Σ} (⟨◯⟩-compute B x)
-      (◯-map λ y → x , y)
+      (◯-map (x ,_))
 
     unpush-sg-split-compute : (x : A) → unpush-sg-split (η x) ≡ abstract-along (⟨◯⟩-compute B x) (◯-map (λ y → x , y))
     unpush-sg-split-compute = ◯-ind-β _ _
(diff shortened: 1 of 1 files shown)


jonsterling:
Not sure why the build is failing on something that I didn't change... Maybe Agda has changed slightly since I started my PR?

@Saizan

Saizan:
@jonsterling it's probably a recent regression in agda/master, I'll try to report it.

Cubical/Modalities/Lex.agda
```diff
@@ -0,0 +1,279 @@
...
+      B : A → Type ℓ′
+      C : Σ A B → Type ℓ′′
+
+  cong-fun : {f g : (x : A) → B x} → f ≡ g → (x : A) → f x ≡ g x
+  cong-fun α x i = α i x
+
+  pair-ext : {p q : Σ A B} (α : p .fst ≡ q .fst) (β : PathP (λ i → B (α i)) (p .snd) (q .snd)) → p ≡ q
+  pair-ext α β i = α i , β i
+
+  λ⟨,⟩_ : ((x : A) (y : B x) → C (x , y)) → (p : Σ A B) → C p
```
mortberg, line 151: This is https://github.com/agda/cubical/blob/master/Cubical/Foundations/Function.agda#L41

Cubical/Modalities/Lex.agda
```diff
@@ -0,0 +1,279 @@
...
+    variable
+      ℓ′′ : Level
+      A A′ : Type ℓ
+      B : A → Type ℓ′
+      C : Σ A B → Type ℓ′′
+
+  cong-fun : {f g : (x : A) → B x} → f ≡ g → (x : A) → f x ≡ g x
+  cong-fun α x i = α i x
+
+  pair-ext : {p q : Σ A B} (α : p .fst ≡ q .fst) (β : PathP (λ i → B (α i)) (p .snd) (q .snd)) → p ≡ q
```
mortberg, line 148: This is https://github.com/agda/cubical/blob/master/Cubical/Data/Sigma/Properties.agda#L37

Cubical/Modalities/Lex.agda
```diff
@@ -0,0 +1,279 @@
...
+
+module Combinators where
+  private
+    variable
+      ℓ′′ : Level
+      A A′ : Type ℓ
+      B : A → Type ℓ′
+      C : Σ A B → Type ℓ′′
+
+  cong-fun : {f g : (x : A) → B x} → f ≡ g → (x : A) → f x ≡ g x
```
mortberg, line 145: This is https://github.com/agda/cubical/blob/master/Cubical/Foundations/Prelude.agda#L201

Cubical/Modalities/Lex.agda
```diff
@@ -0,0 +1,279 @@
...
+
+
+isConnected : Type ℓ → Type ℓ
+isConnected A = isContr (◯ A)
+
+
+
+module FormalDiskBundle {A : Type ℓ} where
+  𝔻 : A → Type ℓ
+  𝔻 a = Σ A (λ x → η a ≡ η x)
```
mortberg, line 279: I think the fancy Sigma notation would make this definition look nicer: https://github.com/agda/cubical/blob/master/Cubical/Core/Primitives.agda#L219

Cubical/Modalities/Lex.agda
jonsterling: It's unfortunate that switching to uncurry requires me to put parentheses around some very large expressions lol

jonsterling: I'll just rename as you suggested!

jonsterling:
OK, I've updated to use the existing combinators from the library, thank you for pointing those out!

Regarding making the modalities file sufficiently polymorphic, I think this is sadly beyond the capabilities of Agda's universe hierarchy -- there is, in particular, no record type in agda's type theory that contains the "polymorphic modalities". One thing that could try is take "\forall l -> Modality l" as a parameter to this module, but the problem is that the eliminators will not have the right levels for their motives and I couldn't figure out how to support this.

Ultimately there may be a more sensible way to do this: rather than defining lex modality as polymorphic in the whole universe hierarchy, just require a couple modalities at different sizes, and require an equivalence involving a certain universe level lift.

Alternatively (?) one can use the formalism of "accessible modalities" from the HoTT book. Anyway, these might be reasonable ideas but I won't have the time to think about them for a little bit. Any suggestions welcome.

mortberg:
Ah ok, I get it now and I don't see how to easily solve it either. I don't think it's too bad as it is, so let's merge it. Thanks for the contribution!

PR #271, opened 2020-04-07
Cleaning of truncation files

mortberg:
I made a pass over the truncation files and cleaned some proofs. I also made the various concrete truncation files export the same kind of lemmas (elim1-elim3, etc.)

### DIFF
--- a/Cubical/Foundations/HLevels.agda
+++ b/Cubical/Foundations/HLevels.agda
@@ -56,6 +56,12 @@ isOfHLevelSuc 0 = isContr→isProp
 isOfHLevelSuc 1 = isProp→isSet
 isOfHLevelSuc (suc (suc n)) h a b = isOfHLevelSuc (suc n) (h a b)
 
+isSet→isGroupoid : isSet A → isGroupoid A
+isSet→isGroupoid = isOfHLevelSuc 2
+
+isGroupoid→is2Groupoid : isGroupoid A → is2Groupoid A
+isGroupoid→is2Groupoid = isOfHLevelSuc 3
+
 isOfHLevelPlus : (m : ℕ) → isOfHLevel n A → isOfHLevel (m + n) A
 isOfHLevelPlus zero hA = hA
 isOfHLevelPlus (suc m) hA = isOfHLevelSuc _ (isOfHLevelPlus m hA)
--- a/Cubical/Foundations/HLevels.agda
+++ b/Cubical/Foundations/HLevels.agda
@@ -27,11 +27,11 @@ open import Cubical.Data.Nat   using (ℕ; zero; suc; _+_; +-zero; +-comm)
 
 private
   variable
-    ℓ ℓ' ℓ'' ℓ''' : Level
+    ℓ ℓ' : Level
     A : Type ℓ
     B : A → Type ℓ
-    C : (x : A) → B x → Type ℓ''
-    D : (x : A) (y : B x) → C x y → Type ℓ''
+    C : (x : A) → B x → Type ℓ
+    D : (x : A) (y : B x) → C x y → Type ℓ
     x y : A
     n : ℕ
 
--- a/Cubical/Foundations/HLevels.agda
+++ b/Cubical/Foundations/HLevels.agda
@@ -107,6 +107,12 @@ isPropIsOfHLevel (suc (suc n)) f g i a b =
 isPropIsSet : isProp (isSet A)
 isPropIsSet = isPropIsOfHLevel 2
 
+isPropIsGroupoid : isProp (isGroupoid A)
+isPropIsGroupoid = isPropIsOfHLevel 3
+
+isPropIs2Groupoid : isProp (is2Groupoid A)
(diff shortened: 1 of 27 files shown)


mortberg:
I also renamed Pi to Π in lemmas about HLevels and moved ΣProp≡ to Data.Sigma. This will cause some conflicts with other PRs once I merge... @m-yac 

I plan to add isPropΠ2 and isPropΠ3 to simplify repeated applications of isPropΠ. Once I'm done with that I'll be ready to merge

mortberg:
Alright, all changes I wanted to do are now done. Comments?

Cubical/Foundations/HLevels.agda
```diff
@@ -22,14 +22,16 @@ open import Cubical.Foundations.Transport
...
 open import Cubical.Data.Nat   using (ℕ; zero; suc; _+_; +-zero; +-comm)
 
 private
   variable
-    ℓ ℓ' : Level
+    ℓ ℓ' ℓ'' ℓ''' : Level
     A : Type ℓ
     B : A → Type ℓ
+    C : (x : A) → B x → Type ℓ''
+    D : (x : A) (y : B x) → C x y → Type ℓ''
```
Saizan, line 34: I'm pretty sure you could just use `ℓ` for `C` and `D` too?

mortberg: Oh, I intended to use l''' for D. But maybe that's not needed? Will Agda maximally generalize everything?

Cubical/HITs/PropositionalTruncation/Properties.agda
```diff
@@ -27,26 +29,41 @@ rec : ∀ {P : Type ℓ} → isProp P → (A → P) → ∥ A ∥ → P
...
+elim3 Pprop g = elim2 (λ _ _ → isPropΠ (λ _ → Pprop _ _ _))
+                      (λ a b → elim (λ _ → Pprop _ _ _) (g a b))
+
+propTruncIsProp : isProp ∥ A ∥
+propTruncIsProp x y = squash x y
 
-propId : isProp A → ∥ A ∥ ≡ A
-propId {A = A} hA = isoToPath (iso (rec hA (idfun A)) (λ x → ∣ x ∣) (λ _ → refl) rinv)
+propTruncId : isProp A → ∥ A ∥ ≡ A
+propTruncId {A = A} hA = isoToPath (iso (rec hA (idfun A)) (λ x → ∣ x ∣) (λ _ → refl) rinv)
```
Saizan, line 57: Should we export the iso too? also probably better to define it by copatterns.

Cubical/HITs/SetTruncation/Properties.agda
```diff
@@ -20,55 +21,51 @@ private
...
 
 setTruncIsSet : isSet ∥ A ∥₀
 setTruncIsSet a b p q = squash₀ a b p q
 
-setId : isSet A → ∥ A ∥₀ ≡ A
-setId {A = A} isset =
-  isoToPath (iso (elim {A = A} (λ _ → isset) (λ x → x)) (λ x → ∣ x ∣₀) (λ b → refl) (λ b → idLemma b))
+setTruncId : isSet A → ∥ A ∥₀ ≡ A
+setTruncId {A = A} hS = isoToPath (iso (rec hS (idfun A)) (λ x → ∣ x ∣₀)
+                                       (λ _ → refl) rinv)
```
Saizan, line 67: `setId` (like `propId`) could be named better maybe? it makes me think of the `Id` type.

Cubical/Foundations/HLevels.agda
Saizan: I'm pretty sure it will. the `ℓ` in the types of `A` and `B` are not constrained to be the same either.

You only need two level `variable`s for something like `F : Set l -> Set l'` where you actually write both in the same type.

Cubical/HITs/SetTruncation/Properties.agda
mortberg: I think this Id stands for idempotent. What about setTruncIdem or setTruncIdempotent?

Cubical/Foundations/HLevels.agda
mortberg: Ah right, I keep forgetting how variables work in Agda... It's so confusing compared to Coq. Will fix this

Cubical/HITs/SetTruncation/Properties.agda
Saizan: Yeah, those would nice!

Cubical/HITs/PropositionalTruncation/Properties.agda
mortberg: I'll export it as an equiv, but define the iso using copatterns in a where

Cubical/HITs/SetTruncation/Properties.agda
mortberg: Alright, I'll go for Idempotent

PR #273, opened 2020-04-08
Multi set structure

mzeuner:
We introduce a simple signature for multi-set structures in the spirit of Okasaki's signature for sets (p. 12 of "Purely Functional Data Structures"). We then define multi-set structures on finite multi-sets and association lists and show that the two are isomorphic.


mortberg:
Nice! Ping @cangiuli

mortberg:
As you're now using the recursor/eliminator and not pattern-matching, shouldn't `with` work so that the helper definitions can be inlined?

Cubical/HITs/AssocList/Properties.agda
```diff
@@ -72,6 +76,108 @@ FMS→AL∘AL→FMS≡id = AL.ElimProp.f (AL.trunc _ _) refl (λ x n {xs} p →
...
+ ALmember a = AL.Rec.f isSetℕ 0 (ALmember-⟨,⟩∷* a) (ALmember-per* a) (ALmember-agg* a) (ALmember-del* a)
+
+
+
+ AL-with-str : Multi-Set A (Discrete→isSet discA)
+ AL-with-str = (AssocList A , ⟨⟩ , ⟨_, 1 ⟩∷_ , ALmember)
+
+
+-- We want to show that Al-with-str ≅ FMS-with-str as multiset-structures
+module _(discA : Discrete A) where
```
mortberg, line 149: Why make a new module here and just repeat the discA assumption?

Cubical/HITs/AssocList/Properties.agda
```diff
@@ -72,6 +76,108 @@ FMS→AL∘AL→FMS≡id = AL.ElimProp.f (AL.trunc _ _) refl (λ x n {xs} p →
...
+    α : ⟨ x , 1 ⟩∷ FMS→AL xs ≡ FMS→AL (x ∷ xs)
+    α = sym (multi-∷-id x 1 xs)
+
+    θ : FMSmember discA a (x ∷ xs) ≡ ALmember discA a (⟨ x , 1 ⟩∷ (FMS→AL xs))
+    θ = χ x (FMSmember discA a xs) (ALmember discA a (FMS→AL xs)) (discA a x) p
+
+
+
+ FMS-with-str≡AL-with-str : FMS-with-str discA ≡ AL-with-str discA
+ FMS-with-str≡AL-with-str = SIP (multi-set-structure A (Discrete→isSet discA))
```
mortberg, line 178: Instead of calling `Discrete→isSet discA` repeatedly just make it a private definition in the model.

mortberg: module :-)

Cubical/HITs/AssocList/Properties.agda
mzeuner: good question, I don't remember why I did that, maybe wanted to do something in between initially...

mzeuner:
Some of the aux-function are needed  in later proofs, see e.g. 
χ :  (x : A) (xs ys : ℕ) (p : Dec (a ≡ x))
      → xs ≡ ys
      → FMSmember-∷*-aux a x p xs ≡ ALmember-⟨,⟩∷*-aux a x p 1 ys
in the proof of FMS→AL-isIso, I guess I could drop all the other aux-function and do it with with, but maybe they will be needed later as well...

mortberg:
> Some of the aux-function are needed in later proofs, see e.g.
> χ : (x : A) (xs ys : ℕ) (p : Dec (a ≡ x))
> → xs ≡ ys
> → FMSmember-∷*-aux a x p xs ≡ ALmember-⟨,⟩∷*-aux a x p 1 ys
> in the proof of FMS→AL-isIso, I guess I could drop all the other aux-function and do it with with, but maybe they will be needed later as well...

Ah I see! I guess it's better to have them like this then.

Merging now.

PR #274, opened 2020-04-09
[ fix #272 ] Defined _×_ in Data.Sigma and ported code over to it.

### LINKED ISSUE #272: Rename `_×_` to `Pair` and `_×Σ_` to `_×_`
I'm much more likely to use the product that's a special case of `Σ` rather than one defined as an inductive type.

What I'm proposing to call `Pair` was defined to get more expanded normal forms for some examples in the paper. That's not often what you want though.

Saizan:
I decided we did not need to rename `_×_` in `Cubical.Data.Prod`.

There's more parts of the library that should be converted to `Cubicla.Data.Sigma`, not sure if in this PR.


mortberg:
I think this PR is a good place to change it everywhere, but I guess this will cause more conflicts with https://github.com/agda/cubical/pull/267. I suggest merging that PR before finishing this one

mortberg:
All the major PRs are now merged. So I guess this one is next in line as soon as all the uses of _×_ have been fixed

PR #276, opened 2020-04-11
Composition, Properties of Pointed Maps

Schippmunk:
This PR adds composition of pointed maps and associativity and unit laws of that operation.

It further moves the loop space operation from HomotopyGroups/Base to Pointed/Properties and adds the natural loop space map.

### DIFF
--- a/Cubical/Foundations/Pointed/Properties.agda
+++ b/Cubical/Foundations/Pointed/Properties.agda
@@ -27,16 +27,20 @@ _∘∙_ : ∀ {ℓA ℓB ℓC} {A : Pointed ℓA} {B : Pointed ℓB} {C : Point
 (g , g∙) ∘∙ (f , f∙) = (λ x → g (f  x)) , ((cong g f∙) ∙ g∙)
 
 {- pointed identity -}
-∙-id : ∀ {ℓA} (A : Pointed ℓA) → (A →∙ A)
-∙-id A = ((λ x → x) , refl)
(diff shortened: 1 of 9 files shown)


Cubical/Foundations/Pointed/Properties.agda
```diff
@@ -13,3 +20,48 @@ open import Cubical.Data.Prod
 
 _×∙_ : ∀ {ℓ ℓ'} (A∙ : Pointed ℓ) (B∙ : Pointed ℓ') → Pointed (ℓ-max ℓ ℓ')
 A∙ ×∙ B∙ = ((typ A∙) × (typ B∙)) , (pt A∙ , pt B∙)
+
+{- composition of pointed maps -}
+_∘*_ : ∀ {ℓA ℓB ℓC} {A : Pointed ℓA} {B : Pointed ℓB} {C : Pointed ℓC}
```
mortberg, line 25: It would be better if all definitions related to pointed things used the same symbol, so please change `*` to `∙` everywhere

mortberg: Oops, this should have been part of the review

Schippmunk: I stuck to Foundations/Pointed/Base, where for pointed functions the asterisk is used, but I can change that aswell.

Schippmunk:
I agree it makes sense to have a separate file for loop spaces.

mortberg:
Thanks for contribution! I'll make an issue about proving more stuff about pointed maps

PR #277, opened 2020-04-12
Divisibility and the Euclidean algorithm

m-yac:
Part of some old code that I cleaned up. Once this is merged I'll make a PR which defines a few different versions of the rationals and proves them equivalent – I decided to break things up to make @mortberg's life a little easier.

Check out `Data.Nat.GCD` for the heart of this PR and `Data.Nat.Divisibility` for the cubical elements. It's been a long time since I've looked at agda-stdlib's version of GCD so I don't remember how this compares.


Cubical/Data/Nat/Divisibility.agda
```diff
@@ -0,0 +1,98 @@
...
+open import Cubical.Data.Nat.Base
+open import Cubical.Data.Nat.Properties
+open import Cubical.Data.Nat.Order
+
+private
+  variable
+    l m n : ℕ
+
+_∣_ : ℕ → ℕ → Type₀
+m ∣ n = ∥ Σ[ c ∈ ℕ ] c * m ≡ n ∥
```
mortberg, line 24: It would be nice to have a notation for this (\exists)

mortberg: I mean for the truncated Sigma

Cubical/Data/Nat/GCD.agda
```diff
@@ -0,0 +1,116 @@
...
+  variable
+    m n d : ℕ
+
+-- common divisors
+
+isCD : ℕ → ℕ → ℕ → Type₀
+isCD m n d = (d ∣ m) × (d ∣ n)
+
+isPropIsCD : isProp (isCD m n d)
+isPropIsCD = isPropΣ isProp∣ (λ _ → isProp∣)
```
mortberg, line 34: You can use https://github.com/agda/cubical/blob/master/Cubical/Foundations/HLevels.agda#L230 (this lemma should be renamed)

Cubical/Data/Nat/GCD.agda
```diff
@@ -0,0 +1,116 @@
...
+-- greatest common divisors
+
+isGCD : ℕ → ℕ → ℕ → Type₀
+isGCD m n d = (isCD m n d) × (∀ d' → isCD m n d' → d' ∣ d)
+
+GCD : ℕ → ℕ → Type₀
+GCD m n = Σ ℕ (isGCD m n)
+
+isPropIsGCD : isProp (isGCD m n d)
+isPropIsGCD = isPropΣ isPropIsCD (λ _ → isPropΠ2 (λ _ _ → isProp∣))
```
mortberg, line 48: here as well

Cubical/Data/Nat/Coprime.agda
```diff
@@ -0,0 +1,47 @@
...
+
+open import Cubical.HITs.PropositionalTruncation as PropTrunc
+
+open import Cubical.Data.Nat.Base
+open import Cubical.Data.Nat.Properties
+open import Cubical.Data.Nat.Divisibility
+open import Cubical.Data.Nat.GCD
+
+areCoprime : ℕ × ℕ₊₁ → Type₀
+areCoprime (m , n) = isGCD m (ℕ₊₁→ℕ n) 1
```
mortberg, line 19: Why not define this for ℕ × ℕ ?

mortberg: This is nicer because the relation is symmetric and no casts are needed

Cubical/Data/Sigma/Properties.agda
```diff
@@ -30,6 +30,9 @@ private
     B B' : (a : A) → Type ℓ
     C : (a : A) (b : B a) → Type ℓ
 
+mapʳ : (∀ {a} → B a → B' a) → Σ A B → Σ A B'
+mapʳ f (a , b) = (a , f b)
```
mortberg, line 34: Should we have map^l as well?

Cubical/Data/Nat/Divisibility.agda
mzeuner: There is the ∃[]-syntax in Foundations.Logic that could be use here

m-yac: The ∃[]-syntax in `Logic` assumes that the RHS of the sigma is a proposition, so I just defined a new version. I also incorporated it in a number of places in the library.

Cubical/Data/Nat/GCD.agda
m-yac: I renamed it and moved it to `Data.Prod.Properties`.

Cubical/Data/Sigma/Properties.agda
m-yac: Added for the case where B is constant, I'm can't imagine a map^l that is more useful than just using an eliminator in the case where is B is non-constant.

Cubical/Data/Nat/Coprime.agda
m-yac: Done. Note that `toCoprime` does need one of its arguments to be nonzero.

Cubical/Data/Nat/Divisibility.agda
mortberg: Nice! We should add unique existence as well (iscontr Sigma), not sure if this should go with exists or in the Prelude

PR #280, opened 2020-04-13
Vectors and n-ary funext

mortberg:
Define vectors and prove funext for n-ary (non-dependent) functions. This lets us generalize InftyMagmas to types equipped with n-ary functions making it easier to instantiate the SIP. I also changed some implicit arguments to make the SIP easier to use.

@mzeuner : Should I remove the InftyMagma file and rewrite Monoids using 2-ary functions? I think this would be cleaner


Cubical/Data/Vec/Properties.agda
```diff
@@ -0,0 +1,19 @@
...
+
+open import Cubical.Data.Nat
+open import Cubical.Data.Vec.Base
+
+private
+  variable
+    ℓ : Level
+
+
+-- This is really cool!
```
mortberg, line 14: @Saizan : didn't you experiment with this a long time ago? It's really neat that PathP simplifies this so much

Cubical/Data/Vec/NAry.agda
```diff
@@ -0,0 +1,50 @@
...
+
+$ⁿ-curryⁿ : ∀ {n} (f : Vec A n → B) → _$ⁿ_ (curryⁿ f) ≡ f
+$ⁿ-curryⁿ {n = zero} f  = funExt λ { [] → refl }
+$ⁿ-curryⁿ {n = suc n} f = funExt λ { (x ∷ xs) i → $ⁿ-curryⁿ {n = n} (λ ys → f (x ∷ ys)) i xs}
+
+curryⁿ-$ⁿ : ∀ {n} (f : nAryOp {ℓ = ℓ} {ℓ' = ℓ'} n A B) → curryⁿ {A = A} {B = B} (_$ⁿ_ f) ≡ f
+curryⁿ-$ⁿ {n = zero} f  = refl
+curryⁿ-$ⁿ {n = suc n} f = funExt λ x → curryⁿ-$ⁿ {n = n} (f x)
+
+nAryOp≃VecFun : ∀ {n} → nAryOp n A B ≃ (Vec A n → B)
```
mortberg, line 43: @Saizan : I couldn't figure out how to apply `ua` to this to get a Path... When trying I got all kinds of yellow that I didn't manage to get rid of

Cubical/Data/Vec/NAry.agda
```diff
@@ -0,0 +1,50 @@
...
+nAryOp≃VecFun {n = n} = isoToEquiv f
+  where
+  f : Iso (nAryOp n A B) (Vec A n → B)
```
potato4444, line 46: I've noticed this pattern of creating a where clause for the purpose of constructing an `Iso` via copatterns. Would it make sense to use the copattern version of [pattern matching lambda](https://agda.readthedocs.io/en/v2.6.1/language/lambda-abstraction.html) like we do [for `Conats`](https://github.com/agda/cubical/blob/4c3a3e2e06c86bce9bdb70d8e863e5ad73b97eb7/Cubical/Codata/Conat/Base.agda#L47)?

Cubical/Data/Vec/NAry.agda
Saizan: I guess the problem is that agda only sees those two types have the same universe level for concrete values of `n`.

Maybe the easiest would be for `nAryOp` to return a type in the max of the two levels and use `Lift` in the `zero` case? Or some other variation that doesn't make the level depend on the `n`.

Cubical/Data/Vec/Properties.agda
jonsterling: beautiful!

Saizan: @mortberg It is really neat! I don't remember doing a lot with vectors though.

mzeuner:
>  Should I remove the InftyMagma file and rewrite Monoids using 2-ary functions? I think this would be cleaner
That sounds like a good idea, I guess the whole point is to generalize infinity magmas.

Cubical/Data/Vec/NAry.agda
mortberg: I guess that should work, but I think it's easier to read this way. I have no strong opinions about this though so if someone wants to rewrite these please do

Cubical/Data/Vec/NAry.agda
mortberg: That might indeed work! But it was so annoying to add Lift everywhere... I have no uses of this equality yet so I'll keep things as they are but with a comment

Cubical/Data/Vec/NAry.agda
Saizan: Pattern matching lambdas have the downside that they will be printed in full most of the time, and otherwise give you an uninformative generated name.

mortberg: I see, that seems like an argument to keep things the way they are

mortberg:
> That sounds like a good idea, I guess the whole point is to generalize infinity magmas.

Yup, did it now and it worked out nicely

PR #284, opened 2020-04-15
Newstructures

borsiemir:
Adding some new structures (semigroups, groups, abelian groups and rngs) and a new folder of Algebra, in order to produce results about these new Algebraic structures.


mortberg:
This PR has some conflicts that need to be resolved. The InftyMagma file has been deleted in favor for the more general https://github.com/agda/cubical/blob/master/Cubical/Structures/NAryOp.agda. To see how to use it take a look at https://github.com/agda/cubical/blob/master/Cubical/Structures/Monoid.agda. 

Once you've fixed the conflicts I'll review the code. Thanks for the contribution!

Cubical/Structures/Monoid.agda
```diff
@@ -43,6 +43,47 @@ monoid-structure = add-to-structure (raw-monoid-structure) monoid-axioms
...
+
+module monoid-syntax where
+  id : (M : Monoids {ℓ}) → ⟨ M ⟩
+  id = monoid-id
+
+  monoid-operation-syntax : (M : Monoids {ℓ}) → ⟨ M ⟩ → ⟨ M ⟩ → ⟨ M ⟩
+  monoid-operation-syntax = monoid-operation
+
+  infixr 20 monoid-operation-syntax
+  syntax monoid-operation-syntax M x y = x ·⟨ M ⟩ y
```
mortberg, line 65: Quick question: would it be possible to use instance arguments to automatically infer `M`? @Saizan

Cubical/Algebra/SymmetricGroup.agda
```diff
@@ -0,0 +1,68 @@
...
+infixr 20 _∘_
+
+_∘_ : {A B C : Type ℓ} → B ≃ C → A ≃ B → A ≃ C
+g ∘ f = compEquiv f g
+
+∘-assoc : {A B C D : Type ℓ} (h : C ≃ D) (g : B ≃ C) (f : A ≃ B)
+        → h ∘ (g ∘ f) ≡ (h ∘ g) ∘ f
+∘-assoc h g f = equivEq _ _ refl
+
+-- Move computation rules of equivalences to Cubical.Foundations.Equiv?
```
mortberg, line 35: Yes, they should be moved here: https://github.com/agda/cubical/blob/master/Cubical/Foundations/Equiv.agda#L97

Cubical/Algebra/SymmetricGroup.agda
```diff
@@ -0,0 +1,68 @@
...
+-- Finite symmetrics groups
+Sym : ℕ → Groups
+Sym n = Symmetric-Group (Fin n) isSetFin
+
+-- Find where this is defined :
+_≢_ : {X : Type ℓ} → X → X → Type ℓ
+x ≢ y = ¬ (x ≡ y)
+
+
+℘ : Type ℓ → Type (ℓ-suc ℓ)
```
mortberg, line 62: This is going to be added by https://github.com/agda/cubical/pull/285 so please remove here

Cubical/Structures/AbGroup.agda
```diff
@@ -0,0 +1,47 @@
...
+open import Cubical.Foundations.SIP renaming (SNS-PathP to SNS)
+
+open import Cubical.Structures.NAryOp
+open import Cubical.Structures.Group
+
+private
+  variable
+    ℓ : Level
+
+raw-abelian-group-structure : Type ℓ → Type ℓ
```
mortberg, line 17: I wouldn't introduce this and instead just use `raw-group-structure`, the same holds for some of the other definitions below. Abelian groups should just be introduced by adding the commutativity axiom to groups

Cubical/Structures/AbGroup.agda
```diff
@@ -0,0 +1,47 @@
...
+raw-abelian-group-is-SNS = raw-group-is-SNS
+
+abelian-group-axioms : (X : Type ℓ) → raw-abelian-group-structure X → Type ℓ
+abelian-group-axioms X _·_ = group-axioms X _·_ ×
+                             ((x y : X) → x · y ≡ y · x)
+
+abelian-group-structure : Type ℓ → Type ℓ
+abelian-group-structure = add-to-structure raw-abelian-group-structure abelian-group-axioms
+
+AbGroups : Type (ℓ-suc ℓ)
```
mortberg, line 30: I would call this `AbGroup` as each element will be an abelian group

Cubical/Structures/Group.agda
```diff
@@ -0,0 +1,171 @@
...
+  variable
+    ℓ ℓ' : Level
+
+raw-group-structure : Type ℓ → Type ℓ
+raw-group-structure = raw-semigroup-structure
+
+raw-group-is-SNS : SNS {ℓ} raw-group-structure _
+raw-group-is-SNS = raw-semigroup-is-SNS
+
+group-axioms : (G : Type ℓ) → raw-group-structure G → Type ℓ
```
mortberg, line 26: Why define groups like this? I think it's better to separate the operations from the axioms as it simplifies the proof that the axioms form a prop later?

Cubical/Structures/Group.agda
```diff
@@ -0,0 +1,171 @@
...
+  where
+  i = semigroup-axioms G _·_
+
+  ii = Σ[ e ∈ G ] ((x : G) → (x · e ≡ x) × (e · x ≡ x)) ×
+                  ((x : G) → Σ[ x' ∈ G ] (x · x' ≡ e) × (x' · x ≡ e))
+
+group-structure : Type ℓ → Type ℓ
+group-structure = add-to-structure raw-group-structure group-axioms
+
+Groups : Type (ℓ-suc ℓ)
```
mortberg, line 38: This should be `Group`

Cubical/Structures/Group.agda
```diff
@@ -0,0 +1,171 @@
...
+
+  α : (e : X) → isProp (is-identity e)
+  α e = isPropΠ (λ _ → isPropΣ (group-is-set 𝒢 _ _) (λ _ → group-is-set 𝒢 _ _))
+
+  β : (e : X) → is-identity e → isProp ((x : X) → Σ[ x' ∈ X ] (x ·⟨ 𝒢 ⟩ x' ≡ e) × (x' ·⟨ 𝒢 ⟩ x ≡ e))
+  β e is-identity-e =
+   isPropΠ λ { x (x' , _ , P) (x'' , Q , _) → ΣProp≡ (λ _ → isPropΣ (group-is-set 𝒢 _ _) λ _ → group-is-set 𝒢 _ _)
+                                                      (inv-lemma ℳ x x' x'' P Q) }
+   where
+    ℳ : Monoids
```
mortberg, line 148: I think `Monoids` should be `Monoid`

Cubical/Structures/Rng.agda
```diff
@@ -0,0 +1,52 @@
+{-# OPTIONS --cubical --safe #-}
+
+module Cubical.Structures.Rng where
```
mortberg, line 3: I don't think we should start with adding Rng but Ring. Is there any reason why you want to do rings without unit?

Cubical/Structures/Group.agda
borsiemir: Well, the main reason is that when you apply SIP with groups with an inverse function (the same applies when you put the neutral element in the raw-structure, as we can see from the Monoid implementation) you get that an isomorphism is an equivalence that preserves multiplication and inversion, while group isomorphisms are defined as preserving only multiplication. I thought an alternative that was to define a second notion of isomorphism (as equivalences that only preserves multiplication) and prove that in fact they were the same (since it can be proven that a function that respect multiplication respects inversion). But I thought it was more difficult that this and actually ,i think more faithful to the mathematical tradition (where Groups are normally defined as having only a multiplication and the inversion and neutral element are defined from axioms).

Cubical/Structures/Rng.agda
borsiemir: I saw them in Escardo's development, I look up for them and I noticed they exist. Since I want to add as many algebraic structures as possible and rings can be defined from rngs I decided to implement them first.

Cubical/Structures/AbGroup.agda
borsiemir: Totally agree on this, I first thought I would be better to have each structure defined in the exact same style but thanks to your comment I totally changed my mind, I'm going to fix it.

Cubical/Algebra/SymmetricGroup.agda
borsiemir: Done

Cubical/Structures/Group.agda
borsiemir: Done

Cubical/Structures/Group.agda
borsiemir: Done

Cubical/Structures/AbGroup.agda
borsiemir: Done

Cubical/Algebra/BasicProp.agda
```diff
@@ -0,0 +1,77 @@
...
+-- Groups basic properties
+---------------------------------------------------------------------
+
+-- We will use the multiplicative notation for groups
+
+module _ (G : Groups {ℓ}) where
+
+  open multiplicative-notation G
+
+  id-is-unique : isContr (Σ[ x ∈ ⟨ G ⟩ ] ∀ (y : ⟨ G ⟩) → (x · y ≡ y) × (y · x ≡ y))
```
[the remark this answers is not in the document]
borsiemir: Should I wait until that notation is added or should I add it?

Cubical/Structures/AbGroup.agda
borsiemir: I changed the only raw-abelian-group-structure for raw-group-structure, should I change both for raw-semigroup-structure (in case we finally stick to the actual raw-group-structure)?

Cubical/Algebra/SymmetricGroup.agda
borsiemir: Done

Cubical/Algebra/SymmetricGroup.agda
borsiemir: Done

Cubical/Algebra/BasicProp.agda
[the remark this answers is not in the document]
mortberg: It should be used in a lot of places, so it would be good to do in a separate PR. I'll make an issue about it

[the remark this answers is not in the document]
mortberg: I actually already did!

Cubical/Structures/Group.agda
mortberg: Ah I see, that makes sense. Can you add a comment about it before `group-axioms`?

Cubical/Structures/Rng.agda
mortberg: What is the reason to add as many algebraic structures as possible? It seems more reasonable to just add the ones that we actually have instances of in order not to pollute the library with lots of structures that have no examples. I would personally skip also rings and go directly for commutative rings

Cubical/Algebra/SymmetricGroup.agda
```diff
@@ -0,0 +1,37 @@
...
+private
+  variable
+    ℓ ℓ' : Level
+
+infixr 20 _∘_
+
+_∘_ : {A B C : Type ℓ} → B ≃ C → A ≃ B → A ≃ C
+g ∘ f = compEquiv f g
+
+∘-assoc : {A B C D : Type ℓ} (h : C ≃ D) (g : B ≃ C) (f : A ≃ B)
```
mortberg, line 24: Can you upstream this result as well? I don't think we need the notation for compEquiv as we already use it for function composition

Cubical/Structures/Rng.agda
borsiemir: Okay, I will eliminate Rng and change them for Ring, although it may be interesting to have non-commutative rings since someone might want them for matrices? (I don't know if they are in the cubical library, neither if they are interesting in this library) Could you send me an email with the rest of Algebraic structures that may be interesting for the library?

Cubical/Structures/Group.agda
borsiemir: -- The neutral element and the inverse function will be derived from the
-- axioms, instead of being defined in the raw-group-structure in order
-- to have that isomorphisms between groups are equivalences that preserves
-- multiplication (so we don't have to show that they also preserve inversion
-- and neutral element, although they will preserve them).

Is this good enough?

Cubical/Algebra/SymmetricGroup.agda
borsiemir: Okay, I will move the result and eliminate the notation.

Cubical/Structures/Rng.agda
guillaumebrunerie: One reason for having intermediate structures is that it makes it easier to deal with the day you do need them. If you start with commutative rings and then one day need rings, then presumably commutative rings will need to be redefined (to make use of rings) and a lot of code about commutative rings will need to be updated.
I don’t know if that applies in this situation though, this is just a random thought.

Cubical/Structures/Group.agda
mortberg: Sounds good!

Cubical/Structures/Rng.agda
mortberg: Fair enough, let's have rings and commutative rings then. :-) With those structures I feel like we have a good amount of algebraic structures for now and the goal should be to develop some theory about what we have and prove that things are instances of them. 

For the design of the library there are a bunch of things I would like to explore before adding more structures: 

- What to do about the definitions using Records? Nested Sigma types are annoying to use and quite inefficient, but the SIP works best for them. I have an issue about this here: https://github.com/agda/cubical/issues/288

- Can we automate boring algebraic reasoning for the structures that we have? I don't think we can get very far unless we have some automation for helping us with algebraic reasoning.

- Can one use instance arguments to make the operations less annoying to use (see https://github.com/agda/cubical/pull/284/files/3c4c177a4000f748cc3ff6f6c9b77628c2517faa#diff-8b6c459df58628fab692db0ebb992ab8R65). My understanding is that instance arguments are Agda's version of type classes so I guess they can be used for this kind of things? I've never used them though so I might be way off, but without some better inference I don't think the library would be fun to use.

Cubical/Algebra/SymmetricGroup.agda
mortberg: Great!

Cubical/Structures/Rng.agda
borsiemir: I have removed Rng, an added Ring and CommRing. Until we can do the definitions with Records should I define extractors as I have done in Group? I see them really useful. If that
is the case, could I add \<_\> as a syntax that may be imported (so that lives inside a module) from Cubical.Foundation.Structure of the function typ? That will solve that \<_\> must be hidden every time you have two algebraic structures in the same file.

Cubical/Structures/Monoid.agda
borsiemir: Today I've read the documentation about instances and I'm note sure if this is what you want but since records and modules are really related in agda (I think I've read that records are modules or modules are records, but don't trust me on this) you can use the modules multiplicative-notation and additive-notation in the next style:
open import multiplicative-notation {{...}} 
and the argument (G : Group) becomes an instance argument.

mortberg: I see. So we won't have to write it explicitly if it's clear from the context which group we are referring to?

Cubical/Structures/Rng.agda
mortberg: Yeah, having extractors is good for now

Cubical/Algebra/BasicProp.agda
```diff
@@ -0,0 +1,77 @@
...
+                         ((x ⁻¹) · x) · z ≡⟨ cong (_· z) (⁻¹· x) ∙ ₁· z ⟩
+                         z ∎
+
+  right-cancel : ∀ (x y z : ⟨ G ⟩) → x · z ≡ y · z → x ≡ y
+  right-cancel x y z eq = x                ≡⟨ sym (cong (x ·_) (·⁻¹ z) ∙ ·₁ x) ⟩
+                          x · (z · (z ⁻¹)) ≡⟨ ·-is-assoc _ _ _ ⟩
+                          (x · z) · (z ⁻¹) ≡⟨ cong (_· (z ⁻¹)) eq ⟩
+                          (y · z) · (z ⁻¹) ≡⟨ sym (·-is-assoc _ _ _) ⟩
+                          y · (z · (z ⁻¹)) ≡⟨ cong (y ·_) (·⁻¹ z) ∙ ·₁ y ⟩
+                          y ∎
```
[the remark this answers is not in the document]
mortberg: @felixwellen This kind of proofs would be nice to automate

mortberg:
@borsiemir I would be happy to merge this now, or is there anything else you want to add?

Cubical/Structures/Monoid.agda
borsiemir: Well, if you have declare an instance and the groups have distinct domain I´m sure it works, I have tried it with multiplicative-notation from Group. But if you want the same syntax for monoids and all the structures I have to define it yet, I can write the code tonight.

borsiemir:
If you wait until tomorrow I will add the extractors that are missing and more notations which can have instance arguments.

mortberg:
> If you wait until tomorrow I will add the extractors that are missing and more notations which can have instance arguments.

That sounds good! Please ping me when you're done

Cubical/Structures/Monoid.agda
borsiemir: With this I'm having two different alternatives:
1. Different structures can't have the same notation at the same time, so if we have _\cdot_ for groups in one file then we can't have _\cdot_ for monoids also in the same file.
2. Every structure have every notation. So if we have a group and a monoid in the same file, and we import the _+_ notation for the Group and _\cdot_ notation for the monoid nothing will stop of us typing x + y where x and y are elements of the group (Note that the same is true for 2 distincts groups in the option 1).
For me the ideal will be a mixture but at the moment I don't know how to implement it.
Any preferences between the above 2 options?

borsiemir: Actually I think I got a solution, sorry for the ping.

borsiemir:
@mortberg The extractors were added and I added a notation for structures that didn't have any. However this notation is very basic and I hope I can improve it in a new PR when I have thought what is the best way of doing it.

mortberg:

PR #285, opened 2020-04-16
Powerset

mzeuner:
We introduce some useful notation à la Escardó for the powerset and subsets of a type. We then prove that the powerset is the subtype classifier.


Cubical/Foundations/Logic.agda
```diff
@@ -273,3 +277,96 @@ Decₚ P = Dec [ P ] , isPropDec (snd P)
...
   ⇐∶ λ pq → (fst ∘ pq ) , (snd ∘ pq)
+
+--------------------------------------------------------------------------------
+-- Introduce the "powerset" of a type in the style of Escardó's lecture notes:
+-- https://www.cs.bham.ac.uk/~mhe/HoTT-UF-in-Agda-Lecture-Notes/HoTT-UF-Agda.html#propositionalextensionality
+
+ℙ : ∀ {ℓ} → Type ℓ → Type (ℓ-suc ℓ)
+ℙ {ℓ} X = X → hProp ℓ
+
+_∈_ : {X : Type ℓ} → X → ℙ X → Type ℓ
```
mortberg, line 288: This makes definitions look so nice! Thanks @martinescardo for coming up with it

Cubical/Foundations/Logic.agda
```diff
@@ -273,3 +277,96 @@ Decₚ P = Dec [ P ] , isPropDec (snd P)
...
+⊆-refl-consequence {X} A B p = subst (λ B → (A ⊆ B)) p (⊆-refl A) , subst (λ A → (B ⊆ A)) (sym p) (⊆-refl B)
+
+
+⊆-extensionality : {X : Type ℓ} (A B : ℙ X)
+                 → (A ⊆ B) × (B ⊆ A) → A ≡ B
+⊆-extensionality A B (φ , ψ) i x = ⇔toPath {P = (A x)} {Q = (B x)} (φ x) (ψ x) i
+
+
+powersets-are-sets : {X : Type ℓ} → isSet (ℙ X)
+powersets-are-sets {X = X} = isSetΠ (λ _ → isSetHProp)
```
mortberg, line 315: Maybe it would be nice to add `isSet→` and friends to avoid having to write `λ _ →` all the time

mortberg: I added an issue about this

PR #287, opened 2020-04-17
Two more definitions of ℚ and arithmetic for ℚ

m-yac:
I don't have time to finish completing and cleaning my old code right now, so I'm just making a small PR adding two more equivalent definitions of the rationals. Proofs of their equivalence and arithmetic operations on them are still needed.


Cubical/HITs/Rational/Base.agda
```diff
@@ -4,15 +4,49 @@ module Cubical.HITs.Rational.Base where
...
+  [ pos a / b ] = let (p , q) = toCoprime (a , b) in ((pos p , q) , toCoprimeAreCoprime (a , b))
+  [ neg a / b ] = let (p , q) = toCoprime (a , b) in ((neg p , q) , toCoprimeAreCoprime (a , b))
+  [ posneg i / b ] = ((posneg i , 1) , toCoprimeAreCoprime (0 , b))
+
+-- ℚ as a set quotient of ℤ × ℕ₊₁ (from the HoTT book)
+
+module Quo where
+
+  ℚ : Type₀
+  ℚ = (ℤ × ℕ₊₁) // (λ (a , 1+ b) (c , 1+ d) → a * pos (suc b) ≡ c * pos (suc d))
```
potato4444, line 37: `b` and `d` seem switched in the relation

```suggestion
  ℚ = (ℤ × ℕ₊₁) // (λ (a , 1+ b) (c , 1+ d) → a * pos (suc d) ≡ c * pos (suc b))
```

mortberg: Yeah I think so as well?

m-yac:
Finally had some time to make my old code work. I was stuck on `isGCD-multʳ`, `*-assoc`, and `*-distrib`. To prove the latter two with such relative brevity (compared to the proofs in the [agda-numerics library](https://github.com/danr/Agda-Numerics/tree/master/Data/Integer/Ring) and [the stdlib](https://github.com/agda/agda-stdlib/blob/master/src/Data/Integer/Properties.agda)), I changed the definition of `HITs.Ints.QuoInt` to use a single constructor `signed` instead of `pos` and `neg`, inspired by the views used [here in the agda-numerics library](https://github.com/danr/Agda-Numerics/blob/master/Data/Integer/Ring/Multiplication.agda#L61).

Sorry for making such a large PR. If it helps the review process, I put below the list of major changes/additions. I would also be happy to break up this PR into multiple smaller ones if that makes things easier.
- add `isGCD-multʳ` to Data.Nat.GCD
- add `toCoprime-idem` and `toCoprime-cancelʳ` to `Data.Nat.Coprime`
- change definition of `ℤ` to use `signed` instead of `pos`, `neg`
- replace `Sign`, `_*S_`, etc. in `HITS.Ints.QuoInt` with renamed `Bool`, `_⊕_`, etc.
- flip argument order of `_+_` in `HITs.Ints.QuoInt` to match `ℕ._+_`
- prove `+-comm`, `+-assoc`, `*-assoc`, `*-distrib` for `QuoInt`
- refactor `HITs.Rational.Base` into `HITs/Rationals/...`
- addition, multiplication, and their properties for `Rationals.QuoQ`
- equivalence of `Rationals.QuoQ` and `Rationals.SigmaQ`

Cubical/HITs/Ints/QuoInt/Base.agda
```diff
@@ -8,17 +8,60 @@ open import Cubical.Foundations.Prelude
...
-open import Cubical.Data.Nat hiding (elim)
+open import Cubical.Data.Int as Int using (Int; sucInt; predInt; discreteInt; isSetInt)
+open import Cubical.Data.Nat as ℕ using (ℕ; zero; suc)
+open import Cubical.Data.Bool as Bool using (Bool; not; notnot)
 
 variable
   l : Level
 
+
+Sign : Type₀
```
mortberg, line 22: Was there any benefit to using Bool for Sign instead of the old Sign datatype?

Cubical/HITs/Rationals/QuoQ/Properties.agda
```diff
@@ -0,0 +1,245 @@
...
+ℚ-cancelˡ {a} {b} c = eq/ _ _
+  (cong (ℤ._* ℕ₊₁→ℤ b) (ℤ.*-comm (ℕ₊₁→ℤ c) a) ∙ sym (ℤ.*-assoc a (ℕ₊₁→ℤ c) (ℕ₊₁→ℤ b)))
+
+ℚ-cancelʳ : ∀ {a b} (c : ℕ₊₁) → [ a ℤ.* ℕ₊₁→ℤ c / b *₊₁ c ] ≡ [ a / b ]
+ℚ-cancelʳ {a} {b} c = eq/ _ _
+  (sym (ℤ.*-assoc a (ℕ₊₁→ℤ c) (ℕ₊₁→ℤ b)) ∙ cong (a ℤ.*_) (ℤ.*-comm (ℕ₊₁→ℤ c) (ℕ₊₁→ℤ b)))
+
+-- useful functions for defining operations on ℚ
+
+onCommonDenom :
```
mortberg, line 28: Wow, this proof is so scary... Makes me think of Brainfuck programming :-)

Cubical/HITs/Rationals/HITQ/Base.agda
```diff
@@ -0,0 +1,36 @@
...
+open import Cubical.Relation.Nullary
+
+open import Cubical.HITs.Ints.QuoInt
+
+open import Cubical.Data.Nat as ℕ hiding (_*_)
+open import Cubical.Data.NatPlusOne
+open import Cubical.Data.Sigma
+
+
+-- ℚ as a higher inductive type
```
mortberg, line 15: It would be great to add an issue about proving that this is equivalent to the other representation(s)!

Cubical/HITs/Rationals/QuoQ/Base.agda
```diff
@@ -0,0 +1,76 @@
+{-# OPTIONS --cubical --safe #-}
```
mortberg, line 1: Can you add a little header explaining what this representation is?

Cubical/HITs/SetQuotients/Properties.agda
```diff
@@ -56,6 +56,18 @@ elimProp Bprop f (squash/ x y p q i j) =
     g = elimProp Bprop f
 elimProp Bprop f (eq/ a b r i) = elimEq/ Bprop (eq/ a b r) (f a) (f b) i
 
+elimProp2 : {P : A / R → A / R → Type ℓ} (Pprop : ∀ x y → isProp (P x y))
```
mortberg, line 59: @felixwellen : Wow, the lemmas I wanted you to add are here!

mortberg:
PS: sorry for my very slow review! I've been really busy this semester...

Cubical/HITs/Ints/QuoInt/Base.agda
m-yac: Yes, I didn't have to redefine `⊕`, `⊕-comm`, and `notnot` for `Sign`. I figured using `Bool` here was a nice way to get rid of some obvious code duplication.

Cubical/HITs/Rationals/HITQ/Base.agda
m-yac: Great, will do once this is merged!

Cubical/HITs/Rationals/QuoQ/Properties.agda
m-yac: Lots of these proofs are very scary looking but have very little content... I wonder if #297 could be used here .

mortberg:
Do all files in this PR have the ` --no-import-sorts` flag? It seems like they don't when I look at the diff.

Apart from that I would be happy to merge

mortberg:
I added "--no-import-sorts" myself so will merge as soon as Travis is done. Thanks for this PR!

PR #289, opened 2020-04-20
Update Cubical.HITs.S1.Base.

favonia:
Some Geometry was lost in translation, it seems.


favonia:
PS: The version with the useless i still works, but is too ugly.

PR #295, opened 2020-04-28
Give a name to the second projection from `hProp`

ayberkt:
The first projection from `hProp` has a name: `[_]`. I think it would be nice to have a name for the second projection as well. This is called [`holds-is-subsingleton`](https://www.cs.bham.ac.uk/~mhe/HoTT-UF-in-Agda-Lecture-Notes/HoTT-UF-Agda.html#249731) in the [HoTT/UF development of Martín Escardó](https://www.cs.bham.ac.uk/~mhe/HoTT-UF-in-Agda-Lecture-Notes/).

### DIFF
--- a/Cubical/Foundations/Logic.agda
+++ b/Cubical/Foundations/Logic.agda
@@ -70,11 +70,11 @@ hProp≡ p = ΣProp≡ (\ _ → isPropIsProp) p
 -- Logical implication of mere propositions
 
 _⇒_ : (A : hProp ℓ) → (B : hProp ℓ') → hProp _
-A ⇒ B = ([ A ] → [ B ]) , isPropΠ λ _ → B .snd
+A ⇒ B = ([ A ] → [ B ]) , isPropΠ λ _ → isProp[] B
 
 ⇔toPath : [ P ⇒ Q ] → [ Q ⇒ P ] → P ≡ Q
 ⇔toPath {P = P} {Q = Q} P⇒Q Q⇒P = hProp≡ (isoToPath
-  (iso P⇒Q Q⇒P (λ b → Q .snd (P⇒Q (Q⇒P b)) b) λ a → P .snd (Q⇒P (P⇒Q a)) a))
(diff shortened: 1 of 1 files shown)


mortberg:
Great, shouldn't this be used throughout the library?

ayberkt:
> Great! Shouldn't this be used throughout the library?

Yes, I guess I should propagate this change to all uses of the second projection on `hProp`s.

Perhaps, we can even keep the type abstract and expose only the two projections (along with an introduction function) to ensure that these names are always used.

mortberg:
> > Great! Shouldn't this be used throughout the library?
> 
> Yes, I guess I should propagate this change to all uses of the second projection on `hProp`s.
> 
> Perhaps, we can even keep the type abstract and expose only the two projections (along with an introduction function) to ensure that these names are always used.

I'm all for propagating the change, but keeping the type abstract and exposing only the projections seems a bit much. But if other people agree with this I have no super strong opinions about it

guillaumebrunerie:
> Perhaps, we can even keep the type abstract and expose only the two projections (along with an introduction function) to ensure that these names are always used.

Note that you don’t need to use the `abstract` machinery for that, you can simply define it as a new record with the names that you want for the two fields.

mortberg:
> > Perhaps, we can even keep the type abstract and expose only the two projections (along with an introduction function) to ensure that these names are always used.
> 
> Note that you don’t need to use the `abstract` machinery for that, you can simply define it as a new record with the names that you want for the two fields.

Oh, of course! That seems much better

ayberkt:
> but keeping the type abstract and exposing only the projections seems a bit much

On second thought, it does seem like an overkill.

I will simply propagate the renaming, which I have already done for the `Logic` module. However, I am not sure what would be an efficient strategy for finding all applications of `snd` to `hprop`s.

mortberg:
I'll merge this one now and make an issue about it instead

ayberkt:
Sorry, I haven't had the time to look at this.

PR #296, opened 2020-04-28
Fix naming in syntax declarations for ∃

ayberkt:
It seemed to me that `∃[]-syntax` and `∃[∶]-syntax` were flipped in the syntax declarations.

https://github.com/agda/cubical/blob/acabbd9595ad4291fd6f4bd2b1e802d285f676ed/Cubical/Foundations/Logic.agda#L171-L172


mortberg:
Thanks!

PR #298, opened 2020-05-02
Fix uncaught problem with addition of Powersets

WorldSEnder:
7b8739cc4cd843bb46c2e27501b1dc22a8a079ed introduced a regression in ListedFiniteSets that was not caught by the CI due to ignoring the result of checking the files. This fixes both the CI process and the problem in LFS. Also removes Algebra/Everything.agda since that file is overwritten anyway when running `make`.


mortberg:
Great, thanks!

PR #299, opened 2020-05-03
Cohomology group structure

aljungstrom:
This PR contains a definition of addition in in n-truncated n-sphere (K n), and, using this, a definition of addition in cohomology groups. I have also given the group laws for these operations. Most of the lemmas in this PR are lemmas leading up to these definitions/laws and some preliminary stuff for the definition of the cup product.

`KcompPrelims.agda` contains a proof of K n ≃ Ω (K (n+1)), which is used to define addition in K n. Most of it is just translations of stuff that has previously been proved for S1 defined using base/loop into the new definition using suspensions. Things compute fairly ok for n = 0 and n = 1, but for n > 1 the map from Ω (K (n+1)) to K n computes horribly. I have played around quite a bit to get things to compute better (with some luck), but the use of Freudenthal in the n > 1 case seems to be problematic.

I also added some Iso-versions of equivalences in the library in order to get better access to the maps involved.


Cubical/HITs/SetTruncation/Properties.agda
```diff
@@ -74,3 +75,20 @@ setTruncIdempotent≃ {A = A} hA = isoToEquiv f
...
+  fun = rec setTruncIsSet λ {(a , p) → ∣ a , ∣ p ∣₀ ∣₀}
+  funinv : ∥ Σ A (λ x → ∥ B x ∥₀) ∥₀ → ∥ Σ A B ∥₀
+  funinv = rec setTruncIsSet (λ {(a , p) → rec setTruncIsSet (λ p → ∣ a , p ∣₀) p})
+  sect : section fun funinv
+  sect = elim (λ _ → isOfHLevelPath 2 setTruncIsSet _ _)
+              λ { (a , p) → elim {B = λ p → fun (funinv ∣ a , p ∣₀) ≡ ∣ a , p ∣₀}
+              (λ p → isOfHLevelPath 2 setTruncIsSet _ _) (λ p → refl) p }
+  retr : retract fun funinv
+  retr = elim (λ _ → isOfHLevelPath 2 setTruncIsSet _ _)
+              λ { (a , p) → refl }
```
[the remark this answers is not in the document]
Saizan: Pattern matching lambdas on records with eta are quite well-behaved, actually, so they shouldn't cause problems.

Cubical/HITs/SmashProduct/Base.agda
```diff
@@ -3,23 +3,45 @@ module Cubical.HITs.SmashProduct.Base where
...
   gluel : (a : typ A) → proj a (pt B) ≡ basel
   gluer : (b : typ B) → proj (pt A) b ≡ baser
 
+private
+  variable
+    ℓ ℓ' ℓ'' ℓ''' : Level
+    A : Pointed ℓ
+    B : Pointed ℓ'
+    C : Pointed ℓ''
+    D : Pointed ℓ'''
```
mortberg, line 27: @Saizan : I'm always confused about how `variable` works in Agda, but I don't think you need all of these levels to be different?

Cubical/HITs/SmashProduct/Base.agda
```diff
@@ -28,18 +50,172 @@ commK (gluer b x) = refl
...
+
+--- Alternative definition
+
+i∧ : ∀ {ℓ ℓ'} {A : Pointed ℓ} {B : Pointed ℓ'} → A ⋁ B → (typ A) × (typ B)
+i∧ {A = A , ptA} {B = B , ptB} (inl x) = x , ptB
+i∧ {A = A , ptA} {B = B , ptB} (inr x) = ptA , x
+i∧ {A = A , ptA} {B = B , ptB} (push tt i) = ptA , ptB
+
+_⋀_ : ∀ {ℓ ℓ'} → Pointed ℓ → Pointed ℓ' → Type (ℓ-max ℓ ℓ')
+_⋀_ A B = Pushout {A = (A ⋁ B)} {B = Unit} {C = (typ A) × (typ B)} (λ _ → tt) i∧
```
mortberg, line 67: Does one really have to give all of these implicit arguments to Pushout?

Cubical/HITs/SmashProduct/Base.agda
```diff
@@ -28,18 +50,172 @@ commK (gluer b x) = refl
...
+           (λ j → (push (inl (fpt (~ j))) ∙ λ i → inr (fpt ((~ j) ∨ i) , (gpt (~ i)))) ∙ λ i → inr ((fpt (~ i)) , (g (pt B)))) ∙
+           (λ j → (push (inl (f (pt A))) ∙ λ i → inr (fpt (i ∧ (~ j)) , gpt ((~ i) ∨ j))) ∙ λ i → inr ((fpt ((~ i) ∧ (~ j))) , (gpt ((~ i) ∧ j)))) ∙
+           (λ j → (rUnit (push (inl (f (pt A)))) (~ j)) ∙ λ i → inr ((f (pt A)) , (gpt (~ i))))
+{- perhaps rewrite this with hcomps ... -}
+
+⋀→Smash : A ⋀ B → Smash A B
+⋀→Smash (inl x) = basel
+⋀→Smash (inr (x , x₁)) = proj x x₁
+⋀→Smash (push (inl x) i) = gluel x (~ i)
+⋀→Smash {A = A} {B = B} (push (inr x) i) = (sym (gluel (snd A)) ∙ gluer (snd B) ∙ sym (gluer x)) i
```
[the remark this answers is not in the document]
mortberg: Yup

Cubical/HITs/SmashProduct/Base.agda
```diff
@@ -28,18 +50,172 @@ commK (gluer b x) = refl
...
+SmashAssociate = comm ∘ Smash-map  (comm , refl) (idfun∙ _) ∘ rearrange
+
+SmashAssociate⁻ : Smash A (Smash∙ B C) → Smash (Smash∙ A B) C
+SmashAssociate⁻ = rearrange ∘ comm ∘ Smash-map (idfun∙ _) (comm , refl)
+
+⋀-associate : (A ⋀∙ B) ⋀ C → A ⋀ (B ⋀∙ C)
+⋀-associate = (idfun∙ _ ⋀⃗ Smash∙→⋀∙) ∘ Smash→⋀ ∘ SmashAssociate ∘ ⋀→Smash ∘ (⋀∙→Smash∙ ⋀⃗ idfun∙ _)
 
+⋀-associate⁻ : A ⋀ (B ⋀∙ C) → (A ⋀∙ B) ⋀ C
+⋀-associate⁻ = (Smash∙→⋀∙ ⋀⃗ idfun∙ _) ∘ Smash→⋀ ∘ SmashAssociate⁻ ∘ ⋀→Smash ∘ (idfun∙ _ ⋀⃗ ⋀∙→Smash∙)
```
mortberg, line 221: Very nice! Now we just need the pengaton! ;-)

Cubical/Homotopy/Loopspace.agda
```diff
@@ -22,3 +24,81 @@ open import Cubical.Foundations.GroupoidLaws
...
+                  ; (z = i0) → ((α ∙ β) ⋆ (α' ∙ β')) i })
+        (bottom z i)
+  where
+  bottom : PathP (λ i → p ∙ p' ≡ (β ⋆ β') (~ i)) ((α ∙ β) ⋆ (α' ∙ β')) (α ⋆ α')
+  bottom i =
+    hcomp (λ k → λ { (i = i0) → (α ∙ β) ⋆ (α' ∙ β')
+                    ; (i = i1) → rUnit α (~ k) ⋆ rUnit α' (~ k)})
+          ((α ∙ λ j → β (~ i ∧ j)) ⋆ (α' ∙ λ j → β' (~ i ∧ j)))
+
+Eckmann-Hilton : ∀ {ℓ} {A : Pointed ℓ} (n : ℕ) (α β : typ ((Ω^ (2 + n)) A))
```
mortberg, line 52: Isn't there a more direct proof of this using J?

(pint @bbentzen)

Cubical/HITs/Sn/Properties.agda
```diff
@@ -0,0 +1,105 @@
...
+SuspBool→S1 (merid false i) = merid south i
+SuspBool→S1 (merid true i) = merid north i
+
+S1→SuspBool : S₊ 1 → Susp Bool
+S1→SuspBool north = north
+S1→SuspBool south = south
+S1→SuspBool (merid north i) = merid true i
+S1→SuspBool (merid south i) = merid false i
+
+SuspBool≃S1 : Susp Bool ≃ S₊ 1
```
mortberg, line 54: Is it necessary to prove this so explicity? There is already https://github.com/agda/cubical/blob/master/Cubical/HITs/Susp/Base.agda#L62

Cubical/HITs/SmashProduct/Base.agda
```diff
@@ -28,18 +50,172 @@ commK (gluer b x) = refl
...
+  helper : Path (Path (C ⋀ D) (inl tt) (inr ((f (pt A)) , (g (pt B)))))
+                (push (inr (g (pt B))) ∙ λ i → inr ((fpt (~ i)) , (g (pt B))))
+                (push (inl (f (pt A))) ∙ λ i → inr ((f (pt A)) , (gpt (~ i))))
+  helper = (λ j → rUnit (push (inr (g (pt B)))) j ∙ λ i → inr ((fpt (~ i)) , (g (pt B))) ) ∙
+           (λ j → (push (inr (gpt j)) ∙ λ i → inr ((pt C) , (gpt ((~ i) ∧ j)))) ∙ λ i → inr ((fpt (~ i)) , (g (pt B)))) ∙
+           (λ j → (cong (push) (push tt) (~ j) ∙ λ i → inr (pt C , gpt (~ i))) ∙ λ i → inr ((fpt (~ i)) , (g (pt B)))) ∙
+           (λ j → (push (inl (fpt (~ j))) ∙ λ i → inr (fpt ((~ j) ∨ i) , (gpt (~ i)))) ∙ λ i → inr ((fpt (~ i)) , (g (pt B)))) ∙
+           (λ j → (push (inl (f (pt A))) ∙ λ i → inr (fpt (i ∧ (~ j)) , gpt ((~ i) ∨ j))) ∙ λ i → inr ((fpt ((~ i) ∧ (~ j))) , (gpt ((~ i) ∧ j)))) ∙
+           (λ j → (rUnit (push (inl (f (pt A)))) (~ j)) ∙ λ i → inr ((f (pt A)) , (gpt (~ i))))
+{- perhaps rewrite this with hcomps ... -}
```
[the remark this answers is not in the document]
aljungstrom: Hehe wups, that was a note for myself that I must have forgotten about... I'll see to this.

Cubical/Homotopy/Connected.agda
```diff
@@ -12,46 +12,123 @@ open import Cubical.Foundations.Path
...
+        (P b .snd)
+        (λ {(a , p) → subst (fst ∘ P) p (t a)})
+
+  isEquivPrecompose : ∀ {ℓ'''} (P : B → HLevel ℓ''' n)
+                   → isHLevelConnectedFun n f
+                   → isEquiv λ(s : (b : B) → P b .fst) → s ∘ f
+  isEquivPrecompose P fConn = isoToIsEquiv (isIsoPrecompose P fConn)
+
+  isConnectedPrecompose : ((∀ {ℓ'''} (P : B → HLevel ℓ''' n)
+                                  → hasSection λ(s : (b : B) → P b .fst) → s ∘ f))
```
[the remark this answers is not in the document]
aljungstrom: Hmm, yes, I remember that I was very annoyed about this too. If I remember correctly, the proof required me to use the assumption for a dependent type of level l''' + 1, which prevented me from fixing the universe level first. It's been quite some time since I wrote this up, so I'll have to double check.

Cubical/Homotopy/Loopspace.agda
aljungstrom: I never thought I'd hear you say that;-) I followed the proof in the HoTT book, but if there is a simpler proof I'd be happy to change it.

aljungstrom:
Thanks @mortberg and @guillaumebrunerie for the comments. Maybe we can discuss some of them on friday and I'll fix everything after that.

Cubical/Homotopy/Loopspace.agda
bbentzen: If there's one thing I learned last year is that there's no shame in using J when necessary! @aljungstrom Maybe you can have a look at Theorem 6.2.1 [here](https://arxiv.org/pdf/1911.05844.pdf). The proof is given informally, so I don't know how easy it would be to mechanize it, but the general idea could be helpful!

Cubical/HITs/SmashProduct/Base.agda
aljungstrom: I can't remember, did we come to a conclusion here? Should I leave it as it is or not?

Cubical/HITs/Sn/Properties.agda
aljungstrom: I did this to avoid having to use transports. I thought this way would compute better, but perhaps it makes no difference. Do you want me to change it or can I keep it?

Cubical/Homotopy/Loopspace.agda
aljungstrom: @bbentzen Sorry for the late reply. Thanks for the tip! I changed to your proof now and things got way shorter.  I interpreted the general statement as a filler of a square (with its corners being paths) instead of a filler of a cube, which I guess it should be (filling cubes can be so scary...). I don't know if this is super pretty, but I thought it made things easier to follow.

The only unfortunate thing is that I had to use a comp instead of an hcomp for the final statement because the general statement says something about refl*refl = refl*refl which does not reduce to refl = refl in cubical. But that should be fine.

I actually never had to use J. Maybe I proved a weaker version of the theorem without realising?

Cubical/HITs/SmashProduct/Base.agda
mortberg: Yeah, you can use \ell for A-D and get the same effect. The only time when you need to have different levels is in statements where you want more than one. So please reduce the number of levels to the actual number that you need in the file.

Cubical/HITs/Sn/Properties.agda
mortberg: I don't think it really matters for this proof, but lets keep it as it is.

Cubical/Homotopy/Loopspace.agda
```diff
@@ -22,3 +24,38 @@ open import Cubical.Foundations.GroupoidLaws
...
+module _ {ℓ : Level} {A : Type ℓ} {a b c : A} {p q : a ≡ b} {r s : b ≡ c} (α : p ≡ q) (β : r ≡ s) where
+  _◾ᵣ_ : (α : p ≡ q) (r : b ≡ c) → p ∙ r ≡ q ∙ r
+  (α ◾ᵣ r) i =
+     α i ∙ r
+
+  _◾ₗ_ : (p : a ≡ b) (β : r ≡ s) → p ∙ r ≡ p ∙ s
+  (p ◾ₗ β) i =
+    p ∙ β i
+
+  generalEH : (α ◾ᵣ r) ∙ (q ◾ₗ β) ≡ (p ◾ₗ β) ∙ (α ◾ᵣ s)
```
mortberg, line 38: Do you really need the ◾ᵣ and ◾ₗ ? The statement more or less the same if you unfold them

Cubical/Homotopy/Loopspace.agda
mortberg: Wow, the proof is really slick and cubical now!

Cubical/Homotopy/Connected.agda
```diff
@@ -93,3 +186,127 @@ isHLevelConnectedPoint n connA a₀ a =
...
+                   → isHLevelConnectedFun m f
+                   → Iso (hLevelTrunc n A) (hLevelTrunc n B)
+connectedTruncIso2 {A = A} {B = B} n m f (x , pf) con =
+  connectedTruncIso n f (isConnectedSubtr n x f (transport (λ i → isHLevelConnectedFun (pf (~ i)) f) con))
+
+connectedTruncEquiv : ∀ {ℓ} {A B : Type ℓ} (n : ℕ) (f : A → B)
+                   → isHLevelConnectedFun n f
+                   → hLevelTrunc n A ≃ hLevelTrunc n B
+connectedTruncEquiv {A = A} {B = B} n f con = isoToEquiv (connectedTruncIso n f con)
+
```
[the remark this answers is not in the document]
aljungstrom: @mortberg @guillaumebrunerie This is the only comment I havent seen to because I am not really sure where to put them. Is it ok to leave this as it is for now, and maybe split things up when we get more theorems in there?

And sorry for being so slow with fixing this PR. I got so caught up in other things..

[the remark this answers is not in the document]
guillaumebrunerie: Yes, I think it’s fine to leave it as is.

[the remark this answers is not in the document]
mortberg: Yeah, you can add a TODO about reorganizing things so that we don't forget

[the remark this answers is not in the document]
aljungstrom: Alright, added. I think that's all the requested changes.

mortberg:
Thanks a lot for this contribution! Looking forward to Mayer-Vietoris

PR #300, opened 2020-05-04
Basic definitions and properties of matrices

mortberg:
I plan to define matrices as vectors of vectors and as maps out of pairs of Fin with an equality between the two representations. So far I have just defined Fin using data and prove that vectors are equal to maps out of Fin. This should be a nice example where we can program using one definition and prove properties using another. Once we have transp for inductive families we can then transport programs and properties between the two representations.

### DIFF
--- a/Cubical/Algebra/Matrix.agda
+++ b/Cubical/Algebra/Matrix.agda
@@ -63,3 +63,22 @@ module _ (R : CommRing {ℓ}) where
   addVecMatrixComm {m} {n} = transport (λ i → (M N : FinMatrix≡VecMatrix ⟨ R ⟩ m n i)
                                             → addMatrixPath i M N ≡ addMatrixPath i N M)
                                        addFinMatrixComm
+
+
+  -- More direct definition of addition for VecMatrix:
+
+  addVec : ∀ {m} → Vec ⟨ R ⟩ m → Vec ⟨ R ⟩ m → Vec ⟨ R ⟩ m
+  addVec [] [] = []
+  addVec (x ∷ xs) (y ∷ ys) = x +⟨ R ⟩ y ∷ addVec xs ys
+
+  addVecMatrix' : ∀ {m n} → VecMatrix ⟨ R ⟩ m n → VecMatrix ⟨ R ⟩ m n → VecMatrix ⟨ R ⟩ m n
+  addVecMatrix' [] [] = []
+  addVecMatrix' (M ∷ MS) (N ∷ NS) = addVec M N ∷ addVecMatrix' MS NS
+
+  -- This proof seems tricky... It might help if transp reduces for Vec
+  addVecMatrixEq : ∀ {m n} → (M N : VecMatrix ⟨ R ⟩ m n) → addVecMatrix M N ≡ addVecMatrix' M N
+  addVecMatrixEq {zero} {n} [] [] = λ i → foo i
+    where
+    foo : transport (λ j → Vec (ua (FinVec≃Vec ⟨ R ⟩ n) j) zero) [] ≡ []
+    foo = {!!}
+  addVecMatrixEq {suc m} {n} (M ∷ MS) (N ∷ NS) = {!!}
--- a/Cubical/Algebra/Matrix.agda
+++ b/Cubical/Algebra/Matrix.agda
@@ -1,0 +1,65 @@
+{-# OPTIONS --cubical --safe #-}
+module Cubical.Algebra.Matrix where
+
+open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.Equiv
+open import Cubical.Foundations.Isomorphism
+open import Cubical.Foundations.Univalence
+
+import Cubical.Data.Empty as ⊥
+open import Cubical.Data.Nat hiding (_+_)
+open import Cubical.Data.Vec
+open import Cubical.Data.FinData
+open import Cubical.Relation.Nullary
+
+open import Cubical.Structures.CommRing
+
+private
+  variable
+    ℓ : Level
+
+
+-- Equivalence between Vec matrix and Fin function matrix
+
+FinMatrix : (A : Type ℓ) (m n : ℕ) → Type ℓ
+FinMatrix A m n = FinVec (FinVec A n) m
+
+VecMatrix : (A : Type ℓ) (m n : ℕ) → Type ℓ
+VecMatrix A m n = Vec (Vec A n) m
+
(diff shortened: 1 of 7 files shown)


Cubical/Algebra/Matrix.agda
```diff
@@ -0,0 +1,65 @@
...
+FinMatrix≡VecMatrix A m n i = FinVec≡Vec (FinVec≡Vec A n i) m i
+
+
+-- Experiment using addition. Transport commutativity from one
+-- representation to the the other.
+module _ (R : CommRing {ℓ}) where
+
+  open commring-operation-syntax
+
+  -- TODO: how to just use +? Seems to work for AbGroup...
```
mortberg, line 40: @borsiemir : Is it somehow possible to just write + for addition of commutative rings? It worked for AbGroup, but for CommRing I don't understand how to get it to work...

Cubical/Algebra/Matrix.agda
```diff
@@ -0,0 +1,84 @@
...
+
+  addVecMatrix' : ∀ {m n} → VecMatrix ⟨ R ⟩ m n → VecMatrix ⟨ R ⟩ m n → VecMatrix ⟨ R ⟩ m n
+  addVecMatrix' [] [] = []
+  addVecMatrix' (M ∷ MS) (N ∷ NS) = addVec M N ∷ addVecMatrix' MS NS
+
+  -- This proof seems tricky... It might help if transp reduces for Vec
+  addVecMatrixEq : ∀ {m n} → (M N : VecMatrix ⟨ R ⟩ m n) → addVecMatrix M N ≡ addVecMatrix' M N
+  addVecMatrixEq {zero} {n} [] [] = λ i → foo i
+    where
+    foo : transport (λ j → Vec (ua (FinVec≃Vec ⟨ R ⟩ n) j) zero) [] ≡ []
```
mortberg, line 82: @Saizan : this goal might be proved by refl if transp would reduce for Vec, or what do you think?

Saizan: Sadly not, because there will be a residual trivial `transp` on the LHS since the `zero` could be a path that depends on `j`.

mortberg: I see... Too bad! 

I guess the way to prove this is to refactor `FinMatrix≡VecMatrix` so that it only uses `ua` once on the outside which would let me use `ua-gluePath`

Cubical/Algebra/Matrix.agda
borsiemir: In this case I would write:

`open commring-\cdotsyntax R`

Since we are not defining R, it is just a parameter, I wouldn't add it to the instance resoluction algorithm. But if you want to do it in that way you could use:

`open commring-\cdotsyntax {{...}}

instance
  _ : CommRing
  _ = R`

and it will also work. However I want to open an issue about what are our exact objectives with notations and improve them, but I'm afraid that this month I'll be short on time.

Cubical/Algebra/Matrix.agda
mortberg: Yay, this refactoring worked!

Cubical/Algebra/Matrix.agda
mortberg: Thanks, this worked!

mortberg:
I don't have time to do more on this right now so I leave a few TODOs in the end of the matrix file and consider this PR ready for review. The example can be seen as proof of concept that we can use one representation of matrices for doing proofs and another one for computing while transporting functions and results between the two (so it's a more complex version of the BinNat example). 

What should we do with FinData? Maybe we should replace the current Fin with this one?

Cubical/Foundations/Univalence.agda
```diff
@@ -215,12 +215,26 @@ univalencePath = ua (compEquiv univalence LiftEquiv)
...
 uaη = J (λ _ q → ua (pathToEquiv q) ≡ q) (cong ua pathToEquivRefl ∙ uaIdEquiv)
 
+-- Useful lemma for unfolding a transported function over ua
+-- If we would have regularity this would be refl
+transportUAop₁ : ∀ {A B : Type ℓ} → (e : A ≃ B) (f : A → A) (x : B)
+               → transport (λ i → ua e i → ua e i) f x ≡ equivFun e (f (invEq e x))
+transportUAop₁ e f x i = transportRefl (equivFun e (f (invEq e (transportRefl x i)))) i
+
+-- Binary version
+transportUAop₂ : ∀ {ℓ} {A B : Type ℓ} → (e : A ≃ B) (f : A → A → A) (x y : B)
```
mortberg, line 231: This lemma is really useful for relating the transported addition function with a concrete definitions for vectors. When I tried to do this directly the goal exploded because of all the trivial transports, but with this lemma I can get rid of them easily and (see the proof of `addVecMatrixEq`). 

It should be possible to use a similar proof when proving that n-ary operations are SNS instead of using EquivJ and n-ary funext (@mzeuner )

Cubical/Algebra/Matrix.agda
```diff
@@ -0,0 +1,146 @@
...
+
+FinMatrix : (A : Type ℓ) (m n : ℕ) → Type ℓ
+FinMatrix A m n = FinVec (FinVec A n) m
+
+VecMatrix : (A : Type ℓ) (m n : ℕ) → Type ℓ
+VecMatrix A m n = Vec (Vec A n) m
+
+FinMatrix→VecMatrix : {m n : ℕ} → FinMatrix A m n → VecMatrix A m n
+FinMatrix→VecMatrix {m = zero}  _ = []
+FinMatrix→VecMatrix {m = suc m} M = FinVec→Vec (M zero) ∷ FinMatrix→VecMatrix (λ x → M (suc x))
```
Saizan, line 33: Have you considered this definition ?
```agda
FinMatrix→VecMatrix M = FinVec→Vec (\ fm -> FinVec→ (\ fn -> M fm fn))
```

Saizan:
Looks great, especially the proof between the transported addition and the direct one is quite elegant.

 `FinData` should indeed replace the current `Fin`, eventually at least.

Cubical/Algebra/Matrix.agda
mortberg: Nope, but it's a lot nicer than mine and everything typechecks so will push it soon! Good catch

mortberg:
> Looks great, especially the proof between the transported addition and the direct one is quite elegant.
> 

Yeah, I was surprised it worked out so smoothly in the end! I had a few failed attempts that were really horrible. When I didn't control the unfolding and simplified using `transportUAop₂` the goal just exploded

PR #302, opened 2020-05-05
Always index by HLevel = ℕ

m-yac:
This PR:
- adds `HLevel = ℕ` to `Foundations.HLevels` and uses it throughout the library where obviously appropriate
- uses `HLevel` for the `∥_∥_` notation in `HITs.Truncation`
- renames `∥_∥₀` to `∥_∥₂`, `∥_∥₁` to `∥_∥₃`, etc. to be consistent with the above
- refactors `isOfHLevelConnected : ℕ → ...` to `isConnected : HLevel → ...`

Two questions:
1. How should we handle `Truncation.FromNegOne`? We could refactor it to index from `ℕ₊₁` instead of from `ℕ₋₁`, but it might be better to just leave it alone. Right now I left a note stating that its indexing is inconsistent.
2. Should we delete `ℕ₋₂` all together? After this PR it would no longer be used anywhere in the library.


mortberg:
Wow, this is amazing! Thanks for fixing this. There are quite a open PRs so it might be good to merge them first as this one will cause lots of conflicts.

Re the questions:

1. Let's leave `Truncation.FromNegOne` as it is but with a comment about the indexing.
2. Let's move the `ℕ₋₂` files to Experiments with a comment in the top explaining why it's not used anymore.

mortberg:
@m-yac Can you please fix the conflicts and move the ℕ₋₂ file to experiments? As soon as this is done I'll merge

m-yac:
Ok, this should be ready to merge! Sorry for taking *so* long to finish this up.

mortberg:
Great, I'm going to merge this now so that it won't have to be updated again. Thanks a lot for fixing this!

PR #309, opened 2020-05-08
Work in progress on proving various things about equality as an inductive family

mortberg:
Using the branch `issue3733` of Agda we get support for transp for inductive families. This file contains various things that can be done using equality defined using `data`. In particular we prove funext, univalence and define some HITs.  The main test is to compute some winding numbers using equality everywhere and the standard HoTT proofs using pattern-matching on `refl`.

@martinescardo : I think this is something you've been wanting for a while. It should be possible to replace all of the postulates in your HoTT/UF library and get things to compute without having to rewrite all definitions using pattern-matching on `refl`


Cubical/Data/Equality.agda
```diff
@@ -1,50 +1,331 @@
...
+
+S¹-recβ : (b : A) (l : b ≡ b) → ap (S¹-rec b l) loop ≡ l
+S¹-recβ b l =
+  ap (S¹-rec b l) loop ≡⟨ ap≡congPath (S¹-rec b l) loopPath ⟩
+  pathToEq (congPath (S¹-rec b l) loopPath) ≡⟨ refl ⟩
+  pathToEq (eqToPath l) ≡⟨ pathToEq (pathToEq-eqToPath l) ⟩
+  l ∎
+
+
+-- This seems hard to prove?
```
mortberg, line 291: @Saizan : any ideas for how to prove this one?

mortberg:
@EgbertRijke : you might be interested in this as well for your book

Cubical/Data/Equality.agda
Saizan: Seems like you need an heterogeneous version of `ap≡congPath`, with a corresponding `pathToEqOver`.

martinescardo:
Yes, thanks, @mortberg. I actually don't have any postulate. I use univalence, function extensionality, propositional truncation etc. as hypotheses in both https://github.com/martinescardo/TypeTopology and https://github.com/martinescardo/HoTT-UF-Agda-Lecture-Notes. What I need is to be able to call my functions with implementations of these hypothesis. The problem, as you say, is that I use the identity type. I will try to understand how this solution you describe here works.

mortberg:
> Yes, thanks, @mortberg. I actually don't have any postulate. I use univalence, function extensionality, propositional truncation etc. as hypotheses in both https://github.com/martinescardo/TypeTopology and https://github.com/martinescardo/HoTT-UF-Agda-Lecture-Notes. What I need is to be able to call my functions with implementations of these hypothesis. The problem, as you say, is that I use the identity type. I will try to understand how this solution you describe here works.

Ok, so you should be able to replace the hypotheses with the implementations that I prove in this PR. The rest of your developments should then typecheck as normal, but you should be able to compute with the definitions. Note that for this to work you need the `issue3733` branch of the development version of Agda installed.  Depending on which variation of univalence you need I might be able to come up with a more direct proof than what's in this PR.

martinescardo:
I will try as soon as I find some time. @mortberg. I am looking forward to using this wonderful new feature,

Saizan:
A word of warning though: currently definitions by pattern matching that rely on injectivity of constructors, such as say `tail` for vectors, will still be stuck when applied to transported elements.

Agda would need to have access to an encode/decode proof for `Nat` to produce the right term for `tail`, and I don't know how to best arrange for that.

mortberg:
> I will try as soon as I find some time. @mortberg. I am looking forward to using this wonderful new feature,

Great! Let us know how it goes. Btw, this new feature was implemented by @Saizan. I just alpha tested it :-)

@Saizan : I wonder how often one might run into these stuck applications to transported elements?

martinescardo:
You may remember, @mortberg , that a while ago @cj-xu and I ported the Agda code of his thesis to cubical Agda, which in particular involved converting all definitions by pattern matching on refl to uses of J. This used a sheaf topos to compute a number. Perhaps this time we can try to use his original code to see if it manages to compute the same number.

martinescardo:
@Saizan , thanks for implementing this. I mostly work with a spartan subset of Agda corresponding to a spartan MLTT. I use Sigma, Pi, Nat, empty type, unit type, universes, identity type. In one occasion I use a W type for accessible relations. I suppose all this would be supported by your current implementation?

martinescardo:
I am getting this while trying to install this branch:
```
[303 of 372] Compiling Agda.TypeChecking.Generalize ( src/full/Agda/TypeChecking/Generalize.hs, dist/build/Agda/TypeChecking/Generalize.o )
[304 of 372] Compiling Agda.TypeChecking.Rules.Data ( src/full/Agda/TypeChecking/Rules/Data.hs, dist/build/Agda/TypeChecking/Rules/Data.o )

src/full/Agda/TypeChecking/Rules/Data.hs:892:21: error:
    Not in scope: type constructor or class ‘MonadFail’
    |
892 |     abstract_trD :: MonadFail m => (Vars m -> Vars m -> Vars m -> NamesT m Telescope) -> NamesT m Telescope
    |                     ^^^^^^^^^

src/full/Agda/TypeChecking/Rules/Data.hs:900:17: error:
    Not in scope: type constructor or class ‘MonadFail’
    |
900 |     bind_trD :: MonadFail m => (ArgVars m -> ArgVars m -> ArgVars m -> NamesT m b) ->
    |                 ^^^^^^^^^

src/full/Agda/TypeChecking/Rules/Data.hs:1132:23: error:
    Not in scope: type constructor or class ‘MonadFail’
     |
1132 |             argLam :: MonadFail m => String -> (Var m -> NamesT m (Arg Term)) -> NamesT m (Arg Term)
     |                       ^^^^^^^^^
cabal: Leaving directory '.'
cabal: Error: some packages failed to install:
Agda-2.6.2-LeCNzEBQdR97fWkX1e26Zx failed during the building phase. The
exception was:
ExitFailure 1
```
I am using Ubuntu 18.04, ghc version 8.6.5.

Saizan:
@martinescardo even with just equality and Nat you could be relying on injectivity of constructors in e.g. a proof like this:

```agda
suc-inj : {x y : Nat} -> suc x ≡ suc y -> x ≡ y
suc-inj refl = refl
```

About `MonadFail` it must be something about differing versions of `ghc`, I will look into what the proper way to support 8.6.5 is.

martinescardo:
Thanks, @Saizan.

(1) I don't think I use such "nasty" things: I try to restrict myself to conventional MLTT modes of definition. But if I inadvertently did something like this, what would happen when I use the branch issue3733 (to compute)?

(2) If you tell me which version of ghc works for you, I can try to use that. Don't bother trying to add support for my version.

Saizan:
@martinescardo 

(1) you would get a `No equiv while splitting on indexed family` warning when you typecheck such a definition, and when you try to compute it will get stuck if applied to an argument that couldn't be reduced to one of the declared constructors because of a `transp`. 

(2) I had to fix this to pass Agda's tests anyway, it should be fine now according to the tests 
https://github.com/agda/agda/runs/669839673
Locally I am using ghc 8.8.3.

martinescardo:
I managed to get this branch of Agda compiled - thanks. How do I need to define the identity type for this to work? If I defined it as 
```
data _≡_ {ℓ} {X : Set ℓ} : X → X → Set ℓ where
 refl : (x : X) → x ≡ x
```
then the following definition gives "No equiv while splitting on indexed family when checking the definition of J":
```
J : ∀ {ℓ ℓ'} (X : Set ℓ) (A : (x y : X) → x ≡ y → Set ℓ')
  → ((x : X) → A x x (refl x))
  → (x y : X) (p : x ≡ y) → A x y p
J X A f x x (refl x) = f x
```
I tried to unravel @mortberg's file but I didn't manage in the short time I have between various meetings these days.

Saizan:
@mortberg has been using the type defined in `Agda.Builtin.Equality`, which is the version that's only indexed on the second element of `A`.

I'm a bit surprised that you are finding problems with your version, thanks for the report.

Saizan:
Oh, I see what's happening, I'm requiring that the `--cubical` flag is on and `Agda.Builtin.Cubical.Path` is imported (not necessarily directly) for pattern matching definitions to be processed properly.

The following module does not generate any warnings:

```agda
{-# OPTIONS --cubical #-}
module Cubical.Data.OtherEq where
import Agda.Builtin.Cubical.Path

data _≡_ {ℓ} {X : Set ℓ} : X → X → Set ℓ where
 refl : (x : X) → x ≡ x

J : ∀ {ℓ ℓ'} (X : Set ℓ) (A : (x y : X) → x ≡ y → Set ℓ')
  → ((x : X) → A x x (refl x))
  → (x y : X) (p : x ≡ y) → A x y p
J X A f x x (refl x) = f x
```

martinescardo:
So I am trying this large file: https://github.com/martinescardo/HoTT-UF-Agda-Lecture-Notes/blob/master/agda/HoTT-UF-Agda.agda

I modified the header to be
```
{-# OPTIONS --cubical --without-K --exact-split --safe #-}

module HoTT-UF-Agda where

import Agda.Builtin.Cubical.Path
```
as in your recipe above. This file has 8868 lines (a modular version is elsewhere). Then

(1) In lines 2057, 2063, 5120, 5227, 5227 (again), 5290 I get "No equiv while splitting on indexed family". They are all pattern matching on refl.

(2) I get a without-K error that doesn't arise in agda 2.6.1 and earlier in line 5475:
```
/home/mhe/feathers/tudo/HoTT-UF-Agda-Lecture-Notes/agda/HoTT-UF-Agda.agda:5475,21-29
I'm not sure if there should be a case for the constructor refl,
because I get stuck when trying to solve the following unification
problems (inferred index ≟ expected index):
  _·_ ≟ _24846 (ua = ua) (_·_ = _·₁_) (_*_ = _·₂_)
        (f = (Σ.x (id-≃ X)))
  _·_ ≟ _24848 (ua = ua) (_·_ = _·₁_) (_*_ = _·₂_)
        (f = (Σ.x (id-≃ X)))
when checking that the pattern refl _·_ has type
homomorphic _·_ _·₁_ (Σ.x (id-≃ X))
``` 
Has the without-K handling been modified?

There are no other errors.

martinescardo:
Oh, actually there are more without-K errors after that that didn't arise in Agda 2.6.1.

Cubical/Data/Equality.agda
nad: This looks like the kind of thing I cover in my paper [Higher Inductive Type Eliminators Without Paths](http://www.cse.chalmers.se/~nad/publications/danielsson-hits-without-paths.html), which supports "arbitrary" notions of equality with J. I would guess that you could finish the proof easily if you import my library and change the definition of `S¹-elim`.

nad: My [proof](http://www.cse.chalmers.se/~nad/listings/equality/Circle.html#elim-loop) of the corresponding property.

Saizan:
@martinescardo 

About the "not sure if there should be a case for the constructor refl" errors, I believe those are caused by "yellow" in previous definitions, caused by `--cubical` overloading lambdas for paths. One solution could be a flag that turns off the overloading.

Some of the "No equiv" warnings are because the matching relies on injectivity of `_,_` for sigma types. Since it's a record type with eta it should be easy to extend the algorithm to handle this case though.

I have a deadline this week, but I will look into both these issues after that.

martinescardo:
Yes, no worries. I have many deadlines too. I will work on this slowly in the background.

nad:
> About the "not sure if there should be a case for the constructor refl" errors, I believe those are caused by "yellow" in previous definitions, caused by `--cubical` overloading lambdas for paths. One solution could be a flag that turns off the overloading.

Or adding some type signatures (like `f : ∀ _ → _`). See also https://github.com/agda/agda/issues/3754.

martinescardo:
(Just in case, I have compiled the main branch of the Agda development version, and it doesn't give the without-K errors reported above with the issue3733 branch (or any errors).)

@nad I already have full type signatures.

nad:
> @nad I already have full type signatures.

If Agda fails to figure out if a lambda expression is a regular lambda or a path lambda, then this can at least sometimes be fixed by giving a type signature for the lambda expression. If you have done this and Agda still fails, then this might be a bug.

martinescardo:
@nad. OK, I will try this.

martinescardo:
The errors I get don't seem to involve any lambda, @nad

Saizan:
This version with annotated the lambdas (by using the `lam` function defined in the module) only reports the "No equiv" warnings:
https://github.com/Saizan/HoTT-UF-Agda-Lecture-Notes/blob/cubical/agda/HoTT-UF-Agda.agda

martinescardo:
I see - thanks! 

There should be a different lambda for dimension abstraction in cubical Agda, anyway, as the interval is not really a type.

martinescardo:
Has there been progress with this? I may want to experiment more with this now.

mortberg:
So what is the current status of inductive families? Should it be possible for me to fix this PR on the current master branch of Agda?

nad:
> So what is the current status of inductive families?

I think that little has happened after the following comment was written: https://github.com/agda/agda/issues/3733#issuecomment-903581647.

Cubical/Data/Equality.agda
```diff
@@ -1,50 +1,332 @@
...
 
-ptoc-ctop : (p : x ≡c y) → ptoc (ctop p) ≡c p
-ptoc-ctop p =
-  J (λ _ h → ptoc (ctop h) ≡c h) (cong ptoc (transportRefl reflp)) p
+  -- These should type check if things would compute:
 
-ctop-ptoc : (p : x ≡p y) → ctop (ptoc p) ≡c p
-ctop-ptoc {x = x} reflp = transportRefl reflp
+  -- test-winding-loop : winding loop ≡ pos 1
+  -- test-winding-loop = refl
```
mortberg, line 325: This does not type check on 2.6.2.2. Any clue why @Saizan ?

mortberg: The LHS evaluates to:

```
transport Cover (transp (λ i → base ≡ loopPath i) i0 refl) (pos zero) != pos 1
```

mortberg:
I'll merge this now to not leave it to bit rot longer, but it seems like some computations are not working on master the way they did on the branch

PR #310, opened 2020-05-08
Add alternative builders for groups.

favonia:


Alizter:
I don't think the *definition* of a group should be changed but rather a better constructor should be used. This is similar to how we can adjointify equivalences but the definition of a half-adjoint equivalence still has the extra data.

WorldSEnder:
I agree with @Alizter . Sometimes you get better computational behaviour when you specify the paths directly. Say you have a group where `comp` computes nicely and `lCancel` is just true by `refl`. Why should one be forced to have `lCancel` be a large path composition - even abstract in this PR! Additionally, it's rather arbitrary to chose `lCancel` instead of `rCancel`.

Saizan:
I would lean towards keeping `lUnit` and `lCancel` in the record and providing a function that builds a group without asking for those.

mortberg:
> I would lean towards keeping `lUnit` and `lCancel` in the record and providing a function that builds a group without asking for those.

Ok, let's do that!  @favonia : can you please implement this?

mortberg:
Great! I really like the names for the group constructors

favonia:
The master branch is broken now.

mortberg:
Yup... And I don't have agda installed on this computer so it will take me a few hours to fix it. Can anyone else do it?

mortberg:
I think the bug comes from me redefining the recursor for truncation without taking into account the cohomology PR...

PR #303, opened 2020-05-05
Add an implementation of frames

ayberkt:

PR #316, opened 2020-05-23
Define the Sierpinski space as an HIT

ayberkt:
I wrote down the Sierpinski set as an HIT as defined by @niccoloveltri [here](http://cs.ioc.ee/ewscs/2017/studsess/veltri-slides.pdf#3).

It is pointed out in the slides that this is the initial σ-frame. I suppose one can use this as an `Ω` for defining the Dedekind reals? It seems simpler than what @m-yac [was working on](https://gist.github.com/m-yac/7d3fd934852fb321446d0c2322f99481).


ayberkt:
Also, a general question: when I add a new file, is it somehow included in the Travis check or do I need it to import it from some other file?

mortberg:
Neat! It would be interesting to also define the partiality monad as in Niccolò's slides.

mortberg:
> Also, a general question: when I add a new file, is it somehow included in the Travis check or do I need it to import it from some other file?

Unless it's in Core, Foundations, Codata or Experiments it will be automatically checked as the Everything file gets automatically generated

ayberkt:
> It would be interesting to also define the partiality monad as in Niccolò's slides.

Good idea! To clarify, are you suggesting that I should do it in this PR for this to merged?

mortberg:
> > It would be interesting to also define the partiality monad as in Niccolò's slides.
> 
> Good idea! To clarify, are you suggesting that I should do it in this PR for this to merged?

That's not necessary. I can merge it as it is, but it's of course nicer to have some properties so that it's not just another HIT. Would you prefer to turn this into a draft PR or should I merge it and we make an issue about extending it?

ayberkt:
> it's of course nicer to have some properties so that it's not just another HIT

That's a good point. It might be better turn this into a draft and revisit it later (maybe even after #303 is merged).

mortberg:
What's the plan for this PR? It seems like the branch has been deleted?

ayberkt:
> What's the plan for this PR?

The plan for this PR is nonexistent; at this point I don't remember what exactly I was doing—sorry for that. Is this needed for something? If that's the case, I can take a look at this again. Also, I added some [related developments](https://www.cs.bham.ac.uk/~mhe/TypeTopology/InitialFrame.html) to `TypeTopology`. Maybe I can make a new PR containing both this and that.

mortberg:
> > What's the plan for this PR?
> 
> The plan for this PR is nonexistent; at this point I don't remember what exactly I was doing—sorry for that. Is this needed for something? If that's the case, I can take a look at this again. Also, I added some [related developments](https://www.cs.bham.ac.uk/~mhe/TypeTopology/InitialFrame.html) to `TypeTopology`. Maybe I can make a new PR containing both this and that.

I don't think anything depends on this, but maybe it could be useful for @mzeuner ?

We're trying to clean up among the old PRs, so the best might be to close and then you can open a new PR with the new content

ayberkt:
> so the best might be to close and then you can open a new PR with the new content

Good idea.

PR #322, opened 2020-05-23
Reintroduce Truncation.rec and fix problems in ZCohomology related to it

WorldSEnder:
Additionally, a workaround in Codata.M.AsLimit.M.Properties is provided that speeds
up type checking. The workaround might not be the best since it introduces a "useless" transport over reflexivity but it forces typechecking to do the morally correct thing.

@Saizan might have additional input for that before this is merged.

### DIFF
--- a/Cubical/Codata/M/AsLimit/M/Properties.agda
+++ b/Cubical/Codata/M/AsLimit/M/Properties.agda
@@ -27,25 +27,35 @@ open import Cubical.Codata.M.AsLimit.Container
 
 open Iso
 
-in-inverse-out : ∀ {ℓ} {S : Container ℓ} -> (in-fun ∘ out-fun {S = S}) ≡ idfun (M S)
-in-inverse-out {S = S} = funExt (rightInv {A = P₀ S (M S)} {B = M S} (shift-iso S))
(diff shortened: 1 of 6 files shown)


Saizan:
@WorldSEnder did you do some analysis that explains why this transport is faster?

My guess would be that Agda no longer reduces the composition on the left hand sides of these lemmas, but rather just the functions appearing in the types of the `idpath` proofs.

WorldSEnder:
Something like that was my thought as well. `funExt (rightInv (shift-iso S))
` has inferred type `(λ x → fun (shift-iso S) (inv (shift-iso S)) x) ≡ (λ x → x)`.

Originally unifying the left side directly with `in-fun {S = S} ∘ out-fun {S = S}` causes agda to (after eta conversion) reduce `in-fun {S = S}` to `fun (shift-iso S)` but since the arguments don't literally match - `(inv (shift-iso S)) x != out-fun {S = S} x` - `fun (shift-iso S)` can and will be further reduced.

Written with transport, the arguments are literally the same, after path application. 2-level-type-theory is a bit far away but may help to guide the typechecker without introducing this transport.

mortberg:
@Saizan : should we merge this?

Issue #317, opened 2020-05-23
Add a contributing guideline


ayberkt:
It might be useful to have a file containing basic information (e.g. [`CONTRIBUTING.md`](https://github.blog/2012-09-17-contributing-guidelines/)) on how to contribute to `cubical`.

mortberg:
Yes! I've been meaning to do this for a while and have a local draft somewhere. I can make a PR with it soon so that we can discuss

ayberkt:
Nice! Here is a question I have that would maybe answered in this guideline: when one adds a new file in a PR, should one include it in some file for it to be included in the CI checks? If yes, how would one find the right file?

ayberkt:
Another question: is there a line limit that contributors should adhere to (e.g. 80 characters)?

mortberg:
> Nice! Here is a question I have that would maybe answered in this guideline: when one adds a new file in a PR, should one include it in some file for it to be included in the CI checks? If yes, how would one find the right file?

As of https://github.com/agda/cubical/pull/268 you don't have to add anything manually unless your file are in Core, Foundations, Codata or Experiments. I'll add it to the contributing file

PR #323, opened 2020-05-23
A CONTRIBUTING.md file

mortberg:
As discussed in https://github.com/agda/cubical/issues/317.

I wrote down everything I could think of but there's surely things I forgot and things that should be reformulated. So comments are very welcome!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,69 @@
+CONTRIBUTING
+============
+
+We very much appreciate contributions in the form of pull requests (PRs).
+When preparing a PR here are some general guidelines:
+
+- Please read through and clean your code before making a PR. Clean
+  code has reasonable line length (<100 characters), good indentation,
+  appropriate amount of comments and consistent naming.
+
+- Local definitions should be put in `where`, `let-in` or `private` so
+  that it is easier to see which are the main results of a file and
+  which are just helper definitions.
+
+- If the code is based on some paper add a reference to the version of
+  the paper that the file follows in a comment at the top of the
+  file. Avoid calling top-level definitions `thm123`, `lem321`, etc.,
+  instead have informative names and put pointers to the theorems and
+  lemmas in comments above the definition.
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
+CONTRIBUTING
+============
+
+We very much appreciate contributions in the form of pull requests (PRs).
+When preparing a PR here are some general guidelines:
+
+- Please read through and clean your code before making a PR. Clean
+  code has reasonable line length (<100 characters), good indentation,
```
mortberg, line 8: I know that 80 +- epsilon is also a common limit for line lengths. I have no really strong opinions about this as long as it's less than 100

mortberg:
Maybe we should say something about structuring files with `Foo/Base.agda`, `Foo/Properties.agda` and then having `Foo.agda` exporting these in some form or another. Wasn't there a discussion started by @ecavallo about what `Foo.agda` should export at some point?

mortberg:
Maybe we should also say something about implicit arguments. I like to have as many things implicit as possible, especially universe levels, but one should not overdo it and make arguments implicit that Agda can't infer.

mortberg:
One more thing: it's not good to have files that take a long time to typecheck. If typechecking a file starts becoming slow it is often a sign that some definitions need to be optimized (maybe using copattern-matching or abstract or rethinking that maths).

ayberkt:
@mortberg It can perhaps be mentioned that `isSetΠn` and `isPropΠn` are to be preferred over multiple applications of their unary counterparts. Same for `funExt`.

This can save some time for the maintainers. I, for instance, was not aware of the existence of the n-ary versions initially.

mortberg:
PRs with postulates should be marked as Draft PRs until the postulates have been fixed

mortberg:
I'm reasonably happy with this now. Any comments before I merge?

mortberg:
Alright, merging this now. Anyone should feel free to make updates

PR #326, opened 2020-05-28
Add some INLINE pragmas

UlfNorell:
to make termination checker not complain if/when we turn off auto-inlining (see agda/agda#4681)


Saizan:
Oh, interesting!

Do these pragmas affect anything other than termination?

UlfNorell:
Also positivity checking and other analyses that look at terms. I think there were only termination errors in the library when turning off auto-inlining though.

PR #328, opened 2020-05-28
[ travis ] use new fix-whitespace repo

Saizan:


mortberg:
Oh, this PR answers my other question. Should we merge so that Travis works again?

Saizan:
We need to get it to work first, but yeah, that's the idea.

Saizan:
Oh, I see, now it's breaking for other reasons.

PR #331, opened 2020-05-29
Algebras and modules

felixwellen:
Algebras and modules over a fixed ring.


mortberg:
Thanks for this code as well! I'll review it when https://github.com/agda/cubical/pull/324 has been merged

felixwellen:
This PR includes the following:
* Left Modules over a fixed ring (with application of SIP)
* Algebras over a fixed ring (without SIP application)
* Definition of morphisms of algebras
* Commutative algebras over a fixed ring (without SIP application)
* The free commutative algebra on a given type
* Universal property of the free commutative algebra

Cubical/Structures/Algebra.agda
```diff
@@ -0,0 +1,231 @@
...
+                  ⋆-lassoc ⋆-rassoc
+
+
+open commonExtractors public
+
+record AlgebraEquiv {R : Ring {ℓ}} (A B : Algebra R) : Type ℓ where
+
+  constructor moduleiso
+
+  instance
```
mortberg, line 125: What are these instances used for?

Cubical/Structures/Algebra.agda
```diff
@@ -0,0 +1,231 @@
+{-# OPTIONS --cubical --safe #-}
```
mortberg, line 1: Please add `--no-import-sorts` here and to the OPTIONS for the other files in this PR

Cubical/Structures/FreeCommAlgebra.agda
```diff
@@ -0,0 +1,398 @@
...
+
+    evaluateAt : Hom
+                 → I → ⟨ A ⟩a
+    evaluateAt (algebrahom f _ _ _ _) x = f (var x)
+
+    mapRetrievable : ∀ (φ : I → ⟨ A ⟩a)
+                     → evaluateAt (inducedHom A φ) ≡ φ
+    mapRetrievable φ = refl
+
+    proveEq : ∀ {X : Type ℓ} (isSetX : isSet X) (f g : ⟨ R [ I ] ⟩a → X)
```
mortberg, line 186: Yikes! This definition was complicated... I wish we could automate the cases for the higher constructors as they're all following the same pattern. @Saizan any thoughts about this after the discussion yesterday?

Cubical/Structures/Algebra.agda
felixwellen: I added them just to use "+" and similar below in the homomorphism properties.

Cubical/Structures/FreeCommAlgebra.agda
felixwellen: It is long but not really complicated. And I wrote it down in a verbose way, with types of all the boundaries-paths of all the squares I used. So most of the code is just there to make it readable (for me).

mortberg: Yeah, it's pretty clear what's going on. I recently learned that you don't have to give these cases in Arend and I just wonder if we can automate them in Agda somehow...

Cubical/Structures/Algebra.agda
mortberg: So that's a cooler version of https://github.com/agda/cubical/blob/master/Cubical/Structures/Monoid.agda#L95?

Cubical/Structures/FreeCommAlgebra.agda
Saizan: It would be nice if we could do this with a tactic, so it can just take the appropriate isOfHLevel proof as argument.

Might need some extension to the reflection machinery though.

mortberg: Actually maybe one can even prove a general theorem for solving this kind of goals parameterized by the hlevel. Doing it for prop, set and groupoid would be a good start.

Cubical/Structures/Algebra.agda
felixwellen: Yes. In particular it is a verion which saves some typing - also in some proofs below. I made the instances private now.

mortberg (approved):

mortberg:
Sorry for the slow reply. Everything looks great, so merging!

PR #334, opened 2020-06-03
Add Cubical Agda paper to README.

cangiuli:
Added the Cubical Agda paper from ICFP 2019 to the README. I wonder whether the HoTT blog post is sufficiently out of date that the link should be removed, but I didn't do that.


mortberg:
This looks good to me! Thanks for updating

PR #335, opened 2020-06-04
funext for dependent paths

ecavallo:
Generalizes the type of funext and related functions to handle dependent paths in a function type, of the form `PathP (λ i → (a : A) → B a i) f g`. This doesn't require any change to the proofs.


mortberg:
Wow, I'm surprised that no changes to other proofs were required. Do you have any use cases in mind for these more general theorems?

ecavallo:
It came up in a structure combinator I was working on.

mortberg:
> It came up in a structure combinator I was working on.

I see. I'll merge this PR now and I'm looking forward to whatever combinators you have in store for us

Issue #315, opened 2020-05-23
Get rid of uses of SNS→UnivalentStr


mortberg:
As discussed in https://github.com/agda/cubical/pull/311

@mzeuner Can you do this?

mortberg:
@ecavallo There are only a few uses left. What do you think about them? Is it worth trying to do this or should I close this issue?

ecavallo:
I don't think it's worth worrying about.

PR #336, opened 2020-06-04
Structure combinators

ecavallo:
- Structures parameterized by an external type A
- Generalize NAryOp so the codomain can be any structure
- Break LeftAction out into its own file, so Multiset doesn't depend on Queue

### DIFF
--- a/Cubical/Structures/Constant.agda
+++ b/Cubical/Structures/Constant.agda
@@ -8,11 +8,11 @@ open import Cubical.Foundations.SIP renaming (SNS-PathP to SNS)
 
 private
   variable
-    ℓ' : Level
+    ℓ ℓ' : Level
 
 -- Constant structure
 
-module _ {ℓ} (A : Type ℓ) where
+module _ (A : Type ℓ) where
 
   constant-structure : Type ℓ' → Type ℓ
   constant-structure _ = A
--- a/Cubical/Structures/Constant.agda
+++ b/Cubical/Structures/Constant.agda
@@ -1,0 +1,24 @@
+{-# OPTIONS --cubical --safe #-}
+module Cubical.Structures.Constant where
+
+open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.Equiv
+
+open import Cubical.Foundations.SIP renaming (SNS-PathP to SNS)
+
+private
+  variable
+    ℓ' : Level
+
+-- Constant structure
+
+module _ {ℓ} (A : Type ℓ) where
+
+  constant-structure : Type ℓ' → Type ℓ
+  constant-structure _ = A
+
+  constant-iso : StrIso {ℓ'} constant-structure ℓ
+  constant-iso (_ , a) (_ , a') _ = a ≡ a'
+
+  constant-is-SNS : SNS {ℓ'} constant-structure constant-iso
+  constant-is-SNS e = idEquiv _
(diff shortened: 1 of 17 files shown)


Cubical/Structures/Constant.agda
```diff
@@ -0,0 +1,24 @@
...
+
+open import Cubical.Foundations.SIP renaming (SNS-PathP to SNS)
+
+private
+  variable
+    ℓ' : Level
+
+-- Constant structure
+
+module _ {ℓ} (A : Type ℓ) where
```
mortberg, line 15: Can't these two be part of the `private variable` declaration?

Cubical/Structures/NAryOp.agda
```diff
@@ -12,62 +12,53 @@ open import Cubical.Data.Vec
...
+  nAryFunSNS : {S : Type ℓ₁ → Type ℓ₂} (n : ℕ) {ℓ₃ : Level}
+    (ι : StrIso S ℓ₃) (θ : SNS S ι)
+    → SNS (nAryFun-structure n S) (nAryFunIso n ι)
+  nAryFunSNS n ι θ =
+    SNS-≡→SNS-PathP (nAryFunIso n ι) λ fX fY →
+    compEquiv
+      (equivPi λ xs → SNS-PathP→SNS-≡ _ ι θ _ _)
+      (nAryFunExtEquiv n fX fY)
 
 -- Some specializations that are not used at the moment, but kept as
```
mortberg, line 36: I think this comment is now outdated?

Cubical/Structures/Ring.agda
```diff
@@ -23,9 +23,12 @@ raw-ring-structure X = (X → X → X) × X × (X → X → X)
 
 -- Maybe this is not the best way? (Suggestions welcome, maybe having raw-monoid-iso defined?)
 raw-ring-is-SNS : SNS {ℓ} raw-ring-structure _
-raw-ring-is-SNS = join-SNS (nAryFunIso 2) (nAryFunSNS 2)
-                           (join-iso pointed-iso (nAryFunIso 2))
-                           (join-SNS pointed-iso pointed-is-SNS (nAryFunIso 2) (nAryFunSNS 2))
+raw-ring-is-SNS = join-SNS (binaryFunIso pointed-iso)
```
mortberg, line 26: Maybe `binaryFunIso pointed-iso` should be stated as a lemma as you use it multiple times? Same for some other arguments here...

Cubical/Structures/NAryOp.agda
```diff
@@ -12,62 +12,53 @@ open import Cubical.Data.Vec
...
-            ((xs : Vec X 1) → (s $ⁿ xs) ≡ (t $ⁿ map (λ x → x) xs))
-    lem {X} {s} {t} = isoToEquiv f
-      where
-      f : Iso ((x : X) → s x ≡ t x)
-              ((xs : Vec X 1) → (s $ⁿ xs) ≡ (t $ⁿ map (λ x → x) xs))
-      Iso.fun f p (x ∷ [])           = p x
-      Iso.inv f p x                  = p (x ∷ [])
-      Iso.rightInv f p _ xs@(x ∷ []) = p xs
-      Iso.leftInv f p _              = p
+  unaryFunSNS : {S : Type ℓ₁ → Type ℓ₂} {ℓ₃ : Level}
```
mortberg, line 46: I might be wrong now, but couldn't this be proved by https://github.com/agda/cubical/blob/master/Cubical/Foundations/Univalence.agda#L226 ? And similarly for the binary version below

ecavallo: I think there should be a proof along those lines, but it's not immediate.

mortberg: Ok, let's investigate this later. I already have an issue about it: https://github.com/agda/cubical/issues/315

PR #337, opened 2020-06-05
Clean up Sigma.Properties a bit

ecavallo:
Mostly factoring the transport characterization of paths in Sigma through the dependent path characterization.


ecavallo:
Btw, there's some duplication of code between `Σ-ap-iso` in this file and `Σ-change-of-variable-Iso` in EscardoSIP.

ecavallo:
Ah, I guess that file is supposed to be self-contained.

Cubical/Data/Sigma/Properties.agda
```diff
@@ -47,107 +49,23 @@ mapˡ f (a , b) = (f a , b)
 Σ-split-iso : {x y : Σ A B}
```
mortberg, line 49: Shouldn't this just be \Sigma-Iso or something to match the naming below?

Cubical/Data/Sigma/Properties.agda
```diff
@@ -47,107 +49,23 @@ mapˡ f (a , b) = (f a , b)
...
      Σ (fst x ≡ fst y) (λ p → PathP (λ i → B (p i)) (snd x) (snd y)) ≃
      (x ≡ y)
-Σ≃ {A = A} {B = B} {x} {y} = isoToEquiv (Σ-split-iso)
+Σ≃ = isoToEquiv (Σ-split-iso)
 
-Σ≡ : {a a' : A} {b : B a} {b' : B a'} → (Σ (a ≡ a') (λ q → PathP (λ i → B (q i)) b b')) ≡ ((a , b) ≡ (a' , b'))
+Σ≡ : {x y : Σ A B} → (Σ (fst x ≡ fst y) (λ q → PathP (λ i → B (q i)) (snd x) (snd y))) ≡ (x ≡ y)
 Σ≡ = isoToPath Σ-split-iso -- ua Σ≃
 
 ΣProp≡ : ((x : A) → isProp (B x)) → {u v : Σ A B}
```
mortberg, line 65: I would swap Prop and \equiv in this name

Cubical/Data/Sigma/Properties.agda
```diff
@@ -47,107 +49,23 @@ mapˡ f (a , b) = (f a , b)
...
-Σ≃ : {x y : Σ A B}  →
+Σ≃ : {x y : Σ A B} →
      Σ (fst x ≡ fst y) (λ p → PathP (λ i → B (p i)) (snd x) (snd y)) ≃
      (x ≡ y)
-Σ≃ {A = A} {B = B} {x} {y} = isoToEquiv (Σ-split-iso)
+Σ≃ = isoToEquiv (Σ-split-iso)
 
-Σ≡ : {a a' : A} {b : B a} {b' : B a'} → (Σ (a ≡ a') (λ q → PathP (λ i → B (q i)) b b')) ≡ ((a , b) ≡ (a' , b'))
+Σ≡ : {x y : Σ A B} → (Σ (fst x ≡ fst y) (λ q → PathP (λ i → B (q i)) (snd x) (snd y))) ≡ (x ≡ y)
 Σ≡ = isoToPath Σ-split-iso -- ua Σ≃
```
mortberg, line 63: Why not use the ua proof that is in a comment?

Cubical/Data/Sigma/Properties.agda
```diff
@@ -160,23 +78,6 @@ discreteΣ {B = B} Adis Bdis (a0 , b0) (a1 , b1) = discreteΣ' (Adis a0 a1)
...
-        cong (λ p → subst B p b) (isProp→isSet (isContr→isProp c) _ _ _ _)
-        ∙ transportRefl _)
-      (λ {(a , b) →
-        sigmaPath→pathSigma _ _ (c .snd a , transportTransport⁻ (cong B (c .snd a)) _)}))
-
--- a special case of the above
-ΣUnit : ∀ {ℓ} (A : Unit → Type ℓ) → Σ Unit A ≃ A tt
-ΣUnit A = isoToEquiv (iso snd (λ { x → (tt , x) }) (λ _ → refl) (λ _ → refl))
-
 assocΣ : (Σ[ (a , b) ∈ Σ A B ] C a b) ≃ (Σ[ a ∈ A ] Σ[ b ∈ B a ] C a b)
```
mortberg, line 81: More naming: shouldn't this be \Sigma-assoc or something match the earlier lemmas in the file?

Cubical/Data/Sigma/Properties.agda
```diff
@@ -199,13 +100,13 @@ swapΣEquiv A B = isoToEquiv (iso (λ x → x .snd , x .fst) (λ z → z .snd ,
 
 Σ-ap-iso₁ : ∀ {ℓ} {ℓ'} {A A' : Type ℓ} {B : A' → Type ℓ'}
```
mortberg, line 101: \ell and \ell' should be variables at the top of the file

mortberg: I don't understand this name and some of the other with ap... They have nothing to do with ap?

Cubical/Data/Sigma/Properties.agda
```diff
@@ -272,3 +164,43 @@ Iso.leftInv (Σ-ap-iso₂ isom) (x , y') = ΣPathP (refl , Iso.leftInv (isom x)
...
   ≡ (Σ X' Y')
 Σ-ap'  {ℓ} {ℓ'} isom isom' = cong₂ (λ (a : Type ℓ) (b : a → Type ℓ') → Σ a λ x → b x) isom isom'
+
+-- Alternative version for path in Σ-types, as in the HoTT book
+
+ΣPathTransport : (a b : Σ A B) → Type _
+ΣPathTransport {B = B} a b =
+  Σ (fst a ≡ fst b) (λ p → transport (λ i → B (p i)) (snd a) ≡ snd b)
+
+_Σ≡T_ : (a b : Σ A B) → Type _
```
mortberg, line 174: I don't like this name... Why not just use \SigmaPathTransport as it's a bit more informative?

Cubical/Data/Sigma/Properties.agda
```diff
@@ -272,3 +164,43 @@ Iso.leftInv (Σ-ap-iso₂ isom) (x , y') = ΣPathP (refl , Iso.leftInv (isom x)
...
+a Σ≡T b = ΣPathTransport a b
+
+ΣPath≃pathΣ : (a b : Σ A B) → (a Σ≡T b) ≃ (a ≡ b)
```
mortberg, line 177: \SigmaPathTransport\sequivPath\Sigma?

mortberg: And similar changes to other lemmas below

Cubical/Data/Sigma/Properties.agda
ecavallo: Any suggestion for `Σ-ap`?
It's ap in the sense that it's the action of `Σ` on paths (or isos...). In that case maybe it should be called `Σ-cong`. It doesn't seem like there's a convention for naming this sort of function yet.

mortberg: Hmm, I guess `Σ-cong` is fine then. How about not calling this lemma `Σ-ap-iso₁` but `Σ-cong-iso-fst` instead?

ecavallo: Shall I change `mapˡ` and `mapʳ` to `map-fst` and `map-snd` as well?

mortberg: That would be great! Maybe we should start drafting a naming convention one of these days...

Cubical/Data/Sigma/Properties.agda
ecavallo: I changed, but actually I think the names of `Σ≃` and `Σ≡` should be changed... they're not very informative.

mortberg: Any suggestions?

ecavallo: Maybe ΣPath-iso, ΣPath≃, ΣPath≡, something like that?

ecavallo: or PathΣ-etc

mortberg: Sounds good to me!

mortberg: Or maybe `ΣPathIso`?

ecavallo: I guess to match the transport versions it should be `ΣPath≃PathΣ` and so on.

Cubical/Data/Sigma/Properties.agda
```diff
@@ -2,7 +2,7 @@
 
 Basic properties about Σ-types
 
-- Characterization of equality in Σ-types using transport ([pathSigma≡sigmaPath])
+- Characterization of equality in Σ-types using transport ([ΣPath≡pathΣ])
```
mortberg, line 5: Is this comment wrong now? It would be great if you could add a few more things to this comment as the file contains a lot of useful results

Cubical/Data/Sigma/Properties.agda
mortberg: Yeah, that sounds good!

PR #338, opened 2020-06-05
Vogt's lemma is iso→HAEquiv

ecavallo:
Eliminating some duplication. There are also two proofs of `iso→HAEquiv` in this file, don't know if we want to keep both of them.


mortberg:
Do we really need this as a definition? I think we could just have it as a comment next to `iso→HAEquiv`

ecavallo:
Got rid of it, and the second proof of iso→HAEquiv as well.

mortberg:
Great! That sorts out some mysterious stuff that I never understood

PR #339, opened 2020-06-05
Polynomials

felixwellen:
The free commutative algebra over a commutative ring as a HIT. Or equivalently: The ring of polynomials over a given commutative ring. The plan is to construct the evaluation map, construct some 'elimProp'-like thing and then to prove the universal property.
I will probably need some help from someone with cubical inside for the elimProp-thing. And there is no definition of homomorphism of algebras yet.


mortberg:
Thank you for this! I'll review once the other PRs have been merged

felixwellen:
I moved everything from this PR to #331

Issue #327, opened 2020-05-28
Rename Set to Type


mortberg:
Now that https://github.com/agda/agda/pull/4629 has been merged I think we can get rid of https://github.com/agda/cubical/blob/master/Cubical/Core/Primitives.agda#L40 and just rename `Set` to `Type` when importing `Agda.Primitive`

mortberg:
@jespercockx : Thanks a lot for implementing this nice change to Agda! Can you just tell me quickly what I need to do to rename Set to Type globally for this library? I'm sure I can find it by reading through that PR, but if you have time I would really appreciate a short explanation

guillaumebrunerie:
The line where you import `Agda.Primitive` should look have an additional renaming of `Set` to `Type`, just like there is one from `Setω` to `Typeω`.

I believe this is enough (and removing the definitions of `Type` and so on, of course).

mortberg:
Wow, that seems simple enough! I guess the only issue is that the library will break for any version prior to when this PR was merged? But @Saizan made a new release recently so I think this is fine.

jespercockx:
You might also want to add the `--no-import-sorts` flag if you actually want to use `Set` for something else (or you might want to wait until https://github.com/agda/agda/issues/3118 is implemented).

mortberg:
> You might also want to add the `--no-import-sorts` flag if you actually want to use `Set` for something else (or you might want to wait until [agda/agda#3118](https://github.com/agda/agda/issues/3118) is implemented).

Yeah, I realized this and added `--no-import-sorts` to all files in the library in my PR (with the help of sed)

PR #341, opened 2020-06-07
Rename Set to Type

mortberg:
This fixes https://github.com/agda/cubical/issues/327 by 

- Renaming Set to Type when importing Set in Cubical.Core.Primitives
- Adding the flag `--no-import-sorts` to all files, including the autogenerated Everything files
- Fix the uses of Set in the library

### DIFF
--- a/Cubical/Algebra/BasicProp.agda
+++ b/Cubical/Algebra/BasicProp.agda
@@ -1,4 +1,4 @@
-{-# OPTIONS --cubical --safe #-}
+{-# OPTIONS --cubical --no-import-sorts --safe #-}
 module Cubical.Algebra.BasicProp where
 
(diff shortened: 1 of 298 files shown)


mortberg:
I'm going to merge this now. There will hopefully not be too many conflicts with the open PRs, but many people might have to update their Agda installation to the latest development version. By merging this now we'll be sure that none of the PRs have the wrong flags at the top of the file.

I guess the people that might be affected are the ones with open PRs, so I'll ping you and the most frequent recent contributors so that you don't miss this and get confused if the library stops compiling suddenly: @felixwellen , @ecavallo , @WorldSEnder , @ayberkt , @aljungstrom , @mzeuner , @cmester0 , @m-yac ,  @Saizan , @cangiuli

PR #344, opened 2020-06-08
Merging both definitions of groups

Edlyr:
This PR merges the definition of group that used to be in Cubical/Data/Group with the SNS one.
A few useful lemmas about groups are also given.


mortberg:
I think you need to merge/rebase on agda/cubical master as I've merged a lot of PRs the last days. I'm pretty sure you'll have some conflicts because of https://github.com/agda/cubical/pull/336 which refactored the SIP code a bit. Let me know if you need some help with the merge/rebase

mortberg:
Would you be up for changing this PR to match the style that is now for Semigroups and which will soon be for monoids when I merge https://github.com/agda/cubical/pull/362/files ? 

The record for groups should look a bit like the one in the standard library when this is done: https://agda.github.io/agda-stdlib/Algebra.Bundles.html#6392

mortberg:
Actually, when working on rewriting rings to be defined using records I realized I have to rewrite Group and AbGroup to be records as well. So I'll do it myself now by just modifying the files in Structures and once that is merged you can update this PR so that everything about groups is in the right place.

mortberg:
I've merged the PR rewriting Group, AbGroup, Ring and CommRing in Structures to use records now. I did not take the time to merge the Data.Group code with the Structures.Group file. Can you do this now? It should mostly be a matter of moving theory from Data.Group to the other file (including the leftist/rightist constructors) and making sure that all files that depended on Data.Group work with Structures.Group.

When doing this can you also move the code in https://github.com/agda/cubical/blob/master/Cubical/Algebra/BasicProp.agda to Structures.Group as well? There might be some code duplication and some parts could maybe be simplified.

Cubical/Foundations/SIP.agda
```diff
@@ -31,17 +31,28 @@ private
 -- a proposition. Indeed this type should correspond to the ways s and t can be identified
 -- as S-structures. This we call a standard notion of structure or SNS.
 -- We will use a different definition, but the two definitions are interchangeable.
+
 SNS-≡ : (S : Type ℓ₁ → Type ℓ₂) (ι : StrIso S ℓ₃) → Type (ℓ-max (ℓ-max (ℓ-suc ℓ₁) ℓ₂) ℓ₃)
 SNS-≡ {ℓ₁} S ι = ∀ {X : Type ℓ₁} (s t : S X) → ι (X , s) (X , t) (idEquiv X) ≃ (s ≡ t)
 
+SNS-≡' : (S : Type ℓ₁ → Type ℓ₂) (ι : StrHom S ℓ₃) → Type _
```
mortberg, line 38: Can you move the changes related to this to a separate PR? It would be good to have a separate discussion about these things

mortberg: By "this" I mean all changes related to StrHom

Cubical/Structures/Group/Base.agda
```diff
@@ -0,0 +1,152 @@
...
+    Carrier : Type ℓ
+    0g      : Carrier
+    _+_     : Carrier → Carrier → Carrier
+    -_      : Carrier → Carrier
+    isGroup : IsGroup 0g _+_ -_
+
+  infix  8 -_
+  infixr 7 _+_
+
+  open IsGroup isGroup public
```
mortberg, line 65: Why is this here? I think it should be part of a separate theory module later in the file

Cubical/Structures/Group/Base.agda
```diff
@@ -0,0 +1,152 @@
...
+            ≡⟨ cong (a +_) (assoc _ _ _ ∙ cong (_+ (- a)) (invr b)) ⟩
+          a + (0g - a)
+            ≡⟨ cong (a +_) (lid (- a)) ∙ invr a ⟩
+          0g ∎
+
+-- Extractor for the carrier type
+⟨_⟩ : Group → Type ℓ
+⟨_⟩ = Group.Carrier
+
+-- TODO: we don't need all of these arguments...
```
mortberg, line 120: Can you move makeGroup-left/right here and remove this comment?

Cubical/Structures/Group/Morphism.agda
```diff
@@ -0,0 +1,248 @@
...
+  isInj-f x y = invEq (_ , isEquiv→isEmbedding eq x y)
+
+invGroupIso : (G : Group {ℓ}) (H : Group {ℓ'}) → GroupIso G H → GroupIso H G
+invGroupIso G H (groupiso f morph) = groupiso (invEquiv f) (isGroupHomInv G H (groupiso f morph))
+
+groupHomEq : (G : Group {ℓ}) (H : Group {ℓ'}) (f g : GroupHom G H) → (fst f ≡ fst g) → f ≡ g
+groupHomEq G H f g p = Σ≡Prop (λ _ → isPropIsGroupHom G H) p
+
+groupIsoEq : (G : Group {ℓ}) (H : Group {ℓ'}) (f g : GroupIso G H) → (GroupIso.eq f ≡ GroupIso.eq g) → f ≡ g
+-- This proof would take one line with Σ≡Prop using Σ-types
```
mortberg, line 99: Why don't you use isPropIsGroupHom? Then it would be one line?

Cubical/Structures/Group/Morphism.agda
```diff
@@ -0,0 +1,248 @@
...
+    GroupIso G H                    ≃⟨ isoToEquiv GroupIsoΣPath ⟩
+    GroupIsoΣ G H                   ≃⟨ GroupΣPath _ _ ⟩
+    Group→GroupΣ G ≡ Group→GroupΣ H ≃⟨ isoToEquiv (invIso (congIso GroupIsoGroupΣ)) ⟩
+    G ≡ H ■
+
+-- Extract the characterization of equality of groups
+GroupPath : (G H : Group {ℓ}) → (GroupIso G H) ≃ (G ≡ H)
+GroupPath = GroupΣ-theory.GroupPath
+
+Group-ua : {G H : Group {ℓ}} → (GroupIso G H) → (G ≡ H)
```
mortberg, line 199: Parantheses are not needed around GroupIso G H and the conclusion

Cubical/Structures/NAryOp.agda
```diff
@@ -17,12 +17,24 @@ module _ {ℓ₁ ℓ₂ : Level} where
     → Type ℓ₁ → Type (nAryLevel ℓ₁ ℓ₂ n)
   nAryFun-structure n S X = nAryOp n X (S X)
 
+{-
```
mortberg, line 20: Can you remove these changes?

Cubical/Structures/Group/Morphism.agda
```diff
@@ -0,0 +1,248 @@
...
+  Σ[ r ∈ PathP (λ i → p i → p i → p i) (_+_ G) (_+_ H) ]
+  Σ[ s ∈ PathP (λ i → p i → p i) (-_ G) (-_ H) ]
+  PathP (λ i → IsGroup (q i) (r i) (s i)) (isGroup G) (isGroup H))
+  ≃ (G ≡ H)
+Group≡ G H = isoToEquiv (iso
+  (λ (p , q , r , s , t) i → group (p i) (q i) (r i) (s i) (t i))
+  (λ p → cong Carrier p , cong 0g p , cong _+_ p , cong -_ p , cong isGroup p)
+  (λ _ → refl) (λ _ → refl))
+
+caracGroup≡ : {G H : Group {ℓ}} (p q : G ≡ H) → cong Group.Carrier p ≡ cong Group.Carrier q → p ≡ q
```
mortberg, line 223: Can't you prove this directly from GroupPath or Group-ua?

Cubical/Structures/Group/Morphism.agda
```diff
@@ -0,0 +1,248 @@
...
+caracGroup-ua : {G H : Group {ℓ}} (f : GroupIso G H) → cong Group.Carrier (Group-ua f) ≡ ua (GroupIso.eq f)
+caracGroup-ua (groupiso f m) =
+  (refl ∙∙ ua f ∙∙ refl)
+    ≡⟨ sym (rUnit (ua f)) ⟩
+  ua f ∎
+
+-- Group-ua functoriality
+open Group
+
+Group≡ : (G H : Group {ℓ}) → (
```
mortberg, line 211: Is this lemma really useful?

aljungstrom:
> Great work!
> 
> @aljungstrom I will merge this before #365, so your PR will need some changes once I merged this

I think I should manage:-) Just a question. The definition of homomorphisms is using a sigma type instead of being defined as a record. Does this have any impact, computation-wise? I am just wondering because I have some pretty heavy proofs of certain maps' being morphisms in my PR.

mortberg:
> > Great work!
> > @aljungstrom I will merge this before #365, so your PR will need some changes once I merged this
> 
> I think I should manage:-) Just a question. The definition of homomorphisms is using a sigma type instead of being defined as a record. Does this have any impact, computation-wise? I am just wondering because I have some pretty heavy proofs of certain maps' being morphisms in my PR.

I thought it wouldn't matter as it's just a two field Sigma type which would be the same as a record, but now I remember about `no-eta-equality` which makes me think that we should explicitly rewrite this as a record as well so that you can easier test to add this flag in case your code is slow. Currently I don't think any of the records is using this flag so it would be great if you could benchmark how long typechecking your PR is without it before you start adding it everywhere (assuming that it's actually possible to typecheck it without a super computer).

@Edlyr: can you please change homomorphisms to be a record as well?

Cubical/Structures/Group/Base.agda
Edlyr: While I agree it would make more structural sense to put this elsewhere, I have found that often in proof related to groups, I always "open" the lemmas about a specific group, as well as the group notations themselves.
So I find it better to be able to access everything with just
`open Group G` or `module G = Group G`
instead of
```
open Group G
open GroupLemmas G
```
every single time.
I noticed that the HoTT-agda library does the same thing (putting the lemmas in the record definition), see for example https://github.com/HoTT/HoTT-Agda/blob/master/core/lib/types/Group.agda#L32

Should I still move this further down in the file ?

mortberg: I personally prefer to keep the theory separate from the definition just to get better control of what is in scope. I also think the file reads easier this way. For Data/HITs we have Base/Properties files for everything and I think the long term goal should be to organize the algebraic structures like this as well. This could be done in a future PR where we also move them to the Algebra package. Anyone else has an opinion about this? I we plan to do this then it will be a little bit easier if the definition and theory is separated clearly in the file.

Cubical/Structures/Group/Morphism.agda
Edlyr: I'm not yet sure (apart from the proof of caracGroup≡).
I thought that some version of ΣPathP for groups might come in handy ?

Cubical/Structures/Group/Base.agda
Edlyr: Another argument against this (at least in the case for groups), is the fact that there may be different definitions, and different ways to "make" a group (for example with makeGroup-left and makeGroup-right), and it is not obvious in this case what should be an axiom and what should be a property.
For example, If I'm doing a proof about multiples groups (say G and H), then I cannot simply write
```
open Group G
open GroupLemmas G
open Group H
open GroupLemmas H
```
because of conflicting names.
So I should instead do something like this :
```
module G = Group G
module Glemmas = GroupLemmas G
module H = Group H
module Hlemmas = GroupLemmas H
```
But then it isn't obvious at all that I should use `G.invr` and `Glemmas.invId` and not the opposite. I think it just adds confusion to the end user experience.

Or maybe we should add all the notations and fields of the record type `Group` to the `GroupLemmas` module.
But I think that too would reduce readability of the code. I think a comment explaining the choice to keep those properties there would make it all clearer and more practical to everyone.

mortberg: > Another argument against this (at least in the case for groups), is the fact that there may be different definitions, and different ways to "make" a group (for example with makeGroup-left and makeGroup-right), and it is not obvious in this case what should be an axiom and what should be a property.
> For example, If I'm doing a proof about multiples groups (say G and H), then I cannot simply write
> 
> ```
> open Group G
> open GroupLemmas G
> open Group H
> open GroupLemmas H
> ```
> 
> because of conflicting names.
> So I should instead do something like this :
> 
> ```
> module G = Group G
> module Glemmas = GroupLemmas G
> module H = Group H
> module Hlemmas = GroupLemmas H
> ```
> 
> But then it isn't obvious at all that I should use `G.invr` and `Glemmas.invId` and not the opposite. I think it just adds confusion to the end user experience.
> 
> Or maybe we should add all the notations and fields of the record type `Group` to the `GroupLemmas` module.
> But I think that too would reduce readability of the code. I think a comment explaining the choice to keep those properties there would make it all clearer and more practical to everyone.

Can't the `GroupLemmas` module just open the `Group` record publically? I think you would then get everything defined in Group when opening GroupLemmas.

Edlyr: Thanks for the remark, I will change this right away.

mortberg: Great! Can you move as much of the other group theory there as well?

Edlyr: I'm not sure I understand by "the other group theory" ?

Cubical/Structures/Group.agda
```diff
@@ -1,4 +1,8 @@
 {-# OPTIONS --cubical --no-import-sorts --safe #-}
+{-<<<<<<< HEAD
```
mortberg, line 2: Oops, seems like the merge didn't go so well...

Cubical/Structures/Group/Morphism.agda
```diff
@@ -0,0 +1,253 @@
...
+isPropIsGroupHom G H {f} = isPropΠ2 λ a b → Group.is-set H _ _
+
+record GroupHom (G : Group {ℓ}) (H : Group {ℓ'}) : Type (ℓ-max ℓ ℓ') where
+  constructor grouphom
+
+  field
+    fun : ⟨ G ⟩ → ⟨ H ⟩
+    isHom : isGroupHom G H fun
+
+record GroupIso (G : Group {ℓ}) (H : Group {ℓ'}) : Type (ℓ-max ℓ ℓ') where
```
mortberg, line 50: @ecavallo Didn't you call this GroupEquiv?

Edlyr: Oops, I forgot to add a few files before committing the merge.

Cubical/Structures/Group/Base.agda
mortberg: That seems like an outdated question from me

mortberg:
Except for the merge problems this is almost finished now! Can you please make sure that all the names in the Group files match the naming scheme that is on the master branch: https://github.com/agda/cubical/blob/master/Cubical/Structures/Group.agda 

(GroupIso should be GroupEquiv, all names should be written using camelCase, etc.)

mortberg:
> Except for the merge problems this is almost finished now! Can you please make sure that all the names in the Group files match the naming scheme that is on the master branch: https://github.com/agda/cubical/blob/master/Cubical/Structures/Group.agda
> 
> (GroupIso should be GroupEquiv, all names should be written using camelCase, etc.)

Oh, seems like you just fixed this! I'll skim through the code as soon as Travis is happy and then merge unless I find something

Cubical/Structures/Group/Morphism.agda
```diff
@@ -0,0 +1,255 @@
...
+groupHomEq G H (grouphom f fm) (grouphom g gm) p i = grouphom (p i) (p-hom i) where
+  p-hom : PathP (λ i → isGroupHom G H (p i)) fm gm
+  p-hom = toPathP (isPropIsGroupHom G H _ _)
+
+groupEquivEq : (G : Group {ℓ}) (H : Group {ℓ'}) (f g : GroupEquiv G H) → (GroupEquiv.eq f ≡ GroupEquiv.eq g) → f ≡ g
+groupEquivEq G H (groupequiv f fm) (groupequiv g gm) p i = groupequiv (p i) (p-hom i) where
+  p-hom : PathP (λ i → isGroupHom G H (p i .fst)) fm gm
+  p-hom = toPathP (isPropIsGroupHom G H _ _)
+
+module GroupΣTheory {ℓ} where
```
mortberg, line 110: I would expect this and the rest of the file to be in the Properties file?

Cubical/Structures/Group/Base.agda
```diff
@@ -0,0 +1,190 @@
...
+open import Cubical.Foundations.Univalence
+open import Cubical.Foundations.Transport
+
+open import Cubical.Data.Sigma
+
+open import Cubical.Structures.Macro
+open import Cubical.Structures.NAryOp
+open import Cubical.Structures.Pointed
+open import Cubical.Structures.Semigroup hiding (⟨_⟩)
+open import Cubical.Structures.Monoid hiding (⟨_⟩)
```
mortberg, line 18: Can you please prune the import list a bit? I don't think the Base file should depend on almost anything except for the Prelude and Monoid. Probably the same can be done for the Morphism file

Cubical/Structures/Group/Base.agda
```diff
@@ -0,0 +1,190 @@
...
+        comp a (comp (inv a) a)
+          ≡⟨ assoc a (inv a) a ⟩
+        comp (comp a (inv a)) a
+          ≡⟨ cong (λ b → comp b a) (rCancel a) ⟩
+        comp id a
+          ≡⟨ lUnit a ⟩
+        a
+          ∎
+
+isPropIsGroup : {G : Type ℓ} (0g : G) (_+_ : G → G → G) (-_ : G → G)
```
mortberg, line 181: Shouldn't this be in the Properties file?

Cubical/Structures/Group/Morphism.agda
```diff
@@ -0,0 +1,255 @@
...
+
+  field
+    eq : ⟨ G ⟩ ≃ ⟨ H ⟩
+    isHom : isGroupHom G H (equivFun eq)
+
+  hom : GroupHom G H
+  hom = grouphom (equivFun eq) isHom
+
+-- Morphism composition
+isGroupHomComp : (F : Group {ℓ}) (G : Group {ℓ'}) (H : Group {ℓ''}) →
```
mortberg, line 62: Can't you make the groups implicit here?

Edlyr: Yes indeed, that's another great reason to define GroupHom as a record type rather than a sigma type.

Cubical/Structures/Group/Morphism.agda
mortberg: Oops, I marked this as resolved too quickly. I think this is the last change and then I'll merge

Edlyr: I moved it to a separate "MorphismProperties" file for readability.

mortberg:
Great! I'll merge once Travis is done

mortberg:
Puh, thanks a lot for fixing this and keeping up with all the merge conflicts we caused by all the refactoring and renaming.

Issue #291, opened 2020-04-25
Add `isSet→` and friends


mortberg:
It would be nice to add `isSet→` and friends to avoid having to write `λ _ →` all the time when using `isSetΠ`

Issue #282, opened 2020-04-14
Add unique existence


mortberg:
We should add unique existence with notation \exists! defined as iscontr Sigma, not sure if this should go with exists in PropositionalTruncation.Properties or in the Prelude

PR #345, opened 2020-06-08
Various small changes

mortberg:
I kind of solved https://github.com/agda/cubical/issues/291 by realizing that it's not very useful to have. Added it for isProp at least

### DIFF
--- a/Cubical/Codata/M/AsLimit/M/Base.agda
+++ b/Cubical/Codata/M/AsLimit/M/Base.agda
@@ -115,30 +115,28 @@ shift-iso S@(A , B) =
         → (p : (n : ℕ) → a (suc n) ≡ a n)
         → (u : (n : ℕ) → B (a n) → Wₙ S n)
         → (n : ℕ)
-        → Iso
-            (PathP (λ x → B (p n x) → Wₙ S n) (πₙ S ∘ u (suc n)) (u n))
-            (πₙ S ∘ (subst (\k -> B k → Wₙ S (suc n)) (α-iso-step-5-Iso-helper0 a p (suc n))) (u (suc n)) ≡ subst (λ k → B k → Wₙ S n) (α-iso-step-5-Iso-helper0 a p n) (u n))
-      α-iso-step-5-Iso-helper1-Iso  a p u n  =
+        → (PathP (λ x → B (p n x) → Wₙ S n) (πₙ S ∘ u (suc n)) (u n)) ≡
+             (πₙ S ∘ (subst (\k -> B k → Wₙ S (suc n)) (α-iso-step-5-Iso-helper0 a p (suc n))) (u (suc n))
+         ≡ subst (λ k → B k → Wₙ S n) (α-iso-step-5-Iso-helper0 a p n) (u n))
+      α-iso-step-5-Iso-helper1-Iso  a p u n =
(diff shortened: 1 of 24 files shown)


mortberg:
I also solved https://github.com/agda/cubical/issues/282 by adding unique existence and move existence to Sigma.Base instead of PropositionalTruncation

Cubical/Codata/M/AsLimit/M/Base.agda
```diff
@@ -115,30 +115,28 @@ shift-iso S@(A , B) =
         → (p : (n : ℕ) → a (suc n) ≡ a n)
         → (u : (n : ℕ) → B (a n) → Wₙ S n)
         → (n : ℕ)
-        → Iso
-            (PathP (λ x → B (p n x) → Wₙ S n) (πₙ S ∘ u (suc n)) (u n))
-            (πₙ S ∘ (subst (\k -> B k → Wₙ S (suc n)) (α-iso-step-5-Iso-helper0 a p (suc n))) (u (suc n)) ≡ subst (λ k → B k → Wₙ S n) (α-iso-step-5-Iso-helper0 a p n) (u n))
-      α-iso-step-5-Iso-helper1-Iso  a p u n  =
+        → (PathP (λ x → B (p n x) → Wₙ S n) (πₙ S ∘ u (suc n)) (u n)) ≡
```
mortberg, line 118: @cmester0 : I simplified shift-iso a little bit. You can probably get rid of all of the other conversion between Iso and path. This should speed up using the proof a lot

mortberg: I think that once you do this you can probably get rid of the lemmas in the helper file. I think most of those are proved in more general form elsewhere in the library now anyway.

Cubical/Codata/M/AsLimit/M/Base.agda
```diff
@@ -151,7 +149,7 @@ shift-iso S@(A , B) =
       α-iso-step-5-Iso =
         Σ-cong-iso (lemma11-Iso {S = S} (λ _ → A) (λ _ x → x)) (λ a,p →
           Σ-cong-iso (pathToIso (cong (λ k → (n : ℕ) → k n) (funExt λ n → cong (λ k → B k → Wₙ S n) (α-iso-step-5-Iso-helper0 (a,p .fst) (a,p .snd) n)))) λ u →
-                              pathToIso (cong (λ k → (n : ℕ) → k n) (funExt λ n → isoToPath (α-iso-step-5-Iso-helper1-Iso (a,p .fst) (a,p .snd) u n))))
+                              pathToIso (cong (λ k → (n : ℕ) → k n) (funExt λ n → α-iso-step-5-Iso-helper1-Iso (a,p .fst) (a,p .snd) u n)))
```
mortberg, line 152: For example these uses of pathToIso could probably be removed. In fact I don't know if you should express this as an Iso at all. It's often easier to work with paths as much as possible

cmester0: I tried writing this as with paths however it was slower than what it is currently for some reason.

mortberg: I see. I guess pathToIso is actually not too bad. The problem is the other way around as it uses univalence which tends to lead to very big terms if you unfold things.

PR #346, opened 2020-06-09
[ refactor ] use with to simplify some code.

Saizan:


mortberg:
PS: for this to work everyone needs to update their Agda installations again...

PR #348, opened 2020-06-09
Ring structure on square matrices

mortberg:


mortberg:
Actually, let me open an issue about this so that we can have a discussion instead

Cubical/Algebra/Matrix.agda
```diff
@@ -70,38 +82,128 @@ FinMatrix≡VecMatrix _ _ _ = ua FinMatrix≃VecMatrix
...
+  mulFinMatrix1r : ∀ {m n} → (M : FinMatrix ⟨ R ⟩ m n) → mulFinMatrix M oneFinMatrix ≡ M
+  mulFinMatrix1r {n = zero} M = funExt₂ λ j → \ { () }
+  mulFinMatrix1r {n = suc n} M = funExt₂ λ j k → {!!}
+
+  mulFinMatrix1l : ∀ {m n} → (M : FinMatrix ⟨ R ⟩ m n) → mulFinMatrix oneFinMatrix M ≡ M
+  mulFinMatrix1l M i j k = {!!}
+
+  mulFinMatrixComm : ∀ {n} → (M N : FinMatrix ⟨ R ⟩ n n) → mulFinMatrix M N ≡ mulFinMatrix N M
+  -- Agda bug?
+  mulFinMatrixComm {n = n} M N i j k = sumVec {n = {!n!}} (λ l → commring·-comm R (M j k) (N k j) i)
```
mortberg, line 147: @Saizan I think I might be hitting some bug here... If I try to input `n` in the hole I get the error:
```
x₁ != x of type Fin n
when checking the definition of mulFinMatrixComm
```

There's now `x` around so I have no clue what is happening. In some other places I also get weird errors where I haven't introduced any `n₁` and I would expect to be able to provide `i` but Agda doesn't like it (despite the mysterious  `n₁` being equal to `n`):
```
Goal: Fin n₁
————————————————————————————————————————————————————————————
j      : Fin n
k      : Fin n
i      : Fin n
n      : ℕ
M      : FinMatrix ⟨ R ⟩ n₁ n₁
n = n₁ : ℕ   (not in scope)
```

Saizan: I can't reproduce the weirdness about `n`, but there's a legitimate reason not to accept that proof. The error message mentioning `x` however is an error reporting bug caused by the internal machinery losing track of the user-generated names.

With the following code

```agda
  mulFinMatrixComm : ∀ {n} → (M N : FinMatrix ⟨ R ⟩ n n) → mulFinMatrix M N ≡ mulFinMatrix N M
  mulFinMatrixComm {n = n} M N i j k = sumVec {n = n} \ l → { }6
```

I get the following constraints for the goal:

```agda
———— Constraints ———————————————————————————————————————————
R .snd .fst .snd .snd (M j l) (N l k) = ?6 (i = i0) : R .fst
R .snd .fst .snd .snd (N j l) (M l k) = ?6 (i = i1) : R .fst
```

notice that while `M` and `N` switch places they also get applied to different indexes, so just commutativity in the `R` ring is not enough to prove this.

mortberg: Oh I see, my bad! I guess Agda is smarter than me sometimes... Is the bug related to losing track of user-generated names reported? I would probably have realized what was wrong if the error message wasn't so weird...

Saizan: I don't think it's reported, would be nice to have a small testcase.

Cubical/Algebra/Matrix.agda
```diff
@@ -69,39 +80,183 @@ FinMatrix≡VecMatrix _ _ _ = ua FinMatrix≃VecMatrix
...
+  addFinMatrixNegMatrixr : ∀ {m n} → (M : FinMatrix R m n) → addFinMatrix M (negFinMatrix M) ≡ zeroFinMatrix
+  addFinMatrixNegMatrixr M i j k = +-rinv (M j k) i
+
+  addFinMatrixNegMatrixl : ∀ {m n} → (M : FinMatrix R m n) → addFinMatrix (negFinMatrix M) M ≡ zeroFinMatrix
+  addFinMatrixNegMatrixl M i j k = +-linv (M j k) i
 
   addFinMatrixComm : ∀ {m n} → (M N : FinMatrix R m n) → addFinMatrix M N ≡ addFinMatrix N M
   addFinMatrixComm M N i k l = +-comm (M k l) (N k l) i
 
+  mulFinMatrixAssoc : ∀ {m n k l} → (M : FinMatrix R m n) → (N : FinMatrix R n k) → (K : FinMatrix R k l)
```
mortberg, line 200: @felixwellen : I managed to do it! :-)

Cubical/Structures/Monoid.agda
```diff
@@ -173,6 +173,66 @@ module MonoidΣTheory {ℓ} where
     M ≡ N ■
 
 
+  -- TODO: clean and generalize the following?
```
mortberg, line 176: @ecavallo : CEHM?

ecavallo: https://github.com/mortberg/cubical/pull/1

mortberg: Lol! Thanks a lot

PR #350, opened 2020-06-09
Simplify proof of SIP

ecavallo:


Cubical/Foundations/SIP.agda
```diff
@@ -92,122 +76,23 @@ SNS-≡→SNS-PathP {S = S} ι θ {A = A} {B = B} e = EquivJ P C e (str A) (str
...
-  where
-  f : Iso (A ≃[ ι ] B) (A ≡ B)
-  Iso.fun f      = sip θ A B
-  Iso.inv f      = sip⁻ θ A B
-  Iso.rightInv f = sip-sip⁻ θ A B
-  Iso.leftInv f  = sip⁻-sip θ A B
 
+  SIP : A ≃[ ι ] B ≃ (A ≡ B)
+  SIP =
+    isoToEquiv (compIso (Σ-cong-iso (equivToIso (invEquiv univalence)) (equivToIso ∘ θ)) ΣPathIsoPathΣ)
```
mortberg, line 89: This proof is amazing! It made me wonder if we can come up with a direct proof of univalence expressed as an Iso...

Cubical/Foundations/SIP.agda
```diff
@@ -92,122 +76,23 @@ SNS-≡→SNS-PathP {S = S} ι θ {A = A} {B = B} e = EquivJ P C e (str A) (str
...
+
+  sip : A ≃[ ι ] B → A ≡ B
```
mortberg, line 91: It this definitionally the same as the old map? Maybe we could keep the old map and just construct the isEquiv part of the above proof so that it is clear what the computational behavior of `sip` is?

Cubical/Foundations/SIP.agda
mortberg: Actually it's already there: https://github.com/agda/cubical/blob/master/Cubical/Foundations/Univalence.agda#L182

Maybe you can give it a name and use here?

mortberg: PS: I mean "it's already **implicitly** there"

mortberg: Would giving this a name and using it simplify the proof of the SIP even more?

Cubical/Foundations/SIP.agda
mortberg: What do you think?

Cubical/Foundations/SIP.agda
ecavallo: Yeah, I can do that.

Cubical/Foundations/SIP.agda
ecavallo: I think it's definitionally the same map, I can expand it out.

PR #352, opened 2020-06-10
Records only

marcinjangrzybowski:
This is ongoing work on records, part of the larger code to generate various coherence conditions, 
temporarily i put some examples in Cubical/Data/Sigma/Record/Examples.agda file,
i was able to generate complicate records (200 fields, with lots of dependence between fields).

there are functions to generate constructors for record where explicitly of the fields can be configurated, those functions have "baked in" paths between constructors
sadly the length of the records often must be provided explicitly. 
I would apriciete advice on some way to fix it.

### DIFF
--- a/Cubical/Data/Sigma/Record/Base.agda
+++ b/Cubical/Data/Sigma/Record/Base.agda
@@ -1,0 +1,436 @@
+{-# OPTIONS --cubical --no-import-sorts --safe #-}
+module Cubical.Data.Sigma.Record.Base where
+
+open import Cubical.Core.Everything
+
+open import Cubical.Data.Nat
+
+open import Cubical.Foundations.Equiv.Fiberwise
+
+open import Cubical.Data.List renaming (_++_ to _++L_)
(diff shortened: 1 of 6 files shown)


Saizan:
Can you explain what the motivation for this work is? And a bit about the design.

marcinjangrzybowski:
in Cubical/Data/Sigma/Record/Examples.agda file there is an example of Σ-assoc-7
associativity of sigmas of any length, this function can generate similar equivalences for any other nested shape of Sigmas. (by "any other" I mean shapes without branching, only different combinations of  Σ Σ _ and  Σ _ Σ )

outside of this PR I was able to define analogues of Square and Cube from Prelude for arbitrary dimensions together with hcomp and hfill, and I belive that its possible (I did not finish it yet) to generate proofs of some coherence conditions of pathComp for arbitrary dimension.

maybe I should wait with this PR until I develop more useful results?
I did not wanted to make to large PR :), I also wanted to get some initial feedback about It

Issue #347, opened 2020-06-09
definition of the n-spheres


pnlph:
Hi,

I do not find the definition `_-sphere` from [Cubical Synthetic](https://dl.acm.org/doi/abs/10.1145/3372885.3373825).

Is the name changed or it is still to be added?

Thanks!

mortberg:
I think it has been renamed to https://github.com/agda/cubical/blob/master/Cubical/HITs/Sn/Base.agda 

Is this what you're looking for?

pnlph:
Yes, I think so, thanks! Then, the notation was changed from `n -sphere` to `S n`.

That also answers another question I had about finding `S⁰`. In the paper `S⁰` seems to be represented directly by `Bool` but now it is also a suspension itself, of the empty type.

I see also now the proof of `Bool≃Susp⊥`. I guess what is called `s2c : 1 -sphere → S¹` is now `SuspBool→S¹ : SuspBool → S¹`.

mortberg:
Yeah, some things have changed in the library since we wrote the paper and we also had to compress some names to fit with the small columns. The changes are mostly just renaming and some minor changes in definitions that turned out to be better for more advanced proofs later on.

mortberg:
Btw, if you want to make a useful contribution you can make a new file in a new folder called "Papers" and import everything from that paper but give it the names that we used for the paper. I guess we should have done this when writing the paper, but as usual there was not much time left when the deadline approached. A similar file for the Cubical Agda paper would of course al be appreciated.

PR #353, opened 2020-06-10
Create Synthetic.agda

pnlph:
First PR. Create Papers folder and the code for the synthetic cubical paper. See issue #347 

Preserved as much as could the notation of the original paper.

Let's see how it goes :-)

### DIFF
--- a/Cubical/Papers/Synthetic.agda
+++ b/Cubical/Papers/Synthetic.agda
@@ -27,6 +27,9 @@ import Cubical.Foundations.Isomorphism        as Isomorphism
 -- 4.1
 import Cubical.HITs.Susp                      as Suspension
 import Cubical.HITs.Sn                        as Sn
+import Agda.Builtin.Nat                       as BNat
+import Agda.Builtin.Cubical.Path              as BPath
+import Agda.Builtin.Bool                      as BBool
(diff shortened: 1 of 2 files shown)


mortberg:
Very nice, thank you! Your file contains some whitespace in the end which Travis complains about. Have a look at https://github.com/agda/cubical/blob/master/CONTRIBUTING.md to see how you can easily fix it.

pnlph:
The test is still not passing, I double-checked that there are not whitespaces at the end of the lines. Could be the problem that I am using Atom? I also turned all the spaces to tabs with the `whitespace` package.

With whitespace is meant leaving one or more empty spaces at the end of a line, right?

mortberg:
Yes, you can't have any empty spaces at the end of lines. Try loading the file with emacs and the run "M-x delete-trailing-whitespace"

guillaumebrunerie:
> I also turned all the spaces to tabs with the `whitespace` package.

I don’t know if that’s the issue, but it’s usually better to use spaces instead of tabs.

pnlph:
Finally Travis is happy, everything fixed. Thank you for the guidance.

mortberg:
Great, thank you for this! If you feel like doing more of it I think it would be useful to have this for the Cubical Agda paper itself as well.

PR #354, opened 2020-06-11
organizing in zigzag

ecavallo:
- Change the definitions of the equivalence relations induced by a bisimulation so that they're prop-valued (but give the same set-quotient).
- Switch things around so that we're not repeating ourselves defining the eq. rels on the left and right types.
- (Not really related) Change the definition of `isEffective` to ask for a specific map to be an equivalence. It's equivalent but I think this definition is more natural.
- Use a record for defining bisimulation.

### DIFF
--- a/Cubical/Relation/ZigZag/Applications/MultiSet.agda
+++ b/Cubical/Relation/ZigZag/Applications/MultiSet.agda
@@ -100,7 +100,7 @@ module Lists&ALists {A : Type ℓ} (discA : Discrete A) where
 
  -- R {List A , Lcount} {AList A , ALcount} is a bisimulation
  isBisimR : isBisimulation (R {List A , Lcount} {AList A , ALcount})
- isBisimR .zigzag r r' r'' a = (r a) ∙∙ sym (r' a) ∙∙ (r'' a)
+ isBisimR .zigzag r r' r'' a = r a ∙∙ sym (r' a) ∙∙ r'' a
  isBisimR .fwd = φ
  isBisimR .fwdRel = η
  isBisimR .bwd = ψ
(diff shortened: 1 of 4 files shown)


Cubical/Relation/ZigZag/Base.agda
```diff
@@ -6,93 +6,84 @@ module Cubical.Relation.ZigZag.Base where
...
-
-isBisimulation : {A B : Type ℓ} (R : A → B → Type ℓ') (f : A → B) (g : B → A) → Type (ℓ-max ℓ ℓ')
-isBisimulation R f g =  isZigZagComplete R
-                      × (∀ a → R a (f a))
-                      × (∀ b → R (g b) b)
-
+isZigZagComplete R = ∀ {a b a' b'} → R a b → R a' b → R a' b' → R a b'
+
+ZigZagRel : (A B : Type ℓ) (ℓ' : Level) → Type (ℓ-max ℓ (ℓ-suc ℓ'))
+ZigZagRel A B ℓ' = Σ (A → B → Type ℓ') isZigZagComplete
```
mortberg, line 24: Can you use the Sigma notation here as well?

Cubical/Relation/ZigZag/Applications/MultiSet.agda
```diff
@@ -94,41 +94,46 @@ module Lists&ALists {A : Type ℓ} (discA : Discrete A) where
...
+ -- Induced quotients and equivalence
 
- -- R {List A , Lcount} {AList A , ALcount} is zigzag-complete
- zigzagR : isZigZagComplete (R {List A , Lcount} {AList A , ALcount})
- zigzagR _ _ _ _ r r' r'' a = (r a) ∙∙ sym (r' a) ∙∙ (r'' a)
+ open isBisimulation
 
+ -- R {List A , Lcount} {AList A , ALcount} is a bisimulation
+ isBisimR : isBisimulation (R {List A , Lcount} {AList A , ALcount})
+ isBisimR .zigzag r r' r'' a = (r a) ∙∙ sym (r' a) ∙∙ (r'' a)
```
mortberg, line 103: Do we really need parantheses around `r a` and `r'' a` ?

ecavallo:
Made the changes.

PR #355, opened 2020-06-12
Option structure

ecavallo:
It would be nice to use this in the queue example, but unfortunately that makes it more complicated. Note the difference between the hand-rolled `deq-iso` in `Structures.Queue` and the definition you would get it using this combinator. The former uses the action of `Unit ⊎_` on functions; the latter uses a `Codes`-like family (the action on relations), which is the best we can do in general.


Cubical/Foundations/Equiv.agda
```diff
@@ -112,8 +112,22 @@ compEquiv-assoc : (f : A ≃ B) (g : B ≃ C) (h : C ≃ D)
                 → compEquiv f (compEquiv g h) ≡ compEquiv (compEquiv f g) h
 compEquiv-assoc f g h = equivEq _ _ refl
 
-LiftEquiv : A ≃ Lift {i = ℓ} {j = ℓ'} A
-LiftEquiv = isoToEquiv LiftIso
+LiftEquiv : {A : Type ℓ} → A ≃ Lift {j = ℓ'} A
```
mortberg, line 115: Why is this proof expanded like this?

ecavallo: I do not know, I'll change it back.

ecavallo:
Somehow I didn't notice that there's already a Maybe data type in the library. I'll switch this (and the queue code) to use that.

mortberg:
> Somehow I didn't notice that there's already a Maybe data type in the library. I'll switch this (and the queue code) to use that.

Great! I forgot that there was a Maybe file in the library as well...

ecavallo:
Ok, Maybe is everywhere.

Cubical/Structures/Option.agda
```diff
@@ -0,0 +1,92 @@
+{-# OPTIONS --cubical --no-import-sorts --no-exact-split --safe #-}
```
mortberg, line 1: Should we really keep this file now that we have Maybe?

ecavallo: No, I just forgot to remove it

Cubical/Foundations/Equiv.agda
```diff
@@ -113,7 +113,21 @@ compEquiv-assoc : (f : A ≃ B) (g : B ≃ C) (h : C ≃ D)
 compEquiv-assoc f g h = equivEq _ _ refl
 
 LiftEquiv : A ≃ Lift {i = ℓ} {j = ℓ'} A
-LiftEquiv = isoToEquiv LiftIso
+LiftEquiv .fst a .lower = a
```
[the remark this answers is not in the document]
ecavallo: It's a smaller proof, but I don't really care

Cubical/Structures/Macro.agda
```diff
@@ -31,6 +32,8 @@ data Desc (ℓ : Level) : Typeω where
   param : ∀ {ℓ'} → (A : Type ℓ') → Desc ℓ  → Desc ℓ
   -- structure S parameterized by variable argument: X ↦ (X → S X)
   recvar : Desc ℓ  → Desc ℓ
+  -- Maybe on a structure S: X ↦ Maybe (S X)
```
mortberg, line 35: Isn't maybe a functorial action?

ecavallo: On its own, yes, but not necessarily in combination with other things. For example X |-> Maybe (X -> X) isn't functorial.

PR #356, opened 2020-06-12
Notion of structure from a functorial action

ecavallo:
This is preferable to #355 for the queue example.


mortberg:
Looks good to me! Merging

PR #358, opened 2020-06-13
Improved I-R univalent universe helper

dolio:
- The higher constructor embeding equivalences and the path
  witnessing its expected computational behavior are sufficient
  to demonstrate that the I-R universe is univalent. No further
  higher constructors are necessary.


Saizan:
Thanks! 

Indeed it would even be possible to contract away the equivalence from the type of both `un` and `comp`, though I wonder if things would compute worse then.

PR #360, opened 2020-06-15
Rewriting Semigroup using records and transporting results from the Sigma definition

mortberg:


mortberg:
Oops, this was supposed to be a draft PR... Comments are very welcome! Can someone see if we can simplify some things to make the transport of results easier?

Cubical/Structures/Semigroup.agda
```diff
@@ -3,80 +3,168 @@ module Cubical.Structures.Semigroup where
...
+    Carrier     : Type ℓ
+    _·_         : Carrier → Carrier → Carrier
+    isSemigroup : IsSemigroup _·_
+
+  open IsSemigroup isSemigroup public
+
+
+record SemigroupIso (M N : Semigroup {ℓ}) : Type ℓ where
+
+  -- How can this be written nicer without all the qualified names?
```
mortberg, line 46: Does anyone know the answer to this? @Saizan maybe?

guillaumebrunerie: Untested, but you should be able to do something like this (adding those two lines between `where` and `field`)
```agda
open Semigroup M renaming (Carrier to [M]; _·_ to _·M_)
open Semigroup N renaming (Carrier to [N]; _·_ to _·N_)
```
and then simply use `[M]`, `[N]`, `_·M_` and `_·N_` (or whatever other names you choose).

guillaumebrunerie: Something else that might work and doesn’t require renaming explicitly all the fields is (also untested)
```agda
module M = Semigroup M
module N = Semigroup N
```
and then use `M.Carrier`, `N.Carrier`, `x M.· y`, and so on.
Not entirely sure it’s ok to reuse the same names `M` and `N` for those modules, but it’s worth trying.

Saizan: the namespace for modules is separate, so you can reuse `M` and `N` for the module names.

Cubical/Structures/Semigroup.agda
```diff
@@ -3,80 +3,168 @@ module Cubical.Structures.Semigroup where
...
-SemigroupPath : (M N : Semigroup {ℓ}) → (M ≃[ semigroup-iso ] N) ≃ (M ≡ N)
-SemigroupPath = SIP semigroup-is-SNS
+-- semigroup-assoc : (G : Semigroup' {ℓ})
+--                 → (x y z : ⟨ G ⟩) → (x ·⟨ G ⟩ (y ·⟨ G ⟩ z)) ≡ ((x ·⟨ G ⟩ y) ·⟨ G ⟩ z)
+-- semigroup-assoc (_ , _ , _ , P) = P
 
--- Semigroup ·syntax
+-- -- Semigroup ·syntax
 
-module semigroup-·syntax (G : Semigroup {ℓ}) where
```
mortberg, line 77: @Saizan @guillaumebrunerie : Do you know if we need this kind of syntax modules when we use records? It seems to me like we don't as we can open the record and name the operator whatever we want?

(My Agda skills are quite rusty and because we've been using nested Sigmas for everything up until now I haven't really figured out how to work most efficiently with records yet)

Cubical/Structures/Semigroup.agda
mortberg: Thanks! I'll try both of these and figure out which I prefer

Cubical/Structures/Semigroup.agda
guillaumebrunerie: Indeed, if `Semigroup` is a record and `G : Semigroup`, then `Semigroup G` is a module that contains all the fields of `G`, so you don’t need to define such a module yourself.

mortberg: Great! Yet another reason to not use nested Sigma types

Cubical/Structures/Semigroup.agda
mortberg: Both worked fine, but I liked the `module M = Semigroup M` solution a bit more

mortberg:
I'm pretty much happy with this now. Any comments?

Cubical/Structures/Semigroup.agda
```diff
@@ -3,80 +3,160 @@ module Cubical.Structures.Semigroup where
...
+⟨_⟩ = Semigroup.Carrier
+
+
+record SemigroupIso (M N : Semigroup {ℓ}) : Type ℓ where
+
+  constructor semigroupiso
+
+  -- Shorter qualified names
+  module M = Semigroup M
+  module N = Semigroup N
```
ecavallo, line 64: You might want to put these definitions in a `private`.

mortberg: Good point! Fixed

PR #361, opened 2020-06-15
Structured relations

ecavallo:
To go along with structured isomorphisms. This is a work in progress.


Cubical/Foundations/RelationalStructure.agda
```diff
@@ -0,0 +1,79 @@
...
+    struct : Type ℓ → Type ℓ'
+    set : ∀ {A} → isSet A → isSet (struct A)
+
+open SetStructure public
+
+-- A notion of structured relation for a structure S assigns a relation on S X and S Y to every relation on X
+-- and Y. We require the output to be proposition-valued when the input is proposition-valued.
+record StrRel (S : Type ℓ → Type ℓ') (ℓ'' : Level) : Type (ℓ-max (ℓ-suc (ℓ-max ℓ ℓ'')) ℓ') where
+  field
+    rel : (A B : Type ℓ) (R : Rel A B ℓ) → Rel (S A) (S B) ℓ''
```
mortberg, line 35: Can't A and B be implicit here? Or does Agda not infer them properly later on?

Cubical/Foundations/RelationalStructure.agda
```diff
@@ -0,0 +1,79 @@
...
+-- A notion of structured relations is standard when
+-- (a) Given a structured type A and equivalence relation R, there is at most one quotient structure on A/R
+-- (b) Any bisimulation descends to the quotients if and only if it is structured
+record isSNRS (S : SetStructure ℓ ℓ') (ρ : StrRel (S .struct) ℓ'') : Type (ℓ-max (ℓ-max (ℓ-suc ℓ) ℓ') ℓ'')
+  where
+  field
+    propQuo : {A : TypeWithStr ℓ (S .struct)}
+      (R : Σ[ R ∈ (typ A → typ A → Type ℓ) ] isEquivRel R)
+      → isProp (QuotientStructure (S .struct) ρ A (R .fst))
+    descends : {A B : TypeWithStr ℓ (S .struct)}
```
mortberg, line 73: I guess nothing would be simpler if descends would be expressed as an equivalence? Maybe one could prove it as a lemma though

Cubical/Structures/Maybe.agda
```diff
@@ -80,6 +83,11 @@ module MaybePathP where
     isom .Iso.rightInv = decodeEncode ox oy
     isom .Iso.leftInv = encodeDecode A ox oy
 
+-- Structured isomorphisms
+
+maybe-structure : (S : Type ℓ → Type ℓ₁) → Type ℓ → Type ℓ₁
+maybe-structure S X = Maybe (S X)
```
mortberg, line 89: This is the kind of thing I was thinking would be a composition of structures. Maybe is functorial structure and S is just a structure... Could this make sense?

Cubical/Structures/Relational/Macro.agda
```diff
@@ -0,0 +1,87 @@
...
+  var : RelDesc ℓ
+  -- join of structures S,T : X ↦ (S X × T X)
+  _,_ : RelDesc ℓ  → RelDesc ℓ  → RelDesc ℓ
+  -- structure S parameterized by constant A : X ↦ (A → S X)
+  param : ∀ {ℓ'} → (A : Type ℓ') → RelDesc ℓ  → RelDesc ℓ
+  -- structure S parameterized by variable argument: X ↦ (X → S X)
+  recvar : RelDesc ℓ  → RelDesc ℓ
+  -- Maybe on a structure S: X ↦ Maybe (S X)
+  maybe : RelDesc ℓ → RelDesc ℓ
+  -- arbitrary structure with notion of structured isomorphism given by functorial action
```
mortberg, line 33: This comment should probably be removed..

Cubical/Foundations/RelationalStructure.agda
```diff
@@ -0,0 +1,79 @@
...
+    quoᴸ : QuotientStructure S ρ A E.Rᴸ
+    quoᴿ : QuotientStructure S ρ B E.Rᴿ
+    path : PathP (λ i → S (ua E.Thm i)) (quoᴸ .fst) (quoᴿ .fst)
+
+open BisimDescends
+
+-- A notion of structured relations is standard when
+-- (a) Given a structured type A and equivalence relation R, there is at most one quotient structure on A/R
+-- (b) Any bisimulation descends to the quotients if and only if it is structured
+record isSNRS (S : SetStructure ℓ ℓ') (ρ : StrRel (S .struct) ℓ'') : Type (ℓ-max (ℓ-max (ℓ-suc ℓ) ℓ') ℓ'')
```
mortberg, line 67: Where do you actually use that S is a SetStructure? It seems like this definition isn't...

ecavallo: It comes up in the proofs for `λ X → (X → S X)`. Roughly, when you're showing that a structure `X → S X` on `X` induces a structure `X/R → S (X/R)` on the quotient `X/R`, you need to know that `S (X/R)` is a set to define it by induction on the set quotient.

mortberg: Maybe one could have this as an additional assumption there instead of assuming it globally? I'm not sure if that would make everything else annoying though

Cubical/Structures/Relational/Macro.agda
ecavallo: whoops

Cubical/Structures/Maybe.agda
ecavallo: My intuition is that it won't make sense, but I'll think about it.

Cubical/Foundations/RelationalStructure.agda
ecavallo: Oh, and the fact that the maps back and forth give you an equivalence in the definition relies on it, so if we remove the assumption we should change the definition to require an equivalence.

mortberg: Ok, sounds like we should keep it then

PR #362, opened 2020-06-15
Rewrite monoids using records

mortberg:
I did a similar rewrite as the one I just merged for Semigroup and everything worked very smoothly

### DIFF
--- a/Cubical/Structures/Monoid.agda
+++ b/Cubical/Structures/Monoid.agda
@@ -8,23 +8,19 @@ open import Cubical.Foundations.SIP renaming (SNS-PathP to SNS)
 
 open import Cubical.Data.Sigma
 
-open import Cubical.Structures.Pointed
-open import Cubical.Structures.NAryOp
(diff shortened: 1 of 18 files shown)


Cubical/Structures/Monoid.agda
```diff
@@ -3,124 +3,180 @@ module Cubical.Structures.Monoid where
...
-
-Monoid : Type (ℓ-suc ℓ)
-Monoid {ℓ} = TypeWithStr ℓ monoid-structure
-
--- Monoid extractors
-
-⟨_⟩ : Monoid {ℓ} → Type ℓ
-⟨ G , _ ⟩ = G
+  monoid-axioms : (A : Type ℓ) → raw-monoid-structure A → Type ℓ
+  monoid-axioms A (e , _·_) = IsSemigroup _·_
```
mortberg, line 91: Note that I used IsSemigroup here and not the nested Sigma type version of this

PR #367, opened 2020-06-18
Cost example using truncated natural numbers

mortberg:
A cute example of programming with HITs due to @cangiuli . We record the number of recursive calls using truncated natural numbers. This is a nice example of truncated types having interesting content

This could probably be done nicer by turning Cost into a monad and programming using do-notation. Is there some reference for how to achieve this in Agda?

### DIFF
--- a/Cubical/HITs/Cost.agda
+++ b/Cubical/HITs/Cost.agda
@@ -1,0 +1,4 @@
+{-# OPTIONS --cubical --no-import-sorts --safe #-}
+module Cubical.HITs.Cost where
+
+open import Cubical.HITs.Cost.Base
--- a/Cubical/HITs/Cost/Base.agda
+++ b/Cubical/HITs/Cost/Base.agda
@@ -1,5 +1,5 @@
 {-# OPTIONS --cubical --no-import-sorts --safe #-}
-module Cubical.HITs.TruncNat.Base where
+module Cubical.HITs.Cost.Base where
 
 open import Cubical.Foundations.Prelude
 open import Cubical.Foundations.Function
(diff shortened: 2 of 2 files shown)


martinescardo:
Nice!

potato4444:
@mortberg Here is the section of the manual that explains do notation: https://agda.readthedocs.io/en/v2.6.1/language/syntactic-sugar.html

My short interprtation of it is functions `_>>=_` and `_>>_` need to be defined.

mortberg:
> @mortberg Here is the section of the manual that explains do notation: https://agda.readthedocs.io/en/v2.6.1/language/syntactic-sugar.html
> 
> My short interprtation of it is functions `_>>=_` and `_>>_` need to be defined.

Thanks! I got it to work now and the code reads nicer. 

I don't think the way I defined >>= and return forms a proper monad though...

potato4444:
@mortberg Using typal equality this is a monad (it's the identity monad). Using definitional equality it fails both the left and right identity laws: 

    (return 42 >>= return) = (42 , | 1 |) /= return 42 = ( 42 , | 0 |)

I haven't looked at the associativity law, but my first guess is it does hold.

mortberg:
> @mortberg Using typal equality this is a monad (it's the identity monad). Using definitional equality it fails both the left and right identity laws:
> 
> ```
> (return 42 >>= return) = (42 , | 1 |) /= return 42 = ( 42 , | 0 |)
> ```
> 
> I haven't looked much at the associativity law, but my first guess is it does hold for concrete values.

Oh right, of course. I'm not sure if counting the number of calls to bind is what we want here, but it makes the examples work as I want. We could of course also just have a state monad with a truncated natural number and explicitly increase it.

mortberg:
> > @mortberg Using typal equality this is a monad (it's the identity monad). Using definitional equality it fails both the left and right identity laws:
> > ```
> > (return 42 >>= return) = (42 , | 1 |) /= return 42 = ( 42 , | 0 |)
> > ```
> > 
> > 
> > I haven't looked much at the associativity law, but my first guess is it does hold for concrete values.
> 
> Oh right, of course. I'm not sure if counting the number of calls to bind is what we want here, but it makes the examples work as I want. We could of course also just have a state monad with a truncated natural number and explicitly increase it.

I checked and all the monad laws hold trivially up to a path.

Issue #278, opened 2020-04-13
Prove more stuff about pointed types


mortberg:
It would be nice to port https://github.com/mortberg/cubicaltt/blob/master/examples/pointedMaps.ctt to Cubical Agda. This might benefit from https://github.com/agda/cubical/blob/master/Cubical/Structures/Pointed.agda

UlrikBuchholtz:
Right now [Π∙](https://github.com/agda/cubical/blob/master/Cubical/Foundations/Pointed/Properties.agda#L11) is defined in terms of a family of pointed types, but that's not the most useful pointed Pi: There we just need the fiber over the base point to be pointed. That way, pointed homotopies become elements of pointed Pi's. Shall I try to rework that? (Of course, for the pointed Pi type itself to be pointed, you need a family of pointed types, but from the latter you can extract the base point in the fiber over the base point.)

mortberg:
Yes, that sounds like a good thing to implement!

PR #369, opened 2020-06-20
Pointed Homotopies, Function Extensionality

Schippmunk:
This PR

- fixes the pointed Pi as noted by @UlrikBuchholtz in #278 (Credit to him for that)
- adds two definitions of pointed homotopies between two functions of the new pointed Pi and proves their equivalence
- adds pointed function extensionality and proves that it is an equivalence
- adds \sim notation for regular homotopies


Cubical/Foundations/Pointed/Properties.agda
```diff
@@ -1,44 +1,60 @@
 {-# OPTIONS --cubical --no-import-sorts --safe #-}
```
mortberg, line 1: This file would be a bit nicer with the levels declared at the top as variables. Can you fix?

mortberg:
Thanks!

PR #370, opened 2020-06-23
retractableIntoSet→isEmbedding, isPropIsHAEquiv and more universe polymorphic definitions in Equiv/Properties

WorldSEnder:
A bit messier than I'd hoped, since the PR contains more unrelated things than I initially expected.

I originally wanted to prove that any retract of a function into a set makes that function an embedding. Then I stumbled upon other stuff to do, mostly with HAEquivs:

- `isHAEquiv→Iso` and `isHAEquiv→isEquiv`
- a more direct proof of `equiv→HAEquiv`
- added `depPostCompEquiv`
- `postCompEquiv` and few other things are more universe polymorphic
- direct proof of `isEquiv→isContrHasRetract` (for universe polymorphism and term size), a bit more complicated than the one for HasSection, but manageable.
- **`isPropIsHAEquiv`** didn't find it elsewhere and thought I'd want to use it. Ended up not using it.
- `hcomp-equivFiller`: this basically proves that `hcomp` has an inverse, but I don't know to straight-forwardly state that. See `slideSquareEquiv` for a usage.
- shortened proof of `substComposite` and `substCommSlice`
- last but not least `retractableIntoSet→isEmbedding : hasRetract f → isSet B → isEmbedding f`

If I should split the PR into more self-contained pieces, just tell me. Other comments are of course welcome too, as always.

### DIFF
--- a/Cubical/Foundations/Equiv/Properties.agda
+++ b/Cubical/Foundations/Equiv/Properties.agda
@@ -184,7 +184,7 @@ isPropIsHAEquiv {f = f} ishaef = goal ishaef where
       -- secondly, convert the path into a dependent path for later convenience
       ≃⟨  Σ-cong-equiv-snd (λ s → Σ-cong-equiv-snd
                              λ η → depPostCompEquiv
-                               λ x → compEquiv (flipSquareEquiv {a₀₀ = f x}) (invEquiv (slideSquareEquiv _ _ _))) ⟩
+                               λ x → compEquiv (flipSquareEquiv {a₀₀ = f x}) (invEquiv slideSquareEquiv)) ⟩
     Σ _ rCoh2
       ≃⟨ Σ-cong-equiv-snd (λ s → invEquiv Σ-Π-≃) ⟩
     Σ _ rCoh3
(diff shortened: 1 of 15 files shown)


Cubical/Foundations/Equiv/Properties.agda
```diff
@@ -18,83 +19,129 @@ open import Cubical.Data.Sigma
...
 
--- There are more universe polymorphic versions of these results in
--- Cubical.Foundations.Equiv.HalfAdjoint that seem to have better
--- computational behavior
-private
-  isEquivCong : ∀ {ℓ} {A B : Type ℓ} {x y : A} (e : A ≃ B) → isEquiv (λ (p : x ≡ y) → (cong (fst e) p))
-  isEquivCong {B = B} e = EquivJ (λ (A' : Type _) (e' : A' ≃ B) →
-                                 (x' y' : A') → isEquiv (λ (p : x' ≡ y') → cong (fst e') p))
-                                 (λ x' y' → idIsEquiv (x' ≡ y')) e _ _
+isEquivCong : {x y : A} (e : A ≃ B) → isEquiv (λ (p : x ≡ y) → cong (e .fst) p)
```
mortberg, line 35: Very nice that this is here now!

mortberg:
Great PR with lots of useful general results! I had no comments so I'll just go ahead and merge. 

A result that I don't think we have is that isIso for sets is a prop? Unless I'm mistaken this should be true

PR #371, opened 2020-06-24
More direct proof of isPropEquiv→Equiv and propositional extensionality

mortberg:
I hope this isn't somewhere else already?


Cubical/Foundations/Univalence.agda
```diff
@@ -38,6 +38,10 @@ ua {A = A} {B = B} e i = Glue B (λ { (i = i0) → (A , e)
 uaIdEquiv : {A : Type ℓ} → ua (idEquiv A) ≡ refl
 uaIdEquiv {A = A} i j = Glue A {φ = i ∨ ~ j ∨ j} (λ _ → A , idEquiv A)
 
+-- Propositional extensionality
+hPropExt : {A B : Type ℓ} → isProp A → isProp B → (A → B) → (B → A) → A ≡ B
+hPropExt Aprop Bprop f g = ua (isPropEquiv→Equiv Aprop Bprop f g)
+
```
WorldSEnder, line 44: There is a similar construction in Logic

https://github.com/agda/cubical/blob/d9e9db0725f5267ed4fe88f767f646a691278f88/Cubical/Foundations/Logic.agda#L75-L77

mortberg: Ah right, we should probably do something about the Logic file. It's not very useful as it is now

PR #378, opened 2020-06-27
Small corrections to files related to groups

Edlyr:
I made a few arguments implicit where I felt they were unnecessary, and made "public" the "open Group" statement in GroupLemmas (which I forgot to do in the previous PR).


mortberg:
Great, thanks!

PR #377, opened 2020-06-26
function structures

ecavallo:
Given structures `S` and `T`, we get a structure `\ X -> (S X -> T X)`. (The already defined structures on `X -> S X` and `A -> S X` are more convenient in those special cases, so we should keep them around.)

PR #379, opened 2020-06-29
Make set preservation a property of suitable relation structure

ecavallo:
Results are a little less sharp, but conceptually simpler.


ecavallo:
Subsumed by #377

PR #380, opened 2020-07-01
Transport structures

ecavallo:
Given a structure `S : Type -> Type` with an action on equivalences that agrees with transport, we get a univalent notion of structure.


Cubical/Foundations/SIP.agda
```diff
@@ -80,6 +80,28 @@ SNS→UnivalentStr {S = S} ι θ {A = A} {B = B} e = EquivJ P C e (str A) (str B
     PathP (λ i → S (ua (idEquiv Y) i)) s t
     ■
 
+TransportStr : {S : Type ℓ → Type ℓ₁} (α : EquivAction S) → Type (ℓ-max (ℓ-suc ℓ) ℓ₁)
+TransportStr {ℓ} {S = S} α =
+  {X Y : Type ℓ} (e : X ≃ Y) (s : S X) → equivFun (α e) s ≡ subst S (ua e) s
+
+TransportStr→UnivalentStr : {S : Type ℓ → Type ℓ₁} (α : EquivAction S)
```
mortberg, line 87: Did you try to prove the converse?

Cubical/Foundations/SIP.agda
```diff
@@ -80,6 +80,28 @@ SNS→UnivalentStr {S = S} ι θ {A = A} {B = B} e = EquivJ P C e (str A) (str B
...
+  subst S (ua e) s ≡ t
+    ≃⟨ invEquiv (PathP≃Path _ _ _) ⟩
+  PathP (λ i → S (ua e i)) s t
+  ■
+
+invTransportStr : {S : Type ℓ → Type ℓ₂} (α : EquivAction S) (τ : TransportStr α)
+  {X Y : Type ℓ} (e : X ≃ Y) (t : S Y) → invEq (α e) t ≡ subst⁻ S (ua e) t
+invTransportStr {S = S} α τ e t =
+  sym (transport⁻Transport (cong S (ua e)) (invEq (α e) t))
+  ∙ sym (cong (subst⁻ S (ua e)) (τ e (invEq (α e) t)))
```
mortberg, line 101: Use "_.._.._"

Cubical/Foundations/SIP.agda
ecavallo: Forgot to, got it now.

ecavallo:
In terms of concrete applications, the `Transp[_]` occurs explicitly in `Cubical.Structures.Queue`. But it's also used whenever we have a function, like `\ X -> X -> X`; `autoDesc` turns that into `function+ var var`, with the first var being a `transpDesc`. This is how the structured equivalences for `(X , f)`,`(Y, g)`,`e` get calculated as

```(x : X) -> e .fst (f x) ≡ g (e .fst x)```

rather than the less convenient

```{x : X} {y : Y} -> (e .fst x ≡ y) -> e .fst (f x) ≡ g y```

PR #384, opened 2020-07-07
truncate choice functions for QER

ecavallo:
Weaken the definition of QER to only require `(a : A) → ∃[ b ∈ B ] R a b` and vice versa, which is all that's needed for the applications. Then we get that R is a QER iff (R . R^-1) and (R^{-1} . R) are equivalence relations (though I didn't prove it explicitly in this PR), which matches the situation for QPERs.

### DIFF
--- a/Cubical/Foundations/RelationalStructure.agda
+++ b/Cubical/Foundations/RelationalStructure.agda
@@ -224,29 +224,13 @@ structuredQER→structuredEquiv : {S : Type ℓ → Type ℓ'} {ρ : StrRel S �
   → QERDescends S ρ A B R
 structuredQER→structuredEquiv {ρ = ρ} θ (X , s) (Y , t) R r .quoᴸ =
   θ .quo (X , s) (QER→EquivRel R)
-    (subst (λ R' → ρ R' s s) correction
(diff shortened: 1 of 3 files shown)


Cubical/Foundations/RelationalStructure.agda
```diff
@@ -270,14 +254,16 @@ structuredQER→structuredEquiv {ρ = ρ} θ (X , s) (Y , t) R r .rel =
...
-                cong (E.Thm .fst) (sym px)
-                ∙ eq/ (R .snd .fwd x) y (R .snd .zigzag (R .snd .bwdRel y) r (R .snd .fwdRel x))
+                (cong (E.Thm .fst) (sym px)
+                ∙ E.relToFwd≡ r)
                 ∙ py})
               qr}))
-        (elimProp
-          {B = λ qx →
-            E.Thm .fst qx ≡ qy → [R] .fst qx qy}
+        (elimProp {B = λ qx → E.Thm .fst qx ≡ qy → [R] .fst qx qy}
```
mortberg, line 261: Can't you use elimProp2?

ecavallo: Not without generalizing it, elimProp2 takes (x y : A / R) but the two arguments here are from different set quotients.

PR #385, opened 2020-07-08
2ltt simpler

Saizan:
Updates for 2ltt


mortberg:
Wow, will there a 2LTT version of Agda soon so that the interval doesn't have to live in Set\omega?

Saizan:
Yes, it's ready to merge into agda/master, just need to confirm cubical/master will compile with it.

Issue #305, opened 2020-05-06
Merge Fin and FinData?


mortberg:
It's probably a good idea to merge https://github.com/agda/cubical/blob/master/Cubical/Data/Fin.agda and https://github.com/agda/cubical/blob/master/Cubical/Data/FinData.agda. I vote for keeping FinData once support for transp for inductive families have been merged into the master branch of Agda

ecavallo:
This is probably not a good idea.

PR #387, opened 2020-07-09
Add Properties to FinData

grahnen:
I added some of the properties defined in Data/Fin to Data/FinData.

I'm using some of these for my master's thesis, and might as well provide them in case they're useful. This might be a step in the direction of #305 if that ends up happening, as I'm basically just copying proof strategies instead of transporting them through an equivalence between the types.

Any comments on the code is greatly appreciated :)

### DIFF
--- a/Cubical/Data/FinData/Base.agda
+++ b/Cubical/Data/FinData/Base.agda
@@ -54,6 +54,6 @@ elim P fz fs {suc k} zero = fz
 elim P fz fs {suc k} (suc fj) = fs (elim P fz fs fj)
 
 
-caseFin : ∀{k} → (a0 aS : A) → Fin k → A
-caseFin a0 aS zero = a0
-caseFin a0 aS (suc x) = aS
+rec : ∀{k} → (a0 aS : A) → Fin k → A
+rec a0 aS zero = a0
+rec a0 aS (suc x) = aS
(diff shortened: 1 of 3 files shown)


Cubical/Data/FinData/Properties.agda
```diff
@@ -0,0 +1,48 @@
...
+
+open import Cubical.Data.FinData.Base as Fin
+import Cubical.Data.Nat as ℕ
+open import Cubical.Data.Empty as Empty
+
+open import Cubical.Relation.Nullary
+
+private
+ variable
+   a b ℓ : Level
```
mortberg, line 17: You only need one level here, please use \ell

Cubical/Data/FinData/Base.agda
```diff
@@ -33,6 +35,25 @@ zero == suc _  = false
...
+  → (∀{k} → P {suc k} zero)
+  → (∀{k} → {fn : Fin k} → P fn → P (suc fn))
+  → {k : ℕ} → (fn : Fin k) → P fn
+
+elim P fz fs {zero} = ⊥.rec ∘ ¬Fin0
+elim P fz fs {suc k} zero = fz
+elim P fz fs {suc k} (suc fj) = fs (elim P fz fs fj)
+
+
+caseFin : ∀{k} → (a0 aS : A) → Fin k → A
```
mortberg, line 57: Can you call this `rec` instead?

mortberg:
Great! Thanks for the contribution

PR #393, opened 2020-07-16
Add definition of finite sets

jorge-jbs:
For now it is very incomplete, since I need some results about `Fin`.

Is `FinSet` okay in `Cubical.Foundations`?

### DIFF
--- a/Cubical/Data/Fin/Base.agda
+++ b/Cubical/Data/Fin/Base.agda
@@ -72,10 +72,7 @@ elim P fz fs {suc k} fj
   ; (inr (fk , p)) → subst P p (fs (elim P fz fs fk))
   }
 
-any?
-  : ∀ {n} {P : Fin n → Type ℓ}
-  → (∀ i → Dec (P i))
-  → Dec (Σ (Fin n) P)
+any? : ∀ {n} {P : Fin n → Type ℓ} → (∀ i → Dec (P i)) → Dec (Σ (Fin n) P)
 any? {n = zero}  {P = _} P? = no (λ (x , _) → ¬Fin0 x)
 any? {n = suc n} {P = P} P? =
   mapDec
@@ -87,9 +84,7 @@ any? {n = suc n} {P = P} P? =
     (λ n h → n (helper h))
     (P? fzero ⊎? any? (P? ∘ fsuc))
   where
-    helper
-      : Σ (Fin (suc n)) P
-      → (P fzero ⊎ Σ (Fin n) (λ z → P (fsuc z)))
+    helper : Σ (Fin (suc n)) P → P fzero ⊎ Σ (Fin n) λ z → P (fsuc z)
     helper (x , Px) with fsplit x
     ... | inl x≡0 = inl (subst P (sym x≡0) Px)
     ... | inr (k , x≡sk) = inr (k , subst P (sym x≡sk) Px)
--- a/Cubical/Data/Fin/Base.agda
+++ b/Cubical/Data/Fin/Base.agda
@@ -10,7 +10,7 @@ import Cubical.Data.Empty as ⊥
 open import Cubical.Data.Nat using (ℕ; zero; suc)
 open import Cubical.Data.Nat.Order
 open import Cubical.Data.Sigma
-open import Cubical.Data.Sum using (_⊎_; inl; inr)
+open import Cubical.Data.Sum using (_⊎_; _⊎?_; inl; inr)
 
 open import Cubical.Relation.Nullary
 
@@ -52,6 +52,9 @@ fsplit
 fsplit (0 , k<sn) = inl (toℕ-injective refl)
 fsplit (suc k , k<sn) = inr ((k , pred-≤-pred k<sn) , toℕ-injective refl)
 
+inject< : ∀ {m n} (m<n : m < n) → Fin m → Fin n
+inject< m<n (k , k<m) = k , <-trans k<m m<n
+
 -- Fin 0 is empty
 ¬Fin0 : ¬ Fin 0
 ¬Fin0 (k , k<0) = ¬-<-zero k<0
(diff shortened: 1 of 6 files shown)


Cubical/Foundations/FinSet.agda
```diff
@@ -0,0 +1,43 @@
...
+open import Cubical.Foundations.Isomorphism
+open import Cubical.Foundations.Univalence
+open import Cubical.Foundations.Equiv
+open import Cubical.HITs.PropositionalTruncation
+open import Cubical.Data.Unit
+open import Cubical.Data.Nat
+open import Cubical.Data.Fin
+
+isFinSet : Type₀ → Type₁
+isFinSet A = ∥ Σ[ n ∈ ℕ ] A ≡ Fin n ∥
```
WorldSEnder, line 18: Could use `isFinSet A = Σ[ n ∈ ℕ ] ∥ A ≃ Fin n ∥` here, which would be stronger and more universe-polymorphic.

martinescardo: There are two suggestions here: (1) To replace equality by equivalence (this makes the definition more universe polymorphic) and (2) to move the truncation inside the Sigma (this would make the definition stronger). But actually, regarding (2), you can prove that the definitions with truncation inside and outside are equivalent, because if `Fin n` and `Fin n'` are equivalent, then `n` and `n'` are equal, and because the natural numbers form a set.

mortberg: I agree with Martín, please use `isFinSet A = ∥ Σ[ n ∈ ℕ ] A ≃ Fin n ∥`. There is a nice notation for this in https://github.com/agda/cubical/blob/master/Cubical/Data/Sigma/Base.agda#L29

This will make it trivial to prove that the property of being finite is a proposition. You should then prove that you can obtain the version that @WorldSEnder suggested

mortberg:
> Is `FinSet` okay in `Cubical.Foundations`?

Please put it in `Cubical.Data`.

jorge-jbs:
I think I solved the comments you mentioned. I added some theorems to `Fin` so comments on that module would also be appreciated.

Cubical/Data/Fin/Properties.agda
```diff
@@ -41,6 +46,22 @@ isContrFin1
...
 
+discreteFin : ∀ {n} → Discrete (Fin n)
+discreteFin {n} (x , hx) (y , hy) with discreteℕ x y
+... | yes prf = yes (Σ≡Prop (λ _ → m≤n-isProp) prf)
+... | no prf = no (λ h → prf (
+    x ≡⟨ refl ⟩
+    fst {B = _< n} (x , hx) ≡⟨ cong fst h ⟩
+    fst {B = _< n} (y , hy) ≡⟨ refl ⟩
+    y ∎
+  ))
```
mortberg, line 57: Am I missing something or isn't this proof just `cong fst h`?

Cubical/Data/Fin/Base.agda
```diff
@@ -69,3 +72,24 @@ elim P fz fs {suc k} fj
...
+      { (inl P0) → fzero , P0
+      ; (inr (x , Px)) → fsuc x , Px
+      }
+    )
+    (λ n h → n (helper h))
+    (P? fzero ⊎? any? (P? ∘ fsuc))
+  where
+    helper
+      : Σ (Fin (suc n)) P
+      → (P fzero ⊎ Σ (Fin n) (λ z → P (fsuc z)))
```
mortberg, line 92: You don't need parentheses around the conclusion. Also, why not put the type on one line? Same applies to the toplevel type.

Cubical/Data/FinSet.agda
```diff
@@ -0,0 +1,110 @@
...
+FinSet : Type (ℓ-suc ℓ)
+FinSet = TypeWithStr _ isFinSet
+
+isFinSetFin : ∀ {n} → isFinSet (Fin n)
+isFinSetFin = ∣ _ , pathToEquiv refl ∣
+
+isFinSetUnit : isFinSet Unit
+isFinSetUnit = ∣ 1 , Unit≃Fin1 ∣
+  where
+    Unit≃Fin1 : Unit ≃ Fin 1
```
mortberg, line 38: Maybe this equivalence should be upstreamed? I can imagine it being useful elsewhere

Cubical/Data/FinSet.agda
```diff
@@ -0,0 +1,110 @@
...
+      isoToEquiv
+        (iso
+          (const fzero)
+          (const tt)
+          (isContrFin1 .snd)
+          (isContrUnit .snd)
+        )
+
+isFinSet′ : Type ℓ → Type ℓ
+isFinSet′ A = Σ[ n ∈ ℕ ] ∥ A ≃ Fin n ∥
```
mortberg, line 49: How about renaming this to something more informative, like `isFinSetΣ`?

Cubical/Data/FinSet.agda
```diff
@@ -0,0 +1,110 @@
...
+    Fin-n≡Fin-m = rec propTruncIsProp (∣_∣ ∘ ua) Fin-n≃Fin-m
+
+    ∥n≡m∥ : ∥ n ≡ m ∥
+    ∥n≡m∥ = rec propTruncIsProp (∣_∣ ∘ Fin-inj n m) Fin-n≡Fin-m
+
+    n≡m : n ≡ m
+    n≡m = rec (isSetℕ n m) (λ p → p) ∥n≡m∥
+
+isFinSet≡isFinSet′ : isFinSet A ≡ isFinSet′ A
+isFinSet≡isFinSet′ {A = A} = ua (isoToEquiv (iso to from to-from from-to))
```
mortberg, line 78: You can use https://github.com/agda/cubical/blob/master/Cubical/Foundations/Univalence.agda#L42 here

mortberg:
PS: can you please also add some headers to the files explaining what they contain?

Cubical/Data/FinSet.agda
jorge-jbs: `FinSet` depends on `Fin` and `Fin` depends on `Unit`. So if I import `FinSet` from `Unit` I get a cycle.

jorge-jbs: Oh, I misunderstood. You only meant the equivalence. But I have tried to only move the equivalence and I still get a cycle, although a less obvious one: `Unit -> Fin -> Nat.Order -> ... -> Unit`.

mortberg: Right, I just meant the equivalence. Can't you put it in Fin?

jorge-jbs: That works

mortberg: Perfect! That seems like a reasonable place to put it

jorge-jbs:
I think everything's done. Thanks for the comments!

mortberg:
Great! Thank you for the contribution

PR #394, opened 2020-07-29
Groupoid quotients and the first Eilenberg–Mac Lane space (group delooping)

UlrikBuchholtz:
This PR introduces groupoid quotients, but most of the results are about the special case of group deloopings. (I looked at the category theory and it's not quite ready for Rezk/groupoid completions yet.)

However, on the way to proving that the loops in the first Eilenberg–Mac Lane space recovers the group, type-checking gets very slow. Any ideas how I can speed things up?

### DIFF
--- a/Cubical/Structures/Group/EilenbergMacLane1.agda
+++ b/Cubical/Structures/Group/EilenbergMacLane1.agda
@@ -1,0 +1,130 @@
+{-# OPTIONS --cubical --no-import-sorts --safe #-}
+
+module Cubical.Structures.Group.EilenbergMacLane1 where
+
+open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.Isomorphism
+open import Cubical.Foundations.Equiv
+open import Cubical.Foundations.Equiv.HalfAdjoint
+open import Cubical.Foundations.GroupoidLaws renaming (assoc to ∙assoc)
+open import Cubical.Foundations.Path
+open import Cubical.Foundations.HLevels
+open import Cubical.Foundations.Transport
+open import Cubical.Foundations.Univalence
+open import Cubical.Foundations.SIP
+open import Cubical.Data.Unit
+open import Cubical.Data.Sigma
+open import Cubical.Relation.Binary.Base
+open import Cubical.Structures.Axioms
+open import Cubical.Structures.Auto
+open import Cubical.Structures.Group.Base
+open import Cubical.Structures.Group.Properties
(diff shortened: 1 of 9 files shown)


Cubical/Structures/Group/EilenbergMacLane1.agda
```diff
@@ -0,0 +1,130 @@
...
+            h + g + - g ≡⟨ cong (h +_) (invr g) ⟩
+                 h + 0g ≡⟨ rid h ⟩ h ∎)
+
+  compRightEquiv : (g h : Carrier)
+    → compEquiv (rightEquiv g) (rightEquiv h) ≡ rightEquiv (g + h)
+  compRightEquiv g h = equivEq _ _ (funExt (λ x → (assoc x g h) ⁻¹))
+
+  Codes : EM₁ → hSet ℓ
+  Codes = rec EM₁Rt (isOfHLevelTypeOfHLevel 2)
+    (λ _ → Carrier , is-set) RE REComp
```
UlrikBuchholtz, line 101: If I leave out the ```REComp``` term and place a hole instead, the speed is acceptable.

UlrikBuchholtz:
The new commit adds just the first EM space as a HIT, and this takes care of the performance issue. There are most things to do, of course. Namely, to prove that this gives an equivalence with the (type/category) of groups.

mortberg:
Nice that the HIT solved the performance issue! 

A small remark: one can often write more direct to definitions of the recursors than what one get by defining them using the eliminators. This could also make things faster.

mortberg:
Should we keep this PR or close it now that you're working on https://github.com/agda/cubical/pull/435 ?

UlrikBuchholtz:
> Should we keep this PR or close it now that you're working on #435 ?

Hm, we decided to split #435 into three parts: (1) The EM stuff, (2) the generic DURG stuff, and (3) the XMod/S2G stuff.

I'll try to merge agda:master into this branch, and add the EM stuff from #435, so that this PR (#394) can comprise (1).

mortberg:
> > Should we keep this PR or close it now that you're working on #435 ?
> 
> Hm, we decided to split #435 into three parts: (1) The EM stuff, (2) the generic DURG stuff, and (3) the XMod/S2G stuff.
> 
> I'll try to merge agda:master into this branch, and add the EM stuff from #435, so that this PR (#394) can comprise (1).

Ok, cool! I wasn't sure how you wanted to do things, but this sounds reasonable to me

PR #395, opened 2020-07-31
Record structure automation

ecavallo:
Automatically generating proofs of `UnivalentStr` for a structure and notion of structured equivalence defined by a pair of records. See the bottom of `Cubical.Structures.Record` for an example. It's a bit slow on my machine but not too bad.

It's not ready for use with the library as is. Currently the record-based structured definitions in the library include the carrier type as a field (instead of a parameter), which doesn't integrate very well with the SIP machinery. I don't know what we want to do about that.

Also there's some duplication in the reflection 'utility' definitions used here and in `Cubical.Structures.Auto` that I should factor out somehow.


mortberg:
Very cool!

Do you think things would be simpler if we turned the carrier type into a parameter? I have nothing against that. Things are the way they are because I just copied things from the standard library, but if it's better to put carrier types as parameters then I support this change (in a separate PR of course).

ecavallo:
The results about induced monoids and so on make it hard to completely eliminate the Sigma-versions of the structure definitions... I'll not worry too much about it for now.

ecavallo:
I'm finding that it's not worth it to try to delete the Sigma-theory definitions yet, because more complicated structures that aren't adjusted yet depend on the semigroup Sigma-theory and so on. So I just replaced the ultimate Semigroup and Monoid results so far and didn't try to delete anything.

I was wrong about Group not being an issue, which I realized because the same problem shows up with Ring. The automation as currently set up doesn't allow prop fields to depend on other prop fields, which is a problem because `IsRing` depends on `0r` and `-_` (which are also props and don't have entries in `RingEquiv`. I'm not sure off-hand whether this is easy to change on the automation side, although one could always just avoid it by moving `0r` and `-_` from `RingStr` to `IsRing`.

I forgot to mention in #430 but structures that depend on parameters (like Algebra) shouldn't require anything new. All of these structures actually have parameters (the universe level), so nothing is special about them.

ecavallo:
I changed something but it doesn't fix the issue with Ring above. Actually I don't know how to arrange things so that it can work with the current definition. Putting `0r` and `-_` in `IsRing` is the only way I can think of that will work.

ecavallo:
To be clear, the problem is that the proof that `0r` is unique depends on the data in `isRing`, while the type of `isRing` depends on `0r`, so there are dependencies going both ways.

I've done what can be done with the definitions as they are, so maybe we should merge this and we can make more structural changes in a new PR?

mortberg:
> I'm finding that it's not worth it to try to delete the Sigma-theory definitions yet, because more complicated structures that aren't adjusted yet depend on the semigroup Sigma-theory and so on. So I just replaced the ultimate Semigroup and Monoid results so far and didn't try to delete anything.
> 
> I was wrong about Group not being an issue, which I realized because the same problem shows up with Ring. The automation as currently set up doesn't allow prop fields to depend on other prop fields, which is a problem because `IsRing` depends on `0r` and `-_` (which are also props and don't have entries in `RingEquiv`). I'm not sure off-hand whether this is easy to change on the automation side, although one could always just avoid it by moving `0r` and `-_` from `RingStr` to `IsRing`.
> 
> I forgot to mention in #430 but structures that depend on parameters (like Algebra) shouldn't require anything new. All of these structures actually have parameters (the universe level), so nothing is special about them.

Ok, I can do the change for algebras and modules if you want!

Cubical/Reflection/Base.agda
```diff
@@ -0,0 +1,45 @@
+{-
+
+Some basic tilities for reflection
```
mortberg, line 3: "u" missing

mortberg:
The only structure that is using the new reflection machinery is Semigroup. How about adding at least monoids as well? That shouldn't be problematic?

ecavallo:
I'm confused, I did add it for monoids.

mortberg:
> I'm confused, I did add it for monoids.

Very weird... My browser must have decided to show an old cached version of the diff or something

mortberg:
Anyway, merging now! This is going to be great when it works for all structures

PR #399, opened 2020-08-18
Localisation

mzeuner:
Started working on localisations, but this is only a first attempt at what things could like. 

One thing that should be discussed before going much further is whether we want the localisation to be defined directly as a HIT (with the zero-divisor in S that annihilates rs'-r's given as an argument to the paths between fractions r/s and r'/s', i.e. as a structure) or do we want to have it as a set-quotient, relating any two fractions r/s and r'/s' for which there merely exists a zero-divisor in S that annihilates rs'-r's?

In any case it should be possible to have a rather direct approach to localisation, although some things like checking well-definedness of addition and multiplication is rather tedious without a ring solver...


Cubical/Algebra/Localisation.agda
```diff
@@ -0,0 +1,214 @@
...
+ open isSubMonoid
+ open CommRing R' renaming (Carrier to R)
+ open Theory (CommRing→Ring R')
+
+
+ data S⁻¹R : Type ℓ where
+  _/_ : (r s : R) → {s ∈ S} → S⁻¹R
+  zd : (r₁ r₂ s s₁ s₂ : R) → {s ∈ S} → {a : s₁ ∈ S} → {b : s₂ ∈ S}
+     → s · ((r₁ · s₂) - (r₂ · s₁)) ≡ 0r
+     → ((r₁ / s₁) {a}) ≡ ((r₂ / s₂) {b})
```
mortberg, line 52: It's annoying to give `{a}` and `{b}` like this. Could one use instance arguments or something to have Agda infer them automatically? Maybe @Saizan or @guillaumebrunerie knows?

@mzeuner : here's the relevant documentation: https://agda.readthedocs.io/en/v2.6.0.1/language/instance-arguments.html

mortberg:
Here's Voevodsky's definition of localizations in UniMath: https://github.com/UniMath/UniMath/blob/master/UniMath/Algebra/RigsAndRings.v#L2486

The code leading up to the definition starts here: https://github.com/UniMath/UniMath/blob/master/UniMath/Algebra/RigsAndRings.v#L2140

Note that the proofs of the algebraic properties of the localization operations involve some hairy algebra. 

If one stares at the definitions long enough one realizes that he's using https://github.com/UniMath/UniMath/blob/master/UniMath/Algebra/Monoids.v#L903 with the multiplicative commutative monoid of R. He's hence quotienting by the relation 
```
(x,a) ~ (x,b) := exists (c : A). (x * b) * c = (y * a) * c
```

Some remarks about this:
- He's using mere existence instead of Sigma. What's the relationship to your definition?
- The equation in the relation is stated in a way that should make some of the goals you get later a lot more manageable (no need to fiddle with distributitivity, minus, etc...)
- He has packaged `s` with `s ∈ S` together and defined submonoid (which I guess would be SubCommRing for you). This might make some of the arguments easier for Agda to infer?

mortberg:
> ```
> (x,a) ~ (x,b) := exists (c : A). (x * b) * c = (y * a) * c
> ```

PS: `a` and `b` are really elements of a Sigma type `(x : A) * (x ∈ S)` but I've left the proof part out to make the definition more readable.

Cubical/Algebra/Localisation.agda
mzeuner: It works quite well with instance arguments, the only time they were explicitly required was in the definition of addition:
[https://github.com/agda/cubical/pull/399/files#diff-bea1144102447f41fb1100cb7dd8f98fR157](url)

Thanks!

mzeuner:
>     * He's using mere existence instead of Sigma. What's the relationship to your definition?
I should probably do both and check equivalence, just to be sure...

>     * He has packaged `s` with `s ∈ S` together and defined submonoid (which I guess would be SubCommRing for you). This might make some of the arguments easier for Agda to infer?
Doesn't that complicate things with all the instance arguments?

mortberg:
> > ```
> > * He's using mere existence instead of Sigma. What's the relationship to your definition?
> > ```
> 
> I should probably do both and check equivalence, just to be sure...
> 

Hmm, if we have a ring with zero divisors can't there be multiple non-equal zero divisors that annihilate the same cross multiplication?

> > ```
> > * He has packaged `s` with `s ∈ S` together and defined submonoid (which I guess would be SubCommRing for you). This might make some of the arguments easier for Agda to infer?
> > ```
> 
> Doesn't that complicate things with all the instance arguments?

Yes, maybe you won't need instance arguments at all if you do that.

mzeuner:
> Hmm, if we have a ring with zero divisors can't there be multiple non-equal zero divisors that annihilate the same cross multiplication?
Yes, but the different instances of zd you get from different zero divisors can be identified by trunc, so the two approaches should be equivalent (I hope)...

> Yes, maybe you won't need instance arguments at all if you do that. It would be interesting to experiment a bit
I'll make a new file that uses Voevodsky's approach and see how it goes...

mortberg:
> > Hmm, if we have a ring with zero divisors can't there be multiple non-equal zero divisors that annihilate the same cross multiplication?
> 
> Yes, but the different instances of zd you get from different zero divisors can be identified by trunc, so the two approaches should be equivalent (I hope)...
> 

Oh right, intriguing! 

> > Yes, maybe you won't need instance arguments at all if you do that. It would be interesting to experiment a bit
> 
> I'll make a new file that uses Voevodsky's approach and see how it goes...

PR #400, opened 2020-08-22
Integers as a CommRing

felixwellen:
This PR is about showing that Int is a CommRing. I chose DiffInt, since I believe that properties will be easier to show here.


felixwellen:
I probably won't continue this one any time soon, since #414 solves what I wanted in a better way.

mortberg:
I'm closing this one now as it has been subsumed by another PR

Issue #351, opened 2020-06-10
Reorganize Structures files


mortberg:
Currently the Structures folder contain various things that I think should be in the Algebra folder.  For instance the definitions of algebraic structures, ideals, etc. should all be in Algebra.

I don't think the `Queue` file belongs there either. I'm not sure where it should go, but as far as I can tell the FMSet structure stuff is not in this folder either. Maybe we should have a folder in Data called Queues with all implementations of queues and similarly for FMSets. This would make sense as these are about datastructures.

What would be left if we do this change are files related to the SIP which should then maybe be moved to a new folder `Foundations.SIP`/`Foundations.Structures`. 

I think this would make the organization a bit more logical, but let's wait with doing this change until https://github.com/agda/cubical/issues/349 has been fixed so that we can avoid conflicts.

ecavallo:
Some other changes that could be made in Structures:
- `add-axioms-SNS` and `join-SNS`, currently in `Foundations.SIP`, could have their own files.
- Perhaps `join-*` should be `product-*`, since `join` already means something else?
- We should pick a convention for formatting the names. Right now we have `pointed-is-SNS`, `join-SNS`, `nAryFunSNS`, ...

mortberg:
I think this might be a good time to fix this issue judging by the open PRs, however I don't have time to do it myself right now. What I think should be done is:

- Move algebraic structures to Algebra package and split into Base/Properties files like what has already been done for groups: https://github.com/agda/cubical/tree/master/Cubical/Structures/Group

- I think specific theories built on top of algebraic structures can stay in the top level part of the Algebra folder (so matrices, polynomials, ideals, etc. should all be on top level). So when moving Ideal and QuotientRing they should end up on the top level.

Sounds good? Anyone up for doing this?

felixwellen:
Yes! Here!
Though I would prefer to do it after #331 is merged.

mortberg:
The Algebra files were moved in https://github.com/agda/cubical/pull/401. Except for maybe Poset?

The remaining files in the Structures folder are related to the SIP and datastructures. I think these should be reorganized somehow as well. Maybe @mzeuner and @ecavallo have some suggestions?

felixwellen:
Ah, ok, Posets don't feel like algebra for me :-)
But of course, I am not against moving it.

mortberg:
> Ah, ok, Posets don't feel like algebra for me :-)
> But of course, I am not against moving it.

[...]

PR #401, opened 2020-08-25
Reorganize structures and algebra

felixwellen:
WIP on #351


felixwellen:
moved all the algebra stuff to 'Algebra' - I just renamed things to make everything check
the only exception is, that I split Ring.agda into Ring/Base.agda and Ring/Properties.agda

mortberg:
Perfect! Merging

PR #409, opened 2020-08-28
Structures for the ring solver

felixwellen:
I added an explanation as a README.md
I started to add the new structures SemiRing and AlmostRing. As I understand it, our example ring solver operates on AlmostRings and it is good to know that the natural numbers are an example.


Cubical/Algebra/RingSolver/AlmostRing.agda
```diff
@@ -0,0 +1,73 @@
...
+                  (0r 1r : R) (_+_ _·_ : R → R → R) (-_ : R → R) : Type ℓ where
+
+  constructor isalmostring
+
+  field
+    +IsMonoid : IsMonoid 0r _+_
+    ·IsMonoid : IsMonoid 1r _·_
+    +Comm : (x y : R) → x + y ≡ y + x
+    ·Comm : (x y : R) → x · y ≡ y · x
+    RightDist : (x y z : R) → x · (y + z) ≡ (x · y) + (x · z)
```
mortberg, line 26: Should this be `·DistR+`?

felixwellen: Yes, sure.

felixwellen: I don't know though, if we agreed what left and right are. I think we didn't so I just use what I usually do.

PR #410, opened 2020-08-28
Almost ring expressions (RingSolver)

felixwellen:


felixwellen:
Subsumed in #417 by now.

Issue #403, opened 2020-08-27
Naming convention for algebra related code


mortberg:
It would be nice if all the algebra would be using some more consistent naming schema. Here's a suggestion that we are more or less following:

- First letter of structure names and types should be uppercase (like `Ring`)
- Use appropriate abbreviation: 
  - Comm = commutative
  - Assoc = associative
  - DistR/DistL = distribute right/left
- Use camelCase as much as possible, also for properties/lemmas related to operations. For example
  - `+Assoc`
  - `·DistR+`

I'm not sure about how to name lemmas like these: https://github.com/agda/cubical/blob/master/Cubical/Structures/Ring.agda#L272

This should be implemented after https://github.com/agda/cubical/pull/401

mchristianl:
Would it be "safe" to rename every occurrence of `_*_` to `_·_` in the `Cubical/Algebra` subdirectory?

mortberg:
> Would it be "safe" to rename every occurrence of `_*_` to `_·_` in the `Cubical/Algebra` subdirectory?

Sure! I think this should be the name for multiplication of numbers as well

mortberg:
(just to be clear, it's \cdot not \.)

mchristianl:
The `*` is currently used in

- **Algebra/CommRing**
- **Data/BinNat** (just a comment)
- **Data/Fin**
- **Data/InfNat**
- **Data/Nat**
- **Data/NatPlusOne**
- Data/Sigma
- Data/Unit
- (Data/Vec)
- HITs/AssocList
- [HITs/Colimit ("lower star functor" should remain `*`)](https://github.com/agda/cubical/blob/367dd3d4cf185693f02aac3deef93888dedc7eeb/Cubical/HITs/Colimit/Base.agda#L38)
- HITs/FiniteMultiset
- (HITs/Hopf)
- **HITs/Ints/BiInvInt**
- **HITs/Ints/QuoInt**
- HITs/ListedFiniteSet
- (HITs/Rationals/HITQ)
- **HITs/Rationals/QuoQ**
- **HITs/Rationals/SigmaQ**
- HITs/RPn
- [HITs/S1 (debatable)](https://github.com/agda/cubical/blob/367dd3d4cf185693f02aac3deef93888dedc7eeb/Cubical/HITs/S1/Base.agda#L396)
- HITs/Truncation
- HITs/Truncation/FromNegTwo
- Homotopy/Connected
- Modalities/Modality
- (Papers/Synthetic)
[...]

PR #414, opened 2020-09-09
Define commutative ring structure of BiInvInt

knrafto:
After doing this, I think BiInvInt (or maybe HAEquivInt) is so far the most natural definition of the integers in cubical type theory. The definitions and proof closely mirror those for ℕ (with an extra case for `pred`) and they satisfy similar definitional equalities (for example `zero + n ≡ n`, `suc m + n ≡ suc (m + n)`, and `pred m + n ≡ pred (m + n)` all hold definitionally). Very cool! The downside is that we have to use recursion/induction principles manually instead of pattern-matching to avoid repeating ourselves too much.

This is my first time contributing to this library, so please let me know how I can match the style better! For example I'm not sure if renaming `BiInvInt` to `ℤ` in this module was a good idea. I also removed the existing `ᴮ` suffix, since I think someone should instead import this module qualified if they need to disambiguate.

### DIFF
--- a/Cubical/HITs/Ints/BiInvInt/Properties.agda
+++ b/Cubical/HITs/Ints/BiInvInt/Properties.agda
@@ -63,8 +63,14 @@ BiInvInt-rec {A = A} z e = φ
   φ (predl n) = Iso.inv e-Iso (φ n)
   φ (predl-suc n i) = Iso.leftInv e-Iso (φ n) i
 
+sucIso : Iso BiInvInt BiInvInt
+Iso.fun sucIso = suc
+Iso.inv sucIso = pred
+Iso.rightInv sucIso = suc-predl
+Iso.leftInv sucIso = predl-suc
+
 sucEquiv : BiInvInt ≃ BiInvInt
-sucEquiv = isoToEquiv (iso suc pred suc-predl predl-suc)
+sucEquiv = isoToEquiv sucIso
 
 -- addition
 
@@ -157,14 +163,14 @@ abs n = Int.abs (bwd n)
 sgn : BiInvInt → Bool
 sgn n = Int.sgn (bwd n)
 
-isEquiv-n+ : ∀ n → isEquiv (n +_)
-isEquiv-n+ n = isoToIsEquiv (iso (n +_) ((- n) +_) sec ret)
-  where
-  sec : ∀ m → n + ((- n) + m) ≡ m
-  sec m = +-assoc n (- n) m ∙ cong (_+ m) (+-invʳ n)
+Iso-n+ : (n : BiInvInt) → Iso BiInvInt BiInvInt
+Iso.fun (Iso-n+ n) = n +_
+Iso.inv (Iso-n+ n) = - n +_
+Iso.rightInv (Iso-n+ n) m = +-assoc n (- n) m ∙ cong (_+ m) (+-invʳ n)
+Iso.leftInv (Iso-n+ n) m = +-assoc (- n) n m ∙ cong (_+ m) (+-invˡ n)
 
-  ret : ∀ m → (- n) + (n + m) ≡ m
-  ret m = +-assoc (- n) n m ∙ cong (_+ m) (+-invˡ n)
+isEquiv-n+ : ∀ n → isEquiv (n +_)
+isEquiv-n+ n = isoToIsEquiv (Iso-n+ n)
 
 -- multiplication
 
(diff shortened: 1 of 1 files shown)


felixwellen:
Nice!
I'm not sure about the renaming and the suffix, but there was a recent agreement on algebra-related names:
https://github.com/agda/cubical/issues/403
I use it when writing new things.

knrafto:
Hmm I feel like in the meantime the property names should be consistent with the names in Cubical.Data.Nat.Properties, since I use those a lot and I would expect working with ℤ to be similar to ℕ. So for things like `+-assoc -> +Assoc` the BiInvInt properties should be renamed at the same time as the Nat properties

Cubical/HITs/Ints/BiInvInt/Properties.agda
```diff
@@ -3,146 +3,240 @@ module Cubical.HITs.Ints.BiInvInt.Properties where
...
-m +ᴮ predr n       = predr (m +ᴮ n)
-m +ᴮ predl n       = predl (m +ᴮ n)
-m +ᴮ suc-predr n i = suc-predr (m +ᴮ n) i
-m +ᴮ predl-suc n i = predl-suc (m +ᴮ n) i
+-- the following equalities hold definitionally:
+--   zero   + n ≡ n
+--   suc m  + n ≡ suc (m + n)
+--   pred m + n ≡ pred (m + n)
+_+_ : BiInvInt → BiInvInt → BiInvInt
+_+_ = BiInvInt-rec (idfun BiInvInt) (postCompEquiv sucEquiv)
```
Saizan, line 77: Has it not been a problem that `m + n` would often get expanded to the `BiInvInt-rec ...` call?

knrafto: I'm not sure what you mean. How would this problem show up?

Saizan: If `m + n` shows up in a goal type, Agda might expand it, it will if you ask for the normal form but in other situations too.

Also, it could have shown up as slower typechecking times.

knrafto: Oh I see. It doesn't get expanded in (non-normalized) goal types, and the file typechecks in 3 seconds. But, when I try `C-u C-u C-c C-t` in the following goal, it takes over 30 seconds and spits out the following 66598-line term: https://gist.github.com/knrafto/ed60e415db86989c0200f9e6397ea125
(I'm not sure how much of that time is just printing though)

```
*-comm : ∀ m n → m * n ≡ n * m
*-comm = BiInvInt-ind-prop (λ _ → isPropΠ λ _ → isSetBiInvInt _ _)
  (λ n → sym (*-zero n))
  (λ m p n → {!!})  -- cong (n +_) (p n) ∙ sym (*-suc n m)
  (λ m p n → cong (- n +_) (p n) ∙ sym (*-pred n m))
```

I guess I never tried to normalize anything because the default goal types were pretty readable already

Saizan: Right, that's also made worse by the use of the `iso` constructor rather than doing a separate definition by copatterns.

Saizan: Can you then make separate copattern definitions for the isos needed in `sucEquiv` and `isEquiv-n+` ?
We are trying to stick to that as a general policy to keep normal forms smaller.

I forgot to thank you for the PR btw! It's a great use of abstraction, and hopefully any problems can be improved on by the tooling.

knrafto: Thank you! And thanks for your work on cubical agda, it's been fun to use.

I created separate definitions for the isos. I haven't really used copatterns before, but I was misled by [the docs](https://agda.readthedocs.io/en/v2.6.1/language/copatterns.html) into thinking copatterns are syntactic sugar for using the constructor. Why does using copatterns result in faster typechecking? Is that a bug?

Saizan: it's not a bug, the reason is that `sucIso` defined by copatterns is already in weak head normal form, so it only reduces when projected from.

Records having eta rules by default still means that the equality is unchanged, but Agda tries to only apply those when useful. 

One can go further with `no-eta-equality` but that's not been necessary yet.

knrafto: Sorry I mean, is it a bug that performance is so much worse for the constructor form? Maybe Agda should be less aggressive when expanding the definitions under the `iso`, for things like definitional equality and printing the normalized goal type.

Saizan: Oh, yeah, Agda should be more careful about expanding definitions in general, and/or there should be good expressive ways to control expansion. it would be a substantial project to improve the situation though.

Issue #392, opened 2020-07-16
Finite types


jorge-jbs:
`LFSet` is defined in the library, but it defines finite **sub**sets of a type. How would you state that a type is finite? I looked at Lean's mathlib and they define it like this:

```agda
record Finite (A : Type₀) : Type₀ where
  field
    elems : LFSet A
    complete : ∀ a → [ a ∈ elems ]
```

Should I make a PR adding this definition? I have already proved that `Fin n` is finite with this definition.

Until now I was using `Finite A = Σ ℕ (λ n → A ≡ Fin n)` and I don't know which one is the right one (they aren't equivalent).

UlrikBuchholtz:
A finite set should be one merely equivalent to Fin n for some n. We should have the pigeonhole principle that shows that it doesn't matter whether the n is inside the truncation or not. This has all been done in the [old HoTT library](https://github.com/HoTT/HoTT-Agda/blob/master/theorems/homotopy/FinSet.agda) and should be easy to port. I think a PR to that effect will be very welcome.

guillaumebrunerie:
Note that Lean does not used homotopy type theory and is moreover based on classical logic, so it is not necessarily a good idea to port definitions from Lean to cubical Agda. I agree with Ulrik’s comment on the definition of finite sets.

mortberg:
Yeah, I think the nicest way to express `Finite` in this library is what @UlrikBuchholtz says. It would also be interesting to add some variations on this, see: https://ncatlab.org/nlab/show/finite+set#Constructivist

If you want to explore subfinite sets then there are some various options for which notion of injection you could use: https://github.com/agda/cubical/blob/master/Cubical/Functions/Embedding.agda  I think all of these are equivalent for sets, but one might be easier to work with than the others.

For finitely indexed sets you can use: https://github.com/agda/cubical/blob/master/Cubical/Functions/Surjection.agda

I think @mzeuner has thought a bit about this for finite multisets and maybe has some thoughts about which versions are easiest to work with.

oisdk:
I'm actually doing my master's thesis on finite sets in cubical Agda, I have quite a few definitions of "finite" worked out here: https://github.com/oisdk/combinatorics-paper.

I probably won't have any time to port much of it into the cubical library over the next few weeks, but I use the cubical library in my code so many of the definitions should be usable as-is.

jorge-jbs:
I opened a PR with the old HoTT library definition #393 

[...]

PR #416, opened 2020-09-13
Add definition of subfinite sets and finitely indexed types

jorge-jbs:
As discussed in #392 and following [nLab](https://ncatlab.org/nlab/show/finite+set#Constructivist), I have defined subfinite sets and finitely indexed types. The latter are usually called finitely indexed sets, but they don't have to be sets in HoTT (for example, the circle is finitely indexed). They are usually called Kuratowski finite, but maybe we can leave that name to the more usual (equivalent) definition with `LFSet`.

### DIFF
--- a/Cubical/Data/FinInd.agda
+++ b/Cubical/Data/FinInd.agda
@@ -1,0 +1,35 @@
+{-# OPTIONS --cubical --no-import-sorts --safe #-}
+
+module Cubical.Data.FinInd where
+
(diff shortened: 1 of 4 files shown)


mortberg (approved):
Nice contribution with clear and well-documented code!

mortberg:
PS: Sorry for the slow review! I've been quite swamped all of September...

PR #417, opened 2020-09-18
Ring solver without reflection

felixwellen:
This PR builds on the other PRs concerning the RingSolver.
Right now, it is possible to solve equations with one variable.  The goal for this PR is to generalize to n variables (unless some objects that this should be done in a new PR).
There is an "Examples" file in "Algebra/RingSolver" with a couple of examples that try to explain how the RingSolver works.


mortberg:
This is great, thanks a lot! 

@mzeuner Check out the examples. I'm very optimistic that most of the long proof about localization can be shortened :-)

PR #420, opened 2020-09-29
Some assorted lemmas

dolio:
These changes add a few simple lemmas.

- Level lifting for `isOfHLevelDep`
  - Also includes lifting from level 1 to a `SquareP`
- A proof of the h-level of Σ types using `isOfHLevelDep` for the second parameter
- Added a module about involutions with some proofs about the paths they induce

### DIFF
--- a/Cubical/Foundations/HLevels.agda
+++ b/Cubical/Foundations/HLevels.agda
@@ -472,7 +472,8 @@ isOfHLevelΣ' 0 Actr Bctr .snd (x , y) i
 isOfHLevelΣ' 1 Alvl Blvl (w , y) (x , z) i .fst = Alvl w x i
 isOfHLevelΣ' 1 Alvl Blvl (w , y) (x , z) i .snd = Blvl y z (Alvl w x) i
 isOfHLevelΣ' {A = A} {B = B} (suc (suc n)) Alvl Blvl (w , y) (x , z)
-  = transport (λ i → isOfHLevel (suc n) (ΣP≡PΣ i))
+  = isOfHLevelRetract (suc n)
+      (λ p → (λ i → p i .fst) , λ i → p i .snd)
+      ΣPathP
+      (λ x → refl)
       (isOfHLevelΣ' (suc n) (Alvl w x) (Blvl y z))
-  where
-  ΣP≡PΣ = ΣPath≡PathΣ {A = λ _ → A} {B = λ _ → B} {x = (w , y)} {y = (x , z)}
--- a/Cubical/Foundations/HLevels.agda
+++ b/Cubical/Foundations/HLevels.agda
@@ -35,7 +35,7 @@ private
     C : (x : A) → B x → Type ℓ
     D : (x : A) (y : B x) → C x y → Type ℓ
     E : (x : A) (y : B x) → (z : C x y) → D x y z → Type ℓ
-    x y : A
+    w x y z : A
     n : HLevel
 
 isOfHLevel : HLevel → Type ℓ → Type ℓ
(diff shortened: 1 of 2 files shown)


Cubical/Foundations/HLevels.agda
```diff
@@ -423,3 +423,56 @@ isOfHLevel→isOfHLevelDep (suc (suc n)) {A = A} {B} h {a0} {a1} b0 b1 =
...
+        (j = i1) → Bprp tw (ts i) (λ k → sq (k ∧ i) (j ∧ k)) k)
+      tw
+
+isOfHLevelΣ' : ∀ n → isOfHLevel n A → isOfHLevelDep n B → isOfHLevel n (Σ A B)
+isOfHLevelΣ' 0 Actr Bctr .fst = (Actr .fst , Bctr .fst)
+isOfHLevelΣ' 0 Actr Bctr .snd (x , y) i
+  = Actr .snd x i , Bctr .snd y (Actr .snd x) i
+isOfHLevelΣ' 1 Alvl Blvl (w , y) (x , z) i .fst = Alvl w x i
+isOfHLevelΣ' 1 Alvl Blvl (w , y) (x , z) i .snd = Blvl y z (Alvl w x) i
+isOfHLevelΣ' {A = A} {B = B} (suc (suc n)) Alvl Blvl (w , y) (x , z)
```
mortberg, line 474: Can you prove this using `isOfHLevelRetract` instead? See the following for inspiration:

https://github.com/agda/cubical/pull/421/files#diff-4b84591f239c56566629cde01652543eR306

When working on that PR @aljungstrom noticed that using `isOfHLevelRetract` seem to produce faster code further down the line

Cubical/Functions/Involution.agda
```diff
@@ -0,0 +1,43 @@
...
+  open Iso
+
+  involIso : Iso A A
+  involIso .fun = f
+  involIso .inv = f
+  involIso .rightInv = invol
+  involIso .leftInv = invol
+
+  involIsEquiv : isEquiv f
+  involIsEquiv = isoToIsEquiv involIso
```
mortberg, line 23: Hmm, isn't there a more direct proof of this which should be simpler than what you get if you unfold `isoToIsEquiv`? I would guess that you can pretty much directly write down the `isEquiv` proof

Cubical/Functions/Involution.agda
```diff
@@ -0,0 +1,43 @@
...
+
+  involEquiv : A ≃ A
+  involEquiv = f , involIsEquiv
+
+  involPath : A ≡ A
+  involPath = ua involEquiv
+
+  involEquivComp : compEquiv involEquiv involEquiv ≡ idEquiv A
+  involEquivComp
+    = equivEq (compEquiv involEquiv involEquiv) (idEquiv A) (λ i x → invol x i)
```
mortberg, line 33: Could the first two arguments of `equivEq` be implicit? I'm kind of puzzled why they're not, but maybe Agda needs them not to be?

Cubical/Functions/Involution.agda
```diff
@@ -0,0 +1,43 @@
...
+
+  involPathComp : involPath ∙ involPath ≡ refl
+  involPathComp
+    = sym (uaCompEquiv involEquiv involEquiv) ∙ cong ua involEquivComp ∙ uaIdEquiv
```
mortberg, line 37: Please use https://github.com/agda/cubical/blob/master/Cubical/Foundations/Prelude.agda#L88

mortberg (changes requested):
Nice, looks like some useful stuff! I had a few comments that I would like to see addressed before merging

Cubical/Functions/Involution.agda
dolio: Apparently they can. I'm not sure why I thought they couldn't.

Cubical/Functions/Involution.agda
dolio: I can try, but personally I find `isEquiv` to be very difficult to prove directly in almost every situation.

Cubical/Functions/Involution.agda
mortberg: Maybe `equivEq` should be redefined so that they're implicit?

dolio: Possibly. I'm not sure if it's used often in situations where the result type is under-determined.

Cubical/Functions/Involution.agda
mortberg: I agree. Filling the square takes some thinking and the feedback one gets from Agda isn't too helpful.

Cubical/Functions/Involution.agda
mortberg: `grep` tells me that it's almost exclusively used with `_` so let's change it, but in a dedicated PR. I'll open an issue about it.

Cubical/Functions/Involution.agda
mortberg: I tried to do it myself, but can't find a simple way... So let's leave it as it is

mortberg (approved):

mortberg:
@dolio Was there anything else you wanted to do or should I merge?

dolio:
Yeah, no additional stuff from me. I couldn't figure out how to prove the `isEquiv` simply, either.

mortberg:
Great! I realized you hadn't answered yet just when I pressed merge... Anyway, thanks for the contribution!

PR #421, opened 2020-09-29
Some improvements to ZCohomology

aljungstrom:
This PR contains quite a lot of stuff, but I think all changes in `Homotopy` and `ZCohomology` can be ignored. The same goes for `FromNegOne`.

All changes are made in order to get the cohomology groups to compute better. The main changes are:

- Redefining n-spheres as suspensions of S1. This removes a lot of annoying translations between the two definitions.
- Using the `FromNegOne` truncation in the definitions of connectedness and cohomology in order to avoid the additional constructors in the usual definition of truncations.
- Avoiding, to an extent as great as possible, transports into univalence in `HLevels`.
- Removing eta from `Iso`

These changes also let me remove most of the abstract in the concerned modules.

After these changes, both the 0th and the 1st cohomology groups compute decently. This confirms my suspicion that Freudenthal is the main culprit (only used for n>1) here. If anyone has a more direct proof of the fact that 
Ω ∥ Sⁿ⁺¹ ∥ₙ₊₁ ≃ ∥ Sⁿ ∥ₙ (especially for n = 2), please tell me:-)


Cubical/Algebra/Group/Algebra.agda
```diff
@@ -202,23 +211,52 @@ vSES→GroupEquiv : {A : Group {ℓ}} {B : Group {ℓ₁}} (leftGr : Group {ℓ
...
-GroupEquiv.isHom (contrGroup≅trivialGroup contr) _ _ = refl
+contrGroup≅trivialGroup contr = GrIsoToGrEquiv (IsoContrGroupTrivialGroup contr)
+
+GroupIso→Iso : {A : Group {ℓ}} {B : Group {ℓ₁}} → GroupIso A B → Iso ⟨ A ⟩ ⟨ B ⟩
+fun (GroupIso→Iso i) = fun (map i)
+inv (GroupIso→Iso i) = inv i
+rightInv (GroupIso→Iso i) = rightInv i
+leftInv (GroupIso→Iso i) = leftInv i
+
+congLemma : ∀ {ℓ} {A : Type ℓ} (_+A_ : A → A → A) (-A_ : A → A)
```
mortberg, line 250: This name is really bad. Please rename or make private where it's used

Cubical/Data/Sigma/Properties.agda
```diff
@@ -78,6 +78,9 @@ module _ {A : I → Type ℓ} {B : (i : I) → A i → Type ℓ'}
               ≡ (PathP (λ i → Σ (A i) (B i)) x y)
   ΣPath≡PathΣ = ua ΣPath≃PathΣ
 
+×≡Prop : isProp A' → {u v : A × A'} → (u .fst ≡ v .fst) → u ≡ v
```
mortberg, line 81: No parentheses needed around first equality argument

Cubical/Data/Sigma/Properties.agda
```diff
@@ -203,17 +206,25 @@ leftInv (Σ-cong-iso-snd isom) (x , y') = ΣPathP (refl , leftInv (isom x) y')
 ΣPathTransport : (a b : Σ A B) → Type _
 ΣPathTransport {B = B} a b = Σ[ p ∈ (fst a ≡ fst b) ] transport (λ i → B (p i)) (snd a) ≡ snd b
 
+IsoΣPathTransportPathΣ : (a b : Σ A B) → Iso (ΣPathTransport a b) (a ≡ b)
+IsoΣPathTransportPathΣ {B = B} a b = compIso (Σ-cong-iso-snd (λ p → invIso (equivToIso (PathP≃Path (λ i → B (p i)) _ _))))
+         ΣPathIsoPathΣ
+
+-- fun (IsoΣPathTransportPathΣ {B = B} a b) (p , q) i = (p i) , (toPathP {A = λ i → B (p i)} q i)
```
mortberg, line 213: Commented code, please delete

Cubical/Data/Unit/Base.agda
```diff
@@ -1,6 +1,11 @@
 {-# OPTIONS --cubical --no-import-sorts --safe #-}
 module Cubical.Data.Unit.Base where
+open import Cubical.Foundations.Prelude
```
mortberg, line 3: Do you need the whole Prelude or just Core?

Cubical/Data/Unit/Base.agda
```diff
@@ -1,6 +1,11 @@
...
 
 -- Obtain Unit
 open import Agda.Builtin.Unit public
   renaming ( ⊤ to Unit )
+
+-- Lifted version
+data Unit* {ℓ : Level} : Type ℓ where
```
mortberg, line 10: Maybe it would be better if this would be the default Unit? Also the commented is strange, why not "universe polymorphic version"?

Cubical/Data/Unit/Properties.agda
```diff
@@ -53,3 +53,16 @@ isContr→≃Unit contr = isoToEquiv (iso (λ _ → tt) (λ _ → fst contr) (λ
 
 isContr→≡Unit : {A : Type₀} → isContr A → A ≡ Unit
 isContr→≡Unit contr = ua (isContr→≃Unit contr)
+
+
+---
```
mortberg, line 58: ---?

Cubical/Data/Unit/Properties.agda
```diff
@@ -53,3 +53,16 @@ isContr→≃Unit contr = isoToEquiv (iso (λ _ → tt) (λ _ → fst contr) (λ
...
+isContrUnit* : ∀ {ℓ} → isContr (Unit* {ℓ})
```
mortberg, line 59: Do you really need to parametrize over \ell here?

mortberg: Same for the lemmas below

Cubical/Foundations/Equiv/HalfAdjoint.agda
```diff
@@ -144,3 +149,15 @@ congIso {x = x} {y} e = goal
                    ; (j = i0) → Iso.leftInv e x (i ∨ k)
                    ; (j = i1) → Iso.leftInv e y (i ∨ k) })
           (Iso.leftInv e (p j) i)
+
+invCongFunct : {x : A} (e : Iso A B) (p : Iso.fun e x ≡ Iso.fun e x) (q : Iso.fun e x ≡ Iso.fun e x)
```
mortberg, line 153: Yikes, this had a scary proof! Can't one do it by `elimIso` (https://github.com/agda/cubical/blob/master/Cubical/Foundations/Univalence.agda#L287) instead?

I guess one potential problem with that lemma is that it's using univalence... But maybe that's still simpler this proof? (which might be using univalence somewhere anyway...)

Cubical/Foundations/HLevels.agda
```diff
@@ -179,30 +183,87 @@ isPropRetract f g h p x y i =
       ; (i = i1) → h y j})
     (g (p (f x) (f y) i))
 
+isSetRetract
```
mortberg, line 186: This type is really strangely indendented. Same for the other similar lemmas

Cubical/Foundations/HLevels.agda
```diff
@@ -179,30 +183,87 @@ isPropRetract f g h p x y i =
...
+is2GroupoidRetract f g h grp x y p q P Q R S i j k l =
+  hcomp (λ r → λ { (i = i0) → h (R j k l) r
+                  ; (i = i1) → h (S j k l) r
+                  ; (j = i0) → h (P k l) r
+                  ; (j = i1) → h (Q k l) r
+                  ; (k = i0) → h (p l) r
+                  ; (k = i1) → h (q l) r
+                  ; (l = i0) → h x r
+                  ; (l = i1) → h y r})
+        (g (grp (f x) (f y) (cong f p) (cong f q)
```
mortberg, line 226: Hmm, wouldn't this be easier to read if you inline the uses of cong?

Cubical/Foundations/HLevels.agda
```diff
@@ -220,15 +281,19 @@ isContrΣ {A = A} {B = B} (a , p) q =
 isContrΣ′ : (ca : isContr A) → isContr (B (fst ca)) → isContr (Σ A B)
 isContrΣ′ ca cb = isContrΣ ca (λ x → subst _ (snd ca x) cb)
 
+Σ≡Prop-sect
```
mortberg, line 284: I would call this `section-Σ≡Prop`

Cubical/Foundations/HLevels.agda
```diff
@@ -220,15 +281,19 @@ isContrΣ {A = A} {B = B} (a , p) q =
...
+  : (pB : (x : A) → isProp (B x)) {u v : Σ A B}
+  → section (Σ≡Prop pB {u} {v}) (cong fst)
+Σ≡Prop-sect {A = A} pB {u} {v} p j i =
+  (p i .fst) , isProp→PathP (λ i → isOfHLevelPath 1 (pB (fst (p i)))
+                                       (Σ≡Prop pB {u} {v} (cong fst p) i .snd)
+                                       (p i .snd) )
+                                       refl refl i j
+
 Σ≡Prop-equiv
```
mortberg, line 293: I would call this `isEquiv-Σ≡Prop`

Cubical/Functions/Fibration.agda
```diff
@@ -67,9 +67,22 @@ module _ {ℓb} (B : Type ℓb) (ℓ : Level) where
...
+
 fiber≡ : ∀ {ℓ ℓ'} {A : Type ℓ} {B : Type ℓ'} {f : A → B} {b : B} (h h' : fiber f b)
   → (h ≡ h') ≡ fiber (cong f) (h .snd ∙∙ refl ∙∙ sym (h' .snd))
-fiber≡ {f = f} h h' =
+fiber≡ {f = f} {b = b} h h' =
   ΣPath≡PathΣ ⁻¹ ∙
-  cong (Σ (h .fst ≡ h' .fst)) (funExt λ p → flipSquarePath ∙ PathP≡doubleCompPathʳ _ _ _ _)
+  fiberPath h h'
 
+-- fiberIso : ∀ {ℓ ℓ'} {A : Type ℓ} {B : Type ℓ'} {f : A → B} {b : B} (h h' : fiber f b)
```
mortberg, line 83: Commented code

Cubical/HITs/Hopf.agda
```diff
@@ -320,3 +322,20 @@ JoinS¹S¹≡TotalHopf = isoToPath (iso JoinS¹S¹→TotalHopf
 
 S³≡TotalHopf : S³ ≡ TotalHopf
 S³≡TotalHopf = S³≡joinS¹S¹ ∙ JoinS¹S¹≡TotalHopf
+
+open Iso renaming (inv to inv')
```
mortberg, line 326: What is the other use on `inv` in scope? Maybe that should be renamed?

Cubical/HITs/Join/Base.agda
```diff
@@ -76,5 +76,12 @@ joinS¹S¹→S³→joinS¹S¹ (push base (loop k) i) l = push base (loop k) (i 
...
+S³IsojoinS¹S¹ : Iso S³ (join S¹ S¹)
+Iso.fun S³IsojoinS¹S¹ = S³→joinS¹S¹
+Iso.inv S³IsojoinS¹S¹ = joinS¹S¹→S³
+Iso.rightInv S³IsojoinS¹S¹ = joinS¹S¹→S³→joinS¹S¹
+Iso.leftInv S³IsojoinS¹S¹ = S³→joinS¹S¹→S³
+
+
 S³≡joinS¹S¹ : S³ ≡ join S¹ S¹
-S³≡joinS¹S¹ = isoToPath (iso S³→joinS¹S¹ joinS¹S¹→S³ joinS¹S¹→S³→joinS¹S¹ S³→joinS¹S¹→S³)
+S³≡joinS¹S¹ = isoToPath (S³IsojoinS¹S¹)
```
mortberg, line 87: Redundant parantheses

Cubical/HITs/S2/Base.agda
```diff
@@ -6,3 +6,8 @@ open import Cubical.Foundations.Prelude
 data S² : Type₀ where
   base : S²
   surf : PathP (λ i → base ≡ base) refl refl
+
+open import Cubical.Foundations.HLevels
```
mortberg, line 10: Move import to the top

Cubical/HITs/S2/Base.agda
```diff
@@ -6,3 +6,8 @@ open import Cubical.Foundations.Prelude
...
+S²ToSetRec : ∀ {ℓ} {A : S² → Type ℓ} → ((x : S²) → isSet (A x)) → (A base) → (x : S²) → A x
```
mortberg, line 11: Redundant parentheses in type and not necessary to pattern-match on {A = A}

Cubical/HITs/Sn/Base.agda
```diff
@@ -7,14 +7,20 @@ open import Cubical.Data.Nat
...
+open import Cubical.Data.Bool
+open import Cubical.HITs.S1
 
 S : ℕ₋₁ → Type₀
 S neg1 = ⊥
 S (ℕ→ℕ₋₁ n) = Susp (S (-1+ n))
 
 S₊ : ℕ → Type₀
-S₊ n = S (ℕ→ℕ₋₁ n)
+S₊ zero = Bool
```
mortberg, line 18: I think you can write 0 and 1 instead of `zero` and `suc zero`

Cubical/HITs/Truncation/FromNegOne/Base.agda
```diff
@@ -13,8 +13,17 @@ open import Cubical.Data.NatMinusOne renaming (suc₋₁ to suc)
...
 open import Cubical.Foundations.HLevels
 open import Cubical.HITs.Sn
+open import Cubical.Data.Nat.Base renaming (suc to sucℕ)
+open import Cubical.Data.Unit.Base
 
-data  ∥_∥_ {ℓ} (A : Type ℓ) (n : ℕ₋₁) : Type ℓ where
-  ∣_∣ : A  → ∥ A ∥ n
-  hub : (f : S (suc n) → ∥ A ∥ n) → ∥ A ∥ n
-  spoke : (f : S (suc n) → ∥ A ∥ n) (x : S (suc n)) → hub f ≡ f x
+data  ∥_∥⁻¹_ {ℓ} (A : Type ℓ) (n : ℕ₋₁) : Type ℓ where
```
mortberg, line 19: I would not introduce the ^-1 here. Maybe better to just call it `HubAndSpoke` or something?

Cubical/HITs/Truncation/FromNegOne/Properties.agda
```diff
@@ -90,93 +99,347 @@ isOfHLevel→isSphereFilled {A = A} (ℕ→ℕ₋₁ n) h = helper λ x y → is
...
-  Iso.fun f        = ind (λ _ → hA) λ a → a
-  Iso.inv f        = ∣_∣
-  Iso.rightInv f _ = refl
-  Iso.leftInv f    = ind (λ _ → isOfHLevelPath (1 + 1+ n) (isOfHLevel∥∥ n) _ _) (λ _ → refl)
+minΣ : (n m : ℕ) → Σ[ x ∈ ℕ ] x + min n m ≡ n
+minΣ zero m = 0 , refl
+minΣ (suc n) zero = (suc n) , cong suc (+-zero n)
+minΣ (suc n) (suc m) = (minΣ n m .fst) , +-suc _ _ ∙ cong suc (minΣ n m .snd)
+
+minComm : (n m : ℕ) → min n m ≡ min m n
```
mortberg, line 162: This should be moved to theory about naturals file

Cubical/HITs/Truncation/FromNegOne/Properties.agda
```diff
@@ -90,93 +99,347 @@ isOfHLevel→isSphereFilled {A = A} (ℕ→ℕ₋₁ n) h = helper λ x y → is
...
 
-truncIdempotent≃ : (n : ℕ₋₁) → isOfHLevel (1 + 1+ n) A → ∥ A ∥ n ≃ A
-truncIdempotent≃ {A = A} n hA = isoToEquiv f
-  where
-  f : Iso (∥ A ∥ n) A
-  Iso.fun f        = ind (λ _ → hA) λ a → a
-  Iso.inv f        = ∣_∣
-  Iso.rightInv f _ = refl
-  Iso.leftInv f    = ind (λ _ → isOfHLevelPath (1 + 1+ n) (isOfHLevel∥∥ n) _ _) (λ _ → refl)
+minΣ : (n m : ℕ) → Σ[ x ∈ ℕ ] x + min n m ≡ n
```
mortberg, line 157: This lemma is a bit strange... It's really just computing `n - min n m`

Cubical/HITs/Truncation/FromNegOne/Properties.agda
```diff
@@ -90,93 +99,347 @@ isOfHLevel→isSphereFilled {A = A} (ℕ→ℕ₋₁ n) h = helper λ x y → is
...
+minComm zero zero = refl
+minComm zero (suc m) = refl
+minComm (suc n) zero = refl
+minComm (suc n) (suc m) = cong suc (minComm n m)
+
+isOfHLevelMin→isOfHLevel : {n m : ℕ} → isOfHLevel (min n m) A → isOfHLevel n A × isOfHLevel m A
+isOfHLevelMin→isOfHLevel {n = zero} {m = m} h = h , isContr→isOfHLevel m h
+isOfHLevelMin→isOfHLevel {n = suc n} {m = zero} h = (isContr→isOfHLevel (suc n) h) , h
+isOfHLevelMin→isOfHLevel {A = A} {n = suc n} {m = suc m} h = (subst (λ x → isOfHLevel x A) (minΣ (suc n) (suc m) .snd) (isOfHLevelPlus _ h))
+                                                           , subst (λ x → isOfHLevel x A) (cong (λ y → minΣ m n .fst + suc y) (minComm n m) ∙ minΣ (suc m) (suc n) .snd) (isOfHLevelPlus _ h)
```
mortberg, line 172: Long line

Cubical/HITs/Truncation/FromNegOne/Properties.agda
```diff
@@ -90,93 +99,347 @@ isOfHLevel→isSphereFilled {A = A} (ℕ→ℕ₋₁ n) h = helper λ x y → is
...
+
+open import Cubical.Data.Nat hiding (elim )
```
mortberg, line 174: Move import up

Cubical/HITs/Truncation/FromNegOne/Properties.agda
```diff
@@ -90,93 +99,347 @@ isOfHLevel→isSphereFilled {A = A} (ℕ→ℕ₋₁ n) h = helper λ x y → is
...
+ΣTruncElim : ∀ {ℓ ℓ' ℓ''} {A : Type ℓ} {n m : ℕ}
+              {B : (x : ∥ A ∥ (suc n)) → Type ℓ'}
+              {C : (Σ[ a ∈ (∥ A ∥ (suc n)) ] (∥ B a ∥ (suc m))) → Type ℓ''}
+              → ((x : (Σ[ a ∈ (∥ A ∥ (suc n)) ] (∥ B a ∥ (suc m)))) → isOfHLevel (min (suc n) (suc m)) (C x))
+              → ((a : A) (b : B (∣ a ∣)) → C (∣ a ∣ , ∣ b ∣))
+              → (x : (Σ[ a ∈ (∥ A ∥ (suc n)) ] (∥ B a ∥ (suc m)))) → (C x)
+ΣTruncElim {n = n} {m = m} {B = B} {C = C} hB g (a , b) = (elim {B = λ a → (b :  (∥ B a ∥ (suc m))) → C (a , b)} (λ x → isOfHLevelΠ (suc n) λ b → isOfHLevelMin→isOfHLevel (hB (x , b)) .fst ) (λ a → elim (λ _ → isOfHLevelMin→isOfHLevel (hB (∣ a ∣ , _)) .snd) λ b → g a b) a) b
```
mortberg, line 181: Long line

Cubical/HITs/Truncation/FromNegOne/Properties.agda
```diff
@@ -90,93 +99,347 @@ isOfHLevel→isSphereFilled {A = A} (ℕ→ℕ₋₁ n) h = helper λ x y → is
...
+
+mapId : {n : HLevel} → ∀ t → map {n = n} (idfun A) t ≡ t
+mapId {n = 0} tt* = refl
+mapId {n = (suc n)} =
+  elim (λ _ → isOfHLevelPath (suc n) (isOfHLevelTrunc (suc n)) _ _) (λ _ → refl)
+
+-- equivalences to prop/set/groupoid truncations
+
+propTruncTrunc1Iso : Iso ∥ A ∥₁ (∥ A ∥ 1)
+Iso.fun propTruncTrunc1Iso = PropTrunc.elim (λ _ → isOfHLevelTrunc 1) ∣_∣
```
mortberg, line 242: Do you really need elim in the fun/inv cases and not just rec? Same for the other similar lemmas

Cubical/HITs/Truncation/FromNegOne/Properties.agda
```diff
@@ -90,93 +99,347 @@ isOfHLevel→isSphereFilled {A = A} (ℕ→ℕ₋₁ n) h = helper λ x y → is
...
+groupoidTrunc≃Trunc3 : ∥ A ∥₃ ≃ ∥ A ∥ 3
+groupoidTrunc≃Trunc3 = isoToEquiv groupoidTrunc≃Trunc3Iso
 
-2GroupoidTrunc≃Trunc2 : ∥ A ∥₂ ≃ ∥ A ∥ 2
-2GroupoidTrunc≃Trunc2 =
+groupoidTrunc≡Trunc3 : ∥ A ∥₃ ≡ ∥ A ∥ 3
+groupoidTrunc≡Trunc3 = ua groupoidTrunc≃Trunc3
+
+
+2GroupoidTrunc≃Trunc4Iso : Iso ∥ A ∥₄ (∥ A ∥ 4)
```
mortberg, line 279: Naming is weird

Cubical/HITs/Truncation/FromNegOne/Properties.agda
```diff
@@ -90,93 +99,347 @@ isOfHLevel→isSphereFilled {A = A} (ℕ→ℕ₋₁ n) h = helper λ x y → is
...
+Iso.fun 2GroupoidTrunc≃Trunc4Iso = 2GpdTrunc.elim (λ _ → isOfHLevelTrunc 4) ∣_∣
+Iso.inv 2GroupoidTrunc≃Trunc4Iso = elim (λ _ → squash₄) ∣_∣₄
+Iso.rightInv 2GroupoidTrunc≃Trunc4Iso = elim (λ _ → isOfHLevelPath 4 (isOfHLevelTrunc 4) _ _) (λ _ → refl)
+Iso.leftInv 2GroupoidTrunc≃Trunc4Iso = 2GpdTrunc.elim (λ _ → isOfHLevelPath 4 squash₄ _ _) (λ _ → refl)
+
+2GroupoidTrunc≃Trunc4 : ∥ A ∥₄ ≃ ∥ A ∥ 4
```
mortberg, line 285: Why not use the lemma above?

Cubical/Homotopy/Connected.agda
```diff
@@ -132,63 +143,61 @@ isOfHLevelPrecomposeConnected : ∀ {ℓ ℓ' ℓ''} (k : HLevel) (n : HLevel)
...
+                                tt)
 
 isConnectedPath : ∀ {ℓ} (n : HLevel) {A : Type ℓ}
   → isConnected (suc n) A
   → (a₀ a₁ : A) → isConnected n (a₀ ≡ a₁)
-isConnectedPath n connA a₀ a₁ =
-  subst isContr (PathIdTrunc _)
+isConnectedPath zero connA a₀ a₁ = isContrUnit*
+isConnectedPath (suc n) connA a₀ a₁ =
+  isContrRetract
```
mortberg, line 183: It seems like this proof pattern is appearing a lot. You apply "isContrRetract" (or similar lemmas) so some iso which you need to decompose. Why not prove this once and for all as a lemma "isContrRetractFromIso" or something?

Cubical/Homotopy/Freudenthal.agda
```diff
@@ -87,7 +94,7 @@ module _ {ℓ} (n : HLevel) {A : Pointed ℓ} (connA : isConnected (suc (suc n))
   encodeMerid : (a : typ A) → encode south (merid a) ≡ ∣ a , refl ∣
   encodeMerid a =
     cong (transport (λ i → gluePath i))
-      (funExt⁻ (WC.left refl a) _ ∙ cong ∣_∣ (cong (a ,_) (lem _ _)))
+      (funExt⁻ (WC.left refl a) _ ∙ λ i → ∣ a , (lem (rCancel' (merid a)) (rCancel' (merid (pt A)))) i ∣)
```
mortberg, line 97: redundant parentheses around lem

Cubical/Homotopy/Freudenthal.agda
```diff
@@ -111,6 +118,11 @@ module _ {ℓ} (n : HLevel) {A : Pointed ℓ} (connA : isConnected (suc (suc n))
   isConnectedσ =
     transport (λ i → isConnectedFun 2n+2 (interpolate (pt A) (~ i))) isConnectedMerid
 
+open import Cubical.Foundations.Pointed
+open import Cubical.Data.Bool
+
+isConn→isConnSusp : ∀ {ℓ} {A : Pointed ℓ} → isConnected 2 (typ A) → isConnected 2 (Susp (typ A))
+isConn→isConnSusp {A = A} iscon = ∣ north ∣ , (trElim (λ _ → isOfHLevelSuc 1 (isOfHLevelTrunc 2 _ _)) (suspToPropRec (pt A) (λ _ → isOfHLevelTrunc 2 _ _) refl))
```
mortberg, line 125: longline

Cubical/Homotopy/Freudenthal.agda
```diff
@@ -111,6 +118,11 @@ module _ {ℓ} (n : HLevel) {A : Pointed ℓ} (connA : isConnected (suc (suc n))
   isConnectedσ =
     transport (λ i → isConnectedFun 2n+2 (interpolate (pt A) (~ i))) isConnectedMerid
 
+open import Cubical.Foundations.Pointed
```
mortberg, line 121: move imports up

mortberg (changes requested):
Very nice that things are faster! I commented on everything except in the ZCohomology files as suggested.

Looking at this code makes me think that FromNegOne should be the default Truncation while the current one should be in a separate folder?

Cubical/Data/Unit/Base.agda
ecavallo: Might be better to define it using the existing `Lift` and `Unit`:
```
Unit* : {ℓ : Level} → Type ℓ
Unit* = Lift Unit

pattern tt* = lift tt
```
I suspect that this would be inconvenient to use as the default `Unit`. For one thing, you're often going to need to specify the level explicitly, `Unit* {ℓ}`, and you'll need more level quantification everywhere.

mortberg: Ah ok, I agree with @ecavallo 's suggestion then!

Cubical/Algebra/Group/Algebra.agda
aljungstrom: Haha wups, didn't mean to keep that name. I think it's a pretty useful lemma, so it should probably be there. But I'll rename it.

Cubical/Data/Unit/Base.agda
aljungstrom: Right, the lifted one is probably better to use with truncations since it inherits definitional equality from Unit. The reason I defined it like this is that I didn't want the definitional equality (I "lock" theorems in ZCohomology by giving them with respect to an element in Unit* and that doesn't work if you have definitional equality). I think I'll add a third Unit type called "lockUnit" or something then.

Cubical/Foundations/Equiv/HalfAdjoint.agda
aljungstrom: Hmm, I'm a bit scared of this proof. But I realised it was super easy to prove with J. This doesn't even rely on the fact that we have an Iso, so that should be much better for computations.

Cubical/Data/Unit/Base.agda
Saizan: Others have used the name `Block` for a unit type used like this. Even giving it a phantom string parameter, `Block : String -> Type` so the `String` can document what you're blocking.

Cubical/HITs/Hopf.agda
aljungstrom: That's the function from S1 to S1 sending loop to sym loop. I renamed it to invLooper

Cubical/Foundations/Equiv/HalfAdjoint.agda
mortberg: I'm confused, what are you using J on to prove it? 

As far as I can tell the proof is still scary btw...

mortberg (approved):

mortberg:
I would be happy to merge once the confusion about `invCongFunct` has been resolved

mortberg:
After merging this the beginning of the output of `make timings` is:

```
Cubical.ZCohomology.MayerVietorisUnreduced       14,023ms 
Cubical.ZCohomology.Properties                   12,158ms 
Cubical.ZCohomology.Groups.Sn                     8,331ms 
Cubical.Data.DescendingList.Examples              7,971ms 
Cubical.ZCohomology.Groups.Torus                  6,908ms 
Cubical.Algebra.Algebra.Base                      6,465ms 
Cubical.ZCohomology.Groups.Prelims                5,337ms 
Cubical.Foundations.GroupoidLaws                  5,008ms 
Cubical.Data.Fin.Properties                       4,479ms 
Cubical.HITs.Hopf                                 4,454ms 
Cubical.Foundations.HLevels                       3,973ms 
Cubical.Foundations.Id                            3,888ms 
Cubical.HITs.Truncation.FromNegOne.Properties     3,477ms 
Cubical.Algebra.FreeCommAlgebra                   3,467ms 
Cubical.Algebra.CommAlgebra.Base                  3,349ms 
Cubical.Homotopy.Connected                        3,243ms 
Cubical.HITs.S1.Base                              3,131ms
```

After running `agda --ignore-interfaces -v profile.definitions:10 Cubical/README.agda` I get the following output:

```
Cubical.Data.DescendingList.Examples.LongerExample                                  9,430ms 
Cubical.ZCohomology.Properties.commₖ                                                3,369ms 
Cubical.ZCohomology.MayerVietorisUnreduced.MV.dHomHelperPath                        2,654ms 
Cubical.Foundations.Id._.f12                                                        2,360ms 
Cubical.HITs.KleinBottle.Properties._                                               2,071ms 
Cubical.Foundations.Pointed.Homotopy._.∙∼≃∙∼P                                       1,569ms 
Cubical.ZCohomology.Groups.Torus._.theIso                                           1,420ms 
Cubical.Homotopy.Freudenthal._.encodeMerid                                          1,328ms 
Cubical.ZCohomology.Groups.Torus.H¹-T²≅ℤ×ℤ                                          1,320ms 
Cubical.ZCohomology.Properties.rUnitlUnit0                                          1,293ms 
Cubical.HITs.Cost.Base._                                                            1,070ms 
Cubical.Homotopy.Freudenthal._.isEquivFwd                                             923ms 
Cubical.ZCohomology.Groups.Prelims._._._.helperFun2                                   868ms 
```

I'll make a PR commenting out the first of these and see if I can do anything about the other slow ones

PR #422, opened 2020-10-01
Expand in-code explanation of Glue

WorldSEnder:
Adds a little bit more prose to the in-code documentation with the goal to increase readability for newcomers. Uses sub-typing to show definitional equality rules. The typing rules aren't numbered in the referenced paper, otherwise I'd add those to the corresponding definitions.

### DIFF
--- a/Cubical/Core/Glue.agda
+++ b/Cubical/Core/Glue.agda
@@ -91,7 +91,7 @@ private
     trans-e :
       ∀ (t : PartialP φ T)
       → Partial φ A
-    trans-e t ϕ1 = e ϕ1 .fst (t ϕ1)
+    trans-e t ϕ1 = equivFun (e ϕ1) (t ϕ1)
 
     -- glue gives a partial element of Glue given an element of A. Note that it "undoes"
     -- the application of the equivalences!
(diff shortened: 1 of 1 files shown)


Cubical/Core/Glue.agda
```diff
@@ -64,22 +64,76 @@ unglue φ = prim^unglue {φ = φ}
...
+    e i φ = Te i φ .snd
+    T0 T1 : Type ℓ'
+    T0 = T i0 1=1
+    T1 = T i1 1=1
+    e0 : T0 ≃ A0
+    e0 = e i0 1=1
+    e1 : T1 ≃ A1
+    e1 = e i1 1=1
+    transportA : A0 → A1
+    transportA = transp (λ i → A i) i0
```
mortberg, line 130: Why not use `transport` here? Same for the `transp` on line 139.

mortberg (changes requested):
Apart from using `transport` instead of `transp` I'm happy with this PR

Cubical/Core/Glue.agda
WorldSEnder: I was afraid of import cycles, but it seems importing from `Cubical.Foundations.Prelude` works fine. Let me just double check I can't also import `invEq`.

WorldSEnder:
Thank you for the quick review :+1:

PR #423, opened 2020-10-02
BiInvInt as a CommRing

felixwellen:
This is a very short PR. A CommRing is defined using the work from #414 .


mortberg (approved):
Great! Will merge once Travis is done and happy

Issue #405, opened 2020-08-27
Turn carrier types into parameters of records for algebraic structures


mortberg:
By doing this the structures will interface much better with the SIP and the automation proposed in https://github.com/agda/cubical/pull/395

This should be done after https://github.com/agda/cubical/pull/401

mortberg:
I'm going to close this now as it has been mostly fixed by https://github.com/agda/cubical/pull/430

PR #424, opened 2020-10-02
Kernel ideals and universal property of ring quotients

felixwellen:
Added RingHoms, kernels of RingHoms as ideals and the universal property of ring quotients.
The history might be confusing, since algebra code was restructered in between. 
The only real updates from this PR are Kernel.agda, some extension of QuotientRing.agda, a new record 'RingHom' in Ring.Base and some minor changes in Ring.Properties.


felixwellen:
No, actually it is also confusing because I messed it up...

felixwellen:
No the changes fit what I wrote.

Cubical/Algebra/Ring/Properties.agda
```diff
@@ -29,11 +29,11 @@ private
   that should become obsolete or subject to change once we have a
   ring solver (see https://github.com/agda/cubical/issues/297)
 -}
-module Theory (R' : Ring {ℓ}) where
+module Theory (R : Ring {ℓ}) where
 
-  open Ring R' renaming ( Carrier to R )
```
mortberg, line 34: Why do this change? I kind of prefer to omit the angle brackets around R...

Cubical/Algebra/Kernel.agda
```diff
@@ -0,0 +1,48 @@
+{-# OPTIONS --cubical --no-import-sorts --safe #-}
+module Cubical.Algebra.Kernel where
```
mortberg, line 2: Calling this file Kernel is maybe not ideal (pun intended!) as it's just about kernels in rings... Maybe we should have a subfolder Algebra.Ring where we have definitions like this? Alternatively calling thing `RingKernel` or something, but I prefer adding a subfolder

mortberg (changes requested):
Looks mostly great to me! But what happened to "Cubical/Algebra/Monoid/Base.agda"? Github claims that it has been completely changed, but I can't see any difference to the actual code...

Cubical/Algebra/Kernel.agda
felixwellen: I didn't think about it, I like the subfolder idea.

felixwellen:
> 
> 
> Looks mostly great to me! But what happened to "Cubical/Algebra/Monoid/Base.agda"? Github claims that it has been completely changed, but I can't see any difference to the actual code...

I'll try to figure out what happened. I missed it when looking at the changes because it was collapsed.

Cubical/Algebra/Ring/Properties.agda
felixwellen: I'll revert that - I don't have a strong preference and I don't remember why I changed that.

Cubical/Algebra/Kernel.agda
felixwellen: I'll also move Ideal and QuotientRing to Ring/

Cubical/Algebra/Ring/Properties.agda
mortberg: This should get better once we go for partially bundled structures as suggested by: https://github.com/agda/cubical/issues/405

felixwellen: Good!

mortberg (approved):

mortberg:
Perfect! Will merge once Travis is finished.

@mzeuner : I guess Localization should go in the Algebra.Ring subfolder once I've merged this

PR #429, opened 2020-10-03
Add "timings" target to makefile.

### LINKED ISSUE #375: Script for getting timings for typechecking files
I would love to have a script that runs `time` on each file when typechecking and then show the N slowest. More fine grained data would also be useful (like which are the slowest definitions to check in these files), but a simple script would be enough to keep track of which files are the slowest.

cangiuli:
Closes #375. Running `make timings` will check all the files then output a sorted list of timings, e.g.:

```
Total                                           275,643ms 
Miscellaneous                                     3,518ms 
Cubical.ZCohomology.Groups.Sn                    26,112ms 
Cubical.ZCohomology.KcompPrelims                 19,645ms 
Cubical.ZCohomology.Groups.Wedge                  7,493ms 
Cubical.ZCohomology.Properties                    6,763ms 
Cubical.Data.DescendingList.Examples              6,420ms 
Cubical.ZCohomology.MayerVietorisUnreduced        5,956ms 
  [ ... ]
```

I also took the opportunity to clean up a couple things nearby in the makefile. In particular, I'm confused why `make check` was invoking agda separately on each dependency of `README.agda` instead of just checking that file directly. If there is a good reason I'm happy to revert that part of the PR. /cc @m-yac

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -4,7 +4,7 @@ AGDA=$(AGDA_EXEC) $(RTS_OPTIONS)
 EVERYTHINGS=runhaskell ./Everythings.hs
 
 .PHONY : all
-all : gen-everythings check
+all : check
 
 .PHONY : test
 test: check-whitespace gen-and-check-everythings check-README check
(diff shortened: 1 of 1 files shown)


mortberg:
Great! Shouldn't the `timings` target depend on `gen-everythings`?

cangiuli:
I did it this way because `check` doesn't currently depend on `gen-everythings`, but IMO it would make sense to add the dependency to both. (`all` calls both `gen-everythings` and `check`.)

mortberg (approved):
Great! These changes seem reasonable to me so I think we should merge as soon as Travis is happy

PR #430, opened 2020-10-05
Parameterize structures by the carrier 

mortberg:
Towards solving: https://github.com/agda/cubical/issues/405

<strike>I've only modified the definition of monoids so far.</strike> Is this the right way to do this so that it works with https://github.com/agda/cubical/pull/395 ? 

(Ping @ecavallo @cangiuli @mzeuner )

### DIFF
--- a/Cubical/Algebra/AbGroup/Base.agda
+++ b/Cubical/Algebra/AbGroup/Base.agda
@@ -8,15 +8,15 @@ open import Cubical.Foundations.HLevels
 open import Cubical.Foundations.Isomorphism
 open import Cubical.Foundations.Univalence
 open import Cubical.Foundations.Transport
-open import Cubical.Foundations.SIP
+open import Cubical.Foundations.SIP hiding (⟨_⟩)
 
 open import Cubical.Data.Sigma
 
 open import Cubical.Structures.Axioms
 open import Cubical.Structures.Macro
 open import Cubical.Structures.Pointed
-open import Cubical.Algebra.Semigroup hiding (⟨_⟩)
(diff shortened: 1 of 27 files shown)


ecavallo:
Yep, that's good.

mortberg:
> Yep, that's good.

Alright, I've started parameterizing the equivalences of structures by the equivalence of types now as well now. Looking at your automation PR I wonder if it's really necessary to package data and laws together for structures?

ecavallo:
No, it's not necessary, it will also work if each law is given its own field in the main structure.

mortberg:
I've done this for AbGroup, CommRing, Group, Monoid, Ring and SemiGroup. I also parameterized the equivalences of these structures by equivalences of the underlying types, except for Group where the equivalences are kind of special (they only need to preserve the composition).

I did not touch Algebra, CommAlgebra and Module yet. I think those can wait as they are also kind of special as they depend on another type.

What does @ecavallo think?

ecavallo:
Sounds like a good-sized testing ground to try the reflection, let's merge and I'll give it a shot. It should be possible to get the right definition of equivalence for the `Group` case just by registering the unit field as a prop.

mortberg:
> Sounds like a good-sized testing ground to try the reflection, let's merge and I'll give it a shot. It should be possible to get the right definition of equivalence for the `Group` case just by registering the unit field as a prop.

Ah right... There's quite a bit of code that depends on GroupEquiv, so changing it will be very painful. I think one should change all of the other fancy morphisms of groups at the same time

mortberg:
Alright, I'm merging this now. Sorry to everyone who will get merge conflicts...

PR #431, opened 2020-10-06
James

aljungstrom:
I'm trying to do some stuff with the James construction but I am having some termination issues that I cannot seem to understand.

I have a HIT defined like this
https://github.com/aljungstrom/cubical/blob/termination/Cubical/HITs/James/Base.agda#L8
(this is one of the HITS from the James construction).

I am trying to define maps from J (S^n) into certain types of fixed hLevel, but I cannot seem to get the higher path cases to terminate. The smallest example I could come up with is this:

https://github.com/aljungstrom/cubical/blob/termination/Cubical/HITs/James/Base.agda#L13

When I try to enter `SquareHelp i j`, Agda tells me:

> Termination checking failed for the following functions:
>   test, _.SquareHelp
> Problematic calls:
>   SquareHelp i j
>   test (αJ (loop i) y)
> when checking that the expression SquareHelp i j has type
> hLevelTrunc 2 S¹

Does anyone have any idea of how I can solve this problem?


aljungstrom:
ping @mortberg @Saizan

Saizan:
The termination checker does not recognize `αJ (loop i) y` as smaller than `αJ (loop i) (αJ base y)` because it's not a subterm.

You might try defining an helper function for the `αJ` case which takes the two arguments separately.

aljungstrom:
> The termination checker does not recognize `αJ (loop i) y` as smaller than `αJ (loop i) (αJ base y)` because it's not a subterm.
> 
> You might try defining an helper function for the `αJ` case which takes the two arguments separately.

Ahh, I see. This (together with some mutual induction) worked! Thanks a lot!

PR #433, opened 2020-10-08
Port proof helpers from stdlib

grahnen:
I added some of the proof-automation methods from `Relation.Nullary.Decidable.Core` in the standard library.


mortberg:
Looks fine to me, but it would be good if these would be used somewhere in the library

PR #434, opened 2020-10-08
create RepresentationIndependence.agda

mzeuner:
Add a file that  contains pointers to the code examples and main results from the paper:

*Internalizing Representation Independence with Univalence*

Carlo Angiuli, Evan Cavallo, Anders Mörtberg, Max Zeuner

Preprint: https://arxiv.org/abs/2009.05547


mortberg:
Great!

PR #435, opened 2020-10-12
Displayed Univalent Reflexive Graphs (related to SIP)

Schippmunk:
I realize this adds a bunch of unrelated things, but we can cherry-pick the relevant parts if we decide to merge.

Relevant is the `DStructures` top-level directory. The definition of (D)URGs is in `DStructures/Base.agda`.
Various operations are in the Meta subdirectory. Tons of examples of structures can be found in the Structures subdirectory.

(This code also contains #394, but that part is broken right now, because I merged a lot of code from the master branch. I am aware that there is some mix-up with the nomenclature of split epis and section retraction pairs.)


mortberg:
In the meeting today we decided to split this PR into 3 parts:

1. EM spaces (will subsume https://github.com/agda/cubical/pull/394 AFAIU)
2. Displayed univalent reflexive graphs (will generalize the current SIP in the library)
3. Equivalence of crossed modules and strict 2-groups (depends on 1 and 2)

For 2 the relationship to the SIP would be clearer if it used an equivalent, but more cubical, definition inspired by https://github.com/agda/cubical/blob/master/Cubical/Foundations/SIP.agda#L45

Open problem: how much of the automation that we already have for the SIP can be ported to DURGs?

PR #436, opened 2020-10-12
WIP on optimizing definitions related to cohomology

mortberg:
Ping @aljungstrom

### DIFF
--- a/Cubical/HITs/SetTruncation/Properties.agda
+++ b/Cubical/HITs/SetTruncation/Properties.agda
@@ -20,7 +20,7 @@ open import Cubical.Data.Sigma
 
 private
   variable
-    ℓ : Level
+    ℓ ℓ' : Level
     A B C D : Type ℓ
 
 rec : isSet B → (A → B) → ∥ A ∥₂ → B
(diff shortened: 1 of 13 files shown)


ecavallo:
I don't know if this would be a significant improvement (or at all), but I'll mention since it occurred to me: it might produce smaller terms to implement functions like `rec2` and `rec3` for the concrete truncations by direct pattern matching, rather than going through `isPropΠ` and so on, like so.
```
rec2' : ∀ {P : Type ℓ} → isProp P → (A → A → P) → ∥ A ∥ → ∥ A ∥ → P
rec2' Pprop f ∣ x ∣ ∣ y ∣ = f x y
rec2' Pprop f ∣ x ∣ (squash t₀ t₁ i) = Pprop (rec2' Pprop f ∣ x ∣ t₀) (rec2' Pprop f ∣ x ∣ t₁) i
rec2' Pprop f (squash t₀ t₁ i) t = Pprop (rec2' Pprop f t₀ t) (rec2' Pprop f t₁ t) i
```

Cubical/ZCohomology/Groups/Wedge.agda
```diff
@@ -45,11 +41,11 @@ module _ {ℓ ℓ'} (A : Pointed ℓ) (B : Pointed ℓ') where
     bijIso : BijectionIso (coHomGr 1 (A ⋁ B)) (×coHomGr 1 (typ A) (typ B))
     BijectionIso.map' bijIso = I.i 1
     BijectionIso.inj bijIso =
-       sElim (λ _ → isOfHLevelΠ 2 λ _ → isOfHLevelPath 2 setTruncIsSet _ _)
-              λ f inker → helper ∣ f ∣₂ (I.Ker-i⊂Im-d 0 ∣ f ∣₂ inker)
+      sElim (λ _ → isSetΠ λ _ → isProp→isSet (setTruncIsSet _ _))
+            λ f inker → helper ∣ f ∣₂ (I.Ker-i⊂Im-d 0 ∣ f ∣₂ inker)
     BijectionIso.surj bijIso =
```
mortberg, line 46: Strangely enough this is now the slowest definition in the whole library... Any ideas how to simplify it @aljungstrom ? Maybe one can prove the base case more directly without going through `BijectionIsoToGroupIso` ?

mortberg:
> I don't know if this would be a significant improvement (or at all), but I'll mention since it occurred to me: it might produce smaller terms to implement functions like `rec2` and `rec3` for the concrete truncations by direct pattern matching, rather than going through `isPropΠ` and so on, like so.
> 
> ```
> rec2' : ∀ {P : Type ℓ} → isProp P → (A → A → P) → ∥ A ∥ → ∥ A ∥ → P
> rec2' Pprop f ∣ x ∣ ∣ y ∣ = f x y
> rec2' Pprop f ∣ x ∣ (squash t₀ t₁ i) = Pprop (rec2' Pprop f ∣ x ∣ t₀) (rec2' Pprop f ∣ x ∣ t₁) i
> rec2' Pprop f (squash t₀ t₁ i) t = Pprop (rec2' Pprop f t₀ t) (rec2' Pprop f t₁ t) i
> ```

Good point! Could be worth trying

Cubical/HITs/SetTruncation/Properties.agda
```diff
@@ -29,24 +29,40 @@ rec Bset f (squash₂ x y p q i j) =
...
         (x : ∥ A ∥₂) (y : ∥ B ∥₂) → C x y
-elim2 Cset f = elim (λ _ → isSetΠ (λ _ → Cset _ _))
-                    (λ a → elim (λ _ → Cset _ _) (f a))
+elim2 Cset f ∣ x ∣₂ ∣ y ∣₂ = f x y
+elim2 Cset f ∣ x ∣₂ (squash₂ y z p q i j) =
+  isOfHLevel→isOfHLevelDep 2 (λ a → Cset ∣ x ∣₂ a) _ _
+     (cong (elim2 Cset f ∣ x ∣₂) p) (cong (elim2 Cset f ∣ x ∣₂) q) (squash₂ y z p q) i j
+elim2 Cset f (squash₂ x y p q i j) z =
+  isOfHLevel→isOfHLevelDep 2 (λ a → Cset a z) _ _
+    (cong (λ a → elim2 Cset f a z) p) (cong (λ a → elim2 Cset f a z) q) (squash₂ x y p q) i j
```
mortberg, line 61: @ecavallo Didn't notice any massive speedup with this change... But I guess it can't hurt having a more direct definition

Cubical/ZCohomology/Groups/Wedge.agda
```diff
@@ -45,11 +41,20 @@ module _ {ℓ ℓ'} (A : Pointed ℓ) (B : Pointed ℓ') where
...
-      sigmaElim (λ _ → isOfHLevelSuc 1 propTruncIsProp)
-                 λ f g → I.Ker-Δ⊂Im-i 1 (∣ f ∣₂ , g) (isOfHLevelSuc 0 (isContrHⁿ-Unit 0) _ _)
+      prodElim (λ _ → isProp→isSet propTruncIsProp)
+                λ f g → ∣ ∣ helpFun f g {!!} ∣₂ , refl ∣₁
```
mortberg, line 48: @aljungstrom I tried to inline the proof and simplify it, but I can't figure out what to do with this hole...

aljungstrom: I'll have a look:-)

Cubical/ZCohomology/Groups/Wedge.agda
aljungstrom: Hmm, I think this is the weird one where I had to eliminatre truncations although I shouldn't have to. Try doing it without elimination on the first thing and maybe it'll go through this time... I think it's hard to do without the bijectionIso since MV only gives us surjective an injective functions (on a set level). But I'll think about it today

Cubical/ZCohomology/Groups/Wedge.agda
aljungstrom: You can skip the elimination. That was caused by some strange issue which seems to have been magically resolved. Try:
BijectionIso.surj bijIso p = I.Ker-Δ⊂Im-i 1 p (isOfHLevelSuc 0 (isContrHⁿ-Unit 0) _ _)

aljungstrom: Actually I think there are many places where one could remove these unnecessary elimminations. I have no idea why it terminates now and not before.

mortberg: Might have to do with me being more careful about rec/elim... Or any of the other many optimizations we've found :-)

mortberg:
Yay, things are now fast enough that I can disable almost all no-eta-equality from Algebra. This is necessary in order to finish https://github.com/agda/cubical/pull/430 without too much pain

I think the main culprit is now compGroupIso. It should have some nicer eq reasoning style notation so that we can give the groups explicitly. I think inferring them is really what makes Agda suffer so much when removing no-eta-equality for IsGroup

PR #437, opened 2020-10-13
More general embedding HLevel proofs

dolio:
This proves that embeddings determine h-levels greater than 1.

I suppose one question is: should this actually use the argument type with the notation? It converts it back to `isEmbedding` to perform the actual proof (like the existing set version), so if you got your `hasPropFibers` from an embedding proof, that's a bit wasteful. Perhaps it should take the premise it actually wants to use?


mortberg:
I'm fine with merging and have no opinions about the question. Should I go ahead and merge?

mortberg (approved):

dolio:
Yeah, might as well merge. Depending on what people think about the question, it probably makes sense to swap all of them or none of them, and is orthogonal to this PR.

PR #438, opened 2020-10-13
HoTT-book truncation

aljungstrom:
This PR changes the main definition of truncations back to the usual hub-and-spoke definition. We lose the nice treatment of n = -2, but computation-wise this should be a good thing.

I've also changed the definition to agree with the new definition of spheres (as iterated suspensions of S¹  with the loop-base constructors). I also threw in a new proof of the n-connectedness of spheres which got much shorter partly due to this change. Right now, I use | base | point of contraction of ∥ Sⁿ⁺¹ ∥ₙ. Using hub | _ | gives as the point of contraction gives a shorter proof (no path composition needed in the proof of contractibility), but I know that maps from truncations usually behave better for 
| base | than for hub |_| ,so I wasn't really sure of which one to use. Perhaps it doesn't matter. Maybe @ecavallo has an idea of which one is best to use?


ecavallo:
No, I don't have any idea which would be better here.

Cubical/HITs/Truncation/Base.agda
```diff
@@ -1,29 +1,34 @@
...
---   ∣_∣ : A → hLevelTrunc n A
---   hub   : (f : S (-1+ n) → hLevelTrunc n A) → hLevelTrunc n A
---   spoke : (f : S (-1+ n) → hLevelTrunc n A) (s : S) → hub f ≡ f s
---   -- two additional constructors needed to ensure that hLevelTrunc 0 A is contractible
---   ≡hub   : ∀ {x y} (p : S (-1+ n) → x ≡ y) → x ≡ y
---   ≡spoke : ∀ {x y} (p : S (-1+ n) → x ≡ y) (s : S (-1+ n)) → ≡hub p ≡ p s
+data HubAndSpoke {ℓ} (A : Type ℓ) (n : ℕ₋₁) : Type ℓ where
+  ∣_∣ : A  → HubAndSpoke A n
+  hub : (f : S' (suc n) → HubAndSpoke A n) → HubAndSpoke A n
+  spoke : (f : S' (suc n) → HubAndSpoke A n) (x : S' (suc n)) → hub f ≡ f x
```
mortberg, line 27: So what it necessary to introduce `S'` and use `suc n` here? It really seems like you can simplify it

aljungstrom: Sorry for the delay, I got distracted by other stuff as usual:-) It's fixed now.

mortberg:
Oops, nu är det konflikter...

Cubical/HITs/Truncation/Base.agda
```diff
@@ -1,29 +1,32 @@
...
---   -- the hub-and-spoke definition in `Truncation.FromNegOne`
---   ∣_∣ : A → hLevelTrunc n A
---   hub   : (f : S (-1+ n) → hLevelTrunc n A) → hLevelTrunc n A
---   spoke : (f : S (-1+ n) → hLevelTrunc n A) (s : S) → hub f ≡ f s
---   -- two additional constructors needed to ensure that hLevelTrunc 0 A is contractible
---   ≡hub   : ∀ {x y} (p : S (-1+ n) → x ≡ y) → x ≡ y
---   ≡spoke : ∀ {x y} (p : S (-1+ n) → x ≡ y) (s : S (-1+ n)) → ≡hub p ≡ p s
+-- this definition is off by one. Use hLevelTrunc or ∥_∥ for truncations
+-- (off by 2 w.r.t. the HoTT-book)
+data HubAndSpoke {ℓ} (A : Type ℓ) (n : ℕ) : Type ℓ where
```
mortberg, line 22: Please change N to HLevel here

aljungstrom: This might be confusing since this is not the HLevel, but HLevel - 1

mortberg: Oh, right! Can you add comment about it

PR #439, opened 2020-10-13
Make equivalence arguments implicit in equivEq

mchristianl:


mortberg (approved):
Excellent! Merging as soon as Travis is finished

PR #441, opened 2020-10-13
rename sec and ret in isHAEquiv to linv and rinv

halfaya:
As discussed in the AIM 33 conversation, this is for consistency with Isomorphism and the HoTT book; it is more clear terminology.

To check this I typechecked README.agda, but several "Everything" files were missing (I commented them out temporarily) so I missed checking some directories. I don't know if there's a better way to do this.

### DIFF
--- a/Cubical/Data/Sigma/Properties.agda
+++ b/Cubical/Data/Sigma/Properties.agda
@@ -149,17 +149,17 @@ leftInv Σ-Π-Iso _     = refl
 fun (Σ-cong-iso-fst isom) x = fun isom (x .fst) , x .snd
 inv (Σ-cong-iso-fst {B = B} isom) x = inv isom (x .fst) , subst B (sym (ε (x .fst))) (x .snd)
   where
-  ε = isHAEquiv.ret (snd (iso→HAEquiv isom))
+  ε = isHAEquiv.rinv (snd (iso→HAEquiv isom))
 rightInv (Σ-cong-iso-fst {B = B} isom) (x , y) = ΣPathP (ε x , toPathP goal)
   where
-  ε = isHAEquiv.ret (snd (iso→HAEquiv isom))
+  ε = isHAEquiv.rinv (snd (iso→HAEquiv isom))
   goal : subst B (ε x) (subst B (sym (ε x)) y) ≡ y
   goal = sym (substComposite B (sym (ε x)) (ε x) y)
       ∙∙ cong (λ x → subst B x y) (lCancel (ε x))
       ∙∙ substRefl {B = B} y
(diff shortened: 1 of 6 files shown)


mortberg:
It seems like your README.agda missed some files. Does it look like this: https://github.com/agda/cubical/blob/master/Cubical/README.agda ?

I think @cangiuli had a similar problem lately...?

halfaya:
> It seems like your README.agda missed some files. Does it look like this: https://github.com/agda/cubical/blob/master/Cubical/README.agda ?
> 
> I think @cangiuli had a similar problem lately...?

Yes, but as I mentioned several "Everything" files are missing. I'll make a separate issue to fix that problem. Meanwhile since Travis found a problem I'll fix that and see if that's sufficient for this PR.

halfaya:
I found running "make" works fine but manually typechecking README.agda in emacs fails. I'll create an issue. I found what are hopefully the final two references.

halfaya:
Ah I think I figured out what's going on. Running "make" creates these files (some of them were there already, perhaps due to things I'd manually typechecked before?), so once I was done they exist. I won't create an issue after all unless there's something you think would be good to clarify.

mortberg:
Right, make generates these files... I'm not sure how we could clarify this. Maybe in the INSTALL instructions?

halfaya:
You're welcome! Make is actually mentioned in INSTALL but I never read that since this is just Agda code and you can just typecheck what you need. It's more an issue for CONTRIBUTING, and although embarrassingly I didn't read that file either it should be added. I'll do that in a separate PR.

cangiuli:
I think it would be better if the Everything.agda files were checked in, and the CI just made sure every module is included. As it stands, I find it weird that you have to run `make` before the Agda files typecheck, and that doing so requires Haskell (although yes, most Agda users probably have Haskell)...

halfaya:
You actually don't need to run make at all as an end user (I never did), but I agree the Everything files are confusing and I didn't realize they were generated by make. It might indeed be better to check them in and just ensure they are correct (they can still be auto-generated by make)...I'm not sure if that ends up being more effort or not, though.

halfaya:
Actually now that I think of it maybe I did run make a long time ago and just forgot, so that would explain why I had some Everything files but not others.

cangiuli:
True, it depends whether you're using any module that imports an Everything file that isn't checked in. Part of the confusion is that _some_ Everything files are checked in and some aren't, because some of them require special `public` or `hiding` directives (e.g., Cubical.Core.Everything).

halfaya:
Oh I hadn't looked but if some Everything files are checked in then yes it would be ideal to just check them all in, and create a make target to update them automatically as needed.

PR #445, opened 2020-10-14
A small example of transporting theory between two equivalent representations of lists

mortberg:
The equivalence is as simple as possible and the goal is just to illustrate how one can transport results either directly or using the SIP

As far as I can tell this is the only example of how to use the SIP machinery which isn't polluted by code relating records and Sigma-types...


Cubical/Experiments/List.agda
```diff
@@ -0,0 +1,204 @@
...
+  -- signature of _++_ and rev.
+  RawStruct : Type → Type
+  RawStruct X = (X → X → X) × (X → X)
+
+  -- Some automated SIP magic
+  RawEquivStr : _
+  RawEquivStr = AutoEquivStr RawStruct
+
+  rawUnivalentStr : UnivalentStr _ RawEquivStr
+  rawUnivalentStr = autoUnivalentStr RawStruct
```
mortberg, line 162: @ecavallo Why is the "a" capitalized in `AutoEquivStr` but not in `autoUnivalentStr`?

Also, maybe these should be combined so that both get generated by one call to automation? I guess `transferAxioms` could take such a pair then

Cubical/Experiments/List.agda
```diff
@@ -0,0 +1,204 @@
...
+  toList'-++ (x ∷ xs) ys i = toList'-++ xs ys i ∷' x
+
+  toList'-rev : (xs : List A) → toList' (rev xs) ≡ rev' (toList' xs)
+  toList'-rev [] = refl
+  toList'-rev (x ∷ xs) = toList'-++ (rev xs) (x ∷ []) ∙ cong (_++' ([] ∷' x)) (toList'-rev xs)
+
+  -- We can now get the property for ++' and rev' via the SIP
+  goal : (xs ys : List' A) → rev' (xs ++' ys) ≡ rev' ys ++' rev' xs
+  goal = transferAxioms rawUnivalentStr List-Struct List'-RawStruct
+                        (ListEquiv , toList'-++ , toList'-rev) .snd .snd
```
mortberg, line 200: Maybe `transferAxioms` should really just output the axioms? We can then have `inducedStructure` instead of the current `transferAxioms`?

Cubical/Experiments/List.agda
```diff
@@ -0,0 +1,204 @@
...
+
+  -- Note that "goal" is really talking about the direct definitions
+  -- of ++' and rev', not the transported operations as in the
+  -- previous attempt.
```
mortberg, line 204: Note that I don't need to assume that A is set for this example

Cubical/Experiments/List.agda
cangiuli: I think the capitalized ones produce types and the lowercase ones don't?

ecavallo: Carlo's right.

I'm not sure what will be the best way of combining record automation with the raw/induced structure stuff.

Cubical/Experiments/List.agda
mortberg: I did this now. What do you think? @cangiuli @ecavallo @mzeuner

Issue #404, opened 2020-08-27
Reorganize files and prove equivalences of the many types of integers and rationals


mortberg:
The library contains a lot of different representations of the integers:

- https://github.com/agda/cubical/tree/master/Cubical/HITs/Ints
- https://github.com/agda/cubical/tree/master/Cubical/Data/DiffInt
- https://github.com/agda/cubical/tree/master/Cubical/Data/Int
- https://github.com/agda/cubical/blob/master/Cubical/HITs/S1/Base.agda#L40

and the rationals:

- https://github.com/agda/cubical/tree/master/Cubical/HITs/Rationals

It would be good to collect these in one place, maybe in a folder in Cubical.Algebra? No matter what the Data/HITs distinction doesn't really make sense. This reorganization should be done in a separate PR.

Once the files are in the same place we should have proofs that all of the different representations are equivalent. This might be easier for some representations than others, but once https://github.com/agda/cubical/pull/400 is done we will then be able to transport the proof that the integers form a commutative ring to all of the other representations using the SIP. Doing one or more of these equivalences would be a good first PR for someone trying to learn Cubical Agda.

mortberg:
These issues are related:
- https://github.com/agda/cubical/issues/116
- https://github.com/agda/cubical/issues/126
- https://github.com/agda/cubical/issues/150

It might be possible to close some of them.

mortberg:
@mchristianl I don't know why, but I can't assign you to this issue...

mchristianl:
That's strange. I got the notification email, so this should be the correct identifier at least

mchristianl:
About the boolean-valuedness of decidable orders `_<_`: we do already have a type-valued order on ℕ:

https://github.com/agda/cubical/blob/367dd3d4cf185693f02aac3deef93888dedc7eeb/Cubical/Data/Nat/Order.agda#L25-L26

and it is also defined likewise (as a relation) in the non-cubical standard library.

Which approach should we follow?

knrafto:
A point in favor of the first definition is that it works nicely for ints as well:

```
_≤_ : ℤ → ℤ → Type₀ 
m ≤ n = Σ[ k ∈ ℕ ] pos k + m ≡ n
```

mortberg:
The question about boolean or non-boolean orders is orthogonal to this issue... But I personally like boolean tests that then get lifted to types using boolean reflection à la SSReflect.  I don't know why other Agda libraries tend to not do this?

[...]

Issue #286, opened 2020-04-17
The type of reals


ayberkt:
I have been looking for a definition of `ℝ` around here but I haven't been able to find one.

Is there a particular reason that this is the case? Otherwise, I can do this and send a PR.

Saizan:
No particular reason as far as I understand, a PR would be great!

UlrikBuchholtz:
Which definition are you planning to use? Probably, some kind of Dedekind reals are a good start, but perhaps parametrized by an arbitrary σ-frame, as suggested in Sec. 11.2 of the HoTT book. Then it can be instantiated with either Prop or the initial σ-frame.

It could also be a good idea to axiomatize the notion of a Caucy-complete Archimedean ordered field, as in [Auke Booij's thesis](https://www.cs.bham.ac.uk/~abb538/thesis-first_submission.pdf), and show that Dedekind reals are an instance. The HoTT reals are then another instance. (Can cubical Agda handle the HoTT reals as a HIIT yet?)

Saizan:
>  (Can cubical Agda handle the HoTT reals as a HIIT yet?)+

I would suggest to start experimenting with them, transp won't compute yet but should be very soon and I could use some examples to test it on.

ayberkt:
@UlrikBuchholtz thanks for the suggestions.

> Which definition are you planning to use? Probably, some kind of Dedekind reals are a good start, but perhaps parametrized by an arbitrary σ-frame, as suggested in Sec. 11.2 of the HoTT book. Then it can be instantiated with either Prop or the initial σ-frame.

Yes, I was planning to follow Sec. 11.2 (option (iv) for Ω) of the book as you suggested. I was planning to use the initial σ-frame instead of parametrising, but it might be better to parametrise. By the way, I have some stuff on frames and their nuclei here: https://ayberkt.gitlab.io/msc-thesis/Frame.html#936. Uses the SNS from this library (an older version).

> It could also be a good idea to axiomatize the notion of a Caucy-complete Archimedean ordered field, as in Auke Booij's thesis, and show that Dedekind reals are an instance. The HoTT reals are then another instance. (Can cubical Agda handle the HoTT reals as a HIIT yet?)

Sounds like a good idea. I need to look at the thesis.

[...]

Issue #116, opened 2019-03-29
Cubical.Data.Rational: define operations and equivalence with HoTT def.


Saizan:
Right now ```Cubical.Data.Rational``` only contains the datatype.

Arithmetic is there to be implemented!

The Book defines rationals as the set quotient ```Z x N / ~``` where 
```(a, b) ~ (c, d) iff a(d + 1) = c(b + 1)```
and it would be good to provide an equivalence.

mortberg:
Closing this as it's superseded by https://github.com/agda/cubical/issues/404

PR #446, opened 2020-10-14
SyntheticReals (draft)

mchristianl:
a few ideas of mine how to "set up" various numbers for the purpose of doing real analysis lateron

see https://github.com/mchristianl/synthetic-reals/blob/master/README.md

to be discussed today

### DIFF
--- a/SyntheticReals/Bundles.agda
+++ b/SyntheticReals/Bundles.agda
@@ -1,0 +1,358 @@
+{-# OPTIONS --cubical --no-import-sorts #-}
+
+module SyntheticReals.Bundles where
+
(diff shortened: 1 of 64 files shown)


mortberg:
I think the conclusion was that this work should be a separate development but that some parts should be upstreamed to the library in the form of smaller PRs? So I think we can close this one?

mchristianl:
Yes. Originally I was motivated because of a comment of @UlrikBuchholtz in https://github.com/agda/cubical/issues/286#issuecomment-615224185 so a new Issue _"Caucy-complete Archimedean ordered field"_ would summarize the parts that should be most suitable for the cubical standard library.

As already suggested in the comments of #286 (The type of reals) we have that #116 is a dependency (operations on rationals), now supersed by #404. I would really like to reach that definition of Caucy-complete Archimedean ordered field at some time.

PR #447, opened 2020-10-14
Equational reasoning for composition of group isos + some cleanup of ZCohomology

aljungstrom:
I guess the title of this PR is pretty self explanatory...

The cleaup of ZCohomology: 
Added equational reasoning.
Shorter for of some trivial cohomology groups using the definition of truncations.

@mortberg I am really not sure what is going on here. The ZCohomology.Groups.Wedge.Agda now takes like 40 seconds to check. I tried making arguments implicit, explicit and defining the vSES using copatterns, but nothing seems to help. Seems weird that this would be caused by the equational reasoning I added in e.g. ZCohomology.Sn


mortberg:
There are conflicts...

aljungstrom:
> There are conflicts...

I think github is just being slow. The files changed doesn't seem to take the last merged branch into account.

mortberg:
Very strange that things got slower. Will try here

mortberg:
New PR coming, so closing this one

PR #450, opened 2020-10-15
refactoring * with · in number and algebra related parts of the library AND commring structure on Cubical.Data.Int 

mchristianl:


mchristianl:
The commring structure with the three commits

- utility for implicational reasoning
- commring structure on Cubical.Data.Int
- fixing broken parts of the library 

should be a draft PR. I'll try to remove these from the current PR ...

PR #451, opened 2020-10-15
Start working on some naming convention guidelines

mortberg:
More things to mention:

- naming for h-level stuff
- names for lemmas characterizing equality in things (`TypeOfHLevel≡`, note that `equivEq` has the wrong name... we should not use `Eq` for paths!)
- using Path vs ≡, equiv/Equiv vs ≃, etc. in names

Anything else that I should add? 

I don't think I want to think about guidelines for indentation and stuff (like https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md). But feel free to try to convince me.

Feedback is very welcome!

Should solve https://github.com/agda/cubical/issues/19 and https://github.com/agda/cubical/issues/403


mortberg:
Finally merging this now!

PR #457, opened 2020-10-19
Assorted proofs about booleans and related types

dolio:
This work contains a bunch of assorted stuff about booleans and types isomorphic to booleans.

Various stuff related to ⊕ is developed, because it's a convenient reflections of paths on booleans.

A module for the special case of finite types equivalent to `Bool` is added. There might be some overlap with the `RPn` modules, because they make use of this, but I didn't really explore that enough to know if it could be refactored or something.

A univalent inductive-recursive 'universe' is defined with one code that classifies `Bool`. However, due to the univalence constructor, this type is equivalent to the larger sub-universe above.

A higher inductive type is defined that should be equivalent to these previous two types. It's an explicit construction of Bℤ₂, so that's what I chose to call it. I proved that mappings can be define to and from the large binary universe, but didn't prove that they form an equivalence yet, because I think the proof would be pretty tedious. I might try some time, though.

I added various lemmas in some other places as I needed them as well.

Let me know if you'd like any revisions.


dolio:
Looks like I need to do something with some `Everything` modules or something?

mortberg:
You need to import `Cubical.Universe.Everything` in `Cubical.README`. 

If you're adding a Universe package maybe https://github.com/agda/cubical/blob/master/Cubical/Foundations/Univalence/Universe.agda should be moved there as well?

dolio:
Sorry about the delay. I fixed up the conflicts and whatnot.

I can move the universe univalence if you want. I initially thought the purpose might be a little different. I was thinking the `Cubical.Universe` namespace might be for the particular universes, similar to `Data` and `HITs`. But it isn't a very strong opinion.

mortberg:
> I can move the universe univalence if you want. I initially thought the purpose might be a little different. I was thinking the `Cubical.Universe` namespace might be for the particular universes, similar to `Data` and `HITs`. But it isn't a very strong opinion.

Right, I guess I think it can be used more generally for results and constructions of universes. I think the Data/HITs distinction is very artificial and should be removed in the future

dolio:
One could probably say that a distinction between universes and data/hits is kind of artificial as well. Would it be better to move my thing over into `Data` instead or something?

Maybe it'd be better to put everything in something like a `Construction` namespace (or a similarly generic name), then subcategorized by _what_ is being constructed, and then further subdivided by the way it's constructed or something. Similar to `Int`, except coalescing the one in `Data` with `HITs` even (and codata versions could fit into this as well when applicable).

mortberg:
Right, maybe the universes should go into Data as well. Feel free to move them there instead of making another package.

I'm not sure I like `Construction` as everything we do is "constructions" :-) The way I think about "Data" is as datatypes/datastructures. I think what should go in there is more the CS/Logic things in the library while the homotopy theoretic HITs should go in the Homotopy package.

dolio:
Okay, I moved the I-R universe version under `Cubical.Data.FinSet` as well.

PR #458, opened 2020-10-20
Add some lemmas and definitions for listed finite sets

jorge-jbs:
Added:

- Commutativity of `_++_`
- Idempotence of `_++_`
- `map`
- Cartesian product
- Disjoint union

Maybe the definitions should go in `Base`, but I saw that `_++_` is already in `Properties`. Should I move `_++_` and the new definitions to `Base`?

### DIFF
--- a/Cubical/HITs/ListedFiniteSet/Base.agda
+++ b/Cubical/HITs/ListedFiniteSet/Base.agda
@@ -3,31 +3,33 @@ module Cubical.HITs.ListedFiniteSet.Base where
 
 open import Cubical.Foundations.Prelude
 open import Cubical.Foundations.HLevels
+open import Cubical.Data.Empty as ⊥
+open import Cubical.Data.Sum as ⊎ using (_⊎_; inl; inr)
 
 open import Cubical.Functions.Logic
 
 private
   variable
-    A : Type₀
+    ℓ : Level
+    A B : Type ℓ
 
 infixr 20 _∷_
 -- infix 30 _∈_
 
-data LFSet (A : Type₀) : Type₀ where
+data LFSet (A : Type ℓ) : Type ℓ where
   []    : LFSet A
   _∷_   : (x : A) → (xs : LFSet A) → LFSet A
   dup   : ∀ x xs   → x ∷ x ∷ xs ≡ x ∷ xs
   comm  : ∀ x y xs → x ∷ y ∷ xs ≡ y ∷ x ∷ xs
   trunc : isSet (LFSet A)
 
-
 -- Membership.
 --
 -- Doing some proofs with equational reasoning adds an extra "_∙ refl"
 -- at the end.
 -- We might want to avoid it, or come up with a more clever equational reasoning.
-_∈_ : A → LFSet A → hProp _
-z ∈ []                  = ⊥
+_∈_ : {A : Type ℓ} → A → LFSet A → hProp ℓ
+z ∈ []                  = Lift ⊥.⊥ , isOfHLevelLift 1 isProp⊥
 z ∈ (y ∷ xs)            = (z ≡ₚ y) ⊔ (z ∈ xs)
 z ∈ dup x xs i          = proof i
   where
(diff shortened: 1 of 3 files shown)


mortberg:
Yes, please move them. Please also make everything universe polymorphic

jorge-jbs:
I've made everything universe polymorphic and moved everything except the cartesian product, since it depends on associativity and commutativity. Maybe the `Properties` and `Base` distinction doesn't make so much sense in HITs.

Saizan:
I think the point is that `.Base` should only have the type declaration with few exceptions, so it can be imported from modules that only need it in type signatures etc. without creating import cycles.

jorge-jbs:
I don't know why the CI build fails. I'm typechecking locally with Agda's master 🤔

Cubical/HITs/ListedFiniteSet/Base.agda
```diff
@@ -89,3 +91,27 @@ module PropElim {ℓ}
     (λ _ _ _ → isOfHLevel→isOfHLevelDep 1 trunc* _ _ _)
     (λ _ _ → isOfHLevel→isOfHLevelDep 1 trunc* _ _ _)
     λ xs → isProp→isSet (trunc* xs)
+
+_++_ : ∀ (xs ys : LFSet A) → LFSet A
```
mortberg, line 95: This should have some fixity, maybe same as for the corresponding list operation?

Cubical/HITs/ListedFiniteSet/Base.agda
```diff
@@ -89,3 +91,27 @@ module PropElim {ℓ}
...
+comm x y xs i       ++ ys = comm x y (xs ++ ys) i
+trunc xs zs p q i j ++ ys = trunc
+  (xs ++ ys) (zs ++ ys) (cong (_++ ys) p) (cong (_++ ys) q) i j
+
+map : (A → B) → LFSet A → LFSet B
+map f [] = []
+map f (x ∷ xs) = f x ∷ map f xs
+map f (dup x xs i) = dup (f x) (map f xs) i
+map f (comm x y xs i) = comm (f x) (f y) (map f xs) i
+map f (trunc xs ys p q i j) = trunc
```
mortberg, line 108: Indentation is weird here, why not put the `trunc` with all of its arguments on the next line instead of splitting it over 6 lines?

mortberg: Similar indentation occurs in the Properties file

mortberg (changes requested):
I had two small comments then I would be happy to merge

Cubical/HITs/ListedFiniteSet/Base.agda
jorge-jbs: In this case it was just because I find it more legible. In the properties file I exceed 100 characters if all the arguments go in the same line (I find 80 characters too long). I have joined some arguments in the same line though.

Agda needs a standard style so bad...

Cubical/HITs/ListedFiniteSet/Base.agda
jorge-jbs: List has `infixr 5 _++_` and that's what I have added.

Cubical/HITs/ListedFiniteSet/Base.agda
mortberg: Ok, it looks better now!

Cubical/HITs/ListedFiniteSet/Properties.agda
```diff
@@ -2,38 +2,103 @@
...
-  = dup x (assoc-++ xs ys zs j) i
-assoc-++ (comm x y xs i) ys zs j
-  = comm x y (assoc-++ xs ys zs j) i
-assoc-++ (trunc xs xs' p q i k) ys zs j
-  = trunc (assoc-++ xs ys zs j) (assoc-++ xs' ys zs j)
-          (cong (\ xs -> assoc-++ xs ys zs j) p) (cong (\ xs -> assoc-++ xs ys zs j) q) i k
+assoc-++ (x ∷ xs) ys zs = cong (x ∷_) (assoc-++ xs ys zs)
+assoc-++ (dup x xs i) ys zs j = dup x (assoc-++ xs ys zs j) i
+assoc-++ (comm x y xs i) ys zs j = comm x y (assoc-++ xs ys zs j) i
+assoc-++ (trunc xs xs' p q i k) ys zs j = trunc
```
mortberg, line 19: I think `trunc` should go on the next line with the arguments indented more. It's currently kind of weird and hard to read because the function is further to the right than its arguments.

The same applies to the calls to `PropElim.f` in the functions below

Cubical/HITs/ListedFiniteSet/Properties.agda
```diff
@@ -2,38 +2,103 @@
...
+  (assoc-++ xs ys zs j) (assoc-++ xs' ys zs j)
+  (cong (λ xs → assoc-++ xs ys zs j) p) (cong (λ xs → assoc-++ xs ys zs j) q)
+  i k
+
+comm-++-[] : ∀ (xs : LFSet A) → xs ++ [] ≡ [] ++ xs
+comm-++-[] xs = PropElim.f
```
mortberg, line 25: How many of these arguments do you have to give explicitly? It seems to me like some of them could be implicit. If they can always be implicit then `PropElim.f` should have more implicit arguments

mortberg:
Great!

PR #459, opened 2020-10-21
CommRing structure on the Integers

mchristianl:
this was mentioned in #400 

I've included the definition of a strict order `_<_` on `Cubical.Data.Int` which still is debated about in that issue https://github.com/agda/cubical/issues/404#issuecomment-709273429

### DIFF
--- a/Cubical/Data/Int/Properties.agda
+++ b/Cubical/Data/Int/Properties.agda
@@ -34,11 +34,22 @@ open import Cubical.Foundations.HLevels
 open import Cubical.Foundations.Isomorphism
 open import Cubical.Foundations.Univalence
 
-open import Cubical.Data.Empty
-open import Cubical.Data.Nat hiding (_+_ ; +-assoc ; +-comm)
-open import Cubical.Data.Bool
+open import Cubical.Data.Empty as ⊥
+open import Cubical.Data.Unit renaming (Unit to ⊤)
+open import Cubical.Data.Nat as ℕ using (ℕ; zero; suc; discreteℕ) renaming
+  ( _·_ to _·ⁿ_
+  ; _+_ to _+ⁿ_
+  ; _⊓_ to _⊓ⁿ_
+  ; _⊔_ to _⊔ⁿ_
+  )
+open import Cubical.Data.Nat.Order as ℕᵒ using () renaming ( _<_ to _<ⁿ_ )
+open import Cubical.Data.Bool hiding (_≟_)
 open import Cubical.Data.Sum
 open import Cubical.Data.Int.Base
+open import Cubical.Functions.Logic using (pathTo⇒; ⊔-comm)
+open import Cubical.HITs.PropositionalTruncation as PropTrunc using (propTruncIsProp)
+open import Cubical.Foundations.Function using (_∘_)
+open import Cubical.Reflection.Base using (_$_) -- TODO: add this to Foundation.Function
 
 open import Cubical.Relation.Nullary
 open import Cubical.Relation.Nullary.DecidableEq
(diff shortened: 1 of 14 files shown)


Cubical/Data/Int/Properties.agda
```diff
@@ -117,14 +169,74 @@ _+negsuc_ : Int → ℕ → Int
...
 m + negsuc n = m +negsuc n
 
+-- TODO: in `Cubical.Papers.RepresentationIndependence` we have
+--   -_ : ℤ → ℤ
+--   - x = 0 - x
+-- there might be an advantage of defining -_ by _-_ and not the other way around
+-- e.g. in Cubical.HITs.QuoInt we have
+--   -_ : ℤ → ℤ
+--   - signed s n = signed (not s) n
+--   - posneg i   = posneg (~ i)
```
mchristianl, line 188: just something that I noticed. The current definition worked out fine so far

Cubical/Data/Int/Properties.agda
```diff
@@ -117,14 +169,74 @@ _+negsuc_ : Int → ℕ → Int
...
+m - n = m + (- n)
+
+_·_ : Int → Int → Int
+x · y  = signed (sgn x ⊕ sgn y) (abs x ·ⁿ abs y)
+
+_<_ : ∀(x y : Int) → Type₀
+pos    x < pos    y = x <ⁿ y
+pos    x < negsuc y = ⊥
+negsuc x < pos    y = ⊤
+negsuc x < negsuc y = y <ⁿ x
```
mchristianl, line 205: definition in terms of `<` on `Cubical.Data.Nat`

Cubical/Data/Int/Properties.agda
```diff
@@ -221,6 +333,197 @@ ind-assoc _·_ f g p q base m n (suc o) =
...
+--        abs (signed (sgn m ⊕ sgn n) (suc mⁿ ·ⁿ suc nⁿ)) ·ⁿ 0 ∎
+-- ... | suc oⁿ | po = cong₂ signed γ₁ γ₂ where
+--   γ₁ = sgn m ⊕ sgn (signed (sgn n ⊕ sgn o) (suc nⁿ ·ⁿ suc oⁿ)) ≡[ i ]⟨ sgn m ⊕ sgn-signed (sgn n ⊕ sgn o) (oⁿ +ⁿ nⁿ ·ⁿ suc oⁿ) i ⟩
+--        sgn m ⊕ (sgn n ⊕ sgn o)                                 ≡⟨ ⊕-assoc (sgn m) (sgn n) (sgn o) ⟩
+--        (sgn m ⊕ sgn n) ⊕ sgn o                                 ≡[ i ]⟨ sgn-signed (sgn m ⊕ sgn n) (nⁿ +ⁿ mⁿ ·ⁿ suc nⁿ) (~ i) ⊕ sgn o ⟩
+--        sgn (signed (sgn m ⊕ sgn n) (suc mⁿ ·ⁿ suc nⁿ)) ⊕ sgn o ∎
+--   γ₂ = suc mⁿ ·ⁿ abs (signed (sgn n ⊕ sgn o) (suc nⁿ ·ⁿ suc oⁿ)) ≡[ i ]⟨ suc mⁿ ·ⁿ abs-signed (sgn n ⊕ sgn o) (suc nⁿ ·ⁿ suc oⁿ) i ⟩
+--        suc mⁿ ·ⁿ (suc nⁿ ·ⁿ suc oⁿ)                              ≡⟨ ℕ.·-assoc (suc mⁿ) (suc nⁿ) (suc oⁿ) ⟩
+--        (suc mⁿ ·ⁿ suc nⁿ) ·ⁿ suc oⁿ                              ≡[ i ]⟨ abs-signed (sgn m ⊕ sgn n) (suc mⁿ ·ⁿ suc nⁿ) (~ i) ·ⁿ suc oⁿ ⟩
+--        abs (signed (sgn m ⊕ sgn n) (suc mⁿ ·ⁿ suc nⁿ)) ·ⁿ suc oⁿ ∎
```
mchristianl, line 442: I think that there are some definitional equalities missing on `Cubical.Data.Int` to perform the same proof as in  `Cubical.HITs.QuoInt.Properties`. Therefore I adapted the proof from the non-cubical standard library (see below).

TODO: I will delete this comment before merge

Cubical/Data/Int/Properties.agda
mchristianl: This is not necessary for the commring-strucure, of course, but I would like to push some order-related properties if that is okay (?)

Cubical/Foundations/Prelude.agda
```diff
@@ -213,6 +213,22 @@ _∎ : (x : A) → x ≡ x
...
+_→⟨_⟩_ : ∀{ℓ ℓ' ℓ''} {Q : Type ℓ'} {R : Type ℓ''} → (P : Type ℓ) → (P → Q) → (Q → R) → (P → R)
+_ →⟨ p→q ⟩ q→r = λ x → q→r (p→q x)
+
+_→⟨⟩_ : ∀{ℓ ℓ'} {Q : Type ℓ'} → (P : Type ℓ) → (P → Q) → (P → Q)
+_ →⟨⟩ p→q = p→q
+
+_◼ : ∀{ℓ} (A : Type ℓ) → A → A
+_ ◼ = λ x → x
+
+-- END
```
mchristianl, line 230: I would like to use this in the proof of `+-reflects-<ʳ` in `Cubical.Data.Int.Properties` but I found it useful for developing proofs in a little bit more verbose way.

Might this be something to add to the standard library? Otherwise I would remove it and collapse the proof.

mchristianl (commented):
a few remarks

mortberg:
I'm confused, wasn't the plan to transport the CommRing structure from BiInvInt to Int? It seems like you define it directly here

mchristianl:
I did that later-on but you are right - some of the proofs can be skipped this way. That would also be better for the libraries overall typechecking performance.

I'll put this in a bigger PR then with other Integer equivalences once that is ready.

Sorry for the confusion :hushed:

mortberg:
> I did that later-on but you are right - some of the proofs can be skipped this way. That would also be better for the libraries overall typechecking performance.
> 
> I'll put this in a bigger PR then with other Integer equivalences once that is ready.
> 
> Sorry for the confusion hushed

Sounds good! The whole point of this is indeed that we can skip a lot of proof for `Int` as we've done them for `BiInvInt`. What I really want is a PR with only the result that `Int` is `CommRing` transported from `BiInvInt`. Other things like the order relations can be added in another PR. 

The reason I want thing split like this is that I want to see exactly what goes into doing this so that we can decide if more lemmas are needed to make it easier and to discuss where to put things. I think it's a bit hard to find the current `CommRing` instance for `BiInvInt`, so it would be good to discuss where to put the instances.

PR #463, opened 2020-10-24
Towards displayed univalent structures

mortberg:
This is closely related to the DURGs of https://github.com/agda/cubical/pull/435, but the definitions are closer to what is currently used by the SIP in the library. I don't develop any theory about displayed univalent structures, but everything in the library still typechecks with the changes in this PR.

An important mathematical difference is that `TypeWithUARel` is weaker than `URGStr`. A way to make them equivalent would be to turn the `uaR` map into an equivalence, but I'm very curious if that's really necessary? Maybe some of the closure properties won't work for the weaker definition? @Schippmunk @UlrikBuchholtz

### DIFF
--- a/Cubical/Foundations/SIP.agda
+++ b/Cubical/Foundations/SIP.agda
@@ -39,13 +39,32 @@ _≃[_]_ : (A : TypeWithStr ℓ₁ S) (ι : StrEquiv S ℓ₂) (B : TypeWithStr 
 A ≃[ ι ] B = Σ[ e ∈ typ A ≃ typ B ] (ι A B e)
 
 
+record TypeWithUARel (A : Type ℓ) (ℓ' : Level) : Type (ℓ-max ℓ (ℓ-suc ℓ')) where
+  no-eta-equality
+  constructor typewithuarel
+  field
+    _≅_ : A → A → Type ℓ'
+    uaR : {a a' : A} → (a ≅ a') → (a ≡ a')
+    -- TODO: do we need more assumptions here? Maybe uaR should be an equiv?
 
--- The following PathP version of SNS is a bit easier to work with
--- for the proof of the SIP
+open TypeWithUARel
+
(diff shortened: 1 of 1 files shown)


mortberg:
Interesting! Thanks for investigating this further. What does @ecavallo think?

ecavallo:
I agree about the weaker version not really working. I'd suspect that the reflexivity witness would be less important if we did more "natively cubical" (as opposed to "HoTT-like") proofs. Maybe I'll take a stab at it on my own and see how it compares.

ecavallo:
Okay, I tried also changing the definition of DURG to give a correspondence with dependent paths:
```
  field
    _≅ᴰ⟨_⟩_ : {a a' : A} → B a → a ≅ a' → B a' → Type ℓ≅B
    uaᴰ : {a a' : A} (b : B a) (p : a ≅ a') (b' : B a') → (b ≅ᴰ⟨ p ⟩ b') ≃ PathP (λ i → B (≅→≡ p i)) b b'
```
With that definition I was able to prove the closure properties in that `Properties` file fairly easily (using some lemmas that we already have from the corresponding SIP code): https://github.com/ecavallo/cubical/blob/durg/Cubical/Displayed/Properties.agda

UlrikBuchholtz:
> Okay, I tried also changing the definition of DURG to give a correspondence with dependent paths: […]
> With that definition I was able to prove the closure properties in that `Properties` file fairly easily (using some lemmas that we already have from the corresponding SIP code): https://github.com/ecavallo/cubical/blob/durg/Cubical/Displayed/Properties.agda

Yes, that's nice. I hadn't realized that much of the structures part had shifted from SNS to UnivalentStr, which is a special case of this. So that's clearly the way to go, then, and should be compatible with the AutoStructure stuff, now?

ecavallo:
Yes, it should be fairly simple to adapt AutoStructure. I think I'll want to change the record automation around a bit; I think it's possible to do a cleaner and more general version with DURGs, and I might also try to factor out some more general macro for converting between records and iterated sigma-types.

ecavallo:
Superseded by #502.

PR #467, opened 2020-10-27
Major speedup in ZCohomology.Groups.Wedge

cangiuli:
I was tinkering with the ZCohomology files to figure out what was so slow, and I found that the Wedge file is **~10x faster** with the `--experimental-lossy-unification` flag, which "enables heuristically unifying `f es = f es'` by unifying `es = es'`, even when it could lose solutions." (The other change in this patch is needed to convince the lossy unifier to accept a definition.)

```
Cubical.ZCohomology.Groups.Wedge                  1,574ms 
```

You may want to consult with @Saizan before using this; I'm not sure if the flag is long for this world.

And no, the other ZCohomology files don't get faster with this flag. /cc @aljungstrom

### DIFF
--- a/Cubical/ZCohomology/Groups/Wedge.agda
+++ b/Cubical/ZCohomology/Groups/Wedge.agda
@@ -1,4 +1,4 @@
-{-# OPTIONS --cubical --no-import-sorts --safe #-}
+{-# OPTIONS --cubical --no-import-sorts --safe --experimental-lossy-unification #-}
 module Cubical.ZCohomology.Groups.Wedge where
 
 open import Cubical.ZCohomology.Base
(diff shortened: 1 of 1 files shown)


mortberg:
Wow!! I vote for merging unless this flag is too experimental. It is compatible with `--safe` though so can't be too bad? @Saizan

Saizan:
Regarding safety the only "danger" is that you might get a meta solved even if there are multiple distinct solutions.

Great to see it's useful! It's good motivation for improving the implementation so we don't have to change things like `1 + n` to `suc n` when we use the flag.

@cangiuli do you have an idea of what was slow to unify without the flag?

aljungstrom:
Wow, this is great!

cangiuli:
@Saizan In the following line, it takes a long time to see that the type of `(isContr→isProp (isContrHⁿ-Unit 0) _ _)` agrees with the domain of the function `I.Ker-Δ⊂Im-i` (namely, `isInKer ...`), and also a long time to see that the codomain of the function (namely, `isInIm ...`) agrees with the type of the declaration.

https://github.com/agda/cubical/blob/e1edc4f07977f0b66216e34876f92398f8d92726/Cubical/ZCohomology/Groups/Wedge.agda#L55

Saizan:
@cangiuli thanks! Might be related to comparing systems, but could also be reductions to get us there.

mortberg:
Interesting! Looking at the comments in the issue on agda/agda I see that @tomdjong wrote:

"As in @WolframKahl's original problem, I developed some abstract stuff and then I instantiate it with a particular concrete (and relatively complicated) example/object. The types then do not match on the nose and Agda starts normalising like crazy, reducing the complicated example/object all the way down, which is slow."

The Wedge file which also becomes faster with the flag is another instance of exactly this. @aljungstrom has developed general result about Z-cohomology which get instantiated with concrete types to compute some cohomology groups of wedges of spheres. The addition operation in the cohomology groups is very complex and I bet that Agda has to work very hard when unifying concrete expressions involving it.

PR #471, opened 2020-11-02
Generalize rec,elim for truncation to 0

ecavallo:
I needed a `∣_∣` that worked for 0 and couldn't think of a name so it's `∣_∣ₕ` (h for hlevel?) but please suggest something better.

My laptop panicked somewhere in `ZCohomology`, so I adjusted to continue using the less general (>0) versions there.


mortberg:
Looks good to me!

@aljungstrom Could review this PR for me?

aljungstrom:
Looks good to me too. I was actually planning on doing this myself, but I've been lazy hehe. My plan was to make your `∣_∣ₕ` the default one (so we just call it `∣_∣` and rename the constructor of the HIT).  Hopefully your laptop won't panic when I'm done fixing ZCohomology. Then I can remove the `rec₊`, `elim₊` , etc and rename `∣_∣ₕ` . So maybe keep `∣_∣ₕ` as it is is for now? (I can't come up with a better name either...)

mortberg:
I can merge once the conflicts are fixed

ecavallo:
They are fixed!

mortberg:
Thanks!

PR #472, opened 2020-11-02
Mayer-Vietoris cofiber sequence

ecavallo:
Translating a result from my master's thesis into cubical type theory. I don't have any endgame for this, I just wanted to try it.

### DIFF
--- a/Cubical/Homotopy/MayerVietorisCofiber.agda
+++ b/Cubical/Homotopy/MayerVietorisCofiber.agda
@@ -1,0 +1,174 @@
+{-
+
+ Mayer-Vietoris cofiber sequence:
+
+  Let X be a pointed type, and let a span B ←[f]- X -[g]→ C be given.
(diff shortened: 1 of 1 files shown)


mortberg:
How does the complexity of the proofs differ from the HoTT proofs?

@aljungstrom : is this something you could have use for?

aljungstrom:
> @aljungstrom : is this something you could have use for?
Ah, yes, possibly!

ecavallo:
> How does the complexity of the proofs differ from the HoTT proofs?

The HoTT-Agda proof is here: https://github.com/HoTT/HoTT-Agda/blob/master/theorems/homotopy/CofiberSequence.agda
Even in HoTT I was using cubical-style box-filling lemmas, so it's not _that_ different.

mortberg:
Merging this now, thanks for the contribution!

PR #478, opened 2020-11-20
Make the ring solver better

felixwellen:
The ring solver from #417 sometimes fails to produce terms that agda can unify. This happens when the almost ring is abstract (e.g. a general CommRing). To remedy this problem, this PR introduces Expressions with constants from a given base-RawRing. The idea is to use Nat or Integer, so agda can unify things like "0 * 1" and "0".


felixwellen:
Works now with everything I tried (which is not too much). I duplicated the existing solver and specialized to commutative rings. I do not see, how the two solvers could now share their code and I guess the old one should be kept to deal with natural numbers.

Cubical/Algebra/CommRing/Localisation/Base.agda
```diff
@@ -76,27 +77,29 @@ module Loc (R' : CommRing {ℓ}) (S' : ℙ (R' .fst)) (SMultClosedSubset : isMul
...
  locTrans : isTrans _≈_
  locTrans (r , s , s∈S') (r' , s' , s'∈S') (r'' , s'' , s''∈S') ((u , u∈S') , p) ((v , v∈S') , q) =
    ((u · v · s') , SMultClosedSubset .multClosed (SMultClosedSubset .multClosed u∈S' v∈S') s'∈S')
    , path
   where
   path : u · v · s' · r · s'' ≡ u · v · s' · r'' · s
-  path = u · v · s' · r · s''   ≡⟨ cong (_· s'') (·-commAssocr _ _ _) ⟩
-         u · v · r · s' · s''   ≡⟨ cong (λ x → x · s' · s'') (·-commAssocr _ _ _) ⟩
-         u · r · v · s' · s''   ≡⟨ cong (_· s'') (·-commAssocr _ _ _) ⟩
+  path = u · v · s' · r · s''   ≡⟨ solve R' (X4 ·' X5 ·' X1 ·' X3 ·' X2)
```
mortberg, line 87: Here's an example how to use the solver @mzeuner 

This would of course be nicer with some reflection, but shortening these proofs using the explicit solver like this is still nice!

Cubical/Algebra/RingSolver/CommRingSolver.agda
```diff
@@ -0,0 +1,221 @@
...
+  X3 : ℤExpr R 5
+  X3 = ∣ (suc (suc Fin.zero))
+
+  X4 : ℤExpr R 5
+  X4 = ∣ (suc (suc (suc Fin.zero)))
+
+  X5 : ℤExpr R 5
+  X5 = ∣ (suc (suc (suc (suc Fin.zero))))
+
+module VarNames6 (R : CommRing {ℓ}) where
```
mortberg, line 204: Couldn't we just have this parametrized by a natural number `n` and have `X_ : N -> ZExpr R n`? (defined by `X_ x = | x % n` where `x % n` is x modulo n viewed as an element of Fin n

mortberg (changes requested):
Looks great to me! Just one small suggestion for how to not hard-code the variables

Cubical/Algebra/RingSolver/CommRingSolver.agda
mortberg: PS: I don't think the `%` operation is defined anywhere, but it should be too hard to add (I hope)

Cubical/Algebra/CommRing/Localisation/Base.agda
mzeuner: Ah nice, it will be interesting to see how much we can simplify in that file.

Cubical/Algebra/RingSolver/CommRingSolver.agda
mortberg: Maybe I should merge this as it is and we make an issue for @felixwellen ? Would be nice if @mzeuner could experiment a bit with simplifying more proofs about localization

felixwellen: Yes. 
I tried what you suggested before making the definitions, but annoyingly and contrary to my intuition, the 'n' in 'Fin n' could not be left implicit. But I could try to figure out why it didn't work or obsolete the issue by building a reflection interface.

felixwellen: Ah, wait. I didn't use modulo arithmetic when I tried. But in any case, I'm for putting it in a separate issue.

mortberg: Yeah, I don't know if the modulo approach is a good idea. It's probably just the old Coq programmer inside me trying to avoid using dependent types... Anyway I'll merge this now and as you say it's probably not worth having a nicer solution as this will be removed once we have some reflection.

PR #479, opened 2020-12-10
Klein

aljungstrom:
This PR contains:
- characterisations of the cohomology groups of Klein bottles
- A proof that any group equivalent to Bool as types is also isomorphic to Bool as groups
- A proof cohomology is the same as reduced cohomology (as types). I still have been a bit too lazy to define the group structure on reduced cohomology groups (this should be easy with the new addition). Using this, we could define it by a transport but I'm not sure how wise this would be.
- characterisations of of the cohomology groups of the real projective plane (using a slightly cuter definition than the general one in HITs.RP). All of this is given for free by the characterisation of the cohomology groups of Klein bottles, but I thought I'd throw it in anyways.
- A fix of the proof in Wedge.agda that makes cohomology groups of wedge sums compute slightly better.


mortberg:
Why would it be unwise to define the group structure using transport?

How hard would it be to prove that your RP^2 is the same as the RP^n instantiated with n=2?

There are some long lines and local lemmas called "cheater" or "lazy". I would rename these to "helper" or "rem/remark" just so that no one thinks that we are actually cheating :-)

mortberg:
Great! I'll merge this as soon as Travis is done. Please remind me if I forget

aljungstrom:
> Great! I'll merge this as soon as Travis is done. Please remind me if I forget

done:-)

PR #482, opened 2020-12-17
Def of reduced cohomology without InducedAbGroupPath

aljungstrom:
I am able to define reduced cohomology directly via InducedAbGroup and now do the previous hack with defining it via InducedAbGroupPath. However, is still have to have coHomGroup≡coHomRedGroup'  and coHomGroup≡coHomRedGroup separate. If I remove coHomGroup≡coHomRedGroup' and write the exact same stuff in the definition of coHomGroup≡coHomRedGroup, it takes ages to type check.


Saizan:
Does `coHomRedGroup` itself get slow to typecheck or later uses?

mortberg:
We figured it out! The solution was to add  `--experimental-lossy-unification`  @Saizan

mortberg:
Without it I think Agda tried to unfold `InducedAbGroup` when trying to fill a goal with `InducedAbGroupPath`. This is a terrible idea and I guess that adding the flag stopped Agda from doing this

aljungstrom:
> Does `coHomRedGroup` itself get slow to typecheck or later uses?

I'm impressed that you managed to understand the description of this PR, because I can barely do so myself...

To answer your question: it actually type checks fast in later uses!

Cubical/ZCohomology/Properties.agda
```diff
@@ -233,23 +233,22 @@ coHomGr≅coHomRedGr : ∀ {ℓ} (n : ℕ) (A : Pointed ℓ)
...
+  InducedAbGroup (coHomGroup (suc n) (typ A))
+                 (coHomRed (suc n) A , _+ₕ∙_)
+                 (isoToEquiv (invIso (Iso-coHom-coHomRed n)))
+                 (homhelp n A)
 
 abstract
   coHomGroup≡coHomRedGroup : ∀ {ℓ} (n : ℕ) (A : Pointed ℓ)
                           → coHomRedGroup (suc n) A ≡ coHomGroup (suc n) (typ A)
-  coHomGroup≡coHomRedGroup = coHomGroup≡coHomRedGroup'
+  coHomGroup≡coHomRedGroup n A =
```
mortberg, line 247: Wait a minute, the name of this lemma is not the same order as in the type. Please swap the type and remove the `sym`

mortberg: ping @aljungstrom

aljungstrom: wups, forgot about this. It's fixed now.

PR #483, opened 2020-12-17
/Categories cleanup

dan-iel-lee:
This commit is purely renaming and refactoring:
- new syntax for hom-sets ( C [ x , y ] )
- \* as sequencing
- \circ as composition
- rename cursive category letters to normal C and D
- camel case where it makes sense (didn't change things like F-hom since I think that looks better than FHom because of capitalization of 'F')
- more level polymorphism in `Core`
- renamed a couple results in `Presheaves` to better reflect their conclusion

### DIFF
--- a/Cubical/Categories/Category.agda
+++ b/Cubical/Categories/Category.agda
@@ -36,23 +36,38 @@ record Precategory ℓ ℓ' : Type (ℓ-suc (ℓ-max ℓ ℓ')) where
     ob : Type ℓ
     Hom[_,_] : ob → ob → Type ℓ'
     id : ∀ x → Hom[ x , x ]
-    _◾_ : ∀ {x y z} (f : Hom[ x , y ]) (g : Hom[ y , z ]) → Hom[ x , z ]
-    seq-λ : ∀ {x y : ob} (f : Hom[ x , y ]) → (id x) ◾ f ≡ f
-    seq-ρ : ∀ {x y} (f : Hom[ x , y ]) → f ◾ (id y) ≡ f
-    seq-α : ∀ {u v w x} (f : Hom[ u , v ]) (g : Hom[ v , w ]) (h : Hom[ w , x ]) → (f ◾ g) ◾ h ≡ f ◾ (g ◾ h)
+    _⋆_ : ∀ {x y z} (f : Hom[ x , y ]) (g : Hom[ y , z ]) → Hom[ x , z ]
+    seq-λ : ∀ {x y : ob} (f : Hom[ x , y ]) → (id x) ⋆ f ≡ f
+    seq-ρ : ∀ {x y} (f : Hom[ x , y ]) → f ⋆ (id y) ≡ f
+    seq-α : ∀ {u v w x} (f : Hom[ u , v ]) (g : Hom[ v , w ]) (h : Hom[ w , x ]) → (f ⋆ g) ⋆ h ≡ f ⋆ (g ⋆ h)
+
+  -- composition: alternative to diagramatic order
+  _∘_ : ∀ {x y z} (g : Hom[ y , z ]) (f : Hom[ x , y ]) → Hom[ x , z ]
+  g ∘ f = f ⋆ g
 
 open Precategory public
 
--- helpful syntax
+
+-- Helpful syntax/notation
+
 _[_,_] : (C : Precategory ℓ ℓ') → (x y : C .ob) → Type ℓ'
 _[_,_] = Hom[_,_]
 
 -- needed to define this in order to be able to make the subsequence syntax declaration
-seq : ∀ (C : Precategory ℓ ℓ') {x y z} (f : C [ x , y ]) (g : C [ y , z ]) → C [ x , z ]
-seq = _◾_
+seq' : ∀ (C : Precategory ℓ ℓ') {x y z} (f : C [ x , y ]) (g : C [ y , z ]) → C [ x , z ]
+seq' = _⋆_
 
-infix 40 seq
-syntax seq C f g = f ◾⟨ C ⟩ g
+infix 15 seq'
+syntax seq' C f g = f ⋆⟨ C ⟩ g
+
+-- composition
+comp' : ∀ (C : Precategory ℓ ℓ') {x y z} (g : C [ y , z ]) (f : C [ x , y ]) → C [ x , z ]
+comp' = _∘_
+
+infix 16 comp'
+syntax comp' C g f = g ∘⟨ C ⟩ f
+
+
 
 -- Categories
 
@@ -68,17 +83,17 @@ record CatIso {C : Precategory ℓ ℓ'} (x y : C .Precategory.ob) : Type ℓ' w
   field
     h : C [ x , y ]
     h⁻¹ : C [ y , x ]
-    sec : (h⁻¹ ◾⟨ C ⟩ h) ≡ C .id y
-    ret : h ◾⟨ C ⟩ h⁻¹ ≡ C .id x
+    sec : (h⁻¹ ⋆⟨ C ⟩ h) ≡ C .id y
+    ret : h ⋆⟨ C ⟩ h⁻¹ ≡ C .id x
 
-pathToIso : {ℓ ℓ' : Level} {C : Precategory ℓ ℓ'} (x y : C .ob) (p : x ≡ y) → CatIso {C = C} x y
+pathToIso : {C : Precategory ℓ ℓ'} (x y : C .ob) (p : x ≡ y) → CatIso {C = C} x y
 pathToIso {C = C} x y p = J (λ z _ → CatIso x z) (catiso (C .id x) idx (C .seq-λ idx) (C .seq-λ idx)) p
   where
     idx = C .id x
 
 -- Univalent Categories
 
-record isUnivalent {ℓ ℓ'} (C : Precategory ℓ ℓ') : Type (ℓ-max ℓ ℓ') where
+record isUnivalent (C : Precategory ℓ ℓ') : Type (ℓ-max ℓ ℓ') where
   field
     univ : (x y : C .ob) → isEquiv (pathToIso {C = C} x y)
 
@@ -86,11 +101,11 @@ open isUnivalent public
 
 -- Opposite Categories
 
-_^op : ∀ {ℓ ℓ'} → Precategory ℓ ℓ' → Precategory ℓ ℓ'
+_^op : Precategory ℓ ℓ' → Precategory ℓ ℓ'
 (C ^op) .ob = C .ob
 (C ^op) .Hom[_,_] x y = C .Hom[_,_] y x
 (C ^op) .id = C .id
-(C ^op) ._◾_ f g = C ._◾_ g f
+(C ^op) ._⋆_ f g = C ._⋆_ g f
 (C ^op) .seq-λ = C .seq-ρ
 (C ^op) .seq-ρ = C .seq-λ
 (C ^op) .seq-α f g h = sym (C .seq-α _ _ _)
--- a/Cubical/Categories/Category.agda
+++ b/Cubical/Categories/Category.agda
@@ -45,7 +45,7 @@ record Precategory ℓ ℓ' : Type (ℓ-suc (ℓ-max ℓ ℓ')) where
   _∘_ : ∀ {x y z} (g : Hom[ y , z ]) (f : Hom[ x , y ]) → Hom[ x , z ]
   g ∘ f = f ⋆ g
 
-open Precategory public
+open Precategory
 
 
 -- Helpful syntax/notation
--- a/Cubical/Categories/Category.agda
+++ b/Cubical/Categories/Category.agda
@@ -24,59 +24,75 @@ module Cubical.Categories.Category where
 open import Cubical.Core.Glue
 open import Cubical.Foundations.Prelude
 
+private
+  variable
+    ℓ ℓ' : Level
+
 -- Precategories
 
 record Precategory ℓ ℓ' : Type (ℓ-suc (ℓ-max ℓ ℓ')) where
   no-eta-equality
   field
     ob : Type ℓ
-    hom : ob → ob → Type ℓ'
-    idn : ∀ x → hom x x
-    seq : ∀ {x y z} (f : hom x y) (g : hom y z) → hom x z
-    seq-λ : ∀ {x y : ob} (f : hom x y) → seq (idn x) f ≡ f
-    seq-ρ : ∀ {x y} (f : hom x y) → seq f (idn y) ≡ f
-    seq-α : ∀ {u v w x} (f : hom u v) (g : hom v w) (h : hom w x) → seq (seq f g) h ≡ seq f (seq g h)
+    Hom[_,_] : ob → ob → Type ℓ'
+    id : ∀ x → Hom[ x , x ]
+    _◾_ : ∀ {x y z} (f : Hom[ x , y ]) (g : Hom[ y , z ]) → Hom[ x , z ]
+    seq-λ : ∀ {x y : ob} (f : Hom[ x , y ]) → (id x) ◾ f ≡ f
+    seq-ρ : ∀ {x y} (f : Hom[ x , y ]) → f ◾ (id y) ≡ f
+    seq-α : ∀ {u v w x} (f : Hom[ u , v ]) (g : Hom[ v , w ]) (h : Hom[ w , x ]) → (f ◾ g) ◾ h ≡ f ◾ (g ◾ h)
 
 open Precategory public
 
+-- helpful syntax
+_[_,_] : (C : Precategory ℓ ℓ') → (x y : C .ob) → Type ℓ'
+_[_,_] = Hom[_,_]
+
+-- needed to define this in order to be able to make the subsequence syntax declaration
+seq : ∀ (C : Precategory ℓ ℓ') {x y z} (f : C [ x , y ]) (g : C [ y , z ]) → C [ x , z ]
+seq = _◾_
+
+infix 40 seq
+syntax seq C f g = f ◾⟨ C ⟩ g
+
 -- Categories
 
-record isCategory {ℓ ℓ'} (𝒞 : Precategory ℓ ℓ') : Type (ℓ-max ℓ ℓ') where
+record isCategory (C : Precategory ℓ ℓ') : Type (ℓ-max ℓ ℓ') where
   field
-    homIsSet : ∀ {x y} → isSet (𝒞 .hom x y)
+    homIsSet : ∀ {x y} → isSet (C [ x , y ])
 
-open isCategory public
 
 -- Isomorphisms and paths in precategories
 
-record CatIso {ℓ ℓ' : Level} {𝒞 : Precategory ℓ ℓ'} (x y : 𝒞 .ob) : Type ℓ' where
+record CatIso {C : Precategory ℓ ℓ'} (x y : C .Precategory.ob) : Type ℓ' where
   constructor catiso
   field
-    h : 𝒞 .hom x y
-    h⁻¹ : 𝒞 .hom y x
-    sec : 𝒞 .seq h⁻¹ h ≡ 𝒞 .idn y
-    ret : 𝒞 .seq h h⁻¹ ≡ 𝒞 .idn x
+    h : C [ x , y ]
+    h⁻¹ : C [ y , x ]
+    sec : (h⁻¹ ◾⟨ C ⟩ h) ≡ C .id y
+    ret : h ◾⟨ C ⟩ h⁻¹ ≡ C .id x
 
-pathToIso : {ℓ ℓ' : Level} {𝒞 : Precategory ℓ ℓ'} (x y : 𝒞 .ob) (p : x ≡ y) → CatIso {𝒞 = 𝒞} x y
-pathToIso {𝒞 = 𝒞} x y p = J (λ z _ → CatIso x z) (catiso (𝒞 .idn x) idx (𝒞 .seq-λ idx) (𝒞 .seq-λ idx)) p
+pathToIso : {ℓ ℓ' : Level} {C : Precategory ℓ ℓ'} (x y : C .ob) (p : x ≡ y) → CatIso {C = C} x y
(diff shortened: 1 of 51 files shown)


Cubical/Categories/Category.agda
```diff
@@ -24,59 +24,90 @@ module Cubical.Categories.Category where
...
-    hom : ob → ob → Type ℓ'
-    idn : ∀ x → hom x x
-    seq : ∀ {x y z} (f : hom x y) (g : hom y z) → hom x z
-    seq-λ : ∀ {x y : ob} (f : hom x y) → seq (idn x) f ≡ f
-    seq-ρ : ∀ {x y} (f : hom x y) → seq f (idn y) ≡ f
-    seq-α : ∀ {u v w x} (f : hom u v) (g : hom v w) (h : hom w x) → seq (seq f g) h ≡ seq f (seq g h)
+    Hom[_,_] : ob → ob → Type ℓ'
+    id : ∀ x → Hom[ x , x ]
+    _⋆_ : ∀ {x y z} (f : Hom[ x , y ]) (g : Hom[ y , z ]) → Hom[ x , z ]
+    seq-λ : ∀ {x y : ob} (f : Hom[ x , y ]) → (id x) ⋆ f ≡ f
```
mortberg, line 40: I would change these names as well. How about `⋆IdL`, `⋆IdR` and `⋆Assoc`? 

This should be in line with what we are doing in Algebra. Or what does @felixwellen think?

Cubical/Categories/Category.agda
```diff
@@ -24,59 +24,90 @@ module Cubical.Categories.Category where
...
 
 open Precategory public
 
+
+-- Helpful syntax/notation
+
+_[_,_] : (C : Precategory ℓ ℓ') → (x y : C .ob) → Type ℓ'
+_[_,_] = Hom[_,_]
+
+-- needed to define this in order to be able to make the subsequence syntax declaration
```
mortberg, line 56: Could one make these private?

Cubical/Categories/Category.agda
```diff
@@ -24,59 +24,90 @@ module Cubical.Categories.Category where
...
+
+
+
 -- Categories
 
-record isCategory {ℓ ℓ'} (𝒞 : Precategory ℓ ℓ') : Type (ℓ-max ℓ ℓ') where
+record isCategory (C : Precategory ℓ ℓ') : Type (ℓ-max ℓ ℓ') where
   field
-    homIsSet : ∀ {x y} → isSet (𝒞 .hom x y)
+    homIsSet : ∀ {x y} → isSet (C [ x , y ])
```
mortberg, line 76: This should be `isSetHom`

Cubical/Categories/Category.agda
```diff
@@ -24,59 +24,90 @@ module Cubical.Categories.Category where
...
+record CatIso {C : Precategory ℓ ℓ'} (x y : C .Precategory.ob) : Type ℓ' where
   constructor catiso
   field
-    h : 𝒞 .hom x y
-    h⁻¹ : 𝒞 .hom y x
-    sec : 𝒞 .seq h⁻¹ h ≡ 𝒞 .idn y
-    ret : 𝒞 .seq h h⁻¹ ≡ 𝒞 .idn x
+    h : C [ x , y ]
+    h⁻¹ : C [ y , x ]
+    sec : (h⁻¹ ⋆⟨ C ⟩ h) ≡ C .id y
```
mortberg, line 86: Why do you need parentheses here? Maybe something should bind differently?

Cubical/Categories/Category.agda
```diff
@@ -24,59 +24,90 @@ module Cubical.Categories.Category where
...
 
-record CatIso {ℓ ℓ' : Level} {𝒞 : Precategory ℓ ℓ'} (x y : 𝒞 .ob) : Type ℓ' where
+record CatIso {C : Precategory ℓ ℓ'} (x y : C .Precategory.ob) : Type ℓ' where
   constructor catiso
   field
-    h : 𝒞 .hom x y
-    h⁻¹ : 𝒞 .hom y x
-    sec : 𝒞 .seq h⁻¹ h ≡ 𝒞 .idn y
-    ret : 𝒞 .seq h h⁻¹ ≡ 𝒞 .idn x
+    h : C [ x , y ]
```
mortberg, line 84: It's probably not good to call this `h`. How about `mor` and `inv` instead? 

I think `sec` and `ret` could be `leftInv` and `rightInv` as well to match https://github.com/agda/cubical/blob/master/Cubical/Foundations/Isomorphism.agda#L35

Cubical/Categories/Functor.agda
```diff
@@ -8,17 +8,18 @@ open import Cubical.Categories.Category
...
+record Functor (C : Precategory ℓC ℓC') (D : Precategory ℓD ℓD') : Type (ℓ-max (ℓ-max ℓC ℓC') (ℓ-max ℓD ℓD')) where
   no-eta-equality
   open Precategory
 
   field
-    F-ob : 𝒞 .ob → 𝒟 .ob
-    F-hom : {x y : 𝒞 .ob} → 𝒞 .hom x y → 𝒟 .hom (F-ob x) (F-ob y)
-    F-idn : {x : 𝒞 .ob} → F-hom (𝒞 .idn x) ≡ 𝒟 .idn (F-ob x)
-    F-seq : {x y z : 𝒞 .ob} (f : 𝒞 .hom x y) (g : 𝒞 .hom y z) → F-hom (𝒞 .seq f g) ≡ 𝒟 .seq (F-hom f) (F-hom g)
+    F-ob : C .ob → D .ob
```
mortberg, line 18: How about calling these `fmapOb`, `fmapHom`, `fmapId` and `fmap⋆`?

Cubical/Categories/NaturalTransformation.agda
```diff
@@ -8,72 +8,79 @@ open import Cubical.Categories.Functor
...
+
+  record NatTrans (F G : Functor C D) : Type (ℓ-max (ℓ-max ℓC ℓC') (ℓ-max ℓD ℓD')) where
     open Precategory
     open Functor
 
     field
-      N-ob : (x : 𝒞 .ob) → 𝒟 .hom (F .F-ob x) (G .F-ob x)
-      N-hom : {x y : 𝒞 .ob} (f : 𝒞 .hom x y) → 𝒟 .seq (F .F-hom f) (N-ob y) ≡ 𝒟 .seq (N-ob x) (G .F-hom f)
+      -- components of the natural transformation
+      N-ob : (x : C .ob) → D [(F .F-ob x) , (G .F-ob x)]
```
mortberg, line 25: How about calling these `transf` and `natural`?

Cubical/Categories/NaturalTransformation.agda
```diff
@@ -8,72 +8,79 @@ open import Cubical.Categories.Functor
...
 
-  FUNCTOR : Precategory (ℓ-max (ℓ-max ℓ𝒞 ℓ𝒞') (ℓ-max ℓ𝒟 ℓ𝒟')) (ℓ-max (ℓ-max ℓ𝒞 ℓ𝒞') (ℓ-max ℓ𝒟 ℓ𝒟'))
-  FUNCTOR .ob = Functor 𝒞 𝒟
-  FUNCTOR .hom = NatTrans
-  FUNCTOR .idn = id-trans
-  FUNCTOR .seq = seq-trans
-  FUNCTOR .seq-λ α = make-nat-trans-path λ i x → 𝒟 .seq-λ (α .N-ob x) i
-  FUNCTOR .seq-ρ α = make-nat-trans-path λ i x → 𝒟 .seq-ρ (α .N-ob x) i
-  FUNCTOR .seq-α α β γ = make-nat-trans-path λ i x → 𝒟 .seq-α (α .N-ob x) (β .N-ob x) (γ .N-ob x) i
+  FUNCTOR : Precategory (ℓ-max (ℓ-max ℓC ℓC') (ℓ-max ℓD ℓD')) (ℓ-max (ℓ-max ℓC ℓC') (ℓ-max ℓD ℓD'))
```
mortberg, line 79: Maybe we should call this `FunctorPrecat`?

Cubical/Categories/Presheaves.agda
```diff
@@ -14,63 +14,63 @@ open import Cubical.Categories.Sets
 
 module _ (ℓ ℓ' : Level) where
   PSH : Precategory ℓ ℓ' → Precategory (ℓ-max (ℓ-suc ℓ) ℓ') (ℓ-max (ℓ-suc ℓ) ℓ')
-  PSH 𝒞 = FUNCTOR (𝒞 ^op) (SET ℓ)
+  PSH C = FUNCTOR (C ^op) (SET ℓ)
```
mortberg, line 17: How about PreShv instead?

Cubical/Categories/Presheaves.agda
```diff
@@ -14,63 +14,63 @@ open import Cubical.Categories.Sets
...
-  yo x .F-seq f g i h = 𝒞 .seq-α g f h i
+  yo : C .ob → Functor (C ^op) (SET ℓ)
+  yo x .F-ob y .fst = C [ y , x ]
+  yo x .F-ob y .snd = C-cat .homIsSet
+  yo x .F-hom f g = f ⋆⟨ C ⟩ g
+  yo x .F-id i f = C .seq-λ f i
+  yo x .F-seq f g i h = C .seq-α g f h i
 
-  YO : Functor 𝒞 (PSH ℓ ℓ 𝒞)
+  YO : Functor C (PSH ℓ ℓ C)
```
mortberg, line 34: `yoFunctor`?

Cubical/Categories/Pullback.agda
```diff
@@ -13,32 +13,32 @@ private
...
 
-record Pullback {𝒞 : Precategory ℓ ℓ'} (cspn : Cospan 𝒞) : Type (ℓ-max ℓ ℓ') where
+record Pullback {C : Precategory ℓ ℓ'} (cspn : Cospan C) : Type (ℓ-max ℓ ℓ') where
   constructor pullback
   field
-    c : ob 𝒞
+    c : ob C
     cn : Cone cspn c
-    universal : {c' : ob 𝒞} (cn' : Cone cspn c') → ∃![ f ∈ 𝒞 .hom c' c ] Σ[ q ∈ Cone.p₁ cn' ≡ 𝒞 .seq f (Cone.p₁ cn) ] (Cone.p₂ cn' ≡ 𝒞 .seq f (Cone.p₂ cn))
+    universal : {c' : ob C} (cn' : Cone cspn c') → ∃![ f ∈ C [ c' , c ] ] Σ[ q ∈ Cone.p₁ cn' ≡ f ⋆⟨ C ⟩ (Cone.p₁ cn) ] (Cone.p₂ cn' ≡ f ⋆⟨ C ⟩ (Cone.p₂ cn))
```
mortberg, line 35: Long line

Cubical/Categories/Pullback.agda
```diff
@@ -13,32 +13,32 @@ private
...
 
 -- whisker the parallel morphisms g and g' with f
-lPrecatWhisker : {𝒞 : Precategory ℓ ℓ'} {x y z : 𝒞 .ob} (f : 𝒞 .hom x y) (g g' : 𝒞 .hom y z) (p : g ≡ g') → 𝒞 .seq f g ≡ 𝒞 .seq f g'
-lPrecatWhisker {𝒞 = 𝒞} f _ _ p = cong (𝒞 .seq f) p
+lPrecatWhisker : {C : Precategory ℓ ℓ'} {x y z : C .ob} (f : C [ x , y ]) (g g' : C [ y , z ]) (p : g ≡ g') → f ⋆⟨ C ⟩ g ≡ f ⋆⟨ C ⟩ g'
```
mortberg, line 38: This general fact should be upstreamed

Cubical/Categories/Sets.agda
```diff
@@ -9,9 +9,9 @@ open import Cubical.Categories.Category
 module _ ℓ where
   SET : Precategory (ℓ-suc ℓ) ℓ
```
mortberg, line 10: `SetPrecat`?

Cubical/Categories/Type.agda
```diff
@@ -8,9 +8,9 @@ open import Cubical.Categories.Category
 module _ ℓ where
   TYPE : Precategory (ℓ-suc ℓ) ℓ
```
mortberg, line 9: `TypePrecat`?

Cubical/Categories/Pullback.agda
```diff
@@ -13,32 +13,32 @@ private
...
+lPrecatWhisker {C = C} f _ _ p = cong (_⋆_ C f) p
 
 -- extend a cone on c by a morphism c'→c using precomposition
-coneMap : {𝒞 : Precategory ℓ ℓ'} {cspn : Cospan 𝒞} {c c' : ob 𝒞} (cn : Cone cspn c) (f : hom 𝒞 c' c) → Cone cspn c'
-coneMap {𝒞 = 𝒞} {cospan _ _ _ s₁ s₂} (cone p₁ p₂ sq) f =
-  cone (𝒞 .seq f p₁)  (𝒞 .seq f p₂) ((𝒞 .seq-α f p₁ s₁) ∙∙ lPrecatWhisker {𝒞 = 𝒞} f (𝒞 .seq p₁ s₁) (𝒞 .seq p₂ s₂) sq ∙∙ sym (𝒞 .seq-α f p₂ s₂))
+coneMap : {C : Precategory ℓ ℓ'} {cspn : Cospan C} {c c' : ob C} (cn : Cone cspn c) (f : C [ c' , c ]) → Cone cspn c'
+coneMap {C = C} {cospan _ _ _ s₁ s₂} (cone p₁ p₂ sq) f =
+  cone (f ⋆⟨ C ⟩ p₁)  (f ⋆⟨ C ⟩ p₂) ((C .seq-α f p₁ s₁) ∙∙ lPrecatWhisker {C = C} f (p₁ ⋆⟨ C ⟩ s₁) (p₂ ⋆⟨ C ⟩ s₂) sq ∙∙ sym (C .seq-α f p₂ s₂))
```
mortberg, line 44: Maybe rewrite using eq reasoning to increase readability?

Cubical/Categories/Pullback.agda
```diff
@@ -13,32 +13,32 @@ private
...
+    p₁ : C [ c , (Cospan.S₁ cspn)]
+    p₂ : C [ c , (Cospan.S₂ cspn)]
+    sq : p₁ ⋆⟨ C ⟩ (Cospan.s₁ cspn) ≡ p₂ ⋆⟨ C ⟩ (Cospan.s₂ cspn)
 
-record Pullback {𝒞 : Precategory ℓ ℓ'} (cspn : Cospan 𝒞) : Type (ℓ-max ℓ ℓ') where
+record Pullback {C : Precategory ℓ ℓ'} (cspn : Cospan C) : Type (ℓ-max ℓ ℓ') where
   constructor pullback
   field
-    c : ob 𝒞
+    c : ob C
```
mortberg, line 33: Maybe `pbObj` and `pbCone`?

Cubical/Categories/Pullback.agda
```diff
@@ -13,32 +13,32 @@ private
...
     ℓ ℓ' : Level
 
-record Cospan (𝒞 : Precategory ℓ ℓ') : Type (ℓ-max ℓ ℓ') where
+record Cospan (C : Precategory ℓ ℓ') : Type (ℓ-max ℓ ℓ') where
   constructor cospan
   field
-    S₁ S₂ vertex : Precategory.ob 𝒞
-    s₁ : hom 𝒞 S₁ vertex
-    s₂ : hom 𝒞 S₂ vertex
+    S₁ S₂ vertex : Precategory.ob C
```
mortberg, line 19: Maybe using `l` and `r` (for left/right) would be better?

mortberg (changes requested):
Thanks a lot for fixing this so quickly! 

One thing I wondered when reading the code is whether we should be using more instance arguments? What does the agda-categories library do? Generally I think we should take inspiration from them about how to organize this library as they probably already had to solve all of the engineering issues we will run into.

Cubical/Categories/Category.agda
felixwellen: Yes, that fits the style of the (new) names in Algebra.

Cubical/Categories/Category.agda
dan-iel-lee: I don't think they can be made private since the syntax isn't a standalone thing. I also tried defining the operator itself as _⋆⟨_⟩_ but the issue there is that the category is the second argument, while the first argument is a morphism of that category, but Agda requires that arguments can only depend on previous arguments (is there some way to get around this)?

Cubical/Categories/Sets.agda
dan-iel-lee: Hm isn't SET sort of like the standard name for this category though?

Cubical/Categories/Type.agda
dan-iel-lee: Same as above, isn't all caps the standard naming?

Cubical/Categories/Pullback.agda
dan-iel-lee: Not sure I completely understand the code for coneMap yet, might have to push this to a later PR

Cubical/Categories/Pullback.agda
dan-iel-lee: done

Cubical/Categories/Pullback.agda
dan-iel-lee: done

Cubical/Categories/Pullback.agda
dan-iel-lee: moved to `Category`

Cubical/Categories/Pullback.agda
dan-iel-lee: split

Cubical/Categories/NaturalTransformation.agda
dan-iel-lee: Same comment as for SET and TYPE, isn't all caps standard?

Cubical/Categories/Presheaves.agda
dan-iel-lee: what about the other embedding then?

Cubical/Categories/Presheaves.agda
dan-iel-lee: done

Cubical/Categories/Functor.agda
dan-iel-lee: Can we keep these for now? I've used these a bunch in later commits already (which I'll be making PRs for soon) and also I do kind of like the current naming. (same for nat).

Cubical/Categories/Category.agda
dan-iel-lee: done

Cubical/Categories/Category.agda
dan-iel-lee: don't need parentheses

Cubical/Categories/Category.agda
dan-iel-lee: changed

Cubical/Categories/Category.agda
dan-iel-lee: changed

Cubical/Categories/Category.agda
mortberg: Ok, let's leave it as it is then

Cubical/Categories/Functor.agda
mortberg: Ok, the F can stand for "Functor" of course :-)

Cubical/Categories/NaturalTransformation.agda
mortberg: No? Both Wikipedia and the nlab uses "Set"

Also, it's better if the name reflects that this is the `Precat` version. There should also be a category and univalent category version of this.

Cubical/Categories/Category.agda
```diff
@@ -24,59 +24,97 @@ module Cubical.Categories.Category where
...
-(𝒞 ^op) .seq-α f g h = sym (𝒞 .seq-α _ _ _)
+_^op : Precategory ℓ ℓ' → Precategory ℓ ℓ'
+(C ^op) .ob = C .ob
+(C ^op) .Hom[_,_] x y = C .Hom[_,_] y x
+(C ^op) .id = C .id
+(C ^op) ._⋆_ f g = C ._⋆_ g f
+(C ^op) .⋆IdL = C .⋆IdR
+(C ^op) .⋆IdR = C .⋆IdL
+(C ^op) .⋆Assoc f g h = sym (C .⋆Assoc _ _ _)
+
```
mortberg, line 112: Can you add a sanity check that C ^op ^op = C? Should be something like

```
involutiveOp : (C : Precategory ℓ ℓ') → C ^op ^op ≡ C
involutiveOp C = refl
```

(maybe you need to pattern match)

dan-iel-lee: Yep, I'd put it in a future PR, I'll add it here as well. Only thing is it I had to enable eta equality for it to go through directly judgmentally.

mortberg: Ah right, can you add a comment about it needing eta?

Cubical/Categories/NaturalTransformation.agda
dan-iel-lee: Hmm yeah you're right; Riehl's book uses **SET** in place of **Set** for when things aren't "small." But I looked it up too and don't see the notation elsewhere.

When you say there will be category and univalent versions, does that mean the versions will actually be different categories? For example, isn't **SET** as defined in `Sets.agda` already a Category, be we just haven't proven that it is an instance of `isCategory` yet.

Cubical/Categories/Category.agda
dan-iel-lee: Yep, it should be there in the latest commit

Cubical/Categories/NaturalTransformation.agda
mortberg: Ah, I was confused and thought that precategory, category and univalent category all had different definitions. But the way things are currently done `isCategory` and `isUnivalent` are just predicates. `

dan-iel-lee: Right right. Are you good with the names as is then? Or would you prefer something like FunctorCat/SetCat?

PR #485, opened 2020-12-21
Slice Category

dan-iel-lee:
(Note this has all the changes from my previous PR so probably should only look at this after that one's merged)
- Added slice category construction
- Added morphism module


dan-iel-lee:
When I rebased with the new master branch, it created a bunch of duplicate commits. I don't want to clutter the commit history; is there a way to get rid of all the intermediate commits?

PR #486, opened 2020-12-22
3rd category PR: Working with things up to path; functor results/constructions

dan-iel-lee:
Two main things in this PR:
- working with morphism and natural transformation sequencing when the shared (co)domain is not judgementally equal
- functors: some notation, functor composition, functors preserve commutative diagrams


Saizan:
I presume this is subsumed by #497 ?

dan-iel-lee:
Yes that's my bad I think I had closed them all :/

Issue #411, opened 2020-08-28
Reflection interface for the ring solver


felixwellen:
The 'agda-ring-solver' has a reflection interface. That means it is easy to use and the code doesn't get cluttered. It should be possible to copy that.
Reflection is needed to represent the endpoints of an equality by the syntax trees under development here: #410 .

felixwellen:
There is also information about this issue in #297

felixwellen:
This should be closed, the ring solver has a reflection interface (and that part worked so far).

PR #490, opened 2021-01-03
Reflection solver

felixwellen:
WIP on a reflection interface for the ring solver.
This is issue #411.


felixwellen:
The reflection interface works!
There are still some issues:
* If the solver cannot solve because the given equation does not hold, the user will see a very unhelpful error message
* There is only a reflection interface for the CommRingSolver, not for the NatSolver.
If you think those things are important to fix, make issues for those issues.

Saizan:
Great! This is very nice.

Is the problem with `_-_` vs. `-_` a known bug or should we report it?

Also, you do get quite a weird error location if the macro goes wrong, it seemed to point to the enclosing module. Do you know if this is a known problem?

felixwellen:
Thanks!

I don't know if any of the problems is known.
In the case of the name clash of the section 'IsCommRing.-_' and 'IsCommRing.-_' I didn't even think it is a bug. Should the section not be a name?

I think the weird error location happens when unification fails in the term I construct with reflection. Maybe I am just using it wrong and should 'catch' the error?

Saizan:
The names `_-_` and `-_` should be distinct. 

Writing `(y -_)` is ambiguous though because it could either mean `\ x -> y - x` or just `y` applied to `-_`.

Saizan:
Indeed we can use `quote (CommRingStr.-_)` to disambiguate. I pushed a commit over at #499 that removes the `*MinusOperator` module.

About the error reporting I'm not sure, we could also merge and open an issue for that, as you suggest. Seems like we want to ask other Agda people about it.

PR #492, opened 2021-01-08
Localisation pr

mzeuner:
This pull request adds the important example of localizing at two elements successively:
given a ring R and two elements f and g we might want to consider the ring R[1/f][1/g] that we obtain
by localizing R at f and then the resulting ring R[1/f] at g (or rather at g/1). We prove that this is equal
to the ring R[1/fg], obtained by localizing R at the product fg. This identity is tacitly used when defining the
structure sheaf of a ring in a lot of algebraic geometry references.

Additionally, some of the proofs are made shorter using the new ring solver...


mortberg (changes requested):
Please use the new fancy solver tactic :-)

mzeuner:
> Please use the new fancy solver tactic :-)

I used it wherever I could and have left comments where the ring solver didn't work so far...

mortberg:
Thanks!

PR #498, opened 2021-01-25
Add literals for Fin

L-TChen:
I just copy the example from the official documentation on [Literal Overloading](https://agda.readthedocs.io/en/v2.6.1.1/language/literal-overloading.html#literal-overloading), as I want to avoid some clumsy symbols for `Fin`.

### DIFF
--- a/Cubical/Data/Fin/Literals.agda
+++ b/Cubical/Data/Fin/Literals.agda
@@ -1,11 +1,12 @@
 {-# OPTIONS --cubical --no-import-sorts --no-exact-split --safe #-}
 module Cubical.Data.Fin.Literals where
 
-open import Cubical.Data.Nat.Literals public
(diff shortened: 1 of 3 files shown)


Cubical/Data/Fin/Literals.agda
```diff
@@ -0,0 +1,17 @@
+{-# OPTIONS --cubical --no-import-sorts --no-exact-split --safe #-}
+module Cubical.Data.Fin.Literals where
+
+open import Cubical.Data.Nat.Literals public
```
Saizan, line 4: Does this need to be imported `public` for the literals to work?

L-TChen: No. I wasn't careful enough.

L-TChen: Removed.

PR #499, opened 2021-01-28
Reflection solver

Saizan:
A tweak to #490


felixwellen:
Thanks, it is great to get rid of the "CommRingMinusOperator"!

I put in a warning of the wrong error locations and a (commented) example in the "Examples.agda".
And I made an issue: #500 . So I think everyone that tries to use the solver is warned and it can be merged.

PR #503, opened 2021-02-04
Converting between iterated record types with reflection

ecavallo:
Generating equivalences between record types and sigma-types using reflection. In the future it should be possible to factor the macros for univalent structures in record types through these.


mortberg (approved):
Wow, very cool!

PR #504, opened 2021-02-08
index structure descriptors by output and equiv structure level

ecavallo:
The `Desc` data type, which describes a grammar for building up compound structures, was parameterized only by the universe level of the induced structure's input (`ℓ` where `S : Type ℓ → Type ℓ'`). This PR adds indices for the universe level of the output and the type of structured equivalences, which were previously calculated after the fact by functions `macroStrLevel : ∀ {ℓ} → Desc ℓ → Level ` and `macroEquivLevel : ∀ {ℓ} → Desc ℓ → Level`. I believe Agda previously didn't allow inductive families indexed by `Level`, but it's possible now.

### DIFF
--- a/Cubical/Experiments/Rng.agda
+++ b/Cubical/Experiments/Rng.agda
@@ -23,7 +23,7 @@ private
     ℓ ℓ' : Level
 
 module _ {ℓ} where
-  rawRngDesc : Desc ℓ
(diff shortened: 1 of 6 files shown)


mortberg (approved):
Looks good to me!

PR #505, opened 2021-02-09
Laxer versions of isOfHLevel≃/≡

dolio:
For positive h-level, we have access to a value of `A ≃ B`, so one of the `isOfHLevel` premises about `A` or `B` is unnecessary.

`isOfHLevel≃` also had an overly restrictive type

I'm not sure I've chosen ideal names for these, but I don't really have better ones. If there are suggestions, I can change it.

### DIFF
--- a/Cubical/Foundations/HLevels.agda
+++ b/Cubical/Foundations/HLevels.agda
@@ -444,7 +444,9 @@ isOfHLevelΠ⁻ (suc (suc n)) h x y z =
 
 -- h-level of A ≃ B and A ≡ B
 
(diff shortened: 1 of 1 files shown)


ecavallo:
Good observation! I would be fine with leaving off the `⁺` in the names, but I don't have a strong opinion about that and don't have any other naming ideas.

I suppose one could avoid the index offset entirely by defining and using `max`:
```
isOfHLevel⁺≃ₗ
  : ∀ n {A : Type ℓ} {B : Type ℓ'}
  → isOfHLevel n A → isOfHLevel (max 1 n) (A ≃ B)
```

PR #507, opened 2021-02-10
Reflection for generating DURGs on records

ecavallo:
Intended to eventually supplant `Cubical.Structures.Record`.

This code just builds a DURG on the record type from provided DURGs (or proofs of propositionality) on the fields. It doesn't generate the DURGs for the fields automatically like `Cubical.Structures.Record`---I think it will be better/more modular to separate these two aspects. Most of the actual reflection work is also now nicely factored through `Cubical.Reflection.RecordEquiv`.

The forms of record allowed are a little less general than in `Cubical.Structures.Record`: the proof that a field is propositional can only use information from _previous_ fields, not from _all_ fields. This is a problem for some algebraic structures currently in the library (see https://github.com/agda/cubical/pull/395#issuecomment-707956974) but it would be complicated enough to cover these cases that I think it would be better to just reorganize those structure definitions.

Like `Cubical.Structures.Record`, non-propositional fields that depend on each other are not currently supported. The increased generality of DURGs should make it easy to capture that case, but for now I just want to get to the point where we can replace existing Structure code with DURGs.


ecavallo:
Travis is getting stuck solving some constraints in `Cubical.Papers.ZCohomology`, which isn't a problem on my machine and which I don't think should be affected by any of my changes in this PR. Maybe an orthogonal issue due to a change in the latest Agda? I'll update my Agda and see if I can track down the issue.

Cubical/Displayed/Record.agda
```diff
@@ -0,0 +1,236 @@
...
+    → R.Term → R.TC Unit
+  𝒮ᴰ-Record = Macro.𝒮ᴰ-Record
+
+-- Example
+
+module Example where
+
+  record Example (A : Type) : Type where
+    field
+      dog : A
```
mortberg, line 221: Nice, so we can do dogegories now?

mortberg (approved):
Looks good to me!

PR #509, opened 2021-02-12
Helper function for building equivalences from strict inverses

ecavallo:
A neat little trick for saving some time.

Often we have functions `f : A → B` and `g : B → A` that are definitionally inverse and want to construct an equivalence from them. In this case there's a direct definition (see for example `idIsEquiv`) that's simpler than what you'd get from `isoToEquiv (iso f g (λ _ → refl) (λ _ → refl))`. But it's a pain to write out each time, and it can't be naively turned into a lemma because of the definitional equality conditions on `f` and `g`.

However, we _can_ write down a weird type `C : {f : A → B} (g : B → A) → Type` such that there is a function `strictContrFibers : ∀ {f} g → C f g` and such that if `f` and `g` _do_ form a strict isomorphism, then `C {f} g` is definitionally equal to `∀ b → isContr (fiber f b)`. Then we can write definitions like so:

```
e : A ≃ B
e .fst = f
e .snd .equiv-proof = strictContrFibers g
```

This is simpler than writing the definition out by hand but produces a simpler term than using `isoToEquiv`.

Here's a poll before I go through the library looking for places to apply this. In many cases in the library there's already a term `isom : Iso A B` defined earlier in the file, and then `e` is defined by applying `isoToEquiv`. The format above then produces a simpler term but is a bit longer to write out. It would easy to define a reflection macro that would grab the two functions from `isom` and fill in the definition above. But I'm wary about compromising the library's readability by involving reflection in such minor tasks. What say you?


Saizan:
I guess the reflection call would generate a pattern matching lambda? Or would it have to use `unquoteDecl` ?

I think macros that let use write the easy thing but generate "efficient" code are a good idea in general, for example we should have one for turning calls to eliminators into their own function definition.

ecavallo:
I was thinking pattern-matching lambda, send `isom` to `λ {.fst → isom .fun; .snd .equiv-proof → strictContrFibers (isom .inv)}`. But I don't know what the tradeoffs are between this and `unquoteDecl`.

ecavallo:
I tried writing some macros (in `Cubical.Reflection.StrictEquiv`, one for building a strict equiv from two functions and one for turning a strict iso into an equiv), but they're very finicky. I don't know whether I'm running into limitations of Agda's reflection or if I just don't know the right way to write them.

It works well if the inputs are defined names of exactly the right type:
```
UnitToType≃ : ∀ {ℓ} (A : Type ℓ) → (Unit → A) ≃ A
UnitToType≃ A = strictEquiv fwd bwd
  where
  fwd : (Unit → A) → A
  fwd f = f _

  bwd : A → (Unit → A)
  bwd = const
```

But even inlining `const` like produces unsolved metavariables:
```
UnitToType≃ : ∀ {ℓ} (A : Type ℓ) → (Unit → A) ≃ A
UnitToType≃ A = strictEquiv fwd const
  where
  fwd : (Unit → A) → A
  fwd f = f _
```

Anonymous functions are OK only if they're completely annotated, which is readable here but not so much in more complicated examples:
```
UnitToType≃ : ∀ {ℓ} (A : Type ℓ) → (Unit → A) ≃ A
UnitToType≃ A = strictEquiv (λ (f : Unit → A) → f _) (λ (a : A) (_ : Unit) → a)
```

ecavallo:
Figured out how to do this the right way, or at least a way that works.

Cubical/Reflection/StrictEquiv.agda
```diff
@@ -0,0 +1,53 @@
...
+        (R.def (quote strictContrFibers) (g v∷ []))
+    ∷ []
+    )
+    []
+
+strictEquivMacro : ∀ {ℓ ℓ'} {A : Type ℓ} {B : Type ℓ'}
+  → (A → B) → (B → A) → R.Term → R.TC Unit
+strictEquivMacro {A = A} {B} f g hole =
+  R.quoteTC (A ≃ B) >>= λ equivTy →
+  R.checkType hole equivTy >>
```
Saizan, line 32: Does it not work without this part? I thought the hole would already have the right type from the surrounding context.

ecavallo: Without that part it still works but is a little less robust. For example in `Cubical.Data.Sigma.Properties` there's a definition 
```
ΣPathP≃PathPΣ = strictIsoToEquiv ΣPathPIsoPathPΣ
```
without a type signature, so there's no surrounding context. Without those lines it becomes necessary to add a type signature.

Saizan: That makes sense, nice to know.

Saizan:
It's unfortunate that with these changes the normal form for e.g. `Σ-assoc-≃` is
```
λ { .fst → fun Σ-assoc-Iso
   ; .snd .equiv-proof → strictContrFibers (inv Σ-assoc-Iso)
   })
```
rather than something more compact like it was before.

This would be solved by an `unquoteDef` version of the macro.

https://agda.readthedocs.io/en/v2.6.1.3/language/reflection.html?highlight=unquoteDecl#id3

ecavallo:
Do you know ìf it's possible to use `unquoteDef` when there's some quantification over arguments before the macro is applied? For example the argument `A` in this case:
```
ΣUnit : ∀ {ℓ} (A : Unit → Type ℓ) → Σ Unit A ≃ A tt
ΣUnit A = strictEquiv snd (λ { x → (tt , x) })
```

Saizan:
You could use a module for this:

```
module _ {ℓ} (A : Unit → Type ℓ) where 
  ΣUnit : Σ Unit A ≃ A tt
  ΣUnit = strictEquiv snd (λ { x → (tt , x) })
```

Saizan:
@ecavallo is this ready to merge then?

ecavallo:
yep!

PR #511, opened 2021-02-16
Finitely presented algebras

felixwellen:


felixwellen:
I forgot to rebase that and its hopeless now. I will throw it away and make a new PR.

PR #520, opened 2021-03-11
Subtraction properties for Int

LuuBluum:
Added some basic properties for subtraction, since nothing was really proven after the definition for Int. Namely, proved anticommutativity and distributivity over addition. Figured that it'd be a nice first project so I can figure out how all of this works.

### DIFF
--- a/Cubical/Data/Int/Properties.agda
+++ b/Cubical/Data/Int/Properties.agda
@@ -396,20 +396,20 @@ pos- (suc m) (suc n) = suc m ℕ- suc n                       ≡⟨ pos- m n �
                                 - (negsuc n₁ - negsuc n) ∎
 
 -Dist+ : ∀ m n → - (m + n) ≡ (- m) + (- n)
--Dist+ (pos n) (pos n₁) =  - (pos  n + pos n₁)       ≡⟨ cong -_ (sym (pos+ n n₁)) ⟩
+-Dist+ (pos n) (pos n₁) =  - (pos  n + pos n₁)      ≡⟨ cong -_ (sym (pos+ n n₁)) ⟩
                           - (pos (n +ℕ n₁))         ≡⟨ -pos (n +ℕ n₁) ⟩
                              neg (n +ℕ n₁)          ≡⟨ neg+ n n₁ ⟩
                             (neg  n) + (neg n₁)     ≡⟨ cong (neg n +_) (sym (-pos n₁)) ⟩
                             (neg  n) + (- (pos n₁)) ≡⟨ cong (_+ (- pos n₁)) (sym (-pos n)) ⟩
                          (-  pos  n) + (- pos n₁) ∎
--Dist+ (pos n) (negsuc n₁) = - (pos n + negsuc n₁)   ≡⟨ sym (-AntiComm (pos (suc n₁)) (pos n)) ⟩
+-Dist+ (pos n) (negsuc n₁) = - (pos n + negsuc n₁)  ≡⟨ sym (-AntiComm (pos (suc n₁)) (pos n)) ⟩
                                pos (suc n₁) - pos n ≡⟨ +-comm (pos (suc n₁)) (- pos n) ⟩
                            (-  pos n) + (- negsuc n₁) ∎
 -Dist+ (negsuc n) (pos n₁) = - (negsuc n + pos n₁) ≡⟨ cong -_ (+-comm (negsuc n) (pos n₁)) ⟩
-                            - (pos n₁ + negsuc n) ≡⟨ sym (-AntiComm (- negsuc n) (pos n₁)) ⟩
+                            - (pos n₁ + negsuc n)  ≡⟨ sym (-AntiComm (- negsuc n) (pos n₁)) ⟩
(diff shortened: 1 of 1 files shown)


LuuBluum:
Somehow I completely missed at the top that predSuc and sucPred are already defined, and this redundantly defines them. I'll go in and fix that.

Cubical/Data/Int/Properties.agda
```diff
@@ -134,9 +134,21 @@ m + negsuc n = m +negsuc n
...
+
+-pos : ∀ n → - (pos n) ≡ neg n
+-pos zero = refl
+-pos (suc n) = refl
+
+-neg : ∀ n → - (neg n) ≡ pos n
+-neg zero = refl
+-neg (suc n) = refl
+
+double- : ∀ z → - (- z) ≡ z
```
mortberg, line 147: Maybe this should be `-Involutive`?

Cubical/Data/Int/Properties.agda
```diff
@@ -300,6 +316,102 @@ private
                                        (minusPlus m)
                                        (plusMinus m))
 
++inv : ∀ z → z - z ≡ pos zero
```
mortberg, line 319: Maybe `-Cancel`?

Cubical/Data/Int/Properties.agda
```diff
@@ -300,6 +316,102 @@ private
...
+                    pos zero - pos (suc n) ∎
+pos- (suc m) zero = refl
+pos- (suc m) (suc n) = suc m ℕ- suc n                       ≡⟨ pos- m n ⟩
+                       pos m - pos n                        ≡⟨ sym (sucPred (pos m - pos n)) ⟩
+                       sucInt (predInt (pos m - pos n))     ≡⟨ cong sucInt (+predInt (pos m) (- pos n)) ⟩
+                       sucInt (pos m + predInt (- (pos n))) ≡⟨ cong sucInt (cong (pos m +_) (predInt-pos n)) ⟩
+                       sucInt (pos m + negsuc n)            ≡⟨ sucInt+negsuc n (pos m) ⟩
+                       pos (suc m) - pos (suc n) ∎
+
+-anticomm : ∀ m n → m - n ≡ - (n - m)
```
mortberg, line 376: I would use camelCase, so `-AntiComm` or `-antiComm`. I'm not sure which is best? @ecavallo

mortberg: Same for the other anticomm above

Cubical/Data/Int/Properties.agda
```diff
@@ -300,6 +316,102 @@ private
...
+                             - (pos (suc n) + pos n₁) ≡⟨ cong -_ (+-comm (pos (suc n)) (pos n₁)) ⟩
+                             - (pos n₁ - negsuc n) ∎
+-anticomm (negsuc n) (negsuc n₁) = negsuc n - negsuc n₁        ≡⟨ +-comm (negsuc n) (pos (suc n₁)) ⟩
+                                   pos (suc n₁) + negsuc n     ≡⟨ sym (pos- (suc n₁) (suc n)) ⟩
+                                   suc n₁ ℕ- suc n             ≡⟨ ℕ-anticomm (suc n₁) (suc n) ⟩
+                                - (suc n ℕ- suc n₁)            ≡⟨ cong -_ (pos- (suc n) (suc n₁)) ⟩
+                                - (pos (suc n) - pos (suc n₁)) ≡⟨ cong -_ (+-comm (pos (suc n)) (negsuc n₁)) ⟩
+                                - (negsuc n₁ - negsuc n) ∎
+
+-dist : ∀ m n → - (m + n) ≡ (- m) + (- n)
```
mortberg, line 398: Maybe `-Dist`?

mortberg (changes requested):
Looks good to me! I just had some comments about naming

Cubical/Data/Int/Properties.agda
mortberg: Actually the name should also include `+` as there are two similar lemmas for multiplication

Cubical/Data/Int/Properties.agda
LuuBluum: Makes sense, and would be more consistent down the line. Done.

Cubical/Data/Int/Properties.agda
LuuBluum: I decided to go with `-AntiComm` so it would fit better with the above. Does make it a bit out-of-step with the naming of `+-assoc` and `+-comm`, but that was going to be an issue anyway.

Cubical/Data/Int/Properties.agda
LuuBluum: Agreed. Went with `-Dist+`.

Cubical/Data/Int/Properties.agda
LuuBluum: Makes sense. +inv would make more sense as part of a group constructor than it does here.

mortberg (approved):

mortberg:
Great! Now we just have to figure out why Agda has stopped building the library...

PR #522, opened 2021-03-12
Use more copatterns in Cubical.Foundations.Pointed

ecavallo:
Also cleaned up some miscellaneous stuff in `Cubical.Foundations.Pointed.Homotopy`.


mortberg:
Thanks!

PR #523, opened 2021-03-13
DiffInt≡Int

LuuBluum:
Contingent on #520 for some subtraction properties in one of the lemmas. Should help with #404.


ecavallo:
Thanks!

PR #524, opened 2021-03-16
Left Kan extension in presheaves

ecavallo:
Kan extension of a functor f : C → D to a functor PreShv C ℓ → PreShv D ℓ that is left adjoint to precomposition by f. Also some related miscellaneous changes:
- Add a universe level parameter to presheaves, `PreShv C ℓ` is presheaves valued in `SET ℓ`. Previously `PreShv C` gave only presheaves valued in `SET ℓ` where `ℓ` was the level of the type of objects of `C`.
- In `Categories.Adjoint`, a helper function for constructing unit-counit adjunctions when the precategories are categories; in this case equality of natural transformations is equality of their behavior on objects, so there is a simpler statement of the triangle equalities.
- Added `Categories.Functor.Compose` with precomposition and postcomposition functors. Perhaps these should just go in `Categories.Functor.Properties`, but I was having trouble with circular imports. Currently `NaturalTransformation.Base` depends on `Functor.Properties` to define whiskering, maybe we should try to untangle that?


Cubical/Categories/Functor/Compose.agda
```diff
@@ -0,0 +1,47 @@
...
+  open Functor
+  open NatTrans
+
+  postcomposeF : Functor (FUNCTOR C D) (FUNCTOR C E)
+  postcomposeF .F-ob F = funcComp G F
+  postcomposeF .F-hom α .N-ob c = G. F-hom (α .N-ob c)
+  postcomposeF .F-hom {F₀} {F₁} α .N-hom {x} {y} f =
+    sym (G .F-seq (F₀ ⟪ f ⟫) (α ⟦ y ⟧))
+    ∙ cong (G ⟪_⟫) (α .N-hom f)
+    ∙ G .F-seq (α ⟦ x ⟧) (F₁ ⟪ f ⟫)
```
mortberg, line 45: Use `_∙∙_∙∙_` here

Cubical/Categories/Presheaf/KanExtension.agda
```diff
@@ -0,0 +1,188 @@
+{-# OPTIONS --cubical --no-import-sorts --safe --experimental-lossy-unification #-}
+
+{-
+  Kan extension of a functor C → D to a functor PreShv C ℓ → PreShv D ℓ.
+  Currently only left Kan extension.
```
mortberg, line 5: Aha, my UniMath formalization that I mentioned the other day was of right Kan extensions: https://unimath.github.io/doc/UniMath/4dd5c17/UniMath.CategoryTheory.RightKanExtension.html

That proof followed  MacLane, CWM, X.3 (p. 233) and was also quite long to formalize

Cubical/Categories/Presheaf/KanExtension.agda
```diff
@@ -0,0 +1,188 @@
...
+  mapL∘ β α d = funExt (elimProp (λ _ → squash/ _ _) (λ _ → refl))
+
+  LanHom : {G G' : Functor (C ^op) (SET ℓ)}
+    → NatTrans G G' → NatTrans (LanOb G) (LanOb G')
+  LanHom α .N-ob = mapL α
+  LanHom α .N-hom = mapLR α
+
+  -- Definition of the left Kan extension functor
+
+  Lan : Functor (FUNCTOR (C ^op) (SET ℓ)) (FUNCTOR (D ^op) (SET (ℓ-max (ℓ-max (ℓ-max ℓC ℓC') ℓD') ℓS)))
```
mortberg, line 127: Can't you use ℓ as level here?

Cubical/Categories/Presheaf/KanExtension.agda
mortberg: It seems like I used Theorem 2 (iv) of Chapter IV.1 of MacLane btw

mortberg (changes requested):
Good stuff! I just had a few small comments

Cubical/Categories/Presheaf/KanExtension.agda
ecavallo: I'm only proving a special case of (the left version of) that result, which probably explains why it's shorter. Since I'm in presheaves I can do concrete constructions with HITs instead of only having categorical universal properties.

Cubical/Categories/Presheaf/KanExtension.agda
ecavallo: Yeah, just missed that one when I introduced the abbreviation.

Cubical/Categories/Presheaf/KanExtension.agda
mortberg: Ah right, good point! Maybe you construction is just an unfolded version of MacLane's? You compute the colimit pointwise in order to define the left Kan extension?

mortberg (approved):

mortberg:
Excellent! Let's merge as soon as Travis is finished

Cubical/Categories/Presheaf/KanExtension.agda
ecavallo: Yeah, I would think it has to be more or less the same thing.

PR #526, opened 2021-03-20
`rec2 ` accepts heterogeneous relations

L-TChen:
`Cubical.HITs.PropositionalTruncation.Properties.rec2 ` accepts heterogeneous relations


mortberg:
Thanks!

PR #527, opened 2021-03-22
Right Kan extension in presheaves

ecavallo:
Companion to #524.


mortberg:
Nice!

mortberg:
PS: the proof looks really slick!

PR #529, opened 2021-03-26
Sum type properties

LuuBluum:
Proved some basic properties about sum types (associativity, commutativity, that summing with bottom adds nothing). Redid naming to mimic that of sigma types.


ecavallo (approved):
Looks good, thanks!

Issue #513, opened 2021-02-20
The RingSolver fails to solve some equations


felixwellen:
I found two examples of equations the RingSolver fails to solve:
x · 0r ≡ 0r
and
x · (y - z) ≡ x · y - x · z
Since the examples are small, it should be easy to find the problem and fix it. My guess is, that in the definition of the multiplication of HornerForms, an additional special case has to be considered.

felixwellen:
Solved by #530.

PR #530, opened 2021-03-26
Fix ring solver (Issue #513)

felixwellen:
This solves #513 . As far as I know, there should be no problems left with equations that should be solvable. Computation of the normal form for expressions in a ring now use more information to decide if something turns out to be zero. The case analysis is better now. I found something I could use in Cubical.Data.Bool.Properties.


mortberg (approved):
Looks good to me! Should I merge?

felixwellen:
Yes, please!

PR #531, opened 2021-03-29
Functions from set truncations

mzeuner:
We prove a special case of the main result of the paper *Functions out of Higher Truncations*
by Capriotti, Kraus and Vezzosi:
https://arxiv.org/abs/1507.01150

Following their "HIT proof", we show that for a type A and a groupoid B a function `f : A -> B`
induces a function from the set-truncation of A to B (i.e. a function `∥ A ∥₂ → B`) if we have
```
∀ (a b : A) (p q : a ≡ b) → cong f p ≡ cong f q
```
This will then be used to obtain the corresponding result for maps out of set-quotients into groupoids.

### DIFF
--- a/Cubical/HITs/SetTruncation/Properties.agda
+++ b/Cubical/HITs/SetTruncation/Properties.agda
@@ -11,6 +11,7 @@ module Cubical.HITs.SetTruncation.Properties where
 open import Cubical.HITs.SetTruncation.Base
 
 open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.GroupoidLaws
 open import Cubical.Foundations.Function
 open import Cubical.Foundations.Isomorphism
 open import Cubical.Foundations.Equiv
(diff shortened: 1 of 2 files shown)


Cubical/HITs/SetTruncation/Properties.agda
```diff
@@ -75,6 +76,121 @@ elim3 Bset g = elim2 (λ _ _ → isSetΠ (λ _ → Bset _ _ _))
...
+                                   (cong (cong fun) α) (cong (cong fun) β)
+                                   (gtrunc x y p q α β) i j k
+
+ module Hrec {C : Type ℓ''} (Cgpd : isGroupoid C)
+             (η* : A → C)
+             (ε* : ∀ (a b : A) → ∥ a ≡ b ∥ → η* a ≡ η* b)
+             (δ* : ∀ (a b : A) (p : a ≡ b) → ε* a b ∣ p ∣ ≡ cong η* p) where
+
+  fun : H → C
+  fun = Helim.fun (λ _ → Cgpd) η* ε* δ*
```
mortberg, line 115: Isn't it often better to define the recursor directly by recursion instead of using the eliminator? I think this gives better computational behavior?

Cubical/HITs/SetTruncation/Properties.agda
```diff
@@ -75,6 +76,121 @@ elim3 Bset g = elim2 (λ _ _ → isSetΠ (λ _ → Bset _ _ _))
...
+   Q→≡ : (x y : H) → Q x .fst → Q y .fst → x ≡ y
+   Q→≡ =     HelimSet.fun (λ _ → isSetΠ3 λ _ _ _ → gtrunc _ _)
+       λ a → HelimSet.fun (λ _ → isSetΠ2 λ _ _ → gtrunc _ _)
+       λ b p q → sym (ε a₀ a p) ∙ ε a₀ b q
+
+ f₂ : ∥ A ∥₂ → H
+ f₂ = rec Hset η
+
+ fun : ∥ A ∥₂ → B
+ fun = f₁ ∘ f₂
```
mortberg, line 192: Maybe restate this at the top-level to make it easier to figure out what the statement really is?

mortberg (changes requested):
Looks good to me! I only had two minor remarks

Cubical/HITs/SetTruncation/Properties.agda
mzeuner: should be clear now, I think

PR #532, opened 2021-04-06
Functions from set quotients

mzeuner:
Building on the previous (https://github.com/agda/cubical/pull/531)
we give a recursor that allows to construct a function  A / R -> B, from a set-quotient A / R into a groupoid B,
from a map A -> B satisfying the conditions
```
(i)   ∀ (a b : A) → R a b → f a ≡ f b
(ii)  ∀ (a b : A) → isProp (f a ≡ f b)
```

Note that one could probably improve this slightly by requiring three (weaker) conditions instead of (ii):
```
(ii')   ∀ (a b : A) (p q : a ≡ b) → cong f p ≡ cong f q
(iii')  ∀ (a b : A) (r r' : R a b) → eq/* a b r ≡ eq/* a b r'
(iv')   ∀ (a b : A) (p : a ≡ b) (r : R a b) → cong f p ≡ eq/* a b r
```
where `eq/* : ∀ (a b : A) → R a b → f a ≡ f b` is the term assumed in (i). 
One can probably proceed analogously to the proof of the recursor for set-truncations into groupoids
using only (i),(ii')-(iv'), but it seems unlikely that this extra bit of generality will lead to any new applications...

### DIFF
--- a/Cubical/HITs/SetTruncation/Properties.agda
+++ b/Cubical/HITs/SetTruncation/Properties.agda
@@ -11,6 +11,7 @@ module Cubical.HITs.SetTruncation.Properties where
 open import Cubical.HITs.SetTruncation.Base
 
 open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.GroupoidLaws
 open import Cubical.Foundations.Function
 open import Cubical.Foundations.Isomorphism
 open import Cubical.Foundations.Equiv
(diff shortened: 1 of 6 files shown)


Cubical/HITs/TypeQuotients/Base.agda
```diff
@@ -0,0 +1,17 @@
...
+
+- Definition of type quotients (i.e. non-truncated quotients)
+
+-}
+{-# OPTIONS --cubical --no-import-sorts --safe #-}
+module Cubical.HITs.TypeQuotients.Base where
+
+open import Cubical.Core.Primitives
+
+-- Set quotients as a higher inductive type:
```
mortberg, line 13: Should be "Type quotients"

Cubical/HITs/TypeQuotients/Properties.agda
```diff
@@ -0,0 +1,60 @@
+{-
+
+Set quotients:
```
mortberg, line 3: Type

Cubical/HITs/SetTruncation/Properties.agda
```diff
@@ -75,6 +76,127 @@ elim3 Bset g = elim2 (λ _ _ → isSetΠ (λ _ → Bset _ _ _))
                      (λ a b → elim (λ _ → Bset _ _ _) (g a b))
 
 
+-- the recursor for maps into groupoids following the "HIT proof" in:
```
mortberg, line 79: Isn't this already merged?

mortberg (changes requested):
Looks good to me except for some Set that should be Type

Cubical/HITs/SetTruncation/Properties.agda
mzeuner: Yeah sorry, I just kept working on the same branch and didn't merge with the updated master branch. Should be fine now.

mortberg (approved):

mortberg:
Great, let's merge as soon as Travis is done checking it

PR #533, opened 2021-04-06
Fix slow checking ZCohomology

aljungstrom:
This is pretty ad hoc... But it fixes some type checking issues in `GroupStructure.agda `. Not sure if this problem started with some previous PR (it still type checked but incredibly slowly, so Travis was probably fine with it) or with some Agda update. I'll leave it to the pros to determine that;-)


Cubical/ZCohomology/GroupStructure.agda
```diff
@@ -211,6 +211,28 @@ assocₖ (suc zero) =
...
+  wedgeConSn-×3 n f g d x =
+    wedgeConSn _ _ (λ _ _ → isOfHLevelPath ((2 + n) + (2 + n)) (wedgeConHLev n) _ _)
+               (f x)
+               (g x)
+               (sphereElim _ {A = λ x → g x (ptSn (suc (suc n))) ≡ f x (ptSn (suc (suc n))) }
+                             (λ _ → isOfHLevelTrunc (4 + n) _ _ _ _)
+                             (sym d) x) .fst
+{-
+This was the original proof for the case n ≥ 2:
+For some reason it doesn't check in reasonable time anymore:
```
mortberg, line 235: Hmm, it's truly mysterious to me why this would be so much slower... @Saizan Any clue?

Saizan: It seems like the type of `helper` has a massive normal form, so if equality checking ends up expanding that type we will get such a slowdown.

The new code avoids this by working with the local variables `f` and `g` which are not going to reduce, and the call to `wedgeConSn-×3` does not seem to involve types as complicated as the one for `helper`.

mortberg: I see, it's scary that the type of helper is so huge...

PR #536, opened 2021-04-15
Prove isPropIsContr without connections

loic-p:
This proof may be slightly cleaner, and it would work for cartesian cubes too.


mortberg:
Nice!

PR #538, opened 2021-04-16
Changes related to a change to Agda's reflection machinery.

nad:
See https://github.com/agda/agda/issues/5317.

### DIFF
--- a/Cubical/Structures/Auto.agda
+++ b/Cubical/Structures/Auto.agda
@@ -46,7 +46,9 @@ private
   infixl 4 _>>=_ _>>_ _<|>_
 
   varg : ∀ {ℓ} {A : Type ℓ} → A → R.Arg A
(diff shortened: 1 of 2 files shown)


ecavallo:
Is there a reason this change is to the [2ltt-simpler](/github/agda/cubical/2ltt-simpler) branch instead of [master](/github/agda/cubical/master)?

Saizan:
I expect that's because what agda/master tests against

nad:
Agda's cubical library submodule is currently at commit e8aa103cc3ac9ff69c8e3ece274b0db17aec50d1, which matches the 2ltt-simpler branch. I don't know if there is any plan behind this.

Given that the code will not work with the current release of Agda I suspect that you do not want to have it on the main branch at the moment.

ecavallo:
Ah, OK.

Issue #539, opened 2021-04-16
Redefine isConnected?


mortberg:
I conjecture that we can simplify many proofs about n-connected things by changing the definition into:

```
isConnected : (n : HLevel) (A : Type ℓ) → Type ℓ
isConnected n A = hLevelTrunc n A × isProp (hLevelTrunc n A)
```

This is equivalent to the old definition `isContr (hLevelTrunc n A)` but having an `isProp` proof is probably easier to work with as we avoid reasoning about Sigma-types. I also conjecture that we can make the second component abstract everywhere without breaking computations using n-connected things, but I might be wrong. With this definition it should at least be easier to experiment with this. My hope is that we can speed things up in Z-cohomology by making more things abstract (or at least use copatterns everywhere when proving `isConnected`).

As an experiment I reproved `isConnectedPath` using the above definition and it got quite slick:

```
isConnectedPath : (n : HLevel) {A : Type ℓ}
                → isConnected (suc n) A
                → (x y : A) → isConnected n (x ≡ y)
fst (isConnectedPath n connA x y) = PathIdTruncIso n .Iso.fun (connA .snd _ _)
snd (isConnectedPath n connA x y) =
  subst isProp (PathIdTrunc n) (isProp→isSet (connA .snd) _ _)
```

For reference here's the old version: https://github.com/agda/cubical/blob/master/Cubical/Homotopy/Connected.agda#L167

I'm not sure if the old version so long and complicated because it has been optimized for computation or if it's just harder to prove things about `isContr`... 

So what do you think, should we try to change the definition and see if all proofs get shorter and things get faster?

ecavallo:
A similarly slick proof is possible with the current definition:
```
isConnectedPath : ∀ {ℓ} (n : HLevel) {A : Type ℓ}
  → isConnected (suc n) A
  → (a₀ a₁ : A) → isConnected n (a₀ ≡ a₁)
isConnectedPath zero connA a₀ a₁ = isContrUnit*
isConnectedPath (suc n) {A = A} connA a₀ a₁ =
  subst isContr (PathIdTrunc (suc n)) (isContr→isContrPath connA _ _)
```
A more optimized but still short version would use the retract hlevel lemma instead of `subst`:
```
isConnectedPath (suc n) {A = A} connA a₀ a₁ =
  isOfHLevelRetractFromIso 0 (invIso (PathIdTruncIso (suc n))) (isContr→isContrPath connA _ _)
```

mortberg:
[...]

PR #540, opened 2021-04-16
Optimizing mostly hlevel

ecavallo:
- Transport `isOfHLevel` proofs using retracts instead of `subst` when possible
- "Smaller" proof of the isomorphism between heterogeneous and homogeneous paths (about the same amount of code as the old "unfolded" proof, but smaller if unfolded all the way)
- Rewrite `isConnectedPath` à la #539 (but without redefining `isConnected`)

### DIFF
--- a/Cubical/Foundations/HLevels.agda
+++ b/Cubical/Foundations/HLevels.agda
@@ -72,16 +72,20 @@ isOfHLevelPlus : (m : HLevel) → isOfHLevel n A → isOfHLevel (m + n) A
 isOfHLevelPlus zero hA = hA
 isOfHLevelPlus (suc m) hA = isOfHLevelSuc _ (isOfHLevelPlus m hA)
 
-isOfHLevelPlus' : (m : HLevel) → isOfHLevel m A → isOfHLevel (m + n) A
-isOfHLevelPlus' {A = A} {n = n} m hA = subst (λ m → isOfHLevel m A) (+-comm n m) (isOfHLevelPlus n hA )
-
 isContr→isOfHLevel : (n : HLevel) → isContr A → isOfHLevel n A
-isContr→isOfHLevel {A = A} n cA = isOfHLevelPlus' 0 cA
+isContr→isOfHLevel zero cA = cA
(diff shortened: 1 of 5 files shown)


Cubical/Foundations/HLevels.agda
```diff
@@ -290,6 +243,54 @@ isContrRetractOfConstFun : {A : Type ℓ} {B : Type ℓ'} (b₀ : B)
...
+                  → isOfHLevel n (A i1)
+                  → (x : A i0) (y : A i1) → isOfHLevel n (PathP A x y)
+isOfHLevelPathP {A = A} n h x y =
+  isOfHLevelRetractFromIso n (PathPIsoPath _ x y) (isOfHLevelPath n h _ _)
+
+-- Fillers for cubes from h-level
+
+isSet→isSet' : isSet A → isSet' A
+isSet→isSet' {A = A} Aset a₀₋ a₁₋ a₋₀ a₋₁ =
+  transport⁻ (PathP≡Path (λ i → a₋₀ i ≡ a₋₁ i) a₀₋ a₁₋) (Aset _ _ _ _)
```
mortberg, line 264: Why not prove this as

```
isSet→isSet' : isSet A → isSet' A
isSet→isSet' {A = A} Aset a₀₋ a₁₋ a₋₀ a₋₁ = toPathP (Aset _ _ _ _)
```
?

Same for other things in this file. It seems really dangerous to transport in this complicated family...

mortberg: Btw, the `PathP≡Path` used here is not `ua` of the proof you simplified in Foundations.Path in this PR

mortberg (changes requested):
Looks good, except for some proofs that can be optimized even more

Cubical/Foundations/HLevels.agda
ecavallo: Oops, I thought I got them all. I don't see any other uses of `PathP≡Path` I missed though.
I think it's best to leave `PathP≡Path` as it is instead of defining it with `ua`... if one really needs the path, then probably the version that avoids `Glue` is better.

mortberg (approved):

PR #541, opened 2021-04-20
Prop truncations over sum types and (non-dependent) sigma types

LuuBluum:
Adds some basic identities for propositional truncations over sum types and non-dependent sigma types. Namely, applying idempotence into both types.

Not entirely happy with the names of these.


mortberg (approved):
Looks good to me and I think the name are OK.

I'm not sure, but it might be worth stating the Iso's on the top level as well

PR #542, opened 2021-04-21
Direct proof of compEquiv

ecavallo:
It's a little grimier than I'd hoped, but it looks like it's still smaller than the naive definition if completely unfolded.

### DIFF
--- a/Cubical/Foundations/Equiv.agda
+++ b/Cubical/Foundations/Equiv.agda
@@ -106,9 +106,38 @@ invEquiv e = isoToEquiv (invIso (equivToIso e))
 invEquivIdEquiv : (A : Type ℓ) → invEquiv (idEquiv A) ≡ idEquiv A
 invEquivIdEquiv _ = equivEq refl
 
--- TODO: there should be a direct proof of this that doesn't use equivToIso
 compEquiv : A ≃ B → B ≃ C → A ≃ C
-compEquiv f g = isoToEquiv (compIso (equivToIso f) (equivToIso g))
+compEquiv f g .fst = g .fst ∘ f .fst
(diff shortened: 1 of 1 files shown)


ecavallo:
As a side note, I was wondering about the following. When we have some string of equivalences we want to turn into a path, we usually try to compose these as equivalences first and then apply `ua` rather than applying `ua` first and then composing. But `Glue` immediately gives us an operator `(A ≃ B) → (B ≡ C) → (A ≡ C)`, so could it be more efficient to use that, building up a path by composing one equivalence onto it at a time? This is definitely better than applying `ua` to each part first and then composing, but I don't know how it would compare with saving `ua` for last.

But I don't know if this pattern comes up often enough in "performance-critical" proofs to be worth figuring out what's better.

mortberg:
> As a side note, I was wondering about the following. When we have some string of equivalences we want to turn into a path, we usually try to compose these as equivalences first and then apply `ua` rather than applying `ua` first and then composing. But `Glue` immediately gives us an operator `(A ≃ B) → (B ≡ C) → (A ≡ C)`, so could it be more efficient to use that, building up a path by composing one equivalence onto it at a time? This is definitely better than applying `ua` to each part first and then composing, but I don't know how it would compare with saving `ua` for last.
> 
> But I don't know if this pattern comes up often enough in "performance-critical" proofs to be worth figuring out what's better.

Ah yeah, good point! In fact, can't we have a version similar to `_∙∙_∙∙_` where we compose a path in the base with two equivalences on the sides?

mortberg (approved):
Looks good to me! The proof is longer, but should be more efficient

ecavallo:
Yeah, that would probably be even better, if more annoying to use.

PR #543, opened 2021-04-21
Some DUARel combinators

ecavallo:
- Define `SubstRel` generalizing the `TransportStr` from the structure code.
- Add some (D)UARel combinators for sigma and function types and organize existing combinators in files according to type former
- Define morphisms of UARels and show that DUARels can be reindexed along these.

(There will be a follow-up PR using this stuff to generalize the auto from Structure.)


Cubical/Displayed/Universe.agda
```diff
@@ -0,0 +1,33 @@
...
+open import Cubical.Displayed.Subst
+
+private
+  variable
+    ℓA ℓ≅A ℓB ℓ≅B ℓP : Level
+
+𝒮-Univ : ∀ ℓ → UARel (Type ℓ) ℓ
+𝒮-Univ ℓ .UARel._≅_ = _≃_
+𝒮-Univ ℓ .UARel.ua _ _ =
+  isoToEquiv (invIso (Univalence.isoThm pathToEquiv pathToEquivRefl))
```
mortberg, line 25: You can use https://github.com/agda/cubical/blob/master/Cubical/Foundations/Univalence.agda#L207

Or take the inverse equiv of univalence?

mortberg (changes requested):
Good stuff! I just had one small comment and then we can merge

Cubical/Displayed/Universe.agda
ecavallo: Ah, I didn't notice `univalenceIso`. I was avoiding inverse equiv to not do unnecessary `isoToEquiv`s, though it probably doesn't matter much here.

mortberg (approved):

mortberg:
Let's merge once Travis is done

Cubical/Displayed/Universe.agda
mortberg: Right, I suspected that was the reason

PR #544, opened 2021-04-21
adapt auto to Displayed

ecavallo:


Cubical/Displayed/Auto.agda
```diff
@@ -0,0 +1,361 @@
...
+      → UARelDesc (𝒮-Π 𝒮-A 𝒮ᴰ-B)
+
+    unit : UARelDesc 𝒮-Unit
+
+  -- Projections from one UARel to another
+  data UARelReindex : ∀ {ℓA ℓ≅A ℓC ℓ≅C}
+    {A : Type ℓA} {𝒮-A : UARel A ℓ≅A}
+    {C : Type ℓC} {𝒮-C : UARel C ℓ≅C}
+    (f : UARelHom 𝒮-A 𝒮-C)
+    → Typeω
```
mortberg, line 64: Why is this `Typeω`?

mortberg (approved):
Looks good to me!

Cubical/Displayed/Auto.agda
Saizan: Because the levels are indexes I would guess, do we need that?

mortberg: Oh right!

ecavallo: They're indices because the level of a compound type expression is built up from the levels of its components, eg. the level of a sigma is the max of the domain and codomain level. I couldn't see away around this.

PR #547, opened 2021-04-27
Reorganize Group files

mortberg:
I've started reorganizing the code about groups. It was quite a mess with lots of definitions in strange places. 

One thing I wonder about is where to put instances? Now I added a subfolder called Instances where I put the trivial group (aka unit), bool and int. I also put the relevant theory about these groups there. Does this make sense or should we organize things differently?


mortberg:
Oops, this was supposed to be a draft PR... Anyway, can @aljungstrom , @ecavallo  and @felixwellen  please take a look? 

I'm just planning to move things between files and not substantially change any definitions

felixwellen:
I like your separation of 'Instances'. So far I called things like that 'Examples'. But I don't have a strong opinion about the name one way or the other.

mortberg:
> I like your separation of 'Instances'. So far I called things like that 'Examples'. But I don't have a strong opinion about the name one way or the other.

Cool! Maybe we should have similar structure for the other algebra folders then. 

About naming, should we really call it "trivialGroup"? I kind of prefer "UnitGroup" to be consistent with other naming (BoolGroup, IntGroup)

felixwellen:
I would drop 'Group', that seems to be noise you can still add if you need to have both (e.g. Bool and BoolGroup).

felixwellen:
'Unit' is good for me. It is not very mathy, but I would say that's not the goal if you are deep in a library.

martinescardo:
Is that the zero object of the category of groups? So it should be called Zero not Unit, I believe.

martinescardo:
https://ncatlab.org/nlab/show/zero+object

felixwellen:
I thought 'Unit' is ok as a name, since it is the Unit type with a/the group structure on it.
While it is technically a zero object, it might be a bit confusing. If someone calls a group Zero, I would think of '{0}' instead of '{1}', which feels like talking about abelian groups. But on the other, we call the operation in a group '+', so this might be a bit superficial...

mortberg:
Right, it's the Unit type with a group structure on it hence the name... I'm up for dropping the "Group" postfix from the instances

mortberg:
Another remark: we are not very consistent with `_+_` vs `_·_`  when talking about groups... The additive notation seems more common, should we use it everywhere for groups (both abelian and non-abelian)?

felixwellen:
If it doesn't make problems, I would very much like using "\_\cdot\_" for not-neccessarily abelian groups.

felixwellen:
I meant to write: '_·_'

felixwellen:
Ohno. Here: \_·\_

mortberg:
Yeah, that would be nicer... It will be a boring rewrite, but this is probably a good time to do it. Should I do it?

What about the inverse? It's currently written `-`, maybe we should call it `inv` to keep it prefix? Renaming it to `_⁻¹ ` will require even more fiddling...

felixwellen:
Oh, I see. 'inv' is certainly better than '-'.

felixwellen:
I don't have time, so I cannot help now, but I think it would be good to do the renaming.

mortberg:
> I don't have time, so I cannot help now, but I think it would be good to do the renaming.

No worries, shouldn't take too long with some emacs magic. I'll do it as part of this PR

felixwellen:
Great! Thanks!

mortberg:
I also renamed the neutral element from `0g` to `id`. Couldn't think of something better and it's short for "identity"

felixwellen:
'id' works for me. The most common math name, I guess, is '1', so '1g' might be a tiny bit better from that perspective and a bit more logical in our naming scheme.

mortberg:
> 'id' works for me. The most common math name, I guess, is '1', so '1g' might be a tiny bit better from that perspective and a bit more logical in our naming scheme.

Good point! I renamed it to `1g`

felixwellen:
Jei!

Cubical/Algebra/Group/MorphismProperties.agda
```diff
@@ -1,283 +1,279 @@
...
+    module G = GroupStr (snd G)
+    module H = GroupStr (snd H)
+
+    idHelper : x G.· G.inv y ≡ G.1g
+    idHelper = h _ (isHom f _ _ ∙
+                    cong (λ a → f .fun x H.· a) (homInv f y) ∙
+                    cong (H._· H.inv (f .fun y)) p ∙
+                    H.invr _)
+
+-- TODO: maybe it would be better to take this as the definition of isInjective?
```
mortberg, line 127: @aljungstrom What do you think about this?

mortberg:
I think this PR is pretty much done now. Apart from lots of reorganization and renaming I added a few results about morphisms of groups (see Group.Morphisms and Group.MorphismProperties)

PR #548, opened 2021-04-29
Subgroups and first isomorphism theorem for groups

mortberg:
Some work on subgroups building on the first definition from @martinescardo 's HoTT/UF lecture notes.

I'm planning to define quotient groups and maybe prove the first isomorphism theorem


mortberg:
I succeeded in proving the first isomorphism theorem without too much work. It was really nice to see how the section-retraction part became trivial because of things reducing nicely (modulo some transport with refl)

Cubical/Algebra/Group/QuotientGroup.agda
```diff
@@ -0,0 +1,105 @@
...
+  isRefl~ x = subst-∈ ⟪ H' ⟫ (sym (invr x)) id-closed
+
+  G/H : Type ℓ
+  G/H = G /s _~_
+
+  1/H : G/H
+  1/H = [ 1g ]
+
+  _·/H_ : G/H → G/H → G/H
+  x ·/H y = setQuotBinOp isRefl~ _·_ rem x y
```
mortberg, line 52: @felixwellen  @mzeuner We should use this function (or the version for commutative operations) in QuotientRing. It should make the definition a bit slicker

PR #550, opened 2021-04-29
Some modular arithmetic

aljungstrom:
This PR contains some basic modular arithmetic (defined on `Fin`). I did this ages ago and haven't thought much about it since, but it seems to work pretty well.

I also threw in a definition of Z/nZ in `Group.Instances`.


aljungstrom:
Wups.. fixed

PR #551, opened 2021-04-29
update definition in reflection for latest Agda

ecavallo:


mortberg:
Nice! So everyone will have to update their Agda's once this is merged?

ecavallo:
Yep

PR #552, opened 2021-04-30
Algebra organization

ecavallo:
This PR contains a draft of guidelines for the organization of `Cubical.Algebra`. It is a wishlist, not a description of the current state of the library. This is just a first stab based on a discussion between @mortberg, @aljungstrom, @mzeuner, and myself, so feel free to suggest changes or additional guidelines.

[link to document](https://github.com/ecavallo/cubical/blob/algebra-contributing/Cubical/Algebra/CONTRIBUTING.md)

### DIFF
--- a/Cubical/Algebra/CONTRIBUTING.md
+++ b/Cubical/Algebra/CONTRIBUTING.md
@@ -19,7 +19,7 @@ CONTRIBUTING
   RawMonoidStr = {! unit and multiplication operators !}
 
   IsMonoidStr : (A : Type ℓ) → RawMonoidStr A → Type ℓ
-  IsMonoidStr = (⋯ associativity and unit laws ⋯)
+  IsMonoidStr = {! associativity and unit laws !}
 
   MonoidStr : Type ℓ → Type ℓ
   MonoidStr A = Σ[ R ∈ RawMonoidStr A ] (IsMonoidStr R)
--- a/Cubical/Algebra/CONTRIBUTING.md
+++ b/Cubical/Algebra/CONTRIBUTING.md
@@ -1,0 +1,91 @@
+CONTRIBUTING
+============
+
+# General
+
+- We divide the definition of an algebraic structure into four parts:
+
+  1. Carrier: This is often `Type`, but might be for example `Type × Type`.
+  2. Properties of the carrier: this is usually `isSet`.
+  3. Operations: Non-propositional operations on the carrier, such as the multiplication and unit for a monoid.
+  4. Properties of the operations: for example, the associativity and unit laws for a monoid.
+
+  We separate 2 and 4 because it makes it easier to avoid redundant assumptions of `isSet` in complex
+  algebraic structures or to drop this assumption when it is not needed. As an example illustrating our naming
+  conventions, we would give the following definitions building up the universe of monoids.
+
+  ```
+  RawMonoidStr : Type ℓ → Type ℓ
+  RawMonoidStr = {! unit and multiplication operators !}
+
+  IsMonoidStr : (A : Type ℓ) → RawMonoidStr A → Type ℓ
+  IsMonoidStr = (⋯ associativity and unit laws ⋯)
+
+  MonoidStr : Type ℓ → Type ℓ
+  MonoidStr A = Σ[ R ∈ RawMonoidStr A ] (IsMonoidStr R)
+
+  Monoid : Type (ℓ-suc ℓ)
+  Monoid = Σ[ A ∈ Type ℓ ] (isSet A × MonoidStr A)
+  ```
+
+  _Is this definition of `Monoid` best, or should it be rearranged to `Σ[ A ∈ hSet ℓ ] MonoidStr (A .fst)`?
+  Should we use a record instead of this four-part `Σ`-type? I'd guess we want sub-groupings of these often
+  enough that a `Σ`-type is preferable._
+
+- We define more complicated instances of 1-4 by nesting simpler instances. For example, the definition of
+  `RawRingStr` structure would be a record or `Σ`-type containing `RawMonoidStr` and `RawAbGroupStr` as fields,
+  and `IsRingStr` would likewise use `IsMonoidStr` and `IsAbGroupStr`.
+
+- We prefer records over nested `Σ`-types for efficiency and readability. For example, the definition of
+  `IsRingStr` could be a record with first `IsMonoidStr` and `IsAbGroupStr` fields and then fields for the
+  laws concerning both the monoid and group operations. Reflection
+  ([`Cubical.Reflection.RecordEquiv`](https://github.com/agda/cubical/blob/master/Cubical/Reflection/RecordEquiv.agda))
+  can be used to derive an equivalent `Σ`-type representation when this is more convenient, for example to
+  prove that some record type is propositional.
+
+- When defining an instance of an algebraic structure, use copattern matching rather than a record
+  constructor, and try to avoid relying on helper functions that take many arguments---this is for efficiency
+  reasons. Because of the heavily nested nature of algebraic structures, it can be more readable to separate
+  out components into separate definitions (local or otherwise), for example defining the group component of a
+  given ring structure separately.
+
+- It is preferred to include redundant structure and properties if there are often more direct proofs in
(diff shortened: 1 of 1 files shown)


Cubical/Algebra/CONTRIBUTING.md
```diff
@@ -0,0 +1,91 @@
...
+# General
+
+- We divide the definition of an algebraic structure into four parts:
+
+  1. Carrier: This is often `Type`, but might be for example `Type × Type`.
+  2. Properties of the carrier: this is usually `isSet`.
+  3. Operations: Non-propositional operations on the carrier, such as the multiplication and unit for a monoid.
+  4. Properties of the operations: for example, the associativity and unit laws for a monoid.
+
+  We separate (ii) and (iv) because it makes it easier to avoid redundant assumptions of `isSet` in complex
```
mortberg, line 13: maybe refer to them as "2." and "4."?

Cubical/Algebra/CONTRIBUTING.md
```diff
@@ -0,0 +1,91 @@
...
+  IsMonoidStr = {! associativity and unit laws !}
+
+  MonoidStr : Type ℓ → Type ℓ
+  MonoidStr A = Σ[ R ∈ RawMonoidStr A ] (IsMonoidStr R)
+
+  Monoid : Type (ℓ-suc ℓ)
+  Monoid = Σ[ A ∈ Type ℓ ] (isSet A × MonoidStr A)
+  ```
+
+  _Is this definition of `Monoid` best, or should it be rearranged to `Σ[ A ∈ hSet ℓ ] MonoidStr (A .fst)`?
```
mortberg, line 31: I think it's often easier to have the isSet assumption separate

Cubical/Algebra/CONTRIBUTING.md
```diff
@@ -0,0 +1,91 @@
...
+  reasons. Because of the heavily nested nature of algebraic structures, it can be more readable to separate
+  out components into separate definitions (local or otherwise), for example defining the group component of a
+  given ring structure separately.
+
+- It is preferred to include redundant structure and properties if there are often more direct proofs in
+  specific instances or if there is no canonical choice of minimal axioms. Helper functions can then be
+  defined to generate the full structure from a subset. For example, the definition of groups includes both a
+  left inverse and right inverse law although this is not strictly necessary. Use your judgment.
+
+- _Where should instances of structure be located?_
```
mortberg, line 57: Maybe having them in subfolders called `Instances` are best? Or we could have an instances file where we define a type (like Int)

Cubical/Algebra/CONTRIBUTING.md
```diff
@@ -0,0 +1,91 @@
...
+- When defining an instance of an algebraic structure, use copattern matching rather than a record
+  constructor, and try to avoid relying on helper functions that take many arguments---this is for efficiency
+  reasons. Because of the heavily nested nature of algebraic structures, it can be more readable to separate
+  out components into separate definitions (local or otherwise), for example defining the group component of a
+  given ring structure separately.
+
+- It is preferred to include redundant structure and properties if there are often more direct proofs in
+  specific instances or if there is no canonical choice of minimal axioms. Helper functions can then be
+  defined to generate the full structure from a subset. For example, the definition of groups includes both a
+  left inverse and right inverse law although this is not strictly necessary. Use your judgment.
```
mortberg, line 55: What was the status of default definitions for fields?

Cubical/Algebra/CONTRIBUTING.md
```diff
@@ -0,0 +1,91 @@
...
+  IsMonoidEquiv (S₀ , _) e (S₁ , _) = IsRawMonoidHom S₀ (e .fst) S₁
+
+  MonoidHom : Monoid ℓ → Monoid ℓ' → Type (ℓ-max ℓ ℓ')
+  MonoidHom (M₀ , _ , S₀) (M₁ , _ S₁) = Σ[ f ∈ (M₀ → M₁) ] IsMonoidHom S₀ f S₁
+
+  MonoidEquiv : Monoid ℓ → Monoid ℓ' → Type (ℓ-max ℓ ℓ')
+  MonoidEquiv = {! as with MonoidHom !}
+  ```
+  _These definitions aren't set up the right way for the current SIP setup, but are correct for the DUARel
+  setup we're ready to replace it with._
```
mortberg, line 87: I think this is different from what I had in mind. Consider what we currently have for groups:

https://github.com/agda/cubical/blob/master/Cubical/Algebra/Group/Morphisms.agda#L48

mortberg: We should also have isomorphisms (and Iso which is a homomorphism)

mortberg (changes requested):
I had some comments and suggestions.

One the we didn't discuss yet is if we should switch off eta for any of these?

Cubical/Algebra/CONTRIBUTING.md
ecavallo: Markdown renders the list items as i-iv and I don't know if one can change this. if only I could \ref...

Cubical/Algebra/CONTRIBUTING.md
ecavallo: I think it will be easier to find things in a subfolder of `Algebra` than if they're placed with particular types, since there isn't a one-to-one correspondence between carrier types and algebra structures. I'm OK with calling it `Instances` if no one has better ideas but it does collide with instance arguments.

Cubical/Algebra/CONTRIBUTING.md
ecavallo: It looks like this is possible to some extent with reflection, I'll try and figure out how it works.

Cubical/Algebra/CONTRIBUTING.md
ecavallo: Is the main difference just that we're currently using a two-field record type instead of a sigma to bundle the function/isomorphism/equivalence with the proof of structuredness? I think we definitely want a sigma so that the SIP/DUARels can be applied.

Saizan:
I would encourage you to use `no-eta-equality` for algebraic structures, and define them by copatterns.
Otherwise working with particular instances is going to be quite painful, e.g. when composing morphisms, because when comparing the source and target you will end comparing the operations and proofs whenever syntactic equality fails.

I was testing @aljungstrom code, and I couldn't get a composition of group isos to typecheck in reasonable time until I turned off eta for `GroupStr`, I had to fix a few proofs but it was straightforward to define the necessary path by copatterns.

Cubical/Algebra/CONTRIBUTING.md
ecavallo: Something that makes this hard to do cleanly with reflection is that we have multiple incompatible ways we want to infer the full structure. For example, we want to infer the left inverse law of a group if we have the right inverse, but also infer the right inverse if we have the left inverse. It seems there's no way to order the pieces defining the record so that both are possible, although we could get one of them.

We could do it by in a more complicated way: have a field `oneInverse : LeftInverse ⊎ RightInverse` and then optional fields `linv : LeftInverse` and `rinv : RightInverse` that are filled in automatically depending on what's given for `oneInverse`. But I suspect this would make other things a pain.

thomas-lamiaux:
I was surprise to see this PR. What it is the current status of this PR ?

If it still in someone mind, I was looking at the contributing file and I was thinking it a bit long and not very informative, especially as there is now a lot of good example in the library.
Maybe it should be shorten like 20/30 lines, and say to to take example on the way Groups or Ring are defined ?
I was also wondering if it should be merged with the Namming file as the seems to be about the similar subject  ?

mortberg:
> I was surprise to see this PR. What it is the current status of this PR ?
> 
> If it still in someone mind, I was looking at the contributing file and I was thinking it a bit long and not very informative, especially as there is now a lot of good example in the library. Maybe it should be shorten like 20/30 lines, and say to to take example on the way Groups or Ring are defined ? I was also wondering if it should be merged with the Namming file as the seems to be about the similar subject ?

These were notes from a meeting we had last year. There are some good ideas for improving the organization of algebraic structures, but many were never implemented (but I think @ecavallo  started doing it?).

mortberg:
@ecavallo Make an issue out of this please

PR #553, opened 2021-04-30
improve RecordEquiv

ecavallo:
Make it so the RecordEquiv reflection utility declares a named function rather than creating an anonymous copattern lambda. Also removed some generality that we might never need.


mortberg (approved):
Looks good to me

PR #555, opened 2021-05-03
Move --cubical and --no-import-sorts flags to agda-lib

ecavallo:
The --cubical and --no-import-sorts flags will be on every file, so we can lift them out to the agda-lib file.


mortberg:
Merging this now. This might cause problems with the open PRs, but those should be easy to fix

PR #558, opened 2021-05-06
Simplify the truncation path space calculation a bit

ecavallo:
Two things:
- We were reproving the truncation level of the Codes fibration, which is already true by construction
- Used `transport-filler` (or corresponding `subst-filler` I added) rather than `J` to avoid some `transportRefl`s


ecavallo:
I also just noticed that `isOfHLevel≡` uses `section univalence`. That's defined using `isoToEquiv` and I think is the direction that's made more complicated by the improvement, so I changed it to use the one straight from the isomorphism.

mortberg (approved):
Looks good to me

PR #560, opened 2021-05-21
Add explicit --guardedness flags to modules that use it

jespercockx:
Soon the `--guardedness` flag will no longer be enabled by default in Agda, so here are the required changes for making the cubical library compatible.


jespercockx:
CI is green and I hope this is not a controversial change, so I'm going ahead and merge this.

PR #561, opened 2021-05-21
Fix Agda version in README.

cangiuli:
The README says something about "Agda 2.6.3", but I think it's supposed to read "Agda 2.6.1.3".

The label of version 0.2 of the library also mentions "Agda 2.6.3", but I don't think I can propose an edit to that in this PR; the maintainers should be able to edit the label themselves.


mortberg:
Oops, good catch! I updated the tag

PR #562, opened 2021-05-26
Basic opens

mzeuner:
This if the first pull request for a formalization of a constructive approach to affine schemes.
We define the basic opens of the Zariski spectrum synthetically as the meet-semilattice of elements of the base ring modulo equating two elements that generate the same radical ideal.
Because our structure (pre-)sheaf is then a map from a set-quotient into a groupoid we have to use a special recursor and write it as a map into commutative algebras over the base ring instead of commutative rings.

### DIFF
--- a/Cubical/Algebra/Algebra/Base.agda
+++ b/Cubical/Algebra/Algebra/Base.agda
@@ -204,6 +204,16 @@ isPropIsAlgebraHom R AS f BS = isOfHLevelRetractFromIso 1 IsAlgebraHomIsoΣ
                                          (isPropΠ λ _ → isSetAlgebra (_ , BS) _ _)
                                          (isPropΠ2 λ _ _ → isSetAlgebra (_ , BS) _ _))
 
+isSetAlgebraHom : {R : Ring ℓ} (M : Algebra R ℓ') (N : Algebra R ℓ'')
+                → isSet (AlgebraHom M N)
+isSetAlgebraHom _ N = isSetΣ (isSetΠ (λ _ → isSetAlgebra N))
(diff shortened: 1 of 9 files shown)


Cubical/Algebra/Algebra/Base.agda
```diff
@@ -178,6 +195,19 @@ isPropIsAlgebra R _ _ _ _ _ _ =
...
+isPropIsAlgebraHom : (R : Ring ℓ) {A : Type ℓ'} {B : Type ℓ''}
+                     (AS : AlgebraStr R A) (f : A → B) (BS : AlgebraStr R B)
+                   → isProp (IsAlgebraHom AS f BS)
+isPropIsAlgebraHom R AS f BS =
+  isOfHLevelRetractFromIso 1 IsAlgebraHomIsoΣ
+   (isProp× (isSetAlgebra (_ , BS) _ _)
+     (isProp× (isSetAlgebra (_ , BS) _ _)
+       (isProp× (isPropΠ2 λ _ _ → isSetAlgebra (_ , BS) _ _)
+         (isProp× (isPropΠ2 λ _ _ → isSetAlgebra (_ , BS) _ _)
+           (isProp× (isPropΠ λ _ → isSetAlgebra (_ , BS) _ _)
```
mortberg, line 207: Can't you use some of the higher arity `isProp×`?

mortberg (approved):
I just had one very minor comment. Then I'd be happy to merge

Cubical/Algebra/Algebra/Base.agda
mzeuner: Yeah that's much nicer. I basically copied this proof from Ring.Base, so I fixed it there as well.

PR #563, opened 2021-05-28
Lemmas on homogeneous types

ecavallo:
Proves that:
- Pointed functions into a homogeneous type are equal if their underlying functions are equal
- A pointed function type into a homogeneous type is homogeneous
- Eilenberg-MacLane spaces (on Z) are homogeneous

May be helpful for @aljungstrom


aljungstrom:
Yes, this was precisely what I was looking for! No more disgusting coherence paths!

Issue #506, opened 2021-02-10
Consistency in section/retraction naming


ecavallo:
In `Cubical.Foundations.Equiv` we have:
```
  secEq : section invEq (w .fst)
  secEq = retIsEq (snd w)

  retEq : retract invEq (w .fst)
  retEq = secIsEq (snd w)
```
That is, the convention for which inverse condition is the section and which is the retraction is reversed between `isEquiv` and `_≃_`. This is very confusing!

PR #564, opened 2021-06-07
Swaps names retEq and secEq

maxdore:
Resolves agda/cubical#506

I hope it makes sense this way around, all right-inverses in sight are now called sections.


mortberg (approved):
Looks good to me! What does @ecavallo  think?

ecavallo:
I made a quick change to the type signatures of `retEq` and `secEq` for clarity, but otherwise looks good!
Maybe we should also change the names of the fields of `Iso` for uniformity, but that can be another PR.

PR #568, opened 2021-06-08
Preservation of hlevel by sections

LuuBluum:
Mirrors the same functions for retracts, both in naming and general structure. Could also be used to prove that isomorphisms preserve hlevels, but that is redundant given the existing proof using retracts.


ecavallo:
Aren't these instances of the retract functions with the arguments reversed? For example:
```
is2GroupoidSection
  : {B : Type ℓ}
  (f : A → B) (g : B → A)
  (h : (y : B) → f (g y) ≡ y)
  → is2Groupoid A → is2Groupoid B
is2GroupoidSection f g h = is2GroupoidRetract g f h
```

PR #571, opened 2021-06-09
Redundant isPropIsSurjection definitions

LuuBluum:
Removed one of the two redundant isPropIsSurjection definitions. Not sure why there were two.

### DIFF
--- a/Cubical/Functions/Embedding.agda
+++ b/Cubical/Functions/Embedding.agda
@@ -44,8 +44,8 @@ private
 isEmbedding : (A → B) → Type _
 isEmbedding f = ∀ w x → isEquiv {A = w ≡ x} (cong f)
 
-isEmbeddingIsProp : isProp (isEmbedding f)
-isEmbeddingIsProp {f = f} = isPropΠ2 λ _ _ → isPropIsEquiv (cong f)
+isPropIsEmbedding : isProp (isEmbedding f)
+isPropIsEmbedding {f = f} = isPropΠ2 λ _ _ → isPropIsEquiv (cong f)
 
 -- If A and B are h-sets, then injective functions between
 -- them are embeddings.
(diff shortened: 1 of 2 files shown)


ecavallo:
It looks like this was my fault :sweat_smile:. Thanks for cleaning up!

LuuBluum:
Er, one moment on this one actually; I managed to get the naming backwards.

ecavallo:
Our convention is actually that it should be called `isPropIsSurjection`, to match the type signature.

LuuBluum:
> Our convention is actually that it should be called `isPropIsSurjection`, to match the type signature.

Oh, I thought I had it backwards when I looked at isEmbeddingIsProp. I guess that one's backwards.

Might as well fix all this naming while I'm at it, then.

ecavallo:
Ah, yeah, we haven't been too careful about enforcing this convention so far. But if you want to fix some it's much appreciated!

mortberg:
> > Our convention is actually that it should be called `isPropIsSurjection`, to match the type signature.
> 
> Oh, I thought I had it backwards when I looked at isEmbeddingIsProp. I guess that one's backwards.
> 
> Might as well fix all this naming while I'm at it, then.

Yeah, the naming is very much not consistent... We should try to make sure that all of the Foundations are consistent. Here is my suggested naming convention: 

https://github.com/agda/cubical/pull/451/files

LuuBluum:
The names should be consistent now, at least across those two. I imagine there's tons of other inconsistencies scattered about, but that would be a bit much for this PR!

mortberg:
> The names should be consistent now, at least across those two. I imagine there's tons of other inconsistencies of this nature scattered about, but that would be a bit much for this PR!

Yeah, @ecavallo  and I have been discussing refactoring/shrinking the Foundations at some point. So I guess we can then make sure that whatever stays in this package has proper names. Anyway, this something for a future PR and I approve the changes in this PR.

mortberg (approved):

PR #574, opened 2021-06-23
[ ci ] switch from Travis to github actions

ecavallo:
Swiping the script from https://github.com/agda/agda-stdlib.


mortberg:
How do we know if this is working?

ecavallo:
I think the simplest way is to just merge it and see what happens.

PR #578, opened 2021-07-07
WIP on defining graded rings

mortberg:
I got pretty much everything formalized except for associativity of multiplication which is very hard to get right because of dependent types

### DIFF
--- a/Cubical/Algebra/Ring/Graded.agda
+++ b/Cubical/Algebra/Ring/Graded.agda
@@ -1,0 +1,124 @@
+{-# OPTIONS --safe #-}
+module Cubical.Algebra.Ring.Graded where
+
+open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.Function
+open import Cubical.Foundations.Powerset
+open import Cubical.Foundations.HLevels
+
+open import Cubical.Relation.Nullary.Base
+
+open import Cubical.Data.Nat renaming (_+_ to _+ℕ_ ; _·_ to _·ℕ_)
+open import Cubical.Data.Nat.Order
+open import Cubical.Data.FinData
+open import Cubical.Data.Bool
+open import Cubical.Data.Sigma
+open import Cubical.Data.Empty.Base
+open import Cubical.Data.Sum.Base hiding (map)
+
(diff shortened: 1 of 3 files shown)


Cubical/Algebra/Ring/Graded.agda
```diff
@@ -0,0 +1,664 @@
...
+                                             ·G z (n ∸ toℕ i)))
+           ≡⟨ refl ⟩
+             (((x , hx) ·⊕G (y , hy)) ·⊕G (z , hz)) .fst n ∎
+      where
+      open MonoidBigOp (Group→Monoid (AbGroup→Group (G n))) renaming (bigOp to ∑)
+
+      helper2 : {n : ℕ} (i : Fin (suc n)) → (j : Fin (suc (n ∸ toℕ i))) → toℕ i +ℕ toℕ j +ℕ (n ∸ toℕ i ∸ toℕ j) ≡ toℕ i +ℕ (n ∸ toℕ i)
+      helper2 {n} i j = (sym (+-assoc (toℕ i) (toℕ j) (n ∸ toℕ i ∸ toℕ j)) ∙ cong (toℕ i +ℕ_) (helper j))
+
+      finalstepGen : (n : ℕ)
```
mortberg, line 376: This is where I'm stuck...

ecavallo:
Would it make things any easier to use a natural number bound instead of a finite set? Like this:
```
  IsFiniteSequence : ((i : ℕ) → G i .fst) → Type ℓ
  IsFiniteSequence f = ∃[ n ∈ ℕ ] (∀ k → n ≤ k → f k ≡ 0G)

  ⊕G : Type ℓ
  ⊕G = Σ[ f ∈ ((i : ℕ) → G i .fst) ] IsFiniteSequence f
```
It at least saves you from some universe level inflation. I experimented with rewriting some but stopped at the hard part: https://github.com/ecavallo/cubical/blob/gradedring/Cubical/Algebra/Ring/Graded.agda Up until that point it doesn't change much, although the finite set lemmas can be avoided.

mortberg:
Closing this now that we have a better version.

@thomas-lamiaux  might want to take a closer look at this for comparison later

PR #586, opened 2021-07-29
Typo and consistent names

HilpAlex:
Fixes a typo and inconsistent naming in Cubical/Algebra/Group/Instances/Bool.agda.


mortberg (approved):
Good catch!

PR #588, opened 2021-07-29
Emacs-lisp code for deleting whitespaces on save

felixwellen:
This PR adds emacs-lisp code for the .emacs to the part of CONTRIBUTING.md, that formerly just claimed that code like this exists.


ecavallo:
Thanks!

PR #590, opened 2021-08-12
Syllepsis

aljungstrom:
This PR contains a proof of the syllepsis (`Eckmann-Hilton p q ≡ (Eckmann-Hilton q p) ⁻¹`) for 3-loops p and q. I think the proof can be shortened a bit by using `EH-alt` as the main definition of `Eckmann-Hilton` (to avoid transports). I'll make this PR a draft for now and get back to it later.

@ecavallo I remember you had another proof of `Eckmann-Hilton`. It's possible that the same idea applied to your proof would make things shorter. Do you still have it somewhere? I seem to have lost the link.


ecavallo:
Here: https://github.com/ecavallo/cubical/blob/a474cd1cb42437011850803490cec442a79d21f0/Cubical/Homotopy/Loopspace.agda#L40

aljungstrom:
@mortberg It doesn't get much shorter than this (for me at least). I think it can be merged.

mortberg (approved):

PR #594, opened 2021-08-14
Delete --guardedness flag

XiaohuWang0921:
Since all library folders except for ```Codata``` do not require this flag


ecavallo:
Good eye, thanks!

PR #595, opened 2021-08-16
Add syntax typeclass for types with carriers.

shlevy:
This allows types not using TypeWithStr to reuse the nice ⟨_⟩ syntax.


shlevy:
If more concrete justification is needed: I'm working on an implementation of ordinals (following the HoTT book) and am using records with meaningful field names instead of multiply nested `TypeWithStr`s.

PR #600, opened 2021-08-24
rotation of Square (by 90deg)

marcinjangrzybowski:
I am not sure where would be a good place for lemma like this, 
maybe  /Foundations/Square/Properties.agda ?
for now, the code is in Experiments.

```
  Equivalence of rotating the Square by 90deg can be easily constructed in two ways:

    - from isomorphism of switching dimensions  (sq i j ⇄ sq (~ j) i)
    - from path obtained via applying Square to 4 paths

  This module shows that those equivalences are equal.
```


marcinjangrzybowski:
> let's put it in `Foundations.Path` for now and we can deal with the reorganization question in a future PR.

@ecavallo I moved the definition as requested, sorry for the late answer :)

Cubical/Foundations/Path.agda
```diff
@@ -351,3 +352,103 @@ compPathR→PathP∙∙ {p = p} {q = q} {r = r} {s = s} P j i =
...
+        pathToEquiv ≡rotateSquare
+  rotateSquare≃-≡rotateSquare-lemma = equivEq (funExt p)
+    where
+
+      p : (x : Square _ _ _ _) → rotateSquare x ≡ transport ≡rotateSquare x
+      p x k i j =
+         hcomp
+           (λ l →
+             primPOr k (~ k ∨ i ∨ ~ i ∨ j ∨ ~ j)
+              (λ _ → coe0→i (λ i' → ≡rotateSquare i') l x i j)
```
mortberg, line 451: Can't you just use a transpFill here and get rid of the dependency on CartesianKanOps?

mortberg (changes requested):

mortberg:
Actually, do you have some use case for this?

mortberg:
Now this has a conflict so I'll close it for now. Feel free to re-open if you fix the conflict and have a use case for the proof! I'm somewhat reluctant to add things that are not of very general interest to the Foundations, so it should probably go somewhere else no matter what

PR #601, opened 2021-09-20
Added uniqueness of initial and final objects

kl-i:
Proved that in a precategory, `hasFinalOb` is a proposition and in the case of univalent categories, one has uniqueness of final objects. Similarly with initial objects.


ecavallo:
Looks good to me, thanks!

PR #587, opened 2021-07-29
Grothendieck Groups of Commutative Monoids

fabianmasato:
This PR introduces commutative monoids and binary products thereof.

PR #603, opened 2021-10-06
define plus to diffInt and prove diffInt is group

hyleIndex:


felixwellen:
Thank you for the contribution. Please note that there is an ongoing effort (#587) to implement the usual general construction of the Grothendieck Group of a commutative monoid (which should turn out to be DiffInt for the monoid (ℕ,+)).

hyleIndex:
> * There are whitespace errors (at least some trainling whitespaces at line endings). You can detect them by running 'make' and most likely, you are using an editor that can remove those whitespaces automatically (see 'CONTRIBUTING.md').
> * Most of the definitions in 'DiffInt/Properties.agda' should be private (with 'private' or by putting everything in a module), so a user can decide if/how to use them. It would be good to hide at least all proofs of group axioms, since that can easily be accessed by opening the term in 'IsGroup DiffInt'.
> * It would be nice to have 'DiffInt' as a term of type 'Group'.

Thanks for your suggestions, now all the issues should be fixed.

Cubical/Data/Int/MoreInts/DiffInt/Properties.agda
```diff
@@ -52,3 +56,172 @@ private
...
+ℤ+-assoc = elimProp3 (λ _ _ _ → ℤ-isSet _ _)
+  (λ { (a , b) (c , d) (e , f) i → [ ℕ.+-assoc a c e i -ℕ' ℕ.+-assoc b d f i ] })
+
+ℤ-isGroup : IsGroup {G = ℤ} [ 0 , 0 ] _+_ -ℤ'_
+IsSemigroup.is-set (IsMonoid.isSemigroup (IsGroup.isMonoid ℤ-isGroup)) = ℤ-isSet
+IsSemigroup.assoc (IsMonoid.isSemigroup (IsGroup.isMonoid ℤ-isGroup)) = ℤ+-assoc
+IsMonoid.identity (IsGroup.isMonoid ℤ-isGroup) = λ x → (zero-identityʳ 0 x  , zero-identityˡ 0 x)
+IsGroup.inverse ℤ-isGroup = λ x → (-ℤ'-invʳ x , -ℤ'-invˡ x)
+
+Int = group ℤ [ 0 , 0 ] _+_ -ℤ'_ ℤ-isGroup
```
felixwellen, line 227: If you add the type, i.e. 'Group ℓ-zero' it will be easier for a user to spot, where the actual group is defined.
Also, if you like, you could add a prove that the multiplication is commutative and turn it into an 'AbGroup'.

Cubical/Data/Int/MoreInts/DiffInt/Properties.agda
```diff
@@ -52,3 +56,172 @@ private
...
+  -ℤ' [ a ] = [ snd a , fst a ]
+  -ℤ' eq/ a b r i = eq/ (snd a , fst a) (snd b , fst b) (tmp a b r) i
+                                                        where
+                                                          tmp : ∀ a b → fst a +ℕ snd b ≡ fst b +ℕ snd a → snd a +ℕ fst b ≡ snd b +ℕ fst a
+                                                          tmp a b r =
+                                                            snd a +ℕ fst b ≡⟨ ℕ.+-comm (snd a) (fst b) ⟩
+                                                            fst b +ℕ snd a ≡⟨ sym r ⟩
+                                                            fst a +ℕ snd b ≡⟨ ℕ.+-comm (fst a) (snd b) ⟩
+                                                            snd b +ℕ fst a ∎
+  -ℤ' squash/ a a₁ p q i i₁ = squash/ (-ℤ' a) (-ℤ' a₁) (cong (λ x → -ℤ' x) p) (cong (λ x → -ℤ' x) q) i i₁
```
felixwellen, line 162: You should avoid the long line by moving the 'where'-block more to the left.

hyleIndex: Thanks for the comment,
I have fixed it now.

Cubical/Data/Int/MoreInts/DiffInt/Properties.agda
hyleIndex: Thanks for the comment,
I have fixed it now.

hyleIndex:
Ah yes, makes sense. Thanks for the suggestions. I have made all the stuff except Int private.

felixwellen:
Looks good to me now, I'll ask a maintainer to merge it.

hyleIndex:
Great! Thank you so much

Cubical/Data/Int/MoreInts/DiffInt/Properties.agda
```diff
@@ -52,3 +61,188 @@ private
...
+  IsSemigroup.is-set (IsMonoid.isSemigroup (IsGroup.isMonoid ℤ-isGroup)) = ℤ-isSet
+  IsSemigroup.assoc (IsMonoid.isSemigroup (IsGroup.isMonoid ℤ-isGroup)) = ℤ+-assoc
+  IsMonoid.identity (IsGroup.isMonoid ℤ-isGroup) = λ x → (zero-identityʳ 0 x  , zero-identityˡ 0 x)
+  IsGroup.inverse ℤ-isGroup = λ x → (-ℤ'-invʳ x , -ℤ'-invˡ x)
+
+  ℤ-isAbGroup : IsAbGroup {G = ℤ} [ 0 , 0 ] _+_ -ℤ'_
+  IsAbGroup.isGroup ℤ-isAbGroup = ℤ-isGroup
+  IsAbGroup.comm ℤ-isAbGroup = ℤ+comm
+
+Int : AbGroup ℓ-zero
```
mortberg, line 246: I think this should be `ℤ`

mortberg (changes requested):
I would suggest moving the Group and AbGroup instances to a subfolder of the respective folders in Algebra. For inspiration see how we did it for CommRing: https://github.com/agda/cubical/blob/master/Cubical/Algebra/CommRing/Instances/

I think this should be the general way of organizing instances so that one can find them easily. At the moment far from all instances are implemented this way. We should fix this in some future PRs

PR #606, opened 2021-10-20
Made the code compatible with the Agda branch issue4786

nad:
This commit makes the code compatible with what might become a development version of Agda, without breaking compatibility with the current development version.

### DIFF
--- a/Cubical/Data/DescendingList/Examples.agda
+++ b/Cubical/Data/DescendingList/Examples.agda
@@ -9,7 +9,7 @@
 -- 2. "sorting" finite multisets by converting into sorted lists.
 ------------------------------------------------------------------------
 
-{-# OPTIONS --safe #-}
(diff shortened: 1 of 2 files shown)


nad:
I disabled an example that type-checks very slowly, if at all, using the variant of Agda from the branch issue4786:

https://github.com/agda/cubical/blob/367f3afbd4cd83521bcbfe18da0932d487c05e90/Cubical/Data/DescendingList/Examples.agda#L91-L92

The slowdown seems to be caused by the changes on that branch. However, another recent development version of Agda (2.6.3-edd8334) also has trouble normalising the type of this example, unlike Agda 2.6.2. Perhaps it would make sense to figure out what, exactly, has made Agda slower.

Saizan:
That's because transports of indexed families are stuck in Agda 2.6.2.

nad:
OK, that explains the slowdown.

PR #610, opened 2021-10-25
Basic results about comonadic modalities

felixwellen:
A couple of basic proofs for the 'flat modality' supported by Agda.


felixwellen:
Put on hold, since it uses flat in a way that is not supported by the latest master of agda.

felixwellen:
This will be incompatible with the --safe mode of agda 2.6.3 -> I will move this stuff to 
https://github.com/felixwellen/synthetic-geometry

felixwellen:
Done:
https://github.com/felixwellen/synthetic-geometry/pull/23
-> Closing

PR #614, opened 2021-11-01
Fundamental theorem of identity types

felixwellen:
A basic characterization of the identity type. This can be used to compute identity types.

### DIFF
--- a/Cubical/Foundations/Equiv/Properties.agda
+++ b/Cubical/Foundations/Equiv/Properties.agda
@@ -195,10 +195,6 @@ compl≡Equiv p q r = congEquiv ((λ s → p ∙ s) , (compPathl-isEquiv p))
 isEquivFromIsContr : {A : Type ℓ} {B : Type ℓ′}
                    → (f : A → B) → isContr A → isContr B
                    → isEquiv f
-equiv-proof (isEquivFromIsContr f isContrA isContrB) y = center , contraction
-  where q = isContr→isProp isContrB (f (fst isContrA)) y
-        center = (fst isContrA) , q
-        contraction = λ {(x , p) i → (snd isContrA x i) ,
-                      isOfHLevel→isOfHLevelDep 1
-                        (λ x' → isOfHLevelPlus 2 isContrB (f x') y)
-                        q p (λ i → snd isContrA x i) i}
+isEquivFromIsContr f isContrA isContrB =
+  subst isEquiv (λ i x → isContr→isProp isContrB (fst B≃A x) (f x) i) (snd B≃A)
+  where B≃A = isContr→Equiv isContrA isContrB
--- a/Cubical/Foundations/Equiv/Properties.agda
+++ b/Cubical/Foundations/Equiv/Properties.agda
@@ -195,9 +195,10 @@ compl≡Equiv p q r = congEquiv ((λ s → p ∙ s) , (compPathl-isEquiv p))
 isEquivFromIsContr : {A : Type ℓ} {B : Type ℓ′}
                    → (f : A → B) → isContr A → isContr B
                    → isEquiv f
-isEquivFromIsContr f isContrA isContrB =
-  isoToIsEquiv
-    (iso f
-         (λ _ → fst isContrA)
-         (λ x → isContr→isProp isContrB (f (fst isContrA)) x)
-         λ y → isContr→isProp isContrA (fst isContrA) y)
+equiv-proof (isEquivFromIsContr f isContrA isContrB) y = center , contraction
+  where q = isContr→isProp isContrB (f (fst isContrA)) y
+        center = (fst isContrA) , q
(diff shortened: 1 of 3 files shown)


mortberg:
Nice! I seem to recall that Egbert used this lemma to derive `Omega(S1) = Z`. Would such a proof be substantially different in any way or is it just a refactoring?

felixwellen:
I guess there will be no substantial difference. The theorem is also not so far from encode-decode, but I haven't thought about the exact relation (to which version...) for some time. For me, the fundamental theorem is just easier to remember because it looks more like a universal property.

mortberg:
> I guess there will be no substantial difference. The theorem is also not so far from encode-decode, but I haven't thought about the exact relation (to which version...) for some time. For me, the fundamental theorem is just easier to remember/think about because it looks more like a universal property.

Right, I guess what I'm really asking is exactly how it relates to encode-decode.

felixwellen:
Ok, let's say we fix `x:A` then `code(y)` should certainly be `Eq(x,y)`. The map
`encode : (y:A) -> x=y -> code(y)`
should be given by 
`encode(relf) := reflEq x : code(x)=Eq(x,x)` 
(assuming with have this ingredient of the fundamental theorem).
And `decode : (y : A) -> code(y) -> x = y` we can define by projecting from the contractibilty proof for `Sigma(y:A) Eq(x,y)` (the second ingredient of the theorem).
Now, if you unpack the contractability of `Sigma(y:A) Eq(x,y)`, that means
- there is a center, but that's just given by `relfEq x`
- any other thing in the sigma-type is equal to that, which can be rewritten as:
  `transport in Eq(x,_) (along some p:x==y) (some q : Eq(x,y)) == (x,reflEq x)`
So `p` is `decode y q` and then (without thinking through all details) this looks like Lemma 8.9.1 in the HoTT-Book.

Cubical/Foundations/Equiv/Properties.agda
```diff
@@ -191,3 +191,13 @@ compr≡Equiv p q r = congEquiv ((λ s → s ∙ r) , compPathr-isEquiv r)
 -- composition on the left induces an equivalence of path types
 compl≡Equiv : {A : Type ℓ} {a b c : A} (p : a ≡ b) (q r : b ≡ c) → (q ≡ r) ≃ (p ∙ q ≡ p ∙ r)
 compl≡Equiv p q r = congEquiv ((λ s → p ∙ s) , (compPathl-isEquiv p))
+
+isEquivFromIsContr : {A : Type ℓ} {B : Type ℓ′}
+                   → (f : A → B) → isContr A → isContr B
+                   → isEquiv f
+isEquivFromIsContr f isContrA isContrB =
```
mortberg, line 198: Hmm, can't one prove that this is an equivalence directly? I feel like it might be easier, but I might be wrong in which case this proof is fine...

mortberg:
> Ok, let's say we fix `x:A` then `code(y)` should certainly be `Eq(x,y)`. The map `encode : (y:A) -> x=y -> code(y)` should be given by `encode(relf) := reflEq x : code(x)=Eq(x,x)` (assuming with have this ingredient of the fundamental theorem). And `decode : (y : A) -> code(y) -> x = y` we can define by projecting from the contractibilty proof for `Sigma(y:A) Eq(x,y)` (the second ingredient of the theorem). Now, if you unpack the contractability of `Sigma(y:A) Eq(x,y)`, that means
> 
>     * there is a center, but that's just given by `relfEq x`
> 
>     * any other thing in the sigma-type is equal to that, which can be rewritten as:
>       `transport in Eq(x,_) (along some p:x==y) (some q : Eq(x,y)) == (x,reflEq x)`
>       So `p` is `decode y q` and then (without thinking through all details) this looks like Lemma 8.9.1 in the HoTT-Book.

Thanks! This was helpful :-)

I would be happy to merge this once the question about this one proof has been answered

mortberg (approved):

Cubical/Foundations/Equiv/Properties.agda
felixwellen: I made a direct proof. It avoids isoToEquiv, but it doesn't look better. But that might be due to me avoiding cubical primitives...

ecavallo: Note there is also a `isContr→Equiv` in `Foundations.Equiv`.

felixwellen: Yes, but I thought that might be a bit awkward to use to show, that some particular function is an equivalence.

felixwellen: But I guess my proof would look nicer, if I use that.

felixwellen: Ok, it does look better.

mortberg: Yeah, that looks good!

PR #615, opened 2021-11-02
Zariski Lattice

mzeuner:
This PR contains a constructive/synthetic definition of  the Zariski lattice (the lattice of open sets of the Zariski topology of a commutative ring) and proves its universal property.
It also contains

- Addition and multiplication of ideals  in a commutative ring and their laws
- The interaction of ideal addition/multiplication with the generators of f.g. ideals
- Some more theory on radical ideals

There are a few things that will be added to `Cubical.Algebra.ZariskiLattice` in the future:

- The Zariski Lattice lives in a higher universe than the base ring but its possible to show that it is equivalent to a "small" type living in the base universe
- The basic opens need to be defined as a subset and proved to a basis in the lattice theoretic-sense
- The structure sheaf on the basic opens should be defined directly, showing that the function that send an element to the localization at that element is `3-Constant`

### DIFF
--- a/Cubical/Algebra/CommRing/FGIdeal.agda
+++ b/Cubical/Algebra/CommRing/FGIdeal.agda
@@ -95,7 +95,6 @@ module _ (Ring@(R , str) : CommRing ℓ) where
                                λ r → isLinearCombinationL· V r
 
 
---open CommIdeal
 open CommIdeal.isCommIdeal
 genIdeal : {n : ℕ} (R : CommRing ℓ) → FinVec (fst R) n → CommIdeal.CommIdeal R
 fst (genIdeal R V) x = isLinearCombination R V x , isPropPropTrunc
@@ -107,7 +106,8 @@ syntax genIdeal R V = ⟨ V ⟩[ R ]
 
 
 FGIdealIn : (R : CommRing ℓ) → Type (ℓ-suc ℓ)
-FGIdealIn R = Σ[ I ∈ CommIdeal.CommIdeal R ] ∃[ n ∈ ℕ ] ∃[ V ∈ FinVec (fst R) n ] I ≡ ⟨ V ⟩[ R ]
+FGIdealIn R = Σ[ I ∈ CommIdeal.CommIdeal R ]
+                   ∃[ nV ∈ Σ[ n ∈ ℕ ] FinVec (fst R) n ] I ≡ ⟨ nV .snd ⟩[ R ]
 
 -- The lattice laws
 module _ (R' : CommRing ℓ) where
@@ -159,8 +159,6 @@ module _ (R' : CommRing ℓ) where
   incl2 x x≡0 = subst-∈ (⟨ replicateFinVec _ 0r ⟩ .fst) (sym x≡0)
                          (⟨ replicateFinVec _ 0r ⟩ .snd .contains0)
 
-
- -- better syntax for ∑ λ i → ... ???
 
  FGIdealAddLemma : {n m : ℕ} (U : FinVec R n) (V : FinVec R m)
                  → ⟨ U ++Fin V ⟩ ≡ ⟨ U ⟩ +i ⟨ V ⟩
--- a/Cubical/Algebra/CommRing/FGIdeal.agda
+++ b/Cubical/Algebra/CommRing/FGIdeal.agda
@@ -8,13 +8,17 @@
 module Cubical.Algebra.CommRing.FGIdeal where
 
 open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.Transport
 open import Cubical.Foundations.Powerset
 open import Cubical.Foundations.HLevels
 
 open import Cubical.Data.Sigma
 open import Cubical.Data.FinData hiding (elim)
-open import Cubical.Data.Nat using (ℕ)
-
+open import Cubical.Data.Nat renaming ( zero to ℕzero ; suc to ℕsuc
+                                      ; _+_ to _+ℕ_ ; _·_ to _·ℕ_
+                                      ; +-assoc to +ℕ-assoc ; +-comm to +ℕ-comm
+                                      ; ·-assoc to ·ℕ-assoc ; ·-comm to ·ℕ-comm)
+                             hiding (elim)
 open import Cubical.HITs.PropositionalTruncation
 
 open import Cubical.Algebra.CommRing
@@ -107,6 +111,8 @@ module _ (R' : CommRing ℓ) where
  private
   R = fst R'
  open CommRingStr (snd R')
+ open Sum (CommRing→Ring R')
+ open KroneckerDelta (CommRing→Ring R')
 
  foo : {n : ℕ} (V : FinVec R n) (I : CommIdeal R')
      → (∀ i → V i ∈ I .fst) → ⟨ V ⟩[ R' ] .fst ⊆ I .fst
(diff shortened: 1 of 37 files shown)


Cubical/Algebra/ZariskiLattice/Base.agda
```diff
@@ -0,0 +1,445 @@
...
+ open ProdFin R'
+
+ private
+  R = fst R'
+  A = Σ[ n ∈ ℕ ] (FinVec R n)
+  ⟨_⟩ : {n : ℕ} → FinVec R n → CommIdeal
+  ⟨ V ⟩ = ⟨ V ⟩[ R' ]
+
+ _∼_ : A → A → Type (ℓ-suc ℓ)
+ (_ , α) ∼ (_ , β) = √i ⟨ α ⟩ ≡ √i ⟨ β ⟩ --replace this by ≡ᴾ := ⊆ × ⊇ to preserve universe level
```
mortberg, line 80: Can probably be made nicer with better notations and some lemmas:

```
 _∼'_ : A → A → Type ℓ
 (_ , α) ∼' (_ , β) = (∀ x → x ∈ √i ⟨ α ⟩ .fst → x ∈ √i ⟨ β ⟩ .fst)
                    × (∀ x → x ∈ √i ⟨ β ⟩ .fst → x ∈ √i ⟨ α ⟩ .fst)

 ZL' : Type ℓ
 ZL' = A / _∼'_

 ~→∼' : {a b : A} → a ∼ b → a ∼' b
 ~→∼' r = (λ x h → subst (λ p → x ∈ p) (cong fst r) h)
        ,  λ x h → subst (λ p → x ∈ p) (cong fst (sym r)) h

 ~'→∼ : {a b : A} → a ∼' b → a ∼ b
 ~'→∼ r = CommIdeal≡Char (fst r) (snd r)

 ZL-Iso-ZL' : Iso ZL ZL'
 fun ZL-Iso-ZL' = SQ.rec squash/ (λ x → [ x ]) (λ a b r → eq/ a b (~→∼' r))
 inv ZL-Iso-ZL' = SQ.rec squash/ (λ x → [ x ]) (λ a b r → eq/ a b (~'→∼ r))
 rightInv ZL-Iso-ZL' = SQ.elimProp (λ x p q → squash/ _ _ p q) (λ x → refl)
 leftInv ZL-Iso-ZL' = SQ.elimProp (λ x p q → squash/ _ _ p q) (λ x → refl)

 ZL≃ZL' : ZL ≃ ZL'
 ZL≃ZL' = isoToEquiv ZL-Iso-ZL'
```

Cubical/Algebra/CommRing/FGIdeal.agda
```diff
@@ -89,8 +95,9 @@ module _ (Ring@(R , str) : CommRing ℓ) where
                                λ r → isLinearCombinationL· V r
 
 
-open isCommIdeal
-genIdeal : {n : ℕ} (R : CommRing ℓ) → FinVec (fst R) n → CommIdeal R
+--open CommIdeal
```
mortberg, line 98: You can remove this comment

Cubical/Algebra/CommRing/FGIdeal.agda
```diff
@@ -100,5 +107,146 @@ syntax genIdeal R V = ⟨ V ⟩[ R ]
 
 
 FGIdealIn : (R : CommRing ℓ) → Type (ℓ-suc ℓ)
-FGIdealIn R = Σ[ I ∈ CommIdeal R ] ∃[ n ∈ ℕ ] ∃[ V ∈ FinVec (fst R) n ] I ≡ ⟨ V ⟩[ R ]
-
+FGIdealIn R = Σ[ I ∈ CommIdeal.CommIdeal R ] ∃[ n ∈ ℕ ] ∃[ V ∈ FinVec (fst R) n ] I ≡ ⟨ V ⟩[ R ]
```
mortberg, line 110: Couldn't it be easier to work with this if you refactored and had only one `∃`? 

So something like: 

```
FGIdealIn R = Σ[ I ∈ CommIdeal.CommIdeal R ] ∃[ nV ∈ Σ[ n ∈ ℕ ] FinVec (fst R) n ]  ] I ≡ ⟨ nV .snd ⟩[ R ]
```

This way you wouldn't have to eliminate twice and could get n and V by pattern-matching

Cubical/Algebra/CommRing/FGIdeal.agda
```diff
@@ -100,5 +107,146 @@ syntax genIdeal R V = ⟨ V ⟩[ R ]
...
+  incl1 x = elim (λ _ → is-set _ _)
+          λ (α , x≡∑α0) → subst-∈ (0Ideal .fst) (sym x≡∑α0) (∑Closed 0Ideal (λ i → α i · 0r)
+          λ i → subst-∈ (0Ideal .fst) (sym (0RightAnnihilates _)) refl)
+
+  incl2 : 0Ideal .fst ⊆ ⟨ replicateFinVec _ 0r ⟩ .fst
+  incl2 x x≡0 = subst-∈ (⟨ replicateFinVec _ 0r ⟩ .fst) (sym x≡0)
+                         (⟨ replicateFinVec _ 0r ⟩ .snd .contains0)
+
+
+ -- better syntax for ∑ λ i → ... ???
```
mortberg, line 163: Remove this comment?

Cubical/Algebra/CommRing/Ideal.agda
```diff
@@ -70,10 +86,239 @@ module _ (R' : CommRing ℓ) where
...
+   inclHelperLeft : (I ·i J) .fst ⊆ (I ·i (J +i K)) .fst
+   inclHelperLeft x' = map (λ (n , (α , β) , α∈I , β∈J , x'≡∑αβ)
+                     → n , (α , β) , α∈I , (λ i → +iLincl J K _ (β∈J i)) , x'≡∑αβ)
+
+   inclHelperRight : (I ·i K) .fst ⊆ (I ·i (J +i K)) .fst
+   inclHelperRight x' = map (λ (n , (α , β) , α∈I , β∈K , x'≡∑αβ)
+                      → n , (α , β) , α∈I , (λ i → +iRincl J K _ (β∈K i)) , x'≡∑αβ)
+
+ -- only one absorption law, i.e. CommIdeal , +i , ·i does not form a dist. lattice
+ ·iAbsorb+i : ∀ (I J : CommIdeal) → I +i (I ·i J) ≡ I
```
mortberg, line 316: Two general remark: 
- Maybe it would be useful to state the various `⊆` lemmas on the top-level? Maybe some statements could even be generalized and the proofs be streamlined that way (e.g. `inclHelperLeft` and `inclHelperRight` just above looks very similar...)
- Maybe you should have a special version of `⊆` for ideals to avoid having to put `fst` on both sides all the time?

Cubical/Algebra/CommRing/RadicalIdeal.agda
```diff
@@ -117,6 +120,116 @@ module _ (R' : CommRing ℓ) where
...
+  incl1 : √i (I ·i √i J) .fst ⊆ √i (I ·i J) .fst
+  incl1 x = PT.elim (λ _ → isPropPropTrunc)
+              (uncurry (λ n → PT.elim (λ _ → isPropPropTrunc)
+                (uncurry4 (curriedIncl1 n))))
+   where
+   curriedIncl1 : (n m : ℕ) (α : FinVec R m × FinVec R m)
+                → (∀ i → α .fst i ∈ I .fst)
+                → (∀ i → α .snd i ∈ √ (J .fst))
+                → (x ^ n ≡ linearCombination R' (α .fst) (α .snd))
+                → x ∈ √i (I ·i J) .fst
```
mortberg, line 189: Maybe we should have a version of `∈` for ideals as well?

Cubical/Algebra/Lattice/Properties.agda
```diff
@@ -55,20 +55,41 @@ module Order (L' : Lattice ℓ) where
  open JoinSemilattice (Lattice→JoinSemilattice L') renaming (_≤_ to _≤j_ ; IndPoset to JoinPoset)
  open MeetSemilattice (Lattice→MeetSemilattice L') renaming (_≤_ to _≤m_ ; IndPoset to MeetPoset)
 
+ ≤j→≤m : ∀ x y → (x ≤j y) → (x ≤m y)
```
mortberg, line 58: redundant parentheses

Cubical/Algebra/Lattice/Properties.agda
```diff
@@ -55,20 +55,41 @@ module Order (L' : Lattice ℓ) where
...
+ ≤j→≤m x y x∨ly≡y = x ∧l y         ≡⟨ cong (x ∧l_) (sym x∨ly≡y) ⟩
+                    x ∧l (x ∨l y) ≡⟨ ∧lAbsorb∨l _ _ ⟩
+                    x ∎
+
+ ≤m→≤j : ∀ x y → (x ≤m y) → (x ≤j y)
```
mortberg, line 63: redundant parentheses

mortberg (changes requested):
Nice stuff! I had some small questions and suggestions, see the comments.

Cubical/Algebra/ZariskiLattice/Base.agda
mzeuner: Thanks, added it in a separate module

Cubical/Algebra/CommRing/Ideal.agda
mzeuner: I'm not too happy with the way things are now either. I think moving up all the interesting inclusions is a good first step. I did start working on replacing the usual path types with the equivalent double inclusion:
https://github.com/mzeuner/cubical/commit/4652ee9a9e12e201e07d36fe372917e88006e735

Maybe it's worthwhile doing that afterall. Adding inclusion and containment of ideals is a good idea to make things more readable. The only issue I'm somewhat unhappy about is that you want to have some sort `cong` for subsets and ideals like this
https://github.com/mzeuner/cubical/blob/Zariski_Lattice/Cubical/Foundations/Powerset.agda#L82

It feels a bit like cheating because you go up a universe level and back down but maybe I'm just being fussy here...

Cubical/Algebra/CommRing/FGIdeal.agda
mzeuner: thanks

Cubical/Algebra/CommRing/FGIdeal.agda
mzeuner: I like that suggestion and copied it. Right now `FGIdealIn` is not used anywhere, but let's keep it and see if it actually ever gets applied...

Cubical/Algebra/Lattice/Properties.agda
mzeuner: true that

Cubical/Algebra/Lattice/Properties.agda
mzeuner: true that

Cubical/Algebra/CommRing/FGIdeal.agda
mzeuner: Yeah, do you have an idea for nice `bigOp` syntax?

Cubical/Algebra/CommRing/Ideal.agda
mortberg: I'm a bit puzzled why you might need these lemmas? You can surely just put the results you have on the top-level without using any lemmas like this. Anyway I'm not sure how useful `≡ᴾ` is and it might be best to just use `≡` when possible but state the containment lemmas separately so that you can reuse them

Cubical/Algebra/CommRing/FGIdeal.agda
mortberg: Not really... I think it's pretty good as is :-)

Cubical/Algebra/CommRing/FGIdeal.agda
mortberg: Ah ok, I wonder if there could be some other places where refactoring like this could lead to less uses of `uncurryN`...

Cubical/Algebra/CommRing/Ideal.agda
mzeuner: Changed everything according to your suggestions.

mortberg (approved):

mortberg:
Great! Will merge as soon as the CI is done

PR #618, opened 2021-11-11
Organize rings and commutative rings into categories

mortberg:
Need to upstream some code, so draft for now. Can any of the lemmas about (comm)ring homomorphisms be found anywhere in the library @mzeuner @felixwellen ?


felixwellen:
I wouldn't know that any of these have been proved elsewhere. Maybe it would be better to move them to 'Ring/Properties' or something like that, so they won't get reproved accidentally.
One small remark about something completely different: It might be better to call the instance of a Category 'Rings', since you can also view a single ring as a category.

mortberg:
Ah, good point! I'll rename the files and upstream the ring lemmas.

mortberg:
I think this is ready for review @felixwellen . Hopefully nothing too contorversial

felixwellen:
Looks good to me. Is there a reason things are different for Ring and CommRing? I means, in the definition of the precategory Rings agda seems to need less information than with CommRings. Also, the lemmas you use look a bit different.

mortberg:
> Looks good to me. Is there a reason things are different for Ring and CommRing? I means, in the definition of the precategory Rings agda seems to need less information than with CommRings. Also, the lemmas you use look a bit different.

The problem is that homomorphisms for CommRings are defined in terms of homomorphisms of Rings. By projecting from CommRing to Ring Agda gets confused sometimes and we need to provide more arguments explicitly... It might be that a direct definition of what a CommRing homomorphism is will be needed if the current definition gets too annoying

PR #619, opened 2021-11-12
Miscellaneous prelude cleaning

ecavallo:


Cubical/Foundations/Prelude.agda
```diff
@@ -286,12 +312,12 @@ module _ (P : ∀ y → x ≡ y → Type ℓ') (d : P x refl) where
 -- Converting to and from a PathP
 
 module _ {A : I → Type ℓ} {x : A i0} {y : A i1} where
-  toPathP : transp (\ i → A i) i0 x ≡ y → PathP A x y
+  toPathP : transp A i0 x ≡ y → PathP A x y
```
mortberg, line 315: Oh, no need to eta-expand anymore? @Saizan

mortberg, line 315: Why not use `transport` here?

Cubical/Foundations/Prelude.agda
```diff
@@ -286,12 +312,12 @@ module _ (P : ∀ y → x ≡ y → Type ℓ') (d : P x refl) where
...
   toPathP p i = hcomp (λ j → λ { (i = i0) → x
                                ; (i = i1) → p j })
                       (transp (λ j → A (i ∧ j)) (~ i) x)
 
-  fromPathP : PathP A x y → transp (\ i → A i) i0 x ≡ y
+  fromPathP : PathP A x y → transp A i0 x ≡ y
```
mortberg, line 320: `transport`?

mortberg (changes requested):
Great! Two small comments, then we can merge

Cubical/Foundations/Prelude.agda
ecavallo: It's necessary to convert between lines and paths, but that's not happening here. Though it would need to expanded if I switch it to `transport`, since that takes a path. (Still it might be conceptually better to use `transport`?)

mortberg: Yeah, maybe nicer with transport?

PR #620, opened 2021-11-16
Combinatorics of Finite Sets

kangrongji:
This PR contains : 

- Closure properties of `FinSet` under several type constructors (e.g., Σ-types, Π-types , type of equivalences); 

- A proof of (axiom of) finite choice; 

- Adding some requisite definitions to the library.

In theory, one can calculate the cardinality of any combinations of these constructors applied to finite sets and many combinatorial problems can be described in this way. For example, the number of elements in `Fin 3 ≃ Fin 3` can be correctly evaluated as `6`. Unfortunately, it seems unfeasible even to compute that for `Fin 4`. 

I guess most of the results can be extended to finite groupoids or even finite categories. So the recent works of Egbert Rijke in [OEIS-A000001](https://github.com/EgbertRijke/OEIS-A000001) can be reproduced in cubical agda.


kangrongji:
I might be too optimistic. There is no way to generalize most of these results to higher types but with one exception, namely the closed-ness under Σ-types. And it is exactly what Rijke needs to construct the number of finite groups.

mortberg:
Nice! Ping @EgbertRijke

Cubical/Data/FinSet/Examples.agda
```diff
@@ -0,0 +1,50 @@
+{-
+
+Computable stuff
+
+-}
+{-# OPTIONS --safe #-}
+
+module Cubical.Data.FinSet.Examples where
```
mortberg, line 8: This would fit better in the Experiments package

Cubical/Data/FinSet/Properties.agda
```diff
@@ -0,0 +1,253 @@
...
+private
+  variable
+    ℓ ℓ' ℓ'' ℓ''' : Level
+
+isFinSet→Discrete : {X : Type ℓ} → isFinSet X → Discrete X
+isFinSet→Discrete p =
+  TruncRec isPropDiscrete (λ (_ , p) → EquivPresDiscrete (invEquiv p) discreteFin) p
+
+isFinSet→isSet : {X : Type ℓ} → isFinSet X → isSet X
+isFinSet→isSet p = Discrete→isSet (isFinSet→Discrete p)
```
mortberg, line 45: Wouldn't it be slicker to prove this from `isProp-isFinSet` ?

Cubical/Data/FinSet/Properties.agda
```diff
@@ -0,0 +1,253 @@
...
+  isFinSetNonEmpty : isFinSet (NonEmpty (X .fst))
+  isFinSetNonEmpty = isFinSet¬ (_ , isFinSet¬ X)
+
+module _
+  {X : Type ℓ }
+  {Y : Type ℓ'}
+  (f : X → Y) where
+
+  isInjective : Type (ℓ-max ℓ ℓ')
+  isInjective = (a b : X) → ¬ (a ≡ b) → ¬ (f a ≡ f b)
```
mortberg, line 200: This is a roundabout way to define injectivity? We should probably upstream this to a suitable place as well

Cubical/Data/FinSet/Properties.agda
```diff
@@ -0,0 +1,253 @@
...
+
+  isSurjective : Type (ℓ-max ℓ ℓ')
```
mortberg, line 202: See https://github.com/agda/cubical/blob/master/Cubical/Functions/Surjection.agda

Cubical/Data/FinSet/Properties.agda
mortberg: A suitable place would be in `Cubical.Functions`

Cubical/Data/FinSet/Properties.agda
```diff
@@ -0,0 +1,253 @@
...
+  isSurjective = (y : Y) → NonEmpty (fiber f y)
+
+  isBijective : Type (ℓ-max ℓ ℓ')
```
mortberg, line 205: upstream this as well

Cubical/Data/FinSet/Properties.agda
```diff
@@ -0,0 +1,253 @@
...
+  isBijective = isInjective × isSurjective
+
+module _
```
mortberg, line 208: Upstream

mortberg (changes requested):
Great! Thanks for the contribution.

A general remark is that we should reorganize FinSet a bit so that we have a Base file and the `Cubical.Data.FinSet` file just re-exports things.

Also, I think `isFinSet` should be called `isFinite`

kangrongji:
> Great! Thanks for the contribution.
> 
> A general remark is that we should reorganize FinSet a bit so that we have a Base file and the `Cubical.Data.FinSet` file just re-exports things.
> 
> Also, I think `isFinSet` should be called `isFinite`

Really necessary to reorganize the files. However, I think we should keep the name `isFinSet`. Finiteness has certain extensions to higher types and someone will probably define something like `isFinGroupoid` in the near future.  Names as `isFinite` would lead to confusion.

Cubical/Data/FinSet/Examples.agda
kangrongji: Got it.

Cubical/Data/FinSet/Properties.agda
kangrongji: This proof is roundabout actually. But how to use `isProp-isFinSet`?

Cubical/Data/FinSet/Properties.agda
kangrongji: Ad-hoc but direct. I think it's what you can find in any basic textbooks but is only working for Sets. Maybe I should turn to those definitions in `Cubical.Functions`.

kangrongji:
Reorganize the files and switch to the library definition of injective/surjective functions. Looks much better now!

Cubical/Data/FinSet/Properties.agda
mortberg: Can't you just use https://github.com/agda/cubical/blob/master/Cubical/Foundations/Prelude.agda#L462 ?

Cubical/Data/FinSet/Properties.agda
```diff
@@ -0,0 +1,151 @@
...
+
+-- useful implications
+
+isFinSetFin : ∀ {n} → isFinSet (Fin n)
+isFinSetFin = ∣ _ , pathToEquiv refl ∣
+
+isFinSetUnit : isFinSet Unit
+isFinSetUnit = ∣ 1 , Unit≃Fin1 ∣
+
+isFinSet→Discrete : {X : Type ℓ} → isFinSet X → Discrete X
```
mortberg, line 47: Can't you replace `X` by `A` here as you already have `A` as a variable? The same comment applies to a few other results in this file

Cubical/Data/FinSet/Properties.agda
mortberg: Maybe we should upstream this to the Base file as it's what motivates calling these `isFinSet` instead of just `isFinite`

Cubical/Data/FinSet/Properties.agda
```diff
@@ -0,0 +1,151 @@
...
+    from (n , squash p q i) = isPropIsFinSet (from (n , p)) (from (n , q)) i
+
+FinSet≡FinSet' : FinSet ℓ ≡ FinSet' ℓ
+FinSet≡FinSet' = ua (isoToEquiv (iso to from to-from from-to))
+  where
+    to : FinSet ℓ → FinSet' ℓ
+    to (A , isFinSetA) = A , transport isFinSet≡isFinSet' isFinSetA
+
+    from : FinSet' ℓ → FinSet ℓ
+    from (A , isFinSet'A) = A , transport (sym isFinSet≡isFinSet') isFinSet'A
```
mortberg, line 122: Looking closer at `to`/`from` here makes me think that we should just put the two `to`/`from` in `isFinSet≡isFinSet'` on the top-level with good names and use them here. Seems silly to package the function up in an equivalence, use univalence, then transport, instead of just applying the functions

mortberg (approved):
I had a few comments which I already posted, once they are fixed I'll be happy to merge

mortberg:
> > Great! Thanks for the contribution.
> > A general remark is that we should reorganize FinSet a bit so that we have a Base file and the `Cubical.Data.FinSet` file just re-exports things.
> > Also, I think `isFinSet` should be called `isFinite`
> 
> Really necessary to reorganize the files. However, I think we should keep the name `isFinSet`. Finiteness has certain extensions to higher types and someone will probably define something like `isFinGroupoid` in the near future. Names as `isFinite` would lead to confusion.

Ok, I buy this argument! Especially if we move the proof that `FinSet A` is actually a set to `Base`

Cubical/Data/FinSet/Properties.agda
mortberg: Right, but a much nicer version for constructive math is to express it positively as: `f a ≡ f b → a ≡ b`

There are many uses of this in the library, but not expressed as `isInjective f`...

kangrongji:
I have moved `isFinSet→isSet` to `Base` with a direct argument. Fixed the variables naming too.

Cubical/Data/FinSet/Properties.agda
```diff
@@ -101,51 +98,55 @@ isPropIsFinSet' {A = A} (n , equivn) (m , equivm) =
...
 
-    from : FinSet' ℓ → FinSet ℓ
-    from (A , isFinSet'A) = A , transport (sym isFinSet≡isFinSet') isFinSet'A
+FinSet→FinSet' : FinSet ℓ → FinSet' ℓ
+FinSet→FinSet' (A , isFinSetA) = A , transport isFinSet≡isFinSet' isFinSetA
 
-    to-from : ∀ A → to (from A) ≡ A
-    to-from A = Σ≡Prop (λ _ → isPropIsFinSet') refl
+FinSet'→FinSet : FinSet' ℓ → FinSet ℓ
+FinSet'→FinSet (A , isFinSet'A) = A , transport (sym isFinSet≡isFinSet') isFinSet'A
```
mortberg, line 118: How about:

```
FinSet'→FinSet (A , isFinSet'A) = A , isFinSet'→isFinSet isFinSet'A
```

Cubical/Data/FinSet/Properties.agda
```diff
@@ -101,51 +98,55 @@ isPropIsFinSet' {A = A} (n , equivn) (m , equivm) =
...
+isFinSet'→isFinSet (n , ∣ isFinSet-A ∣) = ∣ n , isFinSet-A ∣
+isFinSet'→isFinSet (n , squash p q i) = isPropIsFinSet (isFinSet'→isFinSet (n , p)) (isFinSet'→isFinSet (n , q)) i
+
+isFinSet≡isFinSet' : isFinSet A ≡ isFinSet' A
+isFinSet≡isFinSet' {A = A} = hPropExt isPropIsFinSet isPropIsFinSet' isFinSet→isFinSet' isFinSet'→isFinSet
 
-    from : FinSet' ℓ → FinSet ℓ
-    from (A , isFinSet'A) = A , transport (sym isFinSet≡isFinSet') isFinSet'A
+FinSet→FinSet' : FinSet ℓ → FinSet' ℓ
+FinSet→FinSet' (A , isFinSetA) = A , transport isFinSet≡isFinSet' isFinSetA
```
mortberg, line 115: How about:

```
FinSet→FinSet' (A , isFinSetA) = A , isFinSet→isFinSet' isFinSetA
```

Cubical/Data/FinSet/Properties.agda
kangrongji: haha now I got why you asked me to move these functions up.

kangrongji:
Btw there are so many different aspects of finite sets (e.g. `Fin`, `SumFin`, `FinInd`, `FinData`) and the definitions are separated in several files/folders. Maybe we should clean & reorganize them for some time and make things more coherent.

mortberg:
> Btw there are so many different aspects of finite sets (e.g. `Fin`, `SumFin`, `FinInd`, `FinData`) and the definitions are separated in several files/folders. Maybe we should clean & reorganize them for some time and make things more coherent.

Indeed, that would be very welcome in a future PR!

kangrongji:
Wait, I think I should add some facts about cardinality so that others can do combinatorics immediately.

mortberg (approved):
I think we should merge this now and any additions come in another PR. It will be much easier for me to review the new stuff that way.

So should I merge?

kangrongji:
@mortberg It's Ok to merge now. I still need sometime to add new stuff.

mortberg:
Great, thanks for the contribution!

PR #621, opened 2021-11-16
Clean up RingSolver and extract NatSolver

felixwellen:
This PR contains a long overdue clean up of the RingSolver code.
It removes everything from Cubical/Algebra/RingSolver, that is not needed in the latest version of the ring solver. Additionally, an old prototype of the solver, that also worked for semirings is salvaged in the form of the NatSolver - a solver which is specialized to the natural numbers. The idea is, that this will probably be useful and pretty easy to apply. A simple reflection interface is included. 
There is some duplication left between RingSolver and NatSolver (at least some functions in the reflection interfaces), but I think it would be not worth the effort to deduplicate.


Cubical/Algebra/NatSolver/HornerForms.agda
```diff
@@ -0,0 +1,100 @@
...
+  0H : {n : ℕ} → IteratedHornerForms (ℕ.suc n)
+  _·X+_ : {n : ℕ} → IteratedHornerForms (ℕ.suc n) → IteratedHornerForms n
+                  → IteratedHornerForms (ℕ.suc n)
+
+eval : (n : ℕ) (P : IteratedHornerForms n)
+       → Vec ℕ n → ℕ
+eval ℕ.zero (const r) [] = r
+eval .(ℕ.suc _) 0H (_ ∷ _) = 0
+eval (ℕ.suc n) (P ·X+ Q) (x ∷ xs) =
+  (eval (ℕ.suc n) P (x ∷ xs)) · x + eval n Q xs
```
ecavallo, line 44: I think the `ℕ` argument here could be left implicit, as well as in all the lemmas in `NatSolver.EvaluationHomomorphism`. Doesn't really matter, but it would make the code more readable (esp in the latter case). I guess the same change is possible for the ring solver?

Cubical/Algebra/NatSolver/EvaluationHomomorphism.agda
```diff
@@ -1,91 +1,66 @@
 {-# OPTIONS --safe #-}
-module Cubical.Algebra.RingSolver.EvaluationHomomorphism where
+module Cubical.Algebra.NatSolver.EvaluationHomomorphism where
```
ecavallo, line 2: The corresponding module in the ring solver is called `EvalHom`. Can we use the same name for both? (I don't care which one.)

ecavallo (changes requested):

Cubical/Algebra/NatSolver/HornerForms.agda
felixwellen: Ohno. Why didn't I do this?
Thanks - it works for the NatSolver, I'll do the same for the RingSolver now.

felixwellen: Also works for the RingSolver

Cubical/Algebra/RingSolver/HornerEval.agda
```diff
@@ -22,14 +22,14 @@ private
...
 eval : {A : RawAlgebra ℤAsRawRing ℓ′}
-       (n : ℕ) (P : IteratedHornerForms A n)
+       {n : ℕ} (P : IteratedHornerForms A n)
        → Vec ⟨ A ⟩ n → ⟨ A ⟩
-eval {A = A} ℕ.zero (const r) [] = RawAlgebra.scalar A r
-eval {A = A} .(ℕ.suc _) 0H (_ ∷ _) = RawAlgebra.0r A
-eval {A = A} (ℕ.suc n) (P ·X+ Q) (x ∷ xs) =
+eval {A = A} (const r) [] = RawAlgebra.scalar A r
+eval {A = A} {n = .(ℕ.suc _)} 0H (_ ∷ _) = RawAlgebra.0r A
+eval {A = A} {n = ℕ.suc n} (P ·X+ Q) (x ∷ xs) =
```
ecavallo, line 29: Same here.

ecavallo (changes requested):
Almost ready to merge, just noticed a few unnecessarily bound implicit arguments.

ecavallo:
Thanks!

felixwellen:
Thank you for the review!

Issue #625, opened 2021-11-18
Remove the notion of "Precategory"


mortberg:
I think we should remove the notion of "precategory" (categories without the assumption that the hom-types are sets) from the library, or at least decouple them from the theory of categories with hom-sets.

At the moment the hom-set assumption is separate in the form of the `isCategory` predicate which is like in UniMath which has a `has_homsets` predicate. This turned out to be a bad design decision as the extra generality with hom-types doesn't have very interesting theory. Plus, when developing 1-category theory it gets very annoying to pass around the "has homsets" assumption explicitly. Another more major problem is that when one has a category and want to use an operation defined for precategories then one needs to forget the "has homsets" assumption, which then often confuses the system, leading to many arguments having to be passed around explicitly (yet another issue related to this is that Agda doesn't have implicit projections making this even more painful than in Coq...). The problems get super annoying when trying to reason about iterated functor categories or slice categories.

What do other people think?

barrettj12:
I think for sure that `Category` should be the primary notion. I've never seen the word "Precategory" outside this library

kangrongji:
Not really relevant to this issue... But I think what correspond to the usual notion of 1-category is the "univalent category"  or the "category" with Set-objects (in library's  terms), and they should be equivalent someway. Is there any proof about these in library? Also, does the library have defined the so-called Rezk completion? I mean the procedure which makes category into univalent category.

mortberg:
No, we don't have the Rezk completion formalized. It would be a nice contribution.

I'll close this PR now that https://github.com/agda/cubical/pull/629 is merged

anuyts:
As the pull-request above already showed, if anyone is ever going to use precategories more seriously, then this is going to lead to a lot of code duplication.
An alternative could be to define a type `HCategory ℓ ℓ' h h' ` indexed by two levels of type `Level` and two h-levels of type `Maybe ℕ` (with nothing standing for infinity); each time one for objects and one for morphisms. Then we can define `Category ℓ ℓ' = HCategory ℓ ℓ' nothing (just 2)`. This way in a first iteration we can program for categories, but whoever needs precategories, or categories with an h-set of objects, or setoids, can start generalizing rather than duplicating the code.

PR #629, opened 2021-11-19
Decouple Precategory from Category and use Category in the CT library

mortberg:
Precategories are now in a separate file and almost nothing relies on it (I think it's only the precategory of types that does at the moment). 

I did not yet fix cospans and limits. I might redefine the cospan category to rely on Fin instead to get a slicker definition

This PR fixes issue https://github.com/agda/cubical/issues/625

### DIFF
--- a/Cubical/Algebra/Ring/Base.agda
+++ b/Cubical/Algebra/Ring/Base.agda
@@ -132,7 +132,6 @@ makeRing 0r 1r _+_ _·_ -_ is-setR assoc +-rid +-rinv +-comm ·-assoc ·-rid ·-
        (makeIsRing is-setR assoc +-rid +-rinv +-comm
                    ·-assoc ·-rid ·-lid ·-rdist-+ ·-ldist-+ )
 
--- TODO: why are the Ring structures split and organized like this?
 record IsRingHom {A : Type ℓ} {B : Type ℓ'} (R : RingStr A) (f : A → B) (S : RingStr B)
   : Type (ℓ-max ℓ ℓ')
   where
--- a/Cubical/Algebra/Ring/Base.agda
+++ b/Cubical/Algebra/Ring/Base.agda
@@ -190,6 +190,9 @@ isPropIsRingHom R f S = isOfHLevelRetractFromIso 1 IsRingHomIsoΣ
                                   (isPropΠ2 λ _ _ → isSetRing (_ , S) _ _)
                                   (isPropΠ λ _ → isSetRing (_ , S) _ _))
 
+isSetRingHom : (R : Ring ℓ) (S : Ring ℓ') → isSet (RingHom R S)
+isSetRingHom R S = isSetΣSndProp (isSetΠ (λ _ → isSetRing S)) (λ f → isPropIsRingHom (snd R) f (snd S))
+
 RingHomPathP : (R S T : Ring ℓ) (p : S ≡ T) (φ : RingHom R S) (ψ : RingHom R T)
              → PathP (λ i → R .fst → p i .fst) (φ .fst) (ψ .fst)
              → PathP (λ i → RingHom R (p i)) φ ψ
--- a/Cubical/Algebra/Ring/Base.agda
+++ b/Cubical/Algebra/Ring/Base.agda
@@ -132,6 +132,7 @@ makeRing 0r 1r _+_ _·_ -_ is-setR assoc +-rid +-rinv +-comm ·-assoc ·-rid ·-
        (makeIsRing is-setR assoc +-rid +-rinv +-comm
                    ·-assoc ·-rid ·-lid ·-rdist-+ ·-ldist-+ )
 
+-- TODO: why are the Ring structures split and organized like this?
 record IsRingHom {A : Type ℓ} {B : Type ℓ'} (R : RingStr A) (f : A → B) (S : RingStr B)
   : Type (ℓ-max ℓ ℓ')
   where
--- a/Cubical/Categories/Adjoint.agda
+++ b/Cubical/Categories/Adjoint.agda
@@ -12,7 +12,7 @@ open import Cubical.Foundations.Isomorphism
 open Functor
 
 open Iso
-open Precategory
+open Category
 
 {-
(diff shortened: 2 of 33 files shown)


mortberg:
A nice consequence of this is that we do not have to rely on any instance arguments for the hom-set assumptions as they're part of the structure

Cubical/Algebra/Ring/Base.agda
```diff
@@ -132,6 +132,7 @@ makeRing 0r 1r _+_ _·_ -_ is-setR assoc +-rid +-rinv +-comm ·-assoc ·-rid ·-
        (makeIsRing is-setR assoc +-rid +-rinv +-comm
                    ·-assoc ·-rid ·-lid ·-rdist-+ ·-ldist-+ )
 
+-- TODO: why are the Ring structures split and organized like this?
```
mortberg, line 135: @felixwellen  @mzeuner Do you have any clue about this? It seems like a bad idea to me as Agda will have trouble to figure out the ring structures

mortberg: Maybe @ecavallo knows?

felixwellen: Can you explain the problem a bit more? I don't get what problems will be hard to solve for Agda and which aspect of the definition should be the cause. 
If the type of 'dist' is a problem - that could be something I did a long time ago and would do different now.

mortberg: I would like to change this to:
```
record IsRingHom {R : Ring ℓ} {S : Ring ℓ'} (f : R .fst → S .fst) : Type (ℓ-max ℓ ℓ')
```

I think if one keeps the ring structure separate from the type then Agda can sometimes get confused and not automatically infer it, forcing us to give it explicitly. I might be wrong, but I think this could happen more often when we consider homomorphisms between complex rings (e.g. polynomial rings over localized rings, etc.)

felixwellen: Oh - I thought that style of defining structures was the result of a discussion about things like that. In the beginning, it was different and now all algebraic structures (as far as I know) are defined like this.

ecavallo: I think it was my decision. The idea is just that it matches the format of a dependent path type (and consequently the format of a displayed univalent relation).

I don't follow how the change you're suggesting would help with inference (but then there's a lot I don't understand about Agda's inference). Note that in the current definition the structures are explicit arguments. I would expect to have to mention them explicitly as often as one needs to provide the endpoints explicitly in functions manipulating dependent paths.

mortberg: Hmm, OK... Can @ecavallo help me remember why we wanted it like this?  Note that I'm really only asking about what the parameters to homomorphisms should be

mortberg: Ah right, I guess the rings should be explicit in my proposal. No matter what, wouldn't this be slightly nicer:

```
record IsRingHom {A : Type ℓ} {B : Type ℓ'} (R : RingStr A) (S : RingStr B) (f : A → B) 
```

?

ecavallo: I guess the order `{A : Type ℓ} {B : Type ℓ'} (f : A → B) (R : RingStr A) (S : RingStr B)` would actually be the closest to dependent paths. Anyway, I don't really prefer one over the other, but switching to a different convention would involve rewriting a lot of code!

ecavallo: The bigger problem I had with the current organization, by the way, is that because `IsRingEquiv` is just a wrapper around `IsRingHom` (instead of `isRingEquiv` being its own record), Agda often forgets all but the function part of the equivalence argument and needs to be reminded explicitly. I was trying to decide what would be best to do with this when I put the algebra stuff to the side and never came back...

mortberg: Yes, this is exactly the kind of problem I've been trying to explain... Similar problems made `CommRingHom` annoying to use in my other PR

ecavallo: OK, but that seems like a problem with the definition of `CommRingHom` which can't be fixed by changing `IsRingHom`? Or am I misunderstanding?

ecavallo: One option I considered was to parameterize records like `IsRingHom` by an arbitrary notion of structure from which a ring could be projection, sort of a "typeclass" approach. Then you could instantiate at rings or commrings, to get the benefits of working with a record that hangs on to all the structure information without having to write everything multiple times:
``
Ringable : (∀ {ℓ} → Type ℓ → Type ℓ) → Typeω
Ringable S = {ℓ : Level} {A : Type ℓ} → S A → RingStr A

RingableRing : Ringable RingStr
RingableRing = idfun _

record IsRingHomGen {Str : ∀ {ℓ} → Type ℓ → Type ℓ} (π : Ringable Str)
  {ℓ} {ℓ'} {A : Type ℓ} {B : Type ℓ'}
  (R : Str A) (f : A → B) (S : Str B)
  : Type (ℓ-max ℓ ℓ')
  where

  -- Shorter qualified names
  private
    module R = RingStr (π R)
    module S = RingStr (π S)

  field
    pres0 : f R.0r ≡ S.0r
    pres1 : f R.1r ≡ S.1r
    pres+ : (x y : A) → f (x R.+ y) ≡ f x S.+ f y
    pres· : (x y : A) → f (x R.· y) ≡ f x S.· f y
    pres- : (x : A) → f (R.- x) ≡ S.- (f x)

IsRingHom = IsRingHomGen RingableRing
IsCommRingHom = IsRingHomGen ...
``
One could presumably do the same thing with the function/equivalence argument. I didn't spend enough time experimenting with this to know it would be a good idea, though.

mortberg: How about we take this discussion to an issue?

mortberg:
I'll remove the TODO comment which caused the long discussion and then merge

PR #634, opened 2021-11-21
Make the category `C` explicit

### LINKED ISSUE #632: `C` should be explicit in `Cubical.Categories.Morphism`
In `Cubical.Categories.Morphism`, some things are defined inside an anonymous module:
```agda
module _ {C : Precategory ℓ ℓ'} where
  ...
  isMonic : (Hom[ x , y ]) → Type (ℓ-max ℓ ℓ')
```
so that when you access these definitions from the outside world, `C` is added as an implicit argument:
```agda
isMonic : {C : Precategory ℓ ℓ'} → {x y : ob C} → (Hom[ x , y ]) → Type (ℓ-max ℓ ℓ')
```
The issue is that when trying to use `isMonic`, Agda seems unable to work out what `C` should be. I suspect this is related to the fact that `Precategory` is a record type, and I'm not sure if this is a bug or a feature. Nonetheless, when referring to `isMonic`, it seems you *always* have to give `C` as an argument:
```agda
f : C [ x , y ]
...
mon : isMonic {C = C} f
```
Clearly, this defeats the purpose of making `C` implicit in the first place. Hence I propose changing the `Cubical.Categories.Morphism` file to
```agda
module _ (C : Precategory ℓ ℓ') where
```
and all instances of `isMonic` (and the other definitions) to
```agda
isMonic C f
```
This is technically a breaking change, but I don't think its impact will be too great, and it's easily fixed.

barrettj12:
Fixes #632. The following definitions have been changed:

- In `Categories/Morphism.agda`:
  - `isMonic`
  - `isEpic`
  - `isSplitMon`
  - `isSplitEpi`
  - `areInv`
  - `isIso`
- In `Categories/Category/Base.agda`:
  - `CatIso`

### DIFF
--- a/Cubical/Categories/Category/Base.agda
+++ b/Cubical/Categories/Category/Base.agda
@@ -48,7 +48,7 @@ infixr 16 comp'
 syntax comp' C g f = g ∘⟨ C ⟩ f
 
 -- Isomorphisms and paths in categories
-record CatIso {C : Category ℓ ℓ'} (x y : C .ob) : Type ℓ' where
+record CatIso (C : Category ℓ ℓ') (x y : C .ob) : Type ℓ' where
   constructor catiso
   field
     mor : C [ x , y ]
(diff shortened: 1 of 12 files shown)


mortberg (approved):
LGTM

PR #635, opened 2021-11-21
Add Hom as instance of AbGroup

barrettj12:


barrettj12:
@mortberg I thought `HomAbStr` was a better name than `HomIsAb`, thoughts?

Cubical/Algebra/AbGroup/Instances/Hom.agda
```diff
@@ -0,0 +1,147 @@
...
+  HomAdd-zero f = GroupHom≡ (funExt λ a → idrB _)
+
+  -- -ₕ is right inverse
+  HomInv-invr : (f : AbGroupHom A B) → (f +ₕ (-ₕ f)) ≡ zero
+  HomInv-invr f = GroupHom≡ (funExt λ a → invrB _)
+
+
+-- Abelian group structure on AbGroupHom A B
+HomAbStr : (A : AbGroup ℓ) → (B : AbGroup ℓ') → AbGroupStr (AbGroupHom A B)
+HomAbStr A B = abgroupstr
```
mortberg, line 137: Please define using copatterns (I think `makeIsAbGroup` can still be used here, so only the top-level need copatterns)

mortberg (changes requested):
I would call it `HomAbGroupStr`. Apart from this and a comment this looks good to me

mortberg (approved):

PR #636, opened 2021-11-21
Add Ab as instance of category

barrettj12:


mortberg (approved):
LGTM

mortberg:
Wait a minute... To be consistent with naming this should be `AbGroupCategory`.

We might consider better naming for instances at some point... But for now let's try to be consistent with the arguably bad naming

barrettj12:
I can change it if you like. However, I do think there is an argument to be made that we should give them their conventional names (**Set**, **Grp**, **Ring**, **Cat**, **Ab**, etc)

mortberg:
> I can change it if you like. However, I do think there is an argument to be made that we should give them their conventional names (**Set**, **Grp**, **Ring**, **Cat**, **Ab**, etc)

Yup, the name of Set and functor categories currently doesn't follow any convention either. We also have instances for algebraic structure that follow different conventions... We should decide for a convention and rename all of these consistently.

barrettj12:
> We should decide for a convention and rename all of these consistently.

Good idea, I can open an issue on this.

mortberg:
> > We should decide for a convention and rename all of these consistently.
> 
> Good idea, I can open an issue on this.

Yes, please!

PR #637, opened 2021-11-21
Add (co)products

barrettj12:


mortberg:
See how I just refactored the pullbacks: https://github.com/agda/cubical/pull/650 
How about doing the same here?

barrettj12:
@mortberg could you be more specific please? I'm not sure what exactly you'd like me to do - I think my code already looks pretty similar to your refactored code.

mortberg:
State the universal property first just as a definition and not as a record and then prove that it's a property. Then use it to define the product/equalizer structure. Finally, give a definition of what it means for a category to have and merely have products/equalizers: https://github.com/agda/cubical/pull/650/files#diff-9d881e0b34956b31c18090a3faa7d4b14e7adea7aa927fd6914a2ea2cc839ae8R53

barrettj12:
Great, I can do that.

mortberg:
> Great, I can do that.

Perfect! Just so you know: I plan to rewrite the file about terminal objects and limits soon as well.

barrettj12:
Cool. Just so *you* know, I defined coproducts and coequalisers too (as products/equalisers in the opposite category) and was planning to PR them after this one got approved.

mortberg:
I actually don't think it's good to define colimits as limits in C^op. This leads to more headache than its worth (for example record projections will have the wrong names, e.g. the maps into the coproduct will be called projections...). My experience from UniMath is that it's much better to define all limits and colimits as directly as possible.

It's also kind of arbitrary that the colimits are limits in C^op, we could just as well take colimits as primitive and define limits as colimits in C^op. No matter what it's better to keep them separate

barrettj12:
@mortberg have refactored to match your code. I removed equalisers since I don't need them for what I'm doing, and added coproducts.

Cubical/Categories/Colimits/Coproduct.agda
```diff
@@ -0,0 +1,43 @@
...
+
+module _ (C : Category ℓ ℓ') where
+  open Category C
+
+  module _ {x y x+y : ob}
+           (i₁ : Hom[ x , x+y ])
+           (i₂ : Hom[ y , x+y ]) where
+
+    isCoproduct : Type (ℓ-max ℓ ℓ')
+    isCoproduct = ∀ (z : ob) (f₁ : Hom[ x , z ]) (f₂ : Hom[ y , z ]) →
```
mortberg, line 24: Can't the `z` be implicit?

Cubical/Categories/Colimits.agda
```diff
@@ -0,0 +1,4 @@
+{-# OPTIONS --safe #-}
+module Cubical.Categories.Colimits where
```
mortberg, line 2: I would not add another subpackage for colimits but just have them all in limits. A colimit is just a form of limit after all

Cubical/Categories/Colimits/Coproduct.agda
```diff
@@ -0,0 +1,43 @@
...
+        ∃![ f ∈ Hom[ x+y , z ] ] (i₁ ⋆ f ≡ f₁) × (i₂ ⋆ f ≡ f₂)
+
+    isPropIsCoproduct : isProp (isCoproduct)
+    isPropIsCoproduct = isPropΠ3 λ z f₁ f₂ → isPropIsContr
+
+
+  record Coproduct (x y : ob) : Type (ℓ-max ℓ ℓ') where
+    field
+      x+y : ob
```
mortberg, line 33: I would not call this `x+y` (what is `x` and `y`?). Maybe `binCoprodOb` is better?

Cubical/Categories/Colimits/Coproduct.agda
```diff
@@ -0,0 +1,43 @@
...
+
+    isCoproduct : Type (ℓ-max ℓ ℓ')
+    isCoproduct = ∀ (z : ob) (f₁ : Hom[ x , z ]) (f₂ : Hom[ y , z ]) →
+        ∃![ f ∈ Hom[ x+y , z ] ] (i₁ ⋆ f ≡ f₁) × (i₂ ⋆ f ≡ f₂)
+
+    isPropIsCoproduct : isProp (isCoproduct)
+    isPropIsCoproduct = isPropΠ3 λ z f₁ f₂ → isPropIsContr
+
+
+  record Coproduct (x y : ob) : Type (ℓ-max ℓ ℓ') where
```
mortberg, line 31: I would call this as well as the file `BinCoproduct` as it's just the binary version. I would prefer to reserve `Coproduct` for  general indexed coproducts

Cubical/Categories/Limits/Product.agda
```diff
@@ -0,0 +1,46 @@
+-- Binary products
+{-# OPTIONS --safe #-}
+
+module Cubical.Categories.Limits.Product where
```
mortberg, line 4: Please call these `BinProduct`

Cubical/Categories/Colimits/Coproduct.agda
```diff
@@ -0,0 +1,43 @@
...
+    field
+      x+y : ob
+      i₁ : Hom[ x , x+y ]
+      i₂ : Hom[ y , x+y ]
```
mortberg, line 35: I would call these `binCoprodIn₁` and `binCoprodIn₂`

Cubical/Categories/Colimits/Coproduct.agda
```diff
@@ -0,0 +1,43 @@
...
+      univ : isCoproduct i₁ i₂
```
mortberg, line 36: This could be `isBinCoproductBinCoprod` or maybe we use `univProp` and use this for all universal properties...

Cubical/Categories/Limits/Product.agda
```diff
@@ -0,0 +1,46 @@
...
+
+    isProduct : Type (ℓ-max ℓ ℓ')
+    isProduct = ∀ (z : ob) (f₁ : Hom[ z , x ]) (f₂ : Hom[ z , y ]) →
+        ∃![ f ∈ Hom[ z , x×y ] ] (f ⋆ π₁ ≡ f₁) × (f ⋆ π₂ ≡ f₂)
+
+    isPropIsProduct : isProp (isProduct)
+    isPropIsProduct = isPropΠ3 λ z f₁ f₂ → isPropIsContr
+
+
+  record Product (x y : ob) : Type (ℓ-max ℓ ℓ') where
```
mortberg, line 31: I suggest similar renamings as for the binary coproducts

Cubical/Categories/Limits/Product.agda
```diff
@@ -0,0 +1,46 @@
...
+
+
+  Products : Type (ℓ-max ℓ ℓ')
+  Products = (x y : ob) → Product x y
+
+  hasProducts : Type (ℓ-max ℓ ℓ')
+  hasProducts = ∥ Products ∥
+
+-- -- TODO: define products using isLimit from Cubical.Categories.Limits.Base
+-- --   and show this gives the same thing
```
mortberg, line 46: You can remove this comment

mortberg (changes requested):
A lot of suggestions for new names

Cubical/Categories/Colimits/Coproduct.agda
barrettj12: True

Cubical/Categories/Colimits/Coproduct.agda
barrettj12: I think `univProp` is okay

mortberg: Yeah, I agree

mortberg (approved):
Looks good to me!

I would love to get the equalizers as well in another PR

barrettj12:
> I would love to get the equalizers as well in another PR

Sadly I won't have time to put that together, but it should be easy enough to copy the structure of these files.

mortberg:
> > I would love to get the equalizers as well in another PR
> 
> Sadly I won't have time to put that together, but it should be easy enough to copy the structure of these files.

Ok, no problem! I'll do it myself

PR #638, opened 2021-11-21
Add discrete categories

barrettj12:
I've defined discrete categories using `Hom[ x , y ] = x ≡ y`.

I'm now trying to show that a function `f : A → ob C` determines a `Functor (Discrete A) C`, but I'm having trouble showing that this preserves composition. Maybe someone else can help.

### DIFF
--- a/Cubical/Categories/Instances/Discrete.agda
+++ b/Cubical/Categories/Instances/Discrete.agda
@@ -18,23 +18,23 @@ private
 open Category
 
 -- Discrete category
-Discrete : hGroupoid ℓ → Category ℓ ℓ
-Discrete A .ob = A .fst
-Discrete A .Hom[_,_] a a' = a ≡ a'
-Discrete A .id = refl
-Discrete A ._⋆_ = _∙_
-Discrete A .⋆IdL f = sym (lUnit f)
-Discrete A .⋆IdR f = sym (rUnit f)
-Discrete A .⋆Assoc f g h = sym (assoc f g h)
-Discrete A .isSetHom {x} {y} = A .snd x y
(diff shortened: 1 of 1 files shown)


kangrongji:
I think you can use J-rule twice, so you only need to treat composition of identity morphisms.

barrettj12:
Makes sense, but I'm not quite sure how to apply `J` here.

I think the problem might be that given `p : x ≡ y` and `q : y ≡ z`
- `F p` and `F (p ∙ q)` are obtained by transport from `id {F x}`
- `F q` is obtained from `id {F y}`

so it's not clear how to compose them.

Cubical/Categories/Instances/Discrete.agda
```diff
@@ -0,0 +1,55 @@
...
+open import Cubical.Foundations.Transport
+
+private
+  variable
+    ℓ ℓC ℓC' : Level
+
+open Category
+
+-- Discrete category
+Discrete : hGroupoid ℓ → Category ℓ ℓ
```
mortberg, line 21: Rename to `DiscreteCategory` ?

mortberg (changes requested):
Just a name change suggestion

Cubical/Categories/Instances/Discrete.agda
mortberg: My suggestion would be more consistent with https://github.com/agda/cubical/blob/master/Cubical/Categories/Instances/Rings.agda. If someone wants a shorter name they can rename when importing.

barrettj12: Okay, will do.

mortberg (approved):

PR #639, opened 2021-11-22
[ makefile ] added a build target for plain `make` without -W error

Saizan:


mortberg (approved):
So the Github action will run `make test` and just `make` will only run build?

If that's the case then it looks good to me and we can merge once the CI is done!

Saizan:
Exactly, that's what will happen.

andreasabel:
Complemented on the Agda side by: 
- https://github.com/agda/agda/pull/5855

PR #642, opened 2021-11-22
Category of categories

barrettj12:
I thought I'd like to add the category of (small) categories to `Categories/Instances`. Most of the work was already done elsewhere, but I ran into a roadblock trying to prove
```agda
C D : Category ℓ ℓ'
    ⊢ isSet (Functor C D)
```
I don't think I know enough (cubical) type theory to know whether this is even _true_, let alone write a proof in Agda. Maybe someone with more expertise can help.

### DIFF
--- a/Cubical/Categories/Instances/Categories.agda
+++ b/Cubical/Categories/Instances/Categories.agda
@@ -1,0 +1,25 @@
+-- Category of (small) categories
+{-# OPTIONS --safe #-}
+
+module Cubical.Categories.Instances.Categories where
+
+open import Cubical.Categories.Category.Base
+open import Cubical.Categories.Functor.Base
+open import Cubical.Categories.Functor.Properties
+-- open import Cubical.Foundations.HLevels
+open import Cubical.Foundations.Prelude
+
+
+-- Definition of Cat
+module _ (ℓ ℓ' : Level) where
+  open Category
+
+  Cat : Category (ℓ-suc (ℓ-max ℓ ℓ')) (ℓ-max ℓ ℓ')
+  Cat .ob = Category ℓ ℓ'
+  Cat .Hom[_,_] = Functor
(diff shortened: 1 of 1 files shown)


felixwellen:
I don't expect it to be true, since there is no requirement on D to have a 0-type of objects. That's not a recipe for a counterexample, since the type of functors is some Sigma-type based on that. If you want to treat Categories as 0-type based structures, you could ask D to be a 0-type for your construction. Then you should get a 0-type of functors. However, that would mean, that you forget natural transformations and consider naturally isomorphic functors as something completely different.
But in some way you have to do somehting forgetful, if you want to have 'the' 1-category of small 1-categories (since, naturally, I would say it is a 2-category). Another way would be to ask for all the small categories to be univalent, then (not completely sure here...) you would get a 1-type of functors. And you could truncate that to get a 0-type (with all operations and laws you need).

barrettj12:
Hmmm. I think I'll just make it a `Precategory` then. If anyone ends up needing a category of categories later, they can think more about what is required for the homs to be sets.

Cubical/Categories/Instances/Categories.agda
```diff
@@ -0,0 +1,25 @@
...
+open import Cubical.Categories.Category.Precategory
+open import Cubical.Categories.Functor.Base
+open import Cubical.Categories.Functor.Properties
+open import Cubical.Foundations.Prelude
+
+
+module _ (ℓ ℓ' : Level) where
+  open Precategory
+
+  Cat : Precategory (ℓ-suc (ℓ-max ℓ ℓ')) (ℓ-max ℓ ℓ')
```
mortberg, line 16: I would call this `CatPrecategory`

mortberg (changes requested):
Name change suggestion

mortberg (approved):

PR #647, opened 2021-11-25
Abelianization of groups

HilpAlex:
Construct the abelianization of groups as a higher inductive type, like suggested in [https://arxiv.org/abs/2007.05833](https://arxiv.org/abs/2007.05833).


mortberg:
I find the `comm` relation a bit strange, why not just quotient by `a · b ≡ b · a`?  (My German is not good enough to understand the comment after `proofabelian`... :-) ).

No matter what, it's probably easier to define this using a set quotient as you don't have to define the `elimProp` functions: https://github.com/agda/cubical/blob/master/Cubical/HITs/SetQuotients/Properties.agda#L44

Furthermore, there are some helper lemmas that might help you in defining operations on the quotient: https://github.com/agda/cubical/blob/master/Cubical/HITs/SetQuotients/Properties.agda#L207  

For inspiration you can see how @mzeuner defined ring localization as a set quotient in https://github.com/agda/cubical/blob/master/Cubical/Algebra/CommRing/Localisation/Base.agda

mortberg:
PS: an alternative approach would be to construct the commutator subgroup and prove that it's normal. You then get the abelianization from https://github.com/agda/cubical/blob/master/Cubical/Algebra/Group/QuotientGroup.agda

mortberg:
> PS: an alternative approach would be to construct the commutator subgroup and prove that it's normal. You then get the abelianization from https://github.com/agda/cubical/blob/master/Cubical/Algebra/Group/QuotientGroup.agda

PPS: I just saw that they discuss this in the paper and say that "One could mimic the classical definition, describing G^{ab} as the quotient of G by the subgroup generated by commutators, but this is awkward to work with constructively.". Does anyone know why?

felixwellen:
I think you get something else, if you use the simpler comm. We looked into using SetQuotients, but the relation looked awful. So I guess, this is actually simpler.

Alizter:
@mortberg 
Here is the formalization in Coq that I wrote with Dan: https://github.com/HoTT/HoTT/blob/master/theories/Algebra/AbGroups/Abelianization.v

Going the commutator normal subgroup route is a lot of hassle and we were looking for simpler HITs. Here is one that I did end up formalizing more in line with what you are thinking:
```coq
HIT Abel (G : Group) : hSet :=
    | ab : G -> Abel
    | ab_plus : Abel -> Abel -> Abel    | ab_hom : forall x y, ab (x & y) = ab_plus (ab x) (ab y).
    | ab_comm : forall a b, ab_plus a b = ab_plus b a.
    | ab_assoc : forall a b c, ab_plus a (ab_plus b c) = ab_plus (ab_plus a b) c.
```

The assoc constructor is needed and AFAIK cannot be avoided. Dan's suggestion:
```coq
HIT Abel (G : Group) : hSet :=
| ab : G -> Abel
| ab_comm : forall a b c d, ab ((a & (b & c)) & d) = ab ((a & (c & b)) & d).
```

is a super clever way of getting assoc and comm out of the definition. You can also write it as a coequalizer like we do in the HoTT lib.

PR #650, opened 2021-11-26
Refactor pullbacks

mortberg:
The definition of pullbacks was very deeply nested, so I flattened it. I also express the property of being a pullback square separately now as it's very useful to be able to talk about specific squares being pullbacks.

### DIFF
--- a/Cubical/Categories/Limits/Pullback.agda
+++ b/Cubical/Categories/Limits/Pullback.agda
@@ -16,7 +16,7 @@ private
   variable
     ℓ ℓ' : Level
 
-module _ {C : Category ℓ ℓ'} where
+module _ (C : Category ℓ ℓ') where
 
   open Category C
   open Functor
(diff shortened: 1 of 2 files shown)


Cubical/Categories/Limits/Pullback.agda
```diff
@@ -26,37 +28,41 @@ module _ {C : Category ℓ ℓ'} where
       s₁ : C [ l , m ]
       s₂ : C [ r , m ]
```
mortberg, line 30: I'm not sure that it's worth having a separate record for cospans... What do you think @ecavallo ?

ecavallo: I think it's fine to have. I think the category parameter should be explicit, though.

mortberg: I wonder if it should be explicit to all definitions in the module or not... Hmm

mortberg: I made it explicit

mzeuner:
Is there anything you still want to add? If not, would you mind merging it, because I was planning on putting the new definitions to work for finite limits...

mortberg:
> Is there anything you still want to add? If not, would you mind merging it, because I was planning on putting the new definitions to work...
> 
> Everything in this PR looks good to me btw.

Ah yes, I wanted to add an accessor function for the pullback arrow (the map one gets from the universal property) as well as its properties. I'll add this quickly now and then merge. Next thing on my list is to clean the code about initial/terminal objects

mortberg:
Should be merged as soon as CI is done

PR #651, opened 2021-11-26
Add instructions for getting `make` to work on Windows

### LINKED ISSUE #633: Makefile: `runghc` instead of `runhaskell`
To get `make` to run on my Windows 10 machine, I had to change line 5 of GNUmakefile from
```
RUNHASKELL?=runhaskell
```
to
```
RUNHASKELL?=runghc
```
I'm not sure if this is related to Windows, or to new versions of GHC. I'm running GHC version 9.0.1.

Is `runhaskell` still working for those on Unix?

barrettj12:
Fixes #633 (kinda)

Please squash and merge into a single commit.


MAKEWINDOWS.md
```diff
@@ -0,0 +1,28 @@
+# Instructions for getting `make` to work on Windows
+
+The GNU command `make` does not work on Windows out of the box. Here are the steps I took to get it to work on my Windows 10 machine:
```
mortberg, line 3: Please don't use "I" and instead write something generic like: "Here are some possible steps that can make it to work on Windows 10"

INSTALL.md
```diff
@@ -86,7 +86,9 @@ Once this works go to a suitable directory and run
 > make
 ```
 
-This should compile all of the agda/cubical files. To test that it
+This should compile all of the agda/cubical files. (If you're on Windows, see [here](MAKEWINDOWS.md) for instructions on using `make`).
```
mortberg, line 89: Please keep length of lines short (you can use `M-q` in emacs)

INSTALL.md
```diff
@@ -86,7 +86,9 @@ Once this works go to a suitable directory and run
...
+
+To test that it
```
mortberg, line 91: Why is there a line break?

INSTALL.md
```diff
@@ -167,7 +171,9 @@ Once this works go to a suitable directory and run
 > make
 ```
 
-This should compile all of the agda/cubical files. To test that it
+This should compile all of the agda/cubical files. (If you're on Windows, see [here](MAKEWINDOWS.md) for instructions on using `make`).
```
mortberg, line 174: Please add line break

INSTALL.md
```diff
@@ -167,7 +171,9 @@ Once this works go to a suitable directory and run
...
+
+To test that it
```
mortberg, line 176: Weird line break here as well

INSTALL.md
```diff
@@ -251,7 +257,9 @@ Once this works go to a suitable directory and run
 > make
 ```
 
-This should compile all of the agda/cubical files. To test that it
+This should compile all of the agda/cubical files. (If you're on Windows, see [here](MAKEWINDOWS.md) for instructions on using `make`).
```
mortberg, line 260: Long line

INSTALL.md
```diff
@@ -251,7 +257,9 @@ Once this works go to a suitable directory and run
...
+
+To test that it
```
mortberg, line 262: And the line break again :-)

CONTRIBUTING.md
```diff
@@ -8,6 +8,8 @@ When preparing a PR here are some general guidelines:
   which will generate all required `Everything` files in
   `Cubical/README.agda` and then typecheck the latter file.
 
+  - If you're using Windows, [here](MAKEWINDOWS.md) are some instructions to get the `make` command working.
```
mortberg, line 11: Please don't indent this line and add some line breaks

mortberg (changes requested):
Lots of minor stylistic comments. I don't have windows so cannot test, but I think it's good to have these instructions in case someone else has problems

Saizan:
By the way, I don't think `INSTALL.md` should tell users to use `make` for `agda/cubical`, it's not necessary to "install" the library.

In that case if you install Agda through cabal or stack you won't have to run `make` at all.

MAKEWINDOWS.md
barrettj12: Well, the point was that this worked on my machine, but I can't guarantee it's gonna work on anyone else's. But okay, I will rewrite

barrettj12:
> By the way, I don't think `INSTALL.md` should tell users to use `make` for `agda/cubical`, it's not necessary to "install" the library.
> 
> In that case if you install Agda through cabal or stack you won't have to run `make` at all.

True, it's just to test that the installation worked right? I was thinking you could factor this out into a separate section, since this part seems to be repeated verbatim for each installation method.

Saizan:
Yeah, `make` only tests that the installed `agda` can typecheck all the files of the library.

Btw, I still think it's helpful to know how to get make to work on windows, but probably more for someone contributing so that they can run the tests.

barrettj12:
@mortberg fixed. @Saizan had a good point that this is not necessary in the `INSTALL.md` file, so I've reverted this to the current version.

mortberg (approved):

mortberg:
Looks good to me!

PR #660, opened 2021-12-02
Generalize binary set quotient operations + misc cleaning

ecavallo:


barrettj12:
@ecavallo could you also make it so that the levels don't have to all be the same? e.g.
```agda
A : Type ℓA
R : A → A → Type ℓR
B : Type ℓB
...
```

ecavallo:
This is actually already the case, because of the way variables work. See the Note about "Nested generalized variables" here: https://agda.readthedocs.io/en/latest/language/generalization-of-declared-variables.html.

PR #667, opened 2021-12-14
Pointwise ring and algebra structure

felixwellen:
Define a (Comm-) Ring and Algebra structure on types of the form ``X -> A``.

### DIFF
--- a/Cubical/Algebra/CommRing/Instances/Pointwise.agda
+++ b/Cubical/Algebra/CommRing/Instances/Pointwise.agda
@@ -11,22 +11,26 @@ private
     ℓ : Level
 
 pointwiseRing : (X : Type ℓ) (R : CommRing ℓ) → CommRing ℓ
-pointwiseRing X R =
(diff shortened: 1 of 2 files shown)


Cubical/Algebra/CommRing/Instances/Pointwise.agda
```diff
@@ -0,0 +1,32 @@
...
+private
+  variable
+    ℓ : Level
+
+pointwiseRing : (X : Type ℓ) (R : CommRing ℓ) → CommRing ℓ
+pointwiseRing X R =
+  let open CommRingStr (snd R)
+      isSetX→R = isOfHLevelΠ 2 (λ _ → isSetCommRing R)
+  in (X → fst R) ,
+     (commringstr
```
mortberg, line 18: Probably better to use some copatterns here? Maybe at least for the outermost level (so not the `makeIsCommRing` part)

mortberg (changes requested):
Small suggestion

PR #673, opened 2021-12-16
Finite limits

mzeuner:
Start defining finite limits that are needed for sheaves on lattices. 
This is mainly for @mortberg to use this in some other draft PR's, 
but maybe things should be modified a bit, so I kept this a draft as well for now...


kangrongji:
I am not very sure about your notion of _Finite System_... Is it an inductive definition of all finite diagrams or just coproducts of spans?

mortberg:
Indeed, it's a bit confusing to call these "finite diagrams/limits" as it's just finite diagrams/limits of a very particular shape. Let's discuss a better name tomorrow. I also think the code for general limits should be refactored similarly to how I rewrote the code for pullbacks and terminal/initial objects

Probably some of this code could benefit from https://github.com/agda/cubical/pull/666. It would probably be a lot nicer to define a finite graph of the right shape and then just take the free category.

mzeuner:
Yeah sorry, I should have been more clear about that. I fixed the description and maybe you have some recommendations for a better name?

PR #676, opened 2021-12-16
Terminal object and pullbacks in CommRings (i.e. trivial ring and fibered products)

mortberg:
Also rename `·-comm` to `·Comm` and add a nice lemma for proving unique existence (e.g. universal properties)


mortberg:
@mzeuner I renamed `isSetNat` to `isSetNatTrans` as well here now

mortberg:
Will merge once CI is finished

PR #680, opened 2021-12-17
Examples of finite presentations of CommAlgebra's

felixwellen:
The goal of this PR is, to introduce the notion of a finite presentation (instead of just the Prop 'finitelyPresented') and construct some basic finite presentations.
However, I'm stuck here because of a type checking speed issue. It seems that plugging the algebra ``makeFPAlgebra n relations`` into ``CommAlgebraEquiv`` is the cause of the slowdown. Maybe because the fields of this more complicated CommAlgebra get expanded? I don't know...


mortberg:
Strange... I have no good idea why it's so slow. Did you try making some parts abstract in order to isolate where the slowness is coming from?

felixwellen:
I'm not sure if I tried that.
My guess is, that it could be lots of work to actually expand all the fields of 'makeFPAlgebra ...'.

felixwellen:
Hm. Now it seems to be a lot faster with an abstract 'makeFPAlgebra'... Thanks!

felixwellen:
Now I have a new, very much related tc-speed problem. The 'abstract' term I defined is not really usable in the way I want without introducing the same speed problem again...

Cubical/Algebra/CommAlgebra/Properties.agda
```diff
@@ -25,12 +26,34 @@ open import Cubical.Algebra.Ring
...
 open import Cubical.Algebra.CommAlgebra.Base
 
+open import Cubical.Algebra.CommRing using (CommRing→Ring)
+
 open import Cubical.HITs.PropositionalTruncation
 
 private
   variable
-    ℓ ℓ' ℓ'' ℓ''' : Level
+    ℓ ℓ′ ℓ″ ℓ‴ ℓ' ℓ'' ℓ''' : Level
```
mortberg, line 35: Bad levels?

Cubical/Algebra/CommAlgebra/FGIdeal.agda
```diff
@@ -1,22 +1,37 @@
 {-# OPTIONS --safe #-}
 module Cubical.Algebra.CommAlgebra.FGIdeal where
 open import Cubical.Foundations.Prelude
```
mortberg, line 3: empty line

mortberg (changes requested):
small comments

PR #682, opened 2021-12-20
Terminal algebra

felixwellen:
Construct the terminal R-algbera (=zero ring). Use the terminal ring.


mortberg:
Looks good! Thanks for generalizing the terminal ring. I would suggest you add the category of commutative algebras with its terminal object in `Cubical.Categories.Instances.CommAlgebras` as well.

ice1000:
The CI problem, if I'm right, should be fixed by #690

ice1000:
Yaaas!

felixwellen:
Thanks!

PR #685, opened 2021-12-28
Localization pullback

mzeuner:
This PR contains a proof that if two elements of a ring generate the whole ring as an ideal
then we get a special kind of pullback square involving the localization at those elements.

This is fact will be the cornerstone of proving the sheaf property of the structure sheaf.


mortberg:
Nice! Looks like you fixed all the things we discussed? Should I merge?

mzeuner:
Yes, I made all the changes we discussed, so it  should be fine to merge.

PR #687, opened 2022-01-03
Html in gh-pages

ice1000:
Blah

### DIFF
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -123,3 +123,16 @@ jobs:
             AGDA_EXEC='~/.cabal/bin/agda -W error' \
             FIX_WHITESPACE='~/.cabal/bin/fix-whitespace' \
             RUNHASKELL='~/.ghcup/bin/runhaskell'
+
+      - name: Htmlize cubical
+        run: |
+          make listings \
+            AGDA_EXEC='~/.cabal/bin/agda -W error' \
+            FIX_WHITESPACE='~/.cabal/bin/fix-whitespace' \
+            RUNHASKELL='~/.ghcup/bin/runhaskell'
+
+      - name: Deploy to GitHub Pages
+        uses: JamesIves/github-pages-deploy-action@4.1.8
+        with:
+          branch: gh-pages
+          folder: html


ice1000:
image

Any admin can help me clicking this button and select 'gh-pages'? The generation should be set up successfully.

ice1000:
https://github.com/agda/cubical/tree/gh-pages

Saizan:
Done

ice1000:
https://agda.github.io/cubical/Cubical.Algebra.CommAlgebra.Ideal

brendanzab:
yayyyy thanks @ice1000!

brendanzab:
Might be a good idea to put https://agda.github.io/cubical/Cubical.README.html in the repostory metadata now :)

mortberg:
> Might be a good idea to put https://agda.github.io/cubical/Cubical.README.html in the repository metadata now :)

Yes, can someone make a PR?

brendanzab:
Oh, was more suggesting via the adding it to the repo via the web UI:

Screen Shot 2022-01-04 at 9 43 03 pm

felixwellen:
The CI runs on my PRs from today fail with some error message containing "github-pages-deploy-action". Is this related to this PR's changes?
See:
https://github.com/agda/cubical/runs/4704441227?check_suite_focus=true

brendanzab:
Hmm, it's odd that the pages deployment is being done on CI runs - not sure if that is part of it? I generally put it in a separate workflow that only runs when pushing to the default branch (see the different `ci.yml`/`cd.yml` files [here](https://github.com/brendanzab/ocaml-flake-example/tree/main/.github/workflows)). That way you don't randomly see changes whenever tests are run on a PR branch. Not sure that is the cause of the failure though.

ice1000:
> Hmm, it's odd that the pages deployment is being done on CI runs - not sure if that is part of it? I generally put it in a separate workflow that only runs when pushing to the default branch (see the different `ci.yml`/`cd.yml` files [here](https://github.com/brendanzab/ocaml-flake-example/tree/main/.github/workflows)). That way you don't randomly see changes whenever tests are run on a PR branch. Not sure that is the cause of the failure though. 

It is very expensive to install agda so I just changed the test. I think I know the cause and I'll try to fix it

ice1000:
My guess is that the workflow relies on the pusher's access towards the repo, so @brendanzab 's suggestion should fix it if I'm right

brendanzab:
> It is very expensive to install agda so I just changed the test. I think I know the cause and I'll try to fix it

Hmm, I wonder if you could do something like:

```yaml
      - name: Deploy to GitHub Pages
        if: github.event_name == 'push' && github.ref_name == 'master'
        uses: JamesIves/github-pages-deploy-action@4.1.8
        with:
          branch: gh-pages
          folder: html
```

PR #690, opened 2022-01-05
Only htmlize on master branch

ice1000:
Also promote the rendered version in README w/ some adjectives (glorious, clickable) stolen from the stdlib.

Special thanks to @brendanzab

### DIFF
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -132,6 +132,7 @@ jobs:
             RUNHASKELL='~/.ghcup/bin/runhaskell'
 
       - name: Deploy to GitHub Pages
+        if: github.event_name == 'push' && github.ref_name == 'master'
         uses: JamesIves/github-pages-deploy-action@4.1.8
         with:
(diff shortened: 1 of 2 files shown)


brendanzab:
Nice, glad it actually seems to work in practice!

mortberg:
Amazing! Thanks a lot for fixing this so quickly!

ecavallo:
Is there a reason why only the "Deploy to GitHub Pages" step was restricted to master, and not the "Htmlize cubical" step as well?

ice1000:
> Is there a reason why only the "Deploy to GitHub Pages" step was restricted to master, and not the "Htmlize cubical" step as well?

Reason? Yes. Good? No. One may argue that it makes sense to check if it's possible to generate the html files as nobody said type checking guarantees it, but probably the truth is that I forgot to.

ecavallo:
:smile: I think it's safe enough to assume the generation will work, so I'll make a new PR to add the check there. Looks like it will only save a minute or two of CI time, but we may as well.

PR #691, opened 2022-01-07
Use CommIdeal everywhere

felixwellen:
There are currently two definitions of Ideals in commutative Rings. With @mzeuner, I decided to use 'CommIdeal'. 
This PR implements this decision.


mortberg:
Are there still two definitions in the library or did you remove one of them completely?

felixwellen:
Well, it is a bit more complicated. One of the definitions just used the definition of Ideal in a Ring and I changed the definitions building on that, to use the other one. But in principle there could still be places, where we use Ideals in a CommRing instead of CommIdeals. I replaced everywhere I know of.

mortberg:
@mzeuner @felixwellen  Should this be merged?

felixwellen:
Yes

PR #697, opened 2022-01-19
[ ci ] Use the latest Agda but check `cubical` without `NoEquivWhenSplitting` warnings

L-TChen:
The recent commit https://github.com/agda/agda/commit/7bb14e0114b6b101d9a1e7c656ad4e9e711f3dcb breaks the CI, because the CI uses the releaed version `2.6.2` instead of the head of the `master` branch. 

This PR tries to use a more recent version of Agda, although IMO it should use the most recent commit on `master`.

### DIFF
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -87,6 +87,7 @@ jobs:
 
       - name: Check out Agda from GitHub
         uses: actions/checkout@v2
+        if: ${{ !steps.cache.outputs.cache-hit }}
         with:
           repository: agda/agda
           ref: ${{ env.AGDA_COMMIT }}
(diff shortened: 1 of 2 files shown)


L-TChen:
Urgh, I was misguided by the comment. The variable "AGDA_BRANCH" can only be a git branch or a tag, not a commit hash. Therefore, I propose to use the `master` branch of Agda to check the experimental version of the cubical library.

mortberg:
Do we really want the `noNoEquivWhenSplitting` flag to be set? Seems to pollute interaction in emacs

mortberg:
Or wait, is this just introducing a typo and it should be `NoEquivWhenSplitting`? I'm very confused now...

mortberg:
Oh no wait, I seem to not understand what this PR is supposed to be doing... Anyway, whatever it's doing seems to be introducing problems when working in emacs

PR #701, opened 2022-01-24
Revert "[ ci ] Use the latest Agda but check `cubical` without `NoEquivWhenSplitting` warnings"

mortberg:
Reverts agda/cubical#697


mortberg:
I think we should revert this PR now until we understand what it's supposed to do and why it's causing problems. Also, we're intentionally not using the latest version of Agda

PR #704, opened 2022-01-24
Blakers Massey: Pushout-Pullback form 

aljungstrom:
1. Moving `Pushout` out of BlakersMassey.agda and into Pushout.Base.agda and renaming it to `PushoutGen`.
2. Proof that the usual pushout is a special case of the above (`IsoPushoutPushoutGen`).
3. The "usual" formulation Blakers Massey concerning the pullback of a pushout diagram (`toPullbackConnected`).


aljungstrom:
@kangrongji you know the Blakers-Massey proof better than I do, so please tell me if you think the pushout/pullback square version is an immediate corollary, in which case I have overcoded a bit...:-) It took me some time figuring out how to translate things, so I might have done it in an overcomplicated way.

kangrongji:
@aljungstrom Both the type `Q` and the pushout-pullback can be seen as types over `X × Y`, of which the fibers over `(x,y)` are `Q x y` and `inl x  ≡ inr y` respectively. And the map `push` is just the induced map on fibers. The pushout/pullback version follows from the fact that, a map is `n`-connected if and only if the induced maps on fibers are all `n`-connected. 

The equivalences involved are a bit technical but direct. Unfortunately, I don't know the last fact has been already written somewhere in the library. I think this general fact need some work.

aljungstrom:
@kangrongji in my proof i use the fact that the generliased pushout in Blakers-Massey (let's call it `PushoutGen`) can be used to expressed normal pushouts by setting `Q x y = Σ[ a ∈ A ] (f a ≡ x) (g a ≡ c)` (which seems like a useful thing to have in the library anyways). One can get this this construction into Blakers-Massey form and thus get that `push` is `(n+m)`-connected. From that, it's really just a matter of proving that the fibres of `push` are the same as the fibres of the map `h : A -> Pullback inl inr`. Most of it is just refl, but there is still some juggling with paths which makes it a bit longer than I expected.

kangrongji:
Oh I see. Basically you're doing the same thing but combine several smaller steps in one shot, so I won't say that your argument is over-complicated. I think breaking the argument into smaller pieces can make some paths much more simpler.

Btw, the fibres of the map` h : A -> Pullback inl inr` are equivalent to the fibres of `push`, actually.

aljungstrom:
Ah ok, you're right. Currently typing up your suggested proof. No path algebra involved:-) Might be longer in the end, but some of the lemmas I need are useful ones anyways.

> Btw, the fibres of the map` h : A -> Pullback inl inr` are equivalent to the fibres of `push`, actually.
Hehe yeah, but I was lazy proved something using J, which would make the section part of this proof horrible;-)

Cubical/Data/Sigma/Properties.agda
```diff
@@ -130,6 +130,18 @@ discreteΣ {B = B} Adis Bdis (a0 , b0) (a1 , b1) = discreteΣ' (Adis a0 a1)
         ... | (no ¬q) = no (λ r → ¬q (subst (λ X → PathP (λ i → B (X i)) b0 b1) (Discrete→isSet Adis a0 a0 (cong fst r) refl) (cong snd r)))
     discreteΣ' (no ¬p) = no (λ r → ¬p (cong fst r))
 
+lUnit×Iso : ∀ {ℓ} {A : Type ℓ} → Iso (Unit × A) A
```
aljungstrom, line 133: \ell not needed

Cubical/Homotopy/BlakersMassey.agda
```diff
@@ -22,17 +22,24 @@ open import Cubical.Foundations.Transport
...
 open import Cubical.Data.Nat hiding (elim)
 open import Cubical.Data.Sigma
+open import Cubical.Data.Unit
 
 open import Cubical.HITs.Truncation renaming (hLevelTrunc to Trunc)
 
 open import Cubical.Homotopy.Connected
 open import Cubical.Homotopy.WedgeConnectivity
 
+open import Cubical.HITs.Pushout hiding (PushoutGenFib)
```
aljungstrom, line 41: Move up

Cubical/Homotopy/BlakersMassey.agda
```diff
@@ -672,3 +675,133 @@ module BlakersMassey {ℓ₁ ℓ₂ ℓ₃ : Level}
...
+private
+  shuffleFibIso₁ :
+       {ℓ ℓ' ℓ'' : Level} {A : Type ℓ} {B : Type ℓ'} {C : Type ℓ''}
+       (f : A → B) (g : A → C) (b : B)
+     → Iso (Σ[ c ∈ C ] Σ[ a ∈ A ] (f a ≡ b) × (g a ≡ c))
+             (Σ[ a ∈ A ] ((Σ[ c ∈ C ] (g a ≡ c)) × (f a ≡ b)))
+  fun (shuffleFibIso₁ f g x) (c , a , p , q) = a , (c , q) , p
+  inv (shuffleFibIso₁ f g x) (a , (c , q) , p) = c , a , p , q
+  rightInv (shuffleFibIso₁ f g x) _ = refl
+  leftInv (shuffleFibIso₁ f g x) _ = refl
```
aljungstrom, line 713: Check if this can be restated using other lemmas

Cubical/Data/Sigma/Properties.agda
aljungstrom: nor is A:-)

aljungstrom:
@mortberg Ready!

PR #705, opened 2022-01-24
Euclidean Algorithm and Bézout Identity for Integers

kangrongji:
This PR contains what mentioned in the title. If anyone defines the concept of Bézout domain, the proof can be immediately modified to show `ℤ` is among them.

I'm on half way to prove the existence of Smith normal form, based on this little PR.


mortberg:
Nice! For some reason I thought there was already some divisibility theory in the library, but that's maybe for naturals? I'll just take a quick look now and comment more once I have time for a proper readthrough

Cubical/Data/Int/Divisibility.agda
```diff
@@ -0,0 +1,308 @@
...
+-- Bézout and Euclidean Domain
+
+record Bézout (m n : ℤ) : Type where
+  constructor bezout
+  field
+    coef₁ : ℤ
+    coef₂ : ℤ
+    gcd   : ℤ
+    identity : coef₁ · m + coef₂ · n ≡ gcd
+    isGCD : (gcd ∣ m) × (gcd ∣ n)
```
mortberg, line 119: Don't you need to say that `gcd` is the biggest such divisor as well?

kangrongji: That's equivalent to `identity` + `isGCD` and I think this form is simpler and useful. But It seems the name `isGCD` not exactly precise. Should I include the definition of GCD and add a proof of equivalence?

kangrongji:
Yes. The library only has that for natural numbers and no Bézout identity. And also the results are a bit chaotic, so I somewhat packed them up for integers.

Cubical/Data/Int/Divisibility.agda
kangrongji: I add an equivalence proof.

Cubical/Data/Int/Divisibility.agda
```diff
@@ -0,0 +1,436 @@
...
+
+module _
+  (b : Bézout m n) where
+
+  private
+    g = b .gcd
+
+  gcdIsGCD : k ∣ m → k ∣ n → k ∣ g
+  gcdIsGCD {k = k} p q =
+    subst (k ∣_) (b .identity) (∣-+ (∣-right· {n = b .coef₁} p) (∣-right· {n = b .coef₂} q))
```
kangrongji, line 230: I added an equivalence proof.

Cubical/Data/Int/Divisibility.agda
mortberg: Right, maybe calling it `isDivisor` would be better?

Cubical/Data/Int/Divisibility.agda
mortberg: Great!

mortberg (approved):
Just one small renaming needed and then I would be happy to merge

kangrongji:
I rename it to `isCD` for isCommonDivisor, which is used in the files of `ℕ`. @mortberg

PR #461, opened 2020-10-23
CommRing structure on Integers from CommRing structure of BiInvInt (Part 1)

mchristianl:
This PR uses `transferAxioms` from `Cubical.Structures.Axioms` to obtain a `CommRing` structure on

- **`Cubical.HITs.Ints.BiInvInt`** (this CommRing structure is given directly)
- **`Cubical.Data.Int`**
- **`Cubical.Data.DiffInt`**
- **`Cubical.HITs.Ints.QuoInt`**

Not in this PR (and remaining for #404) are `CommRing` structures for

- _(`Cubical.HITs.Ints.DeltaInt`)_
- _(`Cubical.HITs.Ints.HAEquivInt`)_
- _(`Cubical.HITs.S1`)_

I'll need some more time to learn more about the actual "cubical" Agda, especially about hcomp, to adress the remaining ones. To mitigate [upcoming bitrod](https://agda.zulipchat.com/#narrow/stream/260790-cubical/topic/SIP/near/214292567) I've bundled the progress so far.

For getting an overview of the definitions that we would currently have and that might become superfluous, I've made a list for `Int.Properties`, `BiInvInt.Properties`, `QuoInt.Properties` and `DiffInt.Properties`. A `✓` denotes that this definition is directly used by `Cubical.Algebra.CommRing.Integers` and a `(✓)` means that this definition is a dependency of a directly used definition (which I just checked a bit manually).

The changes mostly affect `QuoInt` where a lot of properties like `·-assoc` would not be necessary anymore when importing the `CommRing` structure from `Cubical.Algebra.CommRing.Integers`.

## Int.Properties

||name|signature|
|---|---|---|
|(✓)|`abs`|`Int → ℕ`|
|(✓)|`Sign`|`Type₀`|
|(✓)|`spos`|`= Bool.false`|
|(✓)|`sneg`|`= Bool.true`|
||`_·ˢ_`|`Sign → Sign → Sign`|
|(✓)|`sgn`|`Int → Bool`|
|(✓)|`signed`|`Sign → ℕ → Int`|
|✓|`signed-zero`|`∀ s → signed s 0 ≡ 0`|
|(✓)|`signed-inv`|`∀ a → signed (sgn a) (abs a) ≡ a`|
|(✓)|`sucPred`|`∀ i → sucInt (predInt i) ≡ i`|
|(✓)|`predSuc`|`∀ i → predInt (sucInt i) ≡ i`|
[...]

PR #706, opened 2022-01-24
(WIP) Int is a Commutative Ring.

xekoukou:
I had to make extensive changes  to the Ring , CommRing structures. I haven't checked if it breaks things.

I  will update this  when the structures are stabilized by the authors of this library.

I split  the structures into RawRingStr RingStr and CommRingStr. And I  updated the DUARel accordingly.

Then, I used the  proofs that BiIntZ is a comm. ring , to  do the same for Z, by using the DUARels.

### DIFF
--- a/Cubical/Algebra/CommRing/Base.agda
+++ b/Cubical/Algebra/CommRing/Base.agda
@@ -29,31 +29,23 @@ private
     ℓ ℓ' : Level
 
 record IsCommRing {R : Type ℓ}
-                  (0r 1r : R) (_+_ _·_ : R → R → R) (-_ : R → R) : Type ℓ where
+                  (ringStr : RingStr R) : Type ℓ where
 
   constructor iscommring
 
+  open RawRingStr (RingStr.rawRingStr ringStr) public
+  open IsRing (RingStr.isRing ringStr) public
+  
   field
-    isRing : IsRing 0r 1r _+_ _·_ -_
     ·Comm : (x y : R) → x · y ≡ y · x
-
-  open IsRing isRing public
 
 record CommRingStr (A : Type ℓ) : Type (ℓ-suc ℓ) where
 
   constructor commringstr
 
   field
-    0r         : A
-    1r         : A
-    _+_        : A → A → A
-    _·_        : A → A → A
-    -_         : A → A
-    isCommRing : IsCommRing 0r 1r _+_ _·_ -_
-
-  infix  8 -_
-  infixl 7 _·_
-  infixl 6 _+_
+    ringStr : RingStr A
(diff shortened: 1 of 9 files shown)


xekoukou:
Related https://github.com/agda/cubical/pull/461
and
https://github.com/agda/cubical/pull/461/files#r577129000

felixwellen:
I got an error on running make:
     Checking Cubical.ZCohomology.GroupStructure (/home/felix/repos/cubical/Cubical/ZCohomology/GroupStructure.agda).
/home/felix/repos/cubical/Cubical/ZCohomology/GroupStructure.agda:23,35-57
The module Cubical.Algebra.Group doesn't export the following:
  ℤ

xekoukou:
If you want to look at what I did, only check this [file](https://github.com/xekoukou/cubical/blob/IntCommRing/Cubical/Algebra/Ring/Instances/Int.agda)

and maybe [this](https://github.com/xekoukou/cubical/blob/0552b4966f564ca3a80fe075a5cd7393a589322d/Cubical/Algebra/Ring/Base.agda#L251), where I split the Duarel into two parts.

I didn't update everything , that is why make complains, also there are merge conflicts.

Before merging, someone needs to decide which will be the default structures, rawmonoid, monoid, commmonoid, etc.

I can't do this myself.

mortberg:
Isn't the main result already in the library: https://github.com/agda/cubical/blob/master/Cubical/Algebra/CommRing/Instances/Int.agda#L12 ? 

The reorganization of the structures is something we've been thinking about doing and @ecavallo started doing it for all algebraic structures but never finished...

mortberg:
> Isn't the main result already in the library: https://github.com/agda/cubical/blob/master/Cubical/Algebra/CommRing/Instances/Int.agda#L12 ?
> 
> The reorganization of the structures is something we've been thinking about doing and @ecavallo started doing it for all algebraic structures but never finished...

Oh wait, you're giving the Ring instance. I guess it would be better to make it the same way as the CommRing instead of transporting it. My experience is that the non-higher version of Int is actually the easiest to work with, so I think it should be the default which all other definitions should inherit their properties from

mortberg:
This has been open for very long and I don't think we've reached any consensus about it. The main result of the PR is now in the library, but proved directly so I'm going to close this now. 

The reorganization of the algebraic structures seems like a very good idea, but it should be done in separate PRs after some careful considerations to ensure that inference and automation works properly. Also, for the moment we've added some new lemmas with the current design that should make the proofs in this PR easier to streamline without rewriting the algebraic hierarchy: https://github.com/agda/cubical/blob/master/Cubical/Algebra/Ring/Base.agda#L265

PR #698, opened 2022-01-21
Long exact sequence of homotopy groups + Generator of π₃S²

aljungstrom:
This PR contains

1. The long exact sequence of homotopy groups (and corresponding loop spaces). It also contains some information about how the maps in the sequence look when they are transported to the (Sⁿ→∙A) definition of homotopy groups. Right now exactness is stated in the usual way by having a bunch of separate theorems on the form Ker⊂Im and Im⊂Ker, like in Mayer-Vietrois and Gysin. Anders and I have been discussing adding a reasonable definition of exact sequences for some time now. When that's a reality, things will look nicer.
2. A proof of π₃S²≅ℤ together with the fact that the Hopf map is the generator of π₃S².
3. A proof of πₙSⁿ≅ℤ together with the fact that the identity map is the generator of πₙSⁿ (needed above). The construction of the Iso is somewhat roundabout to make the characterisation of the generator easier. The characterisation of the generator is longish for π₂S² with, but easy in all other cases. Hopefully in the future this step will hold simply by computation.

PR #707, opened 2022-01-25
Clean the summary file for pi_4(S^3) = Z/2Z and prove two small general lemmas that were left as holes

mortberg:
Proving the "mini lemmas" required me to prove some lemmas that I think could be done nicer with some of the theory of https://github.com/agda/cubical/pull/705. Also a lot of the results I proved ended up in ZActions, but should probably be moved and better integrated with the new theory about divisibility for Z. Maybe @kangrongji would be up for doing this?

@aljungstrom : This will cause some conflicts in the Summary file with https://github.com/agda/cubical/pull/698. I think it would be easier to merge this PR first and then just fix the conflicts in this one file of your bigger PR, what do you think?

It could also be nice to keep the summary module with its arguments and just instantiate the main result afterwards. This way the summary file could be used as a connection between the formalization and the main results in Guillaume's thesis. What do you think?

### DIFF
--- a/Cubical/Homotopy/Group/Pi4S3/Summary.agda
+++ b/Cubical/Homotopy/Group/Pi4S3/Summary.agda
@@ -50,15 +50,15 @@ private
 [_]× (f , g) = [ f ∣ g ]π'
 
 -- Some type abbreviations (unproved results)
-π₄S³≅ℤ/something : GroupEquiv (π 3 𝕊²) ℤ → Type₁
-π₄S³≅ℤ/something eq =
+π₄S³≡ℤ/something : GroupEquiv (π 3 𝕊²) ℤ → Type₁
+π₄S³≡ℤ/something eq =
   π 4 𝕊³ ≡ ℤ/ abs (eq .fst .fst [ ∣ idfun∙ _ ∣₂ , ∣ idfun∙ _ ∣₂ ]×)
 
 -- Summary of the last steps of the proof
 module π₄S³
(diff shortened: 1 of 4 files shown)


aljungstrom:
Nice! Merge away, I'll fix the conflicts:-)

Cubical/Homotopy/Group/Pi4S3/Summary.agda
```diff
@@ -1,182 +1,87 @@
...
-open import Cubical.HITs.Sn
-open import Cubical.HITs.SetTruncation
+private
+  variable
+    ℓ : Level
 
-open import Cubical.Algebra.Group
-  renaming (ℤ to ℤGroup ; Bool to BoolGroup ; Unit to UnitGroup)
-open import Cubical.Algebra.Group.ZAction
+-- TODO: ideally this would not be off by one in the definition of π'Gr
```
mortberg, line 39: @aljungstrom How hard would it be to change this?

kangrongji:
I don't think divisibility could help much. Only when a systematic treatment of finitely presented abelian group shows up (even for groups generated by one element), these results would find natural places to settle and they can be deduced quickly from general results...

mortberg:
> I don't think divisibility could help much. Only when a systematic treatment of finitely presented abelian group shows up (even for groups generated by one element), these results would find natural places to settle and they can be deduced quickly from general results...

You're probably right. I'll leave these results in the ZActions file for now then and hopefully one day develop some more general theory

PR #708, opened 2022-01-27
MonoidSolver

fabianmasato:
Implement a solver with reflection interface for monoids similar to the RingSolver. Also uses ideas from the standard library.


ecavallo:
Thanks!

PR #712, opened 2022-02-01
π₄S³≅ℤ/2 (to be cleaned up...)

aljungstrom:
The main result is at the end of `Cubical.Homotopy.Group.Pi4S3.Summary`. The proof is currently far from readable but at least it type-checks... I thought it would at worth creating a draft for now.

I diverged from the summary file a bit in the end. Here comes a small outline of the last steps of the proofs (we are following Brunerie pretty closely at this point). We work entirely on the RHS of the isomorphism π₄S³ ≅ π₃X where X = 1 ← S³ → S².

1. Apply the long exact sequence of homotopy groups and Blaker's Massey to 1 ← S³ → S². This gives us a sequence 
π₃S³≅π₃P → π₃S² → π₃X → 1, where P comes from taking fibres of the inclusions inl, inr : ... → (1 ← S³ → S²) .
2. Note that the induced map π₃S³ → π₃S² maps the generator to the whitehead product [g,g] : π₃S², where g is the generator of π₂S². 
3. Prove that if the Hopf Invariant of [g,g] is ±2, then π₃X ≅π₃S²/2 ≅Z/2.
4. Prove that Hopf Invariant of [g,g] is ±2.

The three first steps uses a lot of a very specific definition of exact sequences G -> H - M -> N with 4 groups and all maps explicitly stated in the type signature. This makes transporting stuff easier, but we should probably develop some more general theory about exact sequences. The same thing applies the third step -- it is done by proving that for any sequence Z -> Z ->  G -> 1 where the first map takes 1 to ±2, we have that G≅Z/2. This should probably be generalised for arbitrary n (that would make it fit better into the original summary file).

I'll clean this up in the coming weeks. Any input on how to structure things is very welcome!


aljungstrom:
I'll keep it closed until it's prettier

kangrongji:
We need more homological algebra, that's the way out.

PR #713, opened 2022-02-02
Add 2 argument version of ua→

shlevy:


mortberg:
Thanks,  but couldn't this easily be generalized to two equivalences instead of just one?

shlevy:
@mortberg Done

mortberg:
Thanks!

PR #718, opened 2022-02-11
James Construction (Only Basics)

kangrongji:
This PR contains the definition of James construction, some operations, and two properties:

- James X ≃ Ω Σ X, for connected pointed type X;
- James X₊ ≃ List X, for any type X, where X₊ means "X together with a new point".

Have you written down these properties already @aljungstrom ? Any plan of more codes on James?


aljungstrom:
Sorry for the late reply! I think I did the loop space bit, but from what I can remember, your code is a lot prettier:-) No plans for now, but I'm still very much interested in it if you want to discuss some details in the future.

kangrongji:
Of course! I'm planning to fully formalize it, actually I failed months ago when I was on my first try.

PR #719, opened 2022-02-18
Free Commutative Monoid

guilhermehas:
I added Free Commutative Monoid in the library.
To prove that it is isomorphic to the associative list, I had to add some auxiliary functions there.


mortberg (approved):
Very nice! Thanks for the contribution

Cubical/HITs/FreeComMonoids/Properties.agda
```diff
@@ -0,0 +1,96 @@
...
+FCM→AL∘AL→FCM≡id = AL.ElimProp.f (trunc _ _) refl
+  λ x n {xs} p → ^-id x n (AL→FCM xs) ∙ cong (⟨ _ , _ ⟩∷_) p
+
+FCMon≃AssocList : AssocList A ≃ FreeComMonoid A
+FCMon≃AssocList = isoToEquiv (iso AL→FCM FCM→AL AL→FCM∘FCM→AL≡id FCM→AL∘AL→FCM≡id)
+
+AssocList≃FCMon : FreeComMonoid A ≃ AssocList A
+AssocList≃FCMon = isoToEquiv (iso FCM→AL AL→FCM FCM→AL∘AL→FCM≡id AL→FCM∘FCM→AL≡id)
+
+FCMon≡AssocList : AssocList A ≡ FreeComMonoid A
```
mortberg, line 95: Wait, this name is a bit weird... It should be `AssocList≡FreeComMonoid`. Similarly for the two definitions above.

mortberg: Great, can you also fix `FCMon≃AssocList`? And remove `AssocList≃FCMon`? I think it's unnecessary to have both when we can just use `invEquiv` to turn one of them around

guilhermehas: I have already made these changes.

PR #720, opened 2022-02-19
Adding natural numbers defined by their associativity property

guilhermehas:
I have defined the natural numbers using their associativity property and proved that they are isomorphic to NatPlusOne in the library. I explained in more detail the data structure on my blog:
https://guilhermehas.github.io/posts/associativity-natural-numbers.html


mortberg (approved):
Looks good to me! Do you have any use cases in mind for this?

guilhermehas:
It is not so useful. But it can be used to define a non-empty vector by its associativity property too.

PR #722, opened 2022-02-22
James Construction (Inductive Version)

kangrongji:
Ahh...Finally all holes filled. This PR contains an inductive definition of James, formalizing the main part of the chapter on James construction in Brunerie's paper. Instead of applying pushouts iteratively, I found a Higher Inductive Family that satisfies exactly the same pushout-formulation. @aljungstrom  @mortberg 

The results are collected in the file `Cubical.HITs.James.Inductive`.

I tried to make a more structural proof of the equivalence between the inductive and non-inductive version. The resulting is rather long... But I think except for a few crucial points, it's not very difficult and it shouldn't be of this length. The problem is basically:

- If one modifies the constructors of an HIT just a little, the result could stay equivalent to the original type. I think this phenomenon is very common in the world of HoTT, since the existence of many paths allows you to do many such modifications. But even if they seem  "obviously" the same, the proof could still be lengthy to show they are the same.

  (~~An example for this is in `Cubical.HITs.James.Inductive.Reduced`, see the comments before the proof of `𝕁 ≃ 𝕁Alt`~~ Update: This part has been resolved)

- Need J rule to construct some cubes otherwise hard to describe by explicit path manipulation. Unfortunately, if one has to apply J rule iteratively, lacking of regularity may lead to some over-complication. It happens both when you apply it or to show that putting refls will give back the initial input. You can see the points in `Cubical.HITs.James.Inductive.Coherence`.

- Also, definitional equality is still lacking, but that's not the key point.  I think the major problem is to find better ways to treat degeneracy. They arise more frequently and having more diverse look when people ascend to higher dimensions. A large portion of the proof is devoted just to correct the endpoints.


kangrongji:
Btw, does indexed family compute correctly in cubical agda now?

kangrongji:
Thanks to @tomjack who makes me realize I've made stupid mistake... The file ``Cubical.HITs.James.Inductive.Reduced`` now has been simplified from 800 lines to less than 200 lines.

kangrongji:
What's with the CI? @mortberg It says `All install methods for ghc 8.6.5 failed`.

ice1000:
> Btw, does indexed family compute correctly in cubical agda now?

It's still wip afaict

ice1000:
> What's with the CI? @mortberg @ecavallo @ice1000 It says `All install methods for ghc 8.6.5 failed`.

Should be a problem of Agda. I'll try to investigate

kangrongji:
I revise the file `Cubical.HITs.James.Inductive.Coherence` so that it seems less horrible.

ecavallo (approved):
Looks good to me! Maybe someone can find more simplifications some day, but for now this is good to have.

PR #724, opened 2022-02-22
Delooping of Group as HIT

marcinjangrzybowski:
code type checks,
closely follows code-decode (Cubical.HITs.S1.Base) and existing delooping of Z_2 (Cubical.HITs.Delooping.Two)
todo : 
 - celanup
 - better naming
 - simplyfing Code definition
 - add abstract defitnions where possible
 - show equivalence with library definition of loopspace (existing eqivalence is to non-tuncated version of loopspace)

### DIFF
--- a/Cubical/Algebra/Group/Instances/SetsAutomorphism.agda
+++ b/Cubical/Algebra/Group/Instances/SetsAutomorphism.agda
@@ -1,0 +1,71 @@
+{-# OPTIONS --safe #-}
+module Cubical.Algebra.Group.Instances.SetsAutomorphism where
+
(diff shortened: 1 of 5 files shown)


kangrongji:
I remember there is some stuff on this in library already. It's in `Cubical.HITs.EilenbergMacLane1`. Is that related to your work?

marcinjangrzybowski:
I was not aware of the whole `Cubical.Algebra.Group.EilenbergMacLane1` where all of this is already proved :) thanks for pointing this out!

kangrongji:
I think you have proved more things than the library. You can add your new results in.

PR #725, opened 2022-02-23
remove obsolete definition of homotopy groups

felixwellen:
There were two definitions of homotopy groups. I removed the one that is never used (and had a wrong index) - there were a couple of unused imports, though.


mortberg:
Thanks!

PR #726, opened 2022-02-23
General cohomology: covariance in coefficients

felixwellen:
The coefficients are dependent spectra and pointwise morphisms of those induce homomorphisms of cohomology groups


felixwellen:
Too outdated -> easier to do from scratch -> closing.

PR #727, opened 2022-02-23
A shorter definition of isHAEquiv.com-op

howsiyu:
Am I missing anything for preferring the shorter definition?


mortberg (approved):
Very nice! Or was there some special reason to write the unfolded version? @ecavallo

ecavallo:
Probably not.

PR #732, opened 2022-03-08
Remove some unused imports

MatthiasHu:
This PR just removes some unused imports in various files.

Most of them were found by applying [agda-unused](https://github.com/msuperdock/agda-unused) to individual files (because it aborts with an error whenever it hits stuff like "record module instance applications"). I did not remove any `open import ... public`.

### DIFF
--- a/Cubical/Algebra/RingSolver/RawAlgebra.agda
+++ b/Cubical/Algebra/RingSolver/RawAlgebra.agda
@@ -2,7 +2,6 @@
 module Cubical.Algebra.RingSolver.RawAlgebra where
 
 open import Cubical.Foundations.Prelude
-open import Cubical.Data.Sigma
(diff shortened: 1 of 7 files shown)


mortberg (approved):
Looks good to me! Will merge once CI is finished

thomas-lamiaux:
@MatthiasHu to have an idea of the importance of the problem in the library, how many files do you thing you have tested ?

MatthiasHu:
I think I ran agda-unused on all .agda files (using find and xargs) and searched the output for "import". But agda-unused aborted immediately for many (most?) files because it hit something it didn't understand (possibly in dependencies of the respective module).


On 19 May 2022 18:02:59 CEST, Thomas Lamiaux ***@***.***> wrote:
***@***.*** to have an idea of the importance of the problem in the library, how many files do you thing you have tested ?

PR #733, opened 2022-03-11
Terminal objects in the categories of CommRings/Algebras

mzeuner:
This PR removes a duplicate definition of the terminal object in the category of R-algebras and makes the definition consistent with the one for commutative rings.

Just one question: Do we want to prove the universal property of the terminal object in `Categories.Instances.Comm...` as it is done now, or move it to `Algebra.Comm... .Instances.Unit` and then just package things up categorically using `Terminal`
in `Categories.Instances`


felixwellen:
Is it on purpose, that you keep the version of the terminal comm algebra, that does not use the terminal ring?

mzeuner:
No, if you prefer the explicit mention of the UnitCommRing I can do that. But in either case the carrier type is Unit* and all the operations and laws are trivial, so I figured it doesn't matter too much.

felixwellen:
Yes, I would prefer that. But, yes, I also think it doesn't matter much ;-) But I guess in general it is good style to remove logical duplication and I guess it gives us one useful definitional equality.

mzeuner:
You're right, changed it accordingly.

mzeuner:
Where should we put the proof of the universal property though?

felixwellen:
Categories/Instances/CommAlgebra/ seems natural to me. But I don't have a strong opinion on this one.

felixwellen:
Categories/Instances/CommAlgebra/ seems natural to me. But I don't have a strong opinion on this one.

PR #737, opened 2022-03-21
Dependent lists

anuyts:
This is not the primary thing that the cubical library was missing, but I needed them so why not contribute :-)
In the near future, I'm likely to make more contributions of similar nature.


xekoukou:
I think, this can be abstracted to heterogeneous lists, like in idris.

[HVEC](https://github.com/idris-lang/Idris-dev/blob/49594c67b4f5d42fda5ee9f8c129d0240a54b825/libs/base/Data/HVect.idr#L10)

anuyts:
I would say heterogeneous lists are a special case, where `B` is the identity.

ecavallo:
Looks good, thanks!

Issue #736, opened 2022-03-21
Product and BinProduct are identical


anuyts:
The files `Cubical.Categories.Constructions.Product` and `Cubical.Categories.Constructions.BinProduct` are identical.

PR #738, opened 2022-03-21
Remove Product, which is a copy of BinProduct

anuyts:
Remove Cubical.Categories.Constructions/Functor.Product, which is a copy of BinProduct. Adapt references accordingly. Solves #736.


ecavallo:
These duplicates keep turning up in Categories... thanks for the catch.

PR #741, opened 2022-03-24
The Brunerie number is ±2 - direct proof

aljungstrom:
This PR contains a direct proof of the fact that the Brunerie number is ±2 (in fact, with my construction, it's -2). The proof itself is relatively straightforward (see Cubical.Homotopy.Group.Pi4S3.QuickProof -- you may suggest a better name...) but relies on some machinery concerning the equivalence `Sⁿ * Sᵐ ≃ Sⁿ⁺ᵐ`. The idea is to formulate π₃(A) as `π*₃(A) := ∥ S¹ * S¹ →∙ A ∥₂` (with an explicitly defined multiplication) and analyse the term `η₁ : π₃*(S²) `defining the Brunerie number as a member of this group. We can then prove directly that it gets mapped to -2 under what is pretty much the usual equivalence `π₃(S²) ≅ ℤ`.

We get a bunch of terms corresponding to η on the way:
`η₁ : π₃*(S²),  η₂ : π₃*(S¹ * S¹),  η₃ : π₃*(S³) , η₄ : π₃(S³)`
which we can try to send to ℤ. This gives equivalent, but hopefully simpler, computations of the Brunerie number (I haven't tried this yet).


mortberg (approved):
Amazing stuff! Merging once the CI is finished

Issue #743, opened 2022-03-26
Reflection API changing in Agda 2.6.3


andreasabel:
The current `master` of `cubical` does not build with Agda `master`, breaking here: 
https://github.com/agda/cubical/blob/43e0bf07f06bc4a713f3c55d97a13922b426813e/Cubical/Reflection/Base.agda#L51
`extendContext` expects one more argument now:
```agda
extendContext    : ∀ {a} {A : Set a} → String → Arg Type → TC A → TC A
```
Agda's reflection API changed in https://github.com/agda/agda/issues/5715 (unreleased, scheduled for Agda 2.6.3).

Maybe you should maintain a branch that builds with latest Agda `master`.  This branch could be merged into `master` once you bump to a new release of Agda (from `master`).  Currently we (Agda devs) cannot really test (from the Agda side) whether `cubical` and `Agda` are still in sync.  If such a branch existed, we could pin our `cubical` submodule to this branch and run Agda on it from our testsuite/CI.

andreasabel:
Looks there was some effort to move to a more recent version of Agda, but it was reverted:
- #697 
- #701

mortberg:
Ah, yes indeed the current version of the library is supposed to build on the latest release of Agda to not force us to update our Agda installations all the time when working on the library. Having a branch that builds on Agda master would indeed be ideal. How about calling it `Agda-master` or something?

MatthewDaggitt:
If you wanted to be consistent with the standard library, we call it `experimental` (see our [docs](https://github.com/agda/agda-stdlib/blob/master/notes/updating-experimental.txt))

andreasabel:
@mortberg @MatthewDaggitt 
> If you wanted to be consistent with the standard library, we call it `experimental` (see our [docs](https://github.com/agda/agda-stdlib/blob/master/notes/updating-experimental.txt?rgh-link-date=2022-03-26T19%3A18%3A25Z))

Maybe `standard-library` could rename it to `Agda-master` instead?
`experimental` is ambiguous, and quite frankly, I am often confused.  It could refer to experimental features either in Agda or the standard-library...
`Agda-master` is a clearly purposed name.
(Too bad `git` doesn't (allow to) attach comments to branches.)

L-TChen:
Having a branch that builds with the latest development version of Agda is ideal, but for this issue, it also makes sense to remove `extend*Context` (#745).

felixwellen:
This is outdated by now, cubical typechecks with agda 2.6.3

Issue #703, opened 2022-01-24
What to do about reverted PRs and compatibility with 2.6.2


mortberg:
@L-TChen I just reverted two of your PRs:

https://github.com/agda/cubical/pull/701
https://github.com/agda/cubical/pull/702

as they broke compatibility with 2.6.2 which is the current version we're using to develop the library. I hope this won't cause any new problems with the CI.

Before merging any PRs that break compatibility with 2.6.2 we should discuss if we want to do this. I don't think we should do this as it's such a headache to make sure that everyone is using the same version of Agda and it's much easier if we're all just using the latest official release. Once 2.6.3 is out we can update the library

L-TChen:
It is not just agda/cubical itself but also agda/agda is checking `cubical` as part of CI. So, if the library on the `master` is not compatible with the `master` branch of [agda/agda](https://github.com/agda/agda), the Agda's CI will break too. 

I can think of at least two possible "solutions"

1. Ignore `cubical` entirely during Agda's CI. 
2. Create an `experimental` branch like the standard library [agda-stdlib](https://github.com/agda/agda-stdlib) so that the `master` maintain the compatibility with `2.6.2` and the `experimental` branch accumulate commits compatible with the lastest commit, i.e. `2.6.3` but the stable release `2.6.2`.

L-TChen:
> Agda developers with that problem just made an ad-hoc branch to reference from agda/master.

If Agda 2.6.3 has breaking changes which are not compatible with any commit on agda/cubical, then it won't work.

In this particular case, though, agda/cubical defines
https://github.com/agda/cubical/blob/1c56f5801ebdcb11563c180ada0d662525c75e7f/Cubical/Reflection/Base.agda#L49-L51
which is not used anywhere else. It is safe to remove this function so that the `agda/agda` on the master branch can still check `agda/cubical` without any issue. 

However, IMO it is wise to have a branch like `experimental` to check if the cubical library is compatible with the latest Agda.

Saizan:
I meant an ad-hoc branch of agda/cubical, containing the commit that agda/master can reference.

But I guess let's agree to call that branch `experimental`, so that if there are multiple compatibility changes to make they are collected in one place, and `cubical` developers can merge it into `master` when the new Agda version releases.

We should probably write about this in `CONTRIBUTING.md`

L-TChen:
Aha, I see. Thanks for the clarification.

> We should probably write about this in CONTRIBUTING.md

[...]

PR #745, opened 2022-03-27
[ re #703, #743 ] Remove `extend*Context`

L-TChen:
This is the only function incompatible with the development version 2.6.3 of Agda. It is not used elsewhere in the library.

Removing it makes the library compatible with Agda 2.6.3 again.

### DIFF
--- a/Cubical/Reflection/Base.agda
+++ b/Cubical/Reflection/Base.agda
@@ -46,10 +46,6 @@ hlam str t = R.lam R.hidden (R.abs str t)
 
 newMeta = R.checkType R.unknown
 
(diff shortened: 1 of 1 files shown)


andreasabel:
> Removing it makes the library compatible with Agda 2.6.3 again.

Having a CI here that tests `cubical` with Agda `master` would be the perfect verification of this claim.

L-TChen:
Here you are. https://github.com/L-TChen/agda/runs/5710346221?check_suite_focus=true

mortberg (approved):
Very good!

PR #746, opened 2022-03-27
Improve and add a few path operators

anuyts:


Cubical/Foundations/Prelude.agda
```diff
@@ -261,7 +261,7 @@ subst2 : ∀ {ℓ' ℓ''} {B : Type ℓ'} {z w : B} (C : A → B → Type ℓ'')
         (p : x ≡ y) (q : z ≡ w) → C x z → C y w
 subst2 B p q b = transport (λ i → B (p i) (q i)) b
 
-substRefl : (px : B x) → subst B refl px ≡ px
+substRefl : ∀ {B : A → Type ℓ} {x} → (px : B x) → subst B refl px ≡ px
```
mortberg, line 264: Why add these to the type?

mortberg (approved):
Just a question, then I would be happy to merge

Cubical/Foundations/Prelude.agda
anuyts: I think I needed to explicitly pass the value for `A` somewhere. The type of `B` defines a name for `A`.

PR #747, opened 2022-03-27
Indexed W-types

anuyts:
Indexed W-types, a characterization of their path structure, and a proof that they inherit their hlevel from the fibres of the shape type.


ecavallo (approved):
Thanks!

PR #754, opened 2022-03-31
Make the ring solver work in more situations

felixwellen:
This PR contains two small upgrades/fixes for the ring solver:
- Adapt the approach from the monoid solver in the stdlib and also match structure which has the same normal form as the record field names of CommRingStr
- Allow the argument of the solver macro to be more things - so far only a *variable* of type CommRing worked...


felixwellen:
To repeat what has been said on zulip here: This problem is now also fixed by the PR.

PR #756, opened 2022-04-04
Some useful transport results from the book

Aqissiaq:
Lemma 2.3.10, theorem 2.11.3 and lemma 2.11.2 from the hott book.

The special cases (lemma 2.11.2) can probably be removed

### DIFF
--- a/Cubical/Foundations/Transport.agda
+++ b/Cubical/Foundations/Transport.agda
@@ -152,3 +152,49 @@ overPathFunct : ∀ {ℓ} {A : Type ℓ} {x y : A} (p q : x ≡ x) (P : x ≡ y)
 overPathFunct p q =
   J (λ y P → transport (λ i → P i ≡ P i) (p ∙ q) ≡ transport (λ i → P i ≡ P i) p ∙ transport (λ i → P i ≡ P i) q)
     (transportRefl (p ∙ q) ∙ cong₂ _∙_ (sym (transportRefl p)) (sym (transportRefl q)))
+
+-- substituting over a composition (P ∘ f) along p is equivalent to
+-- substituting over P along (cong f)
+-- lemma 2.3.10 in The Book
+substFunComp : ∀ {ℓ ℓ'} {A B : Type ℓ} {a a' : A}
+                 → (f : A → B) → (P : B → Type ℓ') → (p : a ≡ a') → (u : P (f a))
+                 → subst (P ∘ f) p u ≡ subst P (cong f p) u
+substFunComp f P p u = J (λ x q → (subst (P ∘ f) q u) ≡ (subst P (cong f q) u))
+  (substRefl {B = P ∘ f} u ∙ sym (substRefl {B = P} u)) p
+
+
+-- substition over families of paths
+-- theorem 2.11.3 in The Book
+substInPaths : ∀ {ℓ} {A B : Type ℓ} {a a' : A}
+                 → (f g : A → B) → (p : a ≡ a') (q : f a ≡ g a)
+                 → subst (λ x → f x ≡ g x) p q ≡ sym (cong f p) ∙ q ∙ cong g p
+substInPaths {a = a} f g p q =
+  J (λ x p' → (subst (λ y → f y ≡ g y) p' q) ≡ (sym (cong f p') ∙ q ∙ cong g p'))
+    p=refl p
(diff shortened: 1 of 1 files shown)


Cubical/Foundations/Transport.agda
```diff
@@ -152,3 +152,49 @@ overPathFunct : ∀ {ℓ} {A : Type ℓ} {x y : A} (p q : x ≡ x) (P : x ≡ y)
...
     (transportRefl (p ∙ q) ∙ cong₂ _∙_ (sym (transportRefl p)) (sym (transportRefl q)))
+
+-- substituting over a composition (P ∘ f) along p is equivalent to
+-- substituting over P along (cong f)
+-- lemma 2.3.10 in The Book
+substFunComp : ∀ {ℓ ℓ'} {A B : Type ℓ} {a a' : A}
+                 → (f : A → B) → (P : B → Type ℓ') → (p : a ≡ a') → (u : P (f a))
+                 → subst (P ∘ f) p u ≡ subst P (cong f p) u
+substFunComp f P p u = J (λ x q → (subst (P ∘ f) q u) ≡ (subst P (cong f q) u))
+  (substRefl {B = P ∘ f} u ∙ sym (substRefl {B = P} u)) p
```
ecavallo, line 163: This is true by `refl` in cubical type theory, because `subst P p u` is defined to be `transport (λ i → P (p i)) u`.

Cubical/Foundations/Transport.agda
```diff
@@ -152,3 +152,49 @@ overPathFunct : ∀ {ℓ} {A : Type ℓ} {x y : A} (p q : x ≡ x) (P : x ≡ y)
...
+-- theorem 2.11.3 in The Book
+substInPaths : ∀ {ℓ} {A B : Type ℓ} {a a' : A}
+                 → (f g : A → B) → (p : a ≡ a') (q : f a ≡ g a)
+                 → subst (λ x → f x ≡ g x) p q ≡ sym (cong f p) ∙ q ∙ cong g p
+substInPaths {a = a} f g p q =
+  J (λ x p' → (subst (λ y → f y ≡ g y) p' q) ≡ (sym (cong f p') ∙ q ∙ cong g p'))
+    p=refl p
+    where
+    p=refl : subst (λ y → f y ≡ g y) refl q
+           ≡ (λ i → f (refl {x = a} (~ i))) ∙ q ∙ (λ i → g (refl {x = a} i))
```
ecavallo, line 176: You can write `a` here in place of `refl {x = a} (~ i)` or `refl {x = a} i`. Actually you could simply write `refl ∙ q ∙ refl` for the right side of this equality.

Cubical/Foundations/Transport.agda
```diff
@@ -152,3 +152,49 @@ overPathFunct : ∀ {ℓ} {A : Type ℓ} {x y : A} (p q : x ≡ x) (P : x ≡ y)
...
+  x=a : (q : x1 ≡ a) → subst (λ x → x ≡ a) p q ≡ sym p ∙ q
+  x=a q = subst (λ x → x ≡ a) p q
+    ≡⟨ substInPaths (λ x → x) (λ _ → a) p q ⟩
+      sym (cong (λ x → x) p) ∙ q ∙ cong (λ _ → a) p
+    ≡⟨ assoc (sym p) q refl ⟩
+      (sym p ∙ q) ∙ refl
+    ≡⟨ sym (rUnit (sym p ∙ q))⟩ sym p ∙ q ∎
+
+  x=x : (q : x1 ≡ x1) → subst (λ x → x ≡ x) p q ≡ sym p ∙ q ∙ p
+  x=x q = substInPaths (λ x → x) (λ x → x) p q
```
ecavallo, line 200: I think the first two of these are worth keeping, though maybe not `x=x`. Let's give them more descriptive names: maybe `substInPathL` and `substInPathR` or something like that?

ecavallo (changes requested):
Thanks for the contribution! One of these is actually trivial in cubical type theory, but the others are good to have around.

Cubical/Foundations/Transport.agda
Aqissiaq: The book notes that these are the "functorial actions of the covariant and contravariant hom-functors", but I think your name suggestions are more descriptive

PR #759, opened 2022-04-08
Cohomology Rings

thomas-lamiaux:
Add a proper definition of the cohomology rings.

Work in progress : compute some cohomology rings


thomas-lamiaux:
To prove that H^n (X) are gradedCommutative, -h _ . _ is introduced. 
However it is not introduced using -_h but by new direct definition to have more specific computations rules.

Hence I have a introduced a -h _ ._ and renamed the other one and the related function by adding a prime.
Prove that it is equal, and added a proof of gradedComm for the "good" -h _ . _ 
This enables to rise it more easily to the cohomology ring H*(X)

Nevertheless, this breaks a lot of dependency because the graded operator is used in several files.
Yet I do believe that it is for the the best, as it described more accurately what is done. 
Plus, it suffices to add a ' to the dependent files to fix the issue, one must just find them all.

thomas-lamiaux:
The state of the project is the current : 
-> Building the idea of a general framework for computing Cohomology Ring if the groups are finitely non trivial and the ring is quotient of Z[X] by a finite vector
-> Computed the Cohomology ring for the Unit using the general framework
-> Computed the Cohomology ring for the sphere S1 !

Currently the files need some clean up, a refresh of Unit with what was learn on S1. 
The next phase is doing S0 and more generally Sn.

Cubical/Algebra/Ring/DirectProd.agda
```diff
@@ -0,0 +1,120 @@
...
+  {X'r@(X' , X'str) : Ring ℓ''}
+  {Y'r@(Y' , Y'str) : Ring ℓ'''}
+  where
+
+  open Iso
+  open IsRingHom
+
+  Coproduct-Equiv-12 : (re : RingEquiv Xr Yr) → (re' : RingEquiv X'r Y'r) →
+                        RingEquiv (DirectProd-Ring Xr X'r) (DirectProd-Ring Yr Y'r)
+  fst (Coproduct-Equiv-12 re re') = isoToEquiv is
```
mortberg, line 98: How about using https://github.com/agda/cubical/blob/master/Cubical/Data/Sigma/Properties.agda#L350 instead?

Cubical/Algebra/Ring/DirectProd.agda
```diff
@@ -0,0 +1,120 @@
...
+    e' = equivToIso (fst re')
+
+    is : Iso (X × X') (Y × Y')
+    fun is (x , x') = (fun e x) , (fun e' x')
+    inv is (y , y') = (inv e y) , inv e' y'
+    rightInv is (x , x') = ≡-× (rightInv e x) (rightInv e' x')
+    leftInv is (y , y') = ≡-× (leftInv e y) (leftInv e' y')
+  snd (Coproduct-Equiv-12 e e') = makeIsRingHom map1 map+ map·
+    where
+    map1 : _
```
mortberg, line 113: It's good to give things explicit types. That can speed things up for Agda as the type doesn't have to be inferred. Also what can happen is that Agda infer a very complicated type where something much simpler could have worked. So try replacing all `foo : _` with explicit types

thomas-lamiaux:
Computation of S0 has been done and the files cleaned and standardized. 
The next step and commits are about computing Sn.

mortberg:
> Computation of S0 has been done and the files cleaned and standardized. The next step and commits are about computing Sn.

So what made it work?

thomas-lamiaux:
The issue was the coexistence of Z[X]/X under two different names and so I thing agda was trying to prove that there a definitionally equal. To do that it was probably unfolding the entire definition as one lambda term and then trying to match the terms. 

To fix it, I have added different instances like Z[X]/X, Z[X]/X2 or Z[X,Y] in CommRing.Instances.
This way they can be used with nice notations in all files.

thomas-lamiaux:
Yet trying to find the bug was good because it made do a nice clean up + made me define some lemma syntactical without using ua.

mortberg:
> The issue was the coexistence of Z[X]/X under two different names and so I thing agda was trying to prove that there a definitionally equal. To do that it was probably unfolding the entire definition as one lambda term and then trying to match the terms.
> 
> To fix it, I have added different instances like Z[X]/X, Z[X]/X2 or Z[X,Y] in CommRing.Instances. This way they can be used with nice notations in all files. [78d5668#diff-88c1e4d6394998e9a4d963cbc0db63b7e307ad554a14fd80eade8e12ad15ca56R56](https://github.com/agda/cubical/commit/78d56683874c0875d13b7571dd06df4bb8cb5d09#diff-88c1e4d6394998e9a4d963cbc0db63b7e307ad554a14fd80eade8e12ad15ca56R56)

Ahh, that explains it! Great that you figured it out. Looking forward to H*(S^n)!

thomas-lamiaux:
An issue that arises with general case like Sn, it that you need to partition N in 0, k < n, k = n, n < k
However, now you coefficient are in coHom k X and not coHom 0 X (in the first case). 
So one would need to add transport to be in the good group   T a   becomes   T ((subst (x -> coHom x X) _ a)

I haven' advanced enough to see if they are going to be annoying or not, however, I have hope for it to work quite correctly.
This should only come to play in section and retraction.

Indeed, you dont 't need subst in the case Z[X] -> H*(X) (not the same X) because Z[X] is the direct sum on Z which is constant.
So it should not be annoying for the cup product as you need to do it on only one side.
And I am doing already doing it the sens Z[X] -> H*(X) for the reason that Z and his product is simpler then coHomk k X

You also loose some computation as n is abstract, it not clear either where it is going to be annoying

thomas-lamiaux:
The content of the Pull request is done.
The only thing left to do is some clean up and standardization of the files

Cubical/Algebra/CommRing/FGIdeal.agda
```diff
@@ -95,6 +99,87 @@ module _ (Ring@(R , str) : CommRing ℓ) where
...
+
+  open CommRingStr Ar using ()
+    renaming
+    ( 0r        to 0A
+    ; 1r        to 1A
+    ; _+_       to _+A_
+    ; -_        to -A_
+    ; _·_       to _·A_ )
+
+  open CommRingStr (snd B') using ()
```
mortberg, line 117: Br

Cubical/Algebra/CommRing/FGIdeal.agda
```diff
@@ -95,6 +99,87 @@ module _ (Ring@(R , str) : CommRing ℓ) where
...
+    ( 0r        to 0B
+    ; 1r        to 1B
+    ; _+_       to _+B_
+    ; -_        to -B_
+    ; _·_       to _·B_ )
+
+  open CommRingStr
+  open IsRingHom
+
+  cancel-linear-combinaison : (n : ℕ) → (a v : FinVec A n) → (fnull : (k : Fin n) → f (v k) ≡ 0B)
```
mortberg, line 128: `cancelLinearCombination`

Cubical/Algebra/CommRing/FGIdeal.agda
```diff
@@ -95,6 +99,87 @@ module _ (Ring@(R , str) : CommRing ℓ) where
...
+  (A'@(A , Astr) : CommRing ℓ)
+  where
+
+  Ar : Ring ℓ
+  Ar = CommRing→Ring A'
+
+  open CommRingStr Astr
+  open RingTheory
+
+  cbn-akbFin-linear-combi : (n : ℕ) → (k : Fin n) → (a : A) → (v : FinVec A n)
```
mortberg, line 155: rename and introduce delta

mortberg (approved):

PR #766, opened 2022-04-28
Eliminators for Rationals HITQ

guilhermehas:
Elim, ElimProp, and Rec eliminators for Rationals HITQ.


mortberg:
Thanks for the contribution!

PR #768, opened 2022-04-29
Isomorphisms of precategories.

anuyts:
Do we want this kind of code duplication though? #625


mortberg:
Not sure how useful this is as precategories are not such a useful concept, but I'll merge anyway in case someone wants to use it. Thanks for the contribution :-)

anuyts:
Well I used it to express isomorphism of categories.

mortberg:
But we already have that?

anuyts:
I mean an isomorphism in the precategory of categories, i.e. an invertible functor.

Issue #774, opened 2022-05-02
Rename the RingSolver to CommRingSolver


felixwellen:
... because it can only solve equations in commutative rings.

felixwellen:
Done in #777

PR #777, opened 2022-05-03
Rename RingSolver to CommRingSolver

felixwellen:


felixwellen:
Resolves #774

mortberg (approved):

kangrongji:
@felixwellen It seems the ring solver still have a bug. I applied the solver to an explicit CommRing, namely a closed term `R : CommRing ℓ` instead of assumption in scope, and it doesn't work. Maybe I should make an issue?

felixwellen:
Yes, opening an issue seems to be a good idea.

Issue #396, opened 2020-08-03
`Cubical.**.Foo` should only import `**.Foo.Base` and `**.Foo.Properties`


Saizan:
I think a `Foo` module should export what you're more likely to need about `Foo`.

If they re-export everything that's been developed for `Foo` then they are not as useful to import, as that slows down typechecking from scratch and bring a lot of names into the namespace.

We should also add this policy to `CONTRIBUTING.md`.

PR #783, opened 2022-05-04
Issue #396: enforce import rules

felixwellen:
Enforce the rule that a re-exporting file Foo.agda should only re-export Foo/Base.agda and Foo/Properties.agda.


mortberg:
Nice! Do you have more things to do or should I merge?

felixwellen:
I would leave it like this for now. Just merge it. I guess we can do more 'Enforce the rules'-PRs in the future, when we find enough stuff to fix...

mortberg:
Ok, I was just not sure as it was marked as a Draft

PR #786, opened 2022-05-08
Fix ZGroup

thomas-lamiaux:
For a weird reason, ZGroup et ZCommRing are not defined with the same definition.
Indeed someone took for the inverse `- x := 0 - x` which is by definition `0 + (- x)`
This poses a lot of trouble when mixing the two instances, indeed some 0 + X pop up sometimes, so one need to rewrite to add 0 to use presinv.
Plus it gets some proof more complicated as one always need to add.

So I am currently rewriting all the files using it


mortberg (approved):
Very good! Are there any other changes you want to make or should I merge?

PR #789, opened 2022-05-09
Clean up library

### LINKED ISSUE #570: Files should not import Everything files unless necessary
Some files import Everything files for no reason which can cause a headache when refactoring or renaming. It would be good to remove unnecessary imports of Everything and instead have more fine grained imports (for example if a file only uses 2-3 modules from Foundations there is no reason to import Foundations.Everything)

thomas-lamiaux:
### This pull request is a one week effort to clean the library
It will focus on : 

- Enforcing some of the naming convention ([global](https://github.com/agda/cubical/blob/master/NAMING.md) and [algebra](https://github.com/agda/cubical/blob/master/Cubical/Algebra/NAMING.md)) decided last week in past files.
In particular the ones about import.
- Renaming some object ill named.
- Split of merge some files to enhanced clarity.

However, the goal is not to add any content to the library.   
Don't hesitate to comment for instance to tell me to look at something 

### Description of the pull request :
- Cleaning Polynomials
- Renaming the PropositionalTruncation (description below)
- Cleaning ZCohomology import, and some other files
- Enforcing naming convention for Group and CommRing
- Getting rid of `open import Cubical.Foundations.Everything` except in [one paper](https://github.com/agda/cubical/blob/master/Cubical/Papers/ZCohomology.agda#L23) in which the import are so messy that I preferred not to touch it. resolve #570 
- Rename Direct-Sum to DirectSum


mortberg (approved):
Great! I think we should merge now, or is there some PR that should be merged first? @ecavallo @felixwellen

mortberg:
Merging!

PR #791, opened 2022-05-10
Build cubical with latest Agda master (2022-05-09)

andreasabel:
I noticed yesterday that the latest `cubical` does not build with the latest `agda`.  This PR should fix this.
- 2 hidden arguments that can no longer be inferred: fixed by providing these arguments
- a regression in the termination checker: fixed by turning the index of `IW` into a parameter

If you agree and merge this, the submodule pin for `cubical` in the `agda` repo can be moved forward.

On another note, it would be good if you would test `cubical` also with the latest `agda`, just to get alerted of breakages.  Maybe you can add such to your CI, using the deploys of Agda (https://github.com/agda/agda/actions/workflows/deploy.yml)


andreasabel:
Build on 2.6.3 (of the merge commit for this PR) broken by:
- https://github.com/agda/cubical/pull/788#issuecomment-1125761025

So, current `master` isn't ready for 2.6.3.

mortberg:
Were we going to make a branch called Agda-dev that is kept up to date with Agda master?

I have no clue how the CI works, so if someone can set up what Andreas suggests that would be great! Maybe @ecavallo  knows how to do it?

ecavallo:
Yeah, I can have a go

PR #793, opened 2022-05-11
Added Free Abelian Group

guilhermehas:
I added Free Abelian Group with their eliminators and with a instance that it is an Abelian Group.


mortberg:
Thanks for the contribution!

PR #796, opened 2022-05-12
Rm ZCohomologyOld

thomas-lamiaux:
Rm ZCohomologyOld


thomas-lamiaux:
I do not know how to fix this the fact that the ZCohomology Paper is using it https://github.com/agda/cubical/blob/master/Cubical/Papers/ZCohomology.agda#L296

I am really in favor to assign papers to a fix commit and never update them after this point

mortberg:
> I do not know how to fix this the fact that the ZCohomology Paper is using it https://github.com/agda/cubical/blob/master/Cubical/Papers/ZCohomology.agda#L296
> 
> I am really in favor to assign papers to a fix commit and never update them after this point

Aha, then this should be fixed when we decide what to do about the Papers folder next week

thomas-lamiaux:
I am going to close this until #815 is decided.
Indeed a decision is going to be made about experiment, at least cleaning it in a deeper way than just removing this folder.

PR #797, opened 2022-05-13
Added Field in Algebra

guilhermehas:
A field is a commutative ring with an inverse operation that one is different than zero.

### DIFF
--- a/Cubical/Algebra/Field/Base.agda
+++ b/Cubical/Algebra/Field/Base.agda
@@ -40,7 +40,7 @@ record IsField {R : Type ℓ}
 
   field
     isCommRing : IsCommRing 0r 1r _+_ _·_ -_
-    ·⁻¹≡1      : (x : R) → x · (x ⁻¹) ≡ 1r
+    ·⁻¹≡1      : (x : R) → ¬ (x ≡ 0r) → x · (x ⁻¹) ≡ 1r
     0≢1        : ¬ (0r ≡ 1r)
 
   open IsCommRing isCommRing public
@@ -81,7 +81,7 @@ makeIsField : {R : Type ℓ} {0r 1r : R} {_+_ _·_ : R → R → R} { -_ _⁻¹ 
                  (·-rid : (x : R) → x · 1r ≡ x)
                  (·-rdist-+ : (x y z : R) → x · (y + z) ≡ (x · y) + (x · z))
                  (·-comm : (x y : R) → x · y ≡ y · x)
-                 (·⁻¹≡1 : (x : R) → x · (x ⁻¹) ≡ 1r)
+                 (·⁻¹≡1 : (x : R) → ¬ (x ≡ 0r) → x · (x ⁻¹) ≡ 1r)
                  (0≢1 : ¬ (0r ≡ 1r))
                → IsField 0r 1r _+_ _·_ -_ _⁻¹
 makeIsField {_+_ = _+_} is-setR +-assoc +-rid +-rinv +-comm ·-assoc ·-rid ·-rdist-+ ·-comm ·⁻¹≡1 0≢1 =
@@ -97,7 +97,7 @@ makeField : {R : Type ℓ} (0r 1r : R) (_+_ _·_ : R → R → R) ( -_ _⁻¹ : 
                  (·-rid : (x : R) → x · 1r ≡ x)
                  (·-rdist-+ : (x y z : R) → x · (y + z) ≡ (x · y) + (x · z))
                  (·-comm : (x y : R) → x · y ≡ y · x)
-                 (·⁻¹≡1 : (x : R) → x · (x ⁻¹) ≡ 1r)
+                 (·⁻¹≡1 : (x : R) → ¬ (x ≡ 0r) → x · (x ⁻¹) ≡ 1r)
                  (0≢1 : ¬ (0r ≡ 1r))
                → Field ℓ
 makeField 0r 1r _+_ _·_ -_ _⁻¹ is-setR +-assoc +-rid +-rinv +-comm ·-assoc ·-rid ·-rdist-+ ·-comm ·⁻¹≡1 0≢1 =
@@ -153,8 +153,8 @@ isPropIsField 0r 1r _+_ _·_ -_ _⁻¹ (isfield RR RC RD) (isfield SR SC SD) =
   isSetR : isSet _
   isSetR =  RR .IsCommRing.·IsMonoid .IsMonoid.isSemigroup .IsSemigroup.is-set
 
-  isPropInv : isProp ((x : _) → x · (x ⁻¹) ≡ 1r)
-  isPropInv = isPropΠ λ _ → isSetR _ _
+  isPropInv : isProp ((x : _) → ¬ (x ≡ 0r) → x · (x ⁻¹) ≡ 1r)
+  isPropInv = isPropΠ2 λ _ _ → isSetR _ _
 
   isProp→⊥ : ∀ {A : Type ℓ} → isProp (A → ⊥)
   isProp→⊥ = isPropΠ λ _ → isProp⊥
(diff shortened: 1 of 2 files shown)


mortberg:
I think the definitions look good, but the only thing I'm unsure about is whether we should call this "Field" or not. Constructively it's very strong... But maybe the best is to call this Field and have other the notions with special names e.g. "ConstructiveField", "ApartnessField", "HeytingField", etc..

I remember there was a long discussion for UniMath as well: https://github.com/UniMath/UniMath/issues/254

The nlab also has an entry about this: https://ncatlab.org/nlab/show/field#constructive

What does @felixwellen  think?

felixwellen:
According to my taste, this should be the default of what we mean by "Field" constructively. I would be happy with using this definition - but we should definitely have a comments which make all readers aware, that the constructive math literature might mean something else with "Field". 
From the discussion in #301, what is defined here might be a denial field, but I'm not sure about {0} being a prime ideal - can we show that? That might also be worth mention in a comment.

felixwellen:
I just checked in [Mines-Richman-Ruitenburg](https://www.springer.com/gp/book/9789401799430) - they don't actually have the prime ideal condition. So we could say in a comment, that this is a special case of their notion of field, for the apratness relation given by inequality.

mortberg:
Ok, good, so we agree that this should be called Field but that it needs to be documented that it's very strong constructively (in particular the reals are not a field constructively with this definition). However, what do you mean by the "prime ideal condition"?

felixwellen:
Forget about the prime ideals. 
It is not the strongest condition, as I just learned from @MatthiasHu , you could also say 1 is different from 0 and every x is zero or invertible.
Maybe it should also be mentioned in the comments, that the reals are not a field with this condition (as you said).

Cubical/Algebra/Field/Base.agda
```diff
@@ -0,0 +1,183 @@
...
+    ℓ ℓ' : Level
+
+record IsField {R : Type ℓ}
+                  (0r 1r : R) (_+_ _·_ : R → R → R) (-_ _⁻¹ : R → R) : Type ℓ where
+
+  constructor isfield
+
+  field
+    isCommRing : IsCommRing 0r 1r _+_ _·_ -_
+    ·⁻¹≡1      : (x : R) → ¬ (x ≡ 0r) → x · (x ⁻¹) ≡ 1r
```
guilhermehas, line 43: I think that if I change this line to this:
```agda
0≡⊎·⁻¹     : (x : R) → (x ≡ 0r) ⊎ (x · (x ⁻¹) ≡ 1r
```

I can't prove this anymore in line 156:
```agda
  isPropInv : isProp ((x : _) → (x ≡ 0r) ⊎ (x · (x ⁻¹) ≡ 1r))
```

MatthiasHu:
I think that the current definition, which asks for an "inverse operation" defined on all elements, including zero, doesn't fit in a constructive development very well! One should talk about invertibility (existence of an inverse) instead.

The current definition of field homomorphisms even asks that the (arbitrarily chosen) "inverse of zero" be preserved. In the case that the underlying set has decidable equality, one could call this a homomorphism of _pointed_ fields. (For example, there would by many non-isomorphic fields with underlying ring the rational numbers.)

MatthiasHu:
Without talking about apartness relations, there are (at least) the following three constructive variations of the field condition:

- For all x, if not (x = 0), then x is invertible.
- For all x, if x is not invertible, then x = 0.
- For all x, x is invertible or x = 0.

The last one is the strongest, the first two are incomparable. To all three, one can (and should?) add elegantly that 1 is not equal to 0: by promoting the implication to logical equivalence, or by promoting the "or" to an exclusive or.

felixwellen:
> I think that the current definition, which asks for an "inverse operation" defined on all elements, including zero, doesn't fit in a constructive development very well! One should talk about invertibility (existence of an inverse) instead.

That should be ok. Note that inverses are unique! That means that the sigma of inverses of some element is a proposition, so it is equivalent to its propositional truncation. That means asking for existence of inverses (for non-zero elements) implies already the existence of the above inverse operation.

felixwellen:
That might also be worth mentioning...

felixwellen:
I would also be ok with merging this PR as it is and collecting all the documentation wishes we have in an issue for now.

MatthiasHu:
@felixwellen I agree that having an inverse operation defined on all non-invertible elements is completely fine. But the current code asks for an operation
```agda
_⁻¹ : R → R
```
such that:
```agda
·⁻¹≡1 : (x : R) → ¬ (x ≡ 0r) → x · (x ⁻¹) ≡ 1r
```
The sigma type combining these two things is certainly not a proposition.

felixwellen:
Ohno! Right - I misread. That is something we certainly shouldn't ask for. The inverse operation should have non-zero elements as its domain.

MatthiasHu:
I'm trying to change the code to use an operation
```agda
≠0→inverse : (x : R) → ¬ (x ≡ 0r) → Σ[ y ∈ R ] x · y ≡ 1r
```
instead of `_⁻¹` and `·⁻¹≡1` now. What is the best way to share these changes here?

felixwellen:
The best way to share your changes afaik, is to make a PR on the PR. In detail: You add guilhermehas fork as a remote and checkout the branch 'field' this PR is based on. Then you basically proceed as with a normal PR, i.e. you make a branch based on 'field', commit your changes to that and push it to your fork. Then github should ask you the right question. Otherwise configure your PR manually to request merging of your branch into 'guilhermehas:field'.
Keep in mind that with default settings of github, @guilhermehas will not be notified of your PR - so you might want to mention that here.

MatthiasHu:
Thank you, done: https://github.com/guilhermehas/cubical/pull/1

guilhermehas:
I merged it, but I think that it is a good idea to add `≠0→inverse` to `FieldStr`.

guilhermehas:
I am changing to have:
```agda
    _⁻¹ : (x : A) → {¬ (x ≡ 0r)} → A
```
And a proof that `x · x ⁻¹ ≡ 1r`

MatthiasHu:
Thank you for merging the meta-PR.

Yes, maybe it is better to split `≠0→inverse` in two parts, as you say.

guilhermehas:
I made the changes that I propose.
However, I had to remove 𝒮ᴰ-Field because I don't know how to code it now.

felixwellen:
I would like to merge this and open two issues, one for documenting the relation to constructive math notions of field, one for applying the SIP. Are you ok with this @mortberg ?

Cubical/Algebra/Field/Base.agda
```diff
@@ -0,0 +1,169 @@
...
+    _[_]⁻¹     : (x : A) → ¬ (x ≡ 0r) → A
+    isField    : IsField 0r 1r _+_ _·_ -_ _[_]⁻¹
+
+  infix  20 _⁻¹
+  infix  8 -_
+  infixl 7 _·_
+  infixl 7 _/_
+  infixl 6 _+_
+
+  _⁻¹ : (x : A) → { ≢0 : ¬ (x ≡ 0r) } → A
```
felixwellen, line 68: Wait - this should not be inferrable in practice. The proof that something is non-zero should be explicit.

Cubical/Algebra/Field/Base.agda
```diff
@@ -0,0 +1,169 @@
...
+  (x ⁻¹) { ≢0 } = x [ ≢0 ]⁻¹
+
+  _/_ : (x y : A) → { ≢0 : ¬ (y ≡ 0r) } → A
```
felixwellen, line 71: Same here. I guess it is better to give the operation a name instead of operator syntax.

Cubical/Algebra/Field/Base.agda
felixwellen: Yes, right, I'm afraid it will not work as intended, so you can just remove it (and the division).

guilhermehas:
I just removed _⁻¹ and _/_

felixwellen:
Thanks! I'll write the issues and merge after that.

felixwellen:
Done (#820, #819).

PR #799, opened 2022-05-14
Congruence over dependent paths of functions with ua in their domain

phijor:
Recently, I've been doing some work involving lots dependent paths over function types. I found the `ua→`-lemmata to quite helpful, but the library is still missing an equivalent of `cong` for such paths. I thought this would make a good first contribution.

Questions:
 - [ ] Is this a duplicate effort?
 - [ ] Is `ua→cong` a good name? (Or `cong-ua→`, `ua→∘`, ...?)
 - [ ] Should the interval argument to `F` be explicit?
 - [ ] What's the documentation supposed to look like?
 - [ ] Would you like to see some example usage?

### DIFF
--- a/Cubical/Foundations/Univalence.agda
+++ b/Cubical/Foundations/Univalence.agda
@@ -256,6 +256,16 @@ ua→ {e = e} {f₀ = f₀} {f₁} h i a =
   lem : ∀ a₁ → e .fst (transport (sym (ua e)) a₁) ≡ a₁
   lem a₁ = secEq e _ ∙ transportRefl _
 
+-- Congruence over dependent paths of functions with ua in their domain
+ua→cong : ∀ {ℓ ℓ' ℓ''} {A₀ A₁ : Type ℓ} {e : A₀ ≃ A₁}
+  {B : (i : I) → Type ℓ'}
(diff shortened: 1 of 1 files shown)


ecavallo:
Thanks for the PR, and sorry for the long wait. I think however that this function is easy enough to write using `congP` that it doesn't need its own definition. Like so:
```
ua→cong F = congP (λ i f → F {i} ∘ f)
```
Or if you want to be extra short:
```
ua→cong F = congP (λ i → F {i} ∘_)
```
Maybe it's helpful to note that `ua` doesn't play an essential role here. You can replace `ua e` with any line `A` and it still works:
```
ua→cong' : ∀ {ℓ ℓ' ℓ''} {A : (i : I) → Type ℓ}
  {B : (i : I) → Type ℓ'}
  {C : (i : I) → Type ℓ''}
  {f₀ : A i0 → B i0} {f₁ : A i1 → B i1}
  (F : {i : I} → B i → C i)
  (p : PathP (λ i → A i → B i) f₀ f₁)
  → PathP (λ i → A i → C i) (F {i0} ∘ f₀) (F {i1} ∘ f₁)
ua→cong' F p = λ i x → F (p i x)
```

PR #801, opened 2022-05-16
Remove an unnecessary argument in aurefl

vikraman:
Noticed an unnecessary argument trying to use this module.


mortberg:
Good catch! I wonder how that happened...

PR #802, opened 2022-05-17
Verification of Brunerie number computation

aljungstrom:
In #763 I made added a small computation to the `Experiments.Brunerie` file which I claimed to be a computation of the Brunerie number. In this PR, I have added a very similar computation to `Homotopy.Group.Pi4S3.QuickProof` and proved that what it computes indeed is the Brunerie number.

For the computation to work, I needed to work with the base/surf definition of S², so I unfortunately had to add a bit of theory about this guy (e.g. inversion and its interaction with the suspension map S² → Ω(Susp S²)). I also had to add a lemma about surjective homs Z -> Z being isos.

I also moved `Brunerie.agda` from `Experiments` into `Homotopy.Group.Pi4S3` and created a properties file for S².


mortberg:
Extremely nice! I can't believe that we've finally proved this by computation  🤯

mortberg:
The few changes I wanted to make can be found in https://github.com/agda/cubical/pull/805  . So merging this now!

PR #805, opened 2022-05-18
Update the summary file for pi_4(S^3) with the direct and computational proofs

mortberg:
This builds on https://github.com/agda/cubical/pull/802 and documents things a bit more to make it clearer that the Agda computation of the new Brunerie number really proves that pi_4(S^3) = Z/2Z 

Also rename QuickProof.agda to DirectProof.agda


ecavallo:
In my opinion we should get rid of `BrunerieExperiments` and move whatever is still necessary to other files. At the very least the strange test definitions should finally be destroyed.

mortberg:
> In my opinion we should get rid of `BrunerieExperiments` and move whatever is still necessary to other files. At the very least the strange test definitions should finally be destroyed.

I deleted some of foo-goo now

mortberg:
Wow, the CI is so fast now that it doesn't recompile everything!

PR #806, opened 2022-05-19
Lookup & tabulate

anuyts:
…for lists and dependent lists.


mortberg:
Thanks a lot and sorry about the slow merging!

PR #807, opened 2022-05-19
Path operators

anuyts:
A few path operators.


Cubical/Foundations/Prelude.agda
```diff
@@ -65,6 +66,10 @@ cong : (f : (a : A) → B a) (p : x ≡ y) →
 cong f p i = f (p i)
 {-# INLINE cong #-}
 
+congS : ∀ {B : Type ℓ} → (f : A → B) (p : x ≡ y) → f x ≡ f y
```
mortberg, line 69: What does the `S` stand for? Also what is the point of having this simpler form?

Cubical/Foundations/Prelude.agda
```diff
@@ -301,6 +306,16 @@ funExt⁻ : {B : A → I → Type ℓ'}
   → ((x : A) → PathP (B x) (f x) (g x))
 funExt⁻ eq x i = eq i x
 
+_≡$_ = funExt⁻
+
+funExtS⁻ : {B : I → Type ℓ'}
```
mortberg, line 311: Same questions about this one. Why the `S` and what is the point of this simpler version?

Cubical/Foundations/Prelude.agda
```diff
@@ -301,6 +306,16 @@ funExt⁻ : {B : A → I → Type ℓ'}
   → ((x : A) → PathP (B x) (f x) (g x))
 funExt⁻ eq x i = eq i x
 
+_≡$_ = funExt⁻
```
mortberg, line 309: How often does one want an infix version of this? Not sure how useful it is, but I'm not strongly opposed to it :-)

mortberg (changes requested):
Very sorry about the slow review! Just a few questions about naming and usability, otherwise it looks fine

Cubical/Foundations/Prelude.agda
anuyts: `S` stands for simple (simply typed), just like `P` seems to stand for dependent? Using `congS` instead of `cong` can simplify some unification problems leading to the resolution of some metavariables that would otherwise remain unsolved.

Cubical/Foundations/Prelude.agda
anuyts: Same answer.

Cubical/Foundations/Prelude.agda
anuyts: What `funExt⁻` really does is applying a path between functions to an element of the functions' domain, so it is really an application operator. In the HoTTbook it is called `happly`, in the agda standard library it is called `cong-app`. In my experience, one almost always wants an infix version of this, and especially when happlying to multiple arguments.

Cubical/Foundations/Prelude.agda
mortberg: Ah ok, I see. I think someone explained this to me before, but I forgot about it. Can you please add a comment about it?

Cubical/Foundations/Prelude.agda
mortberg: Ok, sounds good to me!

mortberg:
This can be merged once there is a comment explaining the point of `congS`

anuyts:
Done!

mortberg (approved):

mortberg:
> Done!

Great!

PR #798, opened 2022-05-14
Direct sum and gradedRing

thomas-lamiaux:
The goal of this PR has changed during the work. This is the updated message
### Overview
The goal of this PR is to add a general notion graded ring.
This for two definitions of the direct sum that shown equivalent.

### HIT Direct Sum
- [x] Add a pseudo normal form for the HIT direct sum if indexed by N
- [x] Add a proof that base is `base k a = base l b => a = 0_k & b = 0_l` for a decidable index
- [x] Add a universal property

### Almost null sequences Direct Sum
- [x] Add another definition of the direct sum using almost null sequences
- [x] Prove the equivalence between the two definition of the direct sum if the index is N

### Graded Ring
- [x] Add a general notion of graded ring using the HIT direct sum
- [x] Add some instances of gradedRing :
- - [x] Trivial Graded  Ring
- - [x]  Polynomial
- - [x] Cohomology Ring
- [x] Replace the following constructions using the new graded ring construction
- - [x] Polynomials
- - [x] Cohomology Ring
- [x] Add the almost null sequences as an instance of the graded Ring when the index is a N monoid.
- [x] Add Instances using this new definition :
- - [x] Polynomials
- - [x] Cohomology Ring

PR #808, opened 2022-05-22
WIP: Coproduct of AbGroups

felixwellen:
Defines the coproduct of a type-indexed family of abelian groups as a HIT. 
WIP: Show the universal property.

### DIFF
--- a/Cubical/Algebra/AbGroup/Coproduct.agda
+++ b/Cubical/Algebra/AbGroup/Coproduct.agda
@@ -17,7 +17,7 @@ module _ {X : Type ℓ} (A : X → Type ℓ') where
   infix 20 _⁻¹
 
 
-  data Coproduct : Type _ where
+  data Coproduct : Type (ℓ-max ℓ ℓ') where
     incl       : (x : X) → A x → Coproduct
     ε         : Coproduct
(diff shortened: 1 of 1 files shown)


thomas-lamiaux:
I think that this something that I have already added to the library as https://github.com/agda/cubical/blob/master/Cubical/Algebra/DirectSum/Base.agda ? With an instantiation here for abelian group here https://github.com/agda/cubical/blob/master/Cubical/Algebra/AbGroup/Instances/DirectSum.agda

incl              = base
epsilon        = 0
⊕               = add
inclPres+    = base-add
assoc         = assoc
comm        = comm
identityᵣ     = IdR
trunc         = trunc
--
⊖_  and invᵣ are not neccesary as they can be lifted from the definition
--
In my definition there is also a base-neutral in file that is useful when working to have only one zero

felixwellen:
Ok - good.

felixwellen:
I'll stop to work on that until #798 is merged. Then I might add the universal property to the definition of 'DirectSum's on that branch/master.

thomas-lamiaux:
In #798 I have added the Universal Property for the direct sum because we needed to prove it to make sure it was the "good direct sum". So I don't know if there is still something you want to add in this PR ?

PR #811, opened 2022-05-23
The abelianization of groups as a coequalizer of sets

HilpAlex:
This PR adds the abelianization of groups as a coequalizer of sets as performed in https://1lab.dev/Algebra.Group.Ab.Free.html    as well as the coequalizer of sets as performed in https://1lab.dev/Data.Set.Coequaliser.html.

Since there are now two definitions of the abelianization of groups, these are compared via their universal properties.


thomas-lamiaux:
Hey, I was looking at the files, the maths seems really good. 
Are you working on implementing the Hurewicz theorem ?

On the origination, currently we are working on cleaning the library.
Looking at the files, I had the following reflection on the organization of the files : 

- The two abelianization should be package in AbGroup.Instances
- Add a short comment in the files "not to use" explaining why one is better / worse than the other
- The base file should include the elim functions.
  These way one can just call the base files and the instances.
  This way fewer definition are imported directly and so it reduces the renaming in case of renaming.
- Separate properties and UniversalMap. 
  This way we have a specific file and we can import it only when needed.
  This reduces the possibility of over lapping names
- Lately we have define convention for [naming convention for algebra](https://github.com/agda/cubical/blob/master/Cubical/Algebra/NAMING.md).
   It still hasn't be propagated to all past files but it would be great if you could rename things like [that](https://github.com/agda/cubical/pull/811/files#diff-adf0e49157ec9144a80dcaaa460745c0977d18fe4f04251a37050cf125936c9dR147) accordingly.
- About the general organization of the folder, as there is now two different definitions, it would be better to have clearer separation of the two. 
   Looking at the code of the folder maybe something like that ?
   
```
  => Abelianization
        => AbelianizationHIT
            => Base.agda
            => Propeties.agda
            => UniversalProperty.agda *
       => AbelianizationAsCoeq
            => Base.agda
            => Properties.agda
            => UniversalProperty.agda *
       => Equiv-HIT-Coeq.agda *
```
* if possible by opening the two abelinization as Instances

felixwellen:
@thomas-lamiaux :

> * The two abelianization should be package in AbGroup.Instances

I disagree. So far, we left it up to judgement of the contributer, if something belongs to 'Instances' and I would like to keep things that way. I do not see a point in moving all the constructions we have (e.g. Quotients...) to Instances - but this is a discussion we can have somewhere else.

> * Add a short comment in the files "not to use" explaining why one is better / worse than the other

No. I think the way a default is picked here, is good. It is easier to find the default by the file name - in this case 'Abelianization'. 

> * The base file should include the elim functions.

I don't think we should have that as a rule.

> * Separate properties and UniversalMap.

I don't think the reasons given justify asking for a refactoring.

I actually like the folder structure and file names as they are.

Where do you see a violation of the naming convention? I think the names for the AbGroup-fields are fine. 
What could be done, is exporting a name 'AbelianizationGroup' like it is suggested in the last point in the naming conventions...

@HilpAlex Thanks for the contribution!

felixwellen:
@thomas-lamiaux : Ok - I can see one of your points. I didn't realize that actually everything is exported from Abelianization/Properties...

Cubical/Algebra/Group/Abelianization/Properties.agda
```diff
@@ -111,29 +104,29 @@ module _ (G : Group ℓ) where
 
 
   -- Definition of the group structure on the abelianization. Here the generality of the comm relation is used.
-  _·Ab_ : Abelianization → Abelianization → Abelianization
+  _·Ab_ : Abelianization G → Abelianization G → Abelianization G
```
felixwellen, line 107: Starting with this definition, things shouldn't be exported. Please put everything that follows in a (named) new module.

Cubical/Algebra/Group/Abelianization/Properties.agda
```diff
@@ -185,7 +177,7 @@ module _ (G : Group ℓ) where
                η (y · x)        ≡⟨ refl ⟩
                (η y) ·Ab (η x) ∎)
 
-  -- The proof that that the abelianization is in fact an abelian group.
+  -- The proof that the abelianization is in fact an abelian group.
   asAbelianGroup : AbGroup ℓ
   asAbelianGroup = makeAbGroup 1Ab _·Ab_ invAb isset assocAb ridAb rinvAb commAb
```
felixwellen, line 183: Here you could end the module and add a definition 'AbelianizationAbGroup' after the module, so this name (which satisfies the new naming conventions) gets exported.

Cubical/Algebra/Group/Abelianization/Properties.agda
```diff
@@ -201,7 +193,7 @@ module _ (G : Group ℓ) where
 
   {- The proof of the universal property of the abelianization.
 
-  G --η--> Abelianization
+  G --η--> abelianization
```
felixwellen, line 196: Put everything that follows in a module. Maybe 'UniversalProperty'?

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
```diff
@@ -0,0 +1,354 @@
...
+            uniqueness to HITuniqueness)
+  hiding (elimProp; elimProp2; elimProp3; elimContr; elimContr2; rec; rec2; _·Ab_; 1Ab; invAb;
+          assocAb; ridAb; rinvAb; commAb)
+
+private
+  variable
+    ℓ : Level
+    G : Group ℓ
+
+module _ (G : Group ℓ) where
```
felixwellen, line 51: You should give that module a name, otherwise all its definitions will be exported. You can then, after the module, make a definition like ```AbelianizationCoeq = MODULENAME.Abelianization``` to export this definition.

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
```diff
@@ -0,0 +1,354 @@
...
+      (λ x y → squash (x ·Ab y) (y ·Ab x))
+      (λ x y → (incAb x) ·Ab (incAb y) ≡⟨ refl ⟩
+               incAb (x · y)           ≡⟨ cong incAb (sym (lid (x · y))) ⟩
+               incAb (1g · (x · y))    ≡⟨ comm 1g x y ⟩
+               incAb (1g · (y · x))    ≡⟨ cong incAb (lid (y · x)) ⟩
+               incAb (y · x)           ≡⟨ refl ⟩
+               (incAb y) ·Ab (incAb x) ∎)
+
+  -- The proof that the abelianization is in fact an abelian group.
+  asAbelianGroup : AbGroup ℓ
```
felixwellen, line 149: This should probably be exported as well.

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
```diff
@@ -0,0 +1,354 @@
...
+  incAbAsGroupHom : GroupHom G (AbGroup→Group asAbelianGroup)
+  incAbAsGroupHom = f , fIsHom
+    where
+    f = λ x → incAb x
+    fIsHom : IsGroupHom (snd G) f (snd (AbGroup→Group asAbelianGroup))
+    IsGroupHom.pres· fIsHom = λ x y → refl
+    IsGroupHom.pres1 fIsHom = refl
+    IsGroupHom.presinv fIsHom = λ x → refl
+
+  {- The proof of the universal property of the abelianization.
```
felixwellen, line 162: The universal property could be in a separate module.

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
```diff
@@ -0,0 +1,354 @@
...
+  G -incAb-> abelianization as coequalizer
+   \         .
+     \       .
+       η   ∃! isomorphism
+         \   .
+           \ .
+             abelianization as HIT
+  commuting diagram
+  -}
+  isomorphism : AbGroupIso asAbelianGroup (HITasAbelianGroup G)
```
felixwellen, line 242: This could be exported directly, if you name it more specifically.

felixwellen (changes requested):
Apart from some namespace/scope issues everything is fine.

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
```diff
@@ -0,0 +1,354 @@
...
+    instance
+      _ = snd G
+    G' = fst G
+
+  {-
+    The definition of the abelianization of a group as a coequalizer of sets. The generality
+    of the glued functions will be needed to define the group structure on the abelianization.
+  -}
+  Abelianization : Type ℓ
+  Abelianization = SetCoequalizer {A = G' × G' × G'} {B = G'} (λ (x , y , z) → x · y · z) (λ (x , y , z) → x · z · y)
```
mortberg, line 64: Long line

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
```diff
@@ -0,0 +1,354 @@
...
+        → fst (compGroupHom (HITηAsGroupHom G) (GroupIso→GroupHom (invGroupIso h))) g
+            ≡⟨ cong
+                 (λ f → f g)
+                 (cong fst (cong (λ f → compGroupHom f (GroupIso→GroupHom (invGroupIso h))) (sym hcomm))) ⟩
+          fst (compGroupHom
+                (compGroupHom incAbAsGroupHom (GroupIso→GroupHom h))
+                (GroupIso→GroupHom (invGroupIso h))) g
+            ≡⟨ cong
+                 (λ f → f g)
+                 (cong fst (compGroupHomAssoc incAbAsGroupHom (GroupIso→GroupHom h) (GroupIso→GroupHom (invGroupIso h)))) ⟩
```
mortberg, line 331: long lines

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
```diff
@@ -0,0 +1,354 @@
...
+          squash
+          HITisset
+          (h .fst)
+          (isomorphism .fst)
+          (λ x → cong
+                   (λ f → f x)
+                   (cong fst (uniqueness (HITasAbelianGroup G) (HITηAsGroupHom G) (GroupIso→GroupHom h) hcomm)))
+          (λ x → cong
+                   (λ f → f x)
+                   (cong fst (HITuniqueness G asAbelianGroup incAbAsGroupHom (GroupIso→GroupHom (invGroupIso h)) isocomm)))
```
mortberg, line 354: long line

Cubical/HITs/SetCoequalizer/Base.agda
```diff
@@ -0,0 +1,23 @@
...
+private
+  variable
+    ℓ ℓ' ℓ'' : Level
+    A : Type ℓ
+    B : Type ℓ'
+
+-- Set coequalizers as a higher inductive type
+data SetCoequalizer {A : Type ℓ} {B : Type ℓ'} (f g : A → B) : Type (ℓ-max ℓ ℓ') where
+  inc    : B → SetCoequalizer f g
+  glue   : (a : A) → inc (f a) ≡ inc (g a)
```
mortberg, line 22: I don't know if I like calling this "glue"... There's already so much "glue" in the library. Any suggestions for a better name?

Cubical/HITs/SetCoequalizer/Properties.agda
```diff
@@ -0,0 +1,116 @@
...
+            {C : SetCoequalizer f g → SetCoequalizer f' g' → Type (ℓ-max ℓ ℓ')}
+          → (Cprop : (x : SetCoequalizer f g) → (y : SetCoequalizer f' g') → isProp (C x y))
+          → (Cinc : (b : B) → (b' : B') → C (inc b) (inc b'))
+          → (x : SetCoequalizer f g) → (y : SetCoequalizer f' g') → C x y
+elimProp2 Cprop Cinc = elimProp (λ x → isPropΠ (λ y → Cprop x y))
+                                (λ x → elimProp (λ y → Cprop (inc x) y) (Cinc x))
+
+elimProp3 : {A' A'' : Type ℓ} {B' B'' : Type ℓ'} {f g : A → B} {f' g' : A' → B'} {f'' g'' : A'' → B''}
+            {C : SetCoequalizer f g → SetCoequalizer f' g' → SetCoequalizer f'' g'' → Type (ℓ-max ℓ ℓ')}
+          → (Cprop : (x : SetCoequalizer f g) → (y : SetCoequalizer f' g') → (z : SetCoequalizer f'' g'') → isProp (C x y z))
```
mortberg, line 48: long line

mortberg (changes requested):
Good stuff! I also had a few minor comments about long lines and I'm not sure if the path constructor should be called `glue` as there's already so many things called glue in the library...

Cubical/HITs/SetCoequalizer/Base.agda
felixwellen: 'eq' because it (co-)equalizes stuff?

mortberg: Yeah, or `coeq`?

felixwellen: Yes. ```coeq``` is the best so far.

Cubical/Algebra/Group/Abelianization/Properties.agda
```diff
@@ -151,20 +144,19 @@ module _ (G : Group ℓ) where
       η ((inv (c · b)) · (inv a))              ≡⟨ cong η (sym (invDistr a (c · b))) ⟩
       η (inv (a · (c · b))) ∎)
 
-  assocAb : (x y z : Abelianization) → x ·Ab (y ·Ab z) ≡ (x ·Ab y) ·Ab z
+  assocAb : (x y z : Abelianization G) → x ·Ab (y ·Ab z) ≡ (x ·Ab y) ·Ab z
```
thomas-lamiaux, line 147: Should now be named `·AbComm` rather then `assocAb` accordng to the new naming guideline ?

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
```diff
@@ -0,0 +1,354 @@
...
+                       incAb (inv (a · (c · b))) ∎)
+
+  assocAb : (x y z : Abelianization) → x ·Ab (y ·Ab z) ≡ (x ·Ab y) ·Ab z
+  assocAb =
+    elimProp3
+      (λ x y z → squash (x ·Ab (y ·Ab z))
+      ((x ·Ab y) ·Ab z))
+      (λ x y z → cong incAb (assoc x y z))
+
+  ridAb : (x : Abelianization) → x ·Ab 1Ab ≡ x
```
thomas-lamiaux, line 121: should be named `·AbIdR` according to the [new guide lines](https://github.com/agda/cubical/blob/master/Cubical/Algebra/NAMING.md) ?

Cubical/Algebra/Group/Abelianization/Properties.agda
felixwellen: With the requested changes, ```assocAb``` will become a local name - I don't think we should apply the naming conventions to local names, only to those that will be used later. The names defined here are packaged as an ```AbGroupStr```...

thomas-lamiaux:
@felixwellen 
This actually a very good question about moving it to instances or not but in that case shouldn't an instantiation ad an AbGroup shoud be created. I didn't thought of it that way. Then it makes less sens to move elim.

You are absolutely right that having a base is better to see which one should be used.
I meant adding a comment on top of the files to explain why the Base one is the recommended one.

For the renaming I was thing of the following Properties](https://github.com/agda/cubical/pull/811/files#diff-adf0e49157ec9144a80dcaaa460745c0977d18fe4f04251a37050cf125936c9dR147) and [AbelianizationAsCoeq](https://github.com/agda/cubical/pull/811/files#diff-522aeacdf4c01b83884f17198f236b1934dd69f0d346ff83f5c9e18a8fb99bfcR121)

Yes everything is exported by Properties. This could be hidden in a module if you don't want to touch at the organization.
Yet a lot of files in Algebra are going to have universal property. (or other properties)
It makes more sens to me to have a specific file for it each time. 
This way we know exactly what it does, we don't have to look for it in Properties or somewhere else.
Plus we export less thing too this way. 

For the reorganizing, I don't think this a good thing to have AbelianizationAsCoeq containing definition, group properties, universal property, and an equivalence with another structure. 
Firstly because it is harder to find chat you are looking for. Secondly because you export a lot of things if it not hidden in modules.That's why I was proposing a structure to separate the files + the universal property issue.

I still think that we should separate the files for the previous reason, but I guess it is subjective.
Plus keeping a base is good have you have stated for clarity.
If one wants to separate things, maybe something like the following could be interesting ?

```
=> Abelianization
     => AbelianizationAsCoeq
           => Base.agda
           => Properties.agda
           => UniversalMap.agda
           => Equiv something
     => Base.agda
     => Properties.agda
     => UniverselMap.agda
```

Cubical/Algebra/Group/Abelianization/Properties.agda
thomas-lamiaux: This a good question. 
To me the point of the naming were to have uniform naming in files, even for local names.
Maybe it should precise in NamingAlgebra.

felixwellen:
@thomas-lamiaux 

> I meant adding a comment on top of the files to explain why the Base one is the recommended one.

*If* there is a reason to prefer one over the other, then it could be good to have a comment about that. But this is not something I want to request and we have mentioned it enough now to make the contributor aware of the point.

> Yet a lot of files in Algebra are going to have universal property. (or other properties)
> It makes more sens to me to have a specific file for it each time.

That might be the case and we might come up with a way of standardizing universal properties. But this is not the right place to have this discussion. If you like, you can write an issue for that and you could label it with the new 'discuss' label so we know it is supposed to be a point we can discuss in the next meeting.

> For the reorganizing, I don't think this a good thing to have AbelianizationAsCoeq containing definition, [...]

Keep in mind that it is possible to refactor things later. Right now, we don't even know if the contents of this file will ever be used - it is an interesting remark that tells us a mathematical fact. I already requested changes, that ensure that the file can be imported without getting to much names - that is actually already a bit more than enough.

> If one wants to separate things, maybe something like the following could be interesting ?

I don't think it makes a relevant difference to have either a module in a file or a separate file. We can leave that decision to the contributer. And again, if it *actually turns out* to be more useful either way, we can still refactor it.

Cubical/Algebra/Group/Abelianization/Properties.agda
felixwellen: I don't think we want a convention for local names, but that should also be discussed somewhere else.

thomas-lamiaux: Yes sure, it can always be renamed once this has been decided.

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
HilpAlex: Line is now shorter

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
HilpAlex: Line is now shorter

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
HilpAlex: Line is now shorter

Cubical/HITs/SetCoequalizer/Base.agda
HilpAlex: renamed to coeq

Cubical/HITs/SetCoequalizer/Properties.agda
HilpAlex: Line is now shorter

Cubical/Algebra/Group/Abelianization/Properties.agda
HilpAlex: added module and definition

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
```diff
@@ -0,0 +1,398 @@
...
+
+  1Ab : Abelianization
+  1Ab = incAb 1g
+
+  invAb : Abelianization → Abelianization
+  invAb =
+    rec
+      squash
+      (λ x → incAb (inv x))
+      (λ (a , b , c) → incAb (inv (a · (b · c)))              ≡⟨ cong incAb (invDistr a (b · c)) ⟩
```
felixwellen, line 104: lines could easily be shortened, by line breaking before the arrow...

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
```diff
@@ -0,0 +1,398 @@
...
+      (λ (x , y , z) → x · z · y)
+
+  -- some convenient relabellings, so it looks more similar to the abelianization as a HIT.
+  incAb : G' → Abelianization
+  incAb = inc
+
+  comm : (x y z : G') → incAb (x · (y · z)) ≡ incAb (x · (z · y))
+  comm = λ x y z → coeq (x , y , z)
+
+  -- Definition of the group structure on the abelianization. Here the generality of the coequalized functions is used.
```
felixwellen, line 76: This line could easily be shortened

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
```diff
@@ -0,0 +1,398 @@
...
+              _ = snd (AbGroup→Group H)
+            f' : fst G → fst H
+            f' = fst f
+            g : Abelianization → fst H
+            g = rec
+                  (isSetAbGroup H)
+                  (λ x → (f') x)
+                  (λ (a , b , c) → f' (a · b · c)     ≡⟨ (snd f).pres· a (b · c) ⟩
+                            (f' a) · (f' (b · c))    ≡⟨ cong (λ x → (f' a) · x) ((snd f).pres· b c) ⟩
+                            (f' a) · (f' b) · (f' c) ≡⟨ cong (λ x → (f' a) · x) ((snd H).AbGroupStr.comm (f' b) (f' c)) ⟩
```
felixwellen, line 207: Please line break here... (you can just put "((snd ..." under "(lambda x -> ..." )

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
HilpAlex: done

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
HilpAlex: done

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
HilpAlex: done

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
HilpAlex: is this still necessary, with the proof being its own module? i feel like i should then export the other two lemmas too. i am open for renaming it though

Cubical/Algebra/Group/Abelianization/Properties.agda
HilpAlex: if the name should be changed, then to ·AbAssoc (as this is asscociativity and not commutativity). Since the name won't get exported i'll resolve the conversation though

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
HilpAlex: same here

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
HilpAlex: done

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
HilpAlex: done

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
HilpAlex: done

HilpAlex:
lines now all have less then 101 characters

Cubical/Algebra/Group/Abelianization/AbelianizationAsCoeq.agda
HilpAlex: exported the isomorphism directly

mortberg (approved):

PR #816, opened 2022-05-24
Describe kernel of ring and algebra quotient maps

MatthiasHu:
Show that the kernel of the canonical map to the ring quotient, is actually the ideal the ring was quotiented by. Transfer the result to CommRing and CommAlgebra.


felixwellen:
This replaces #810

felixwellen:
I'm happy with this PR and will merge in the evening, if there are no objections.

PR #817, opened 2022-05-25
improve group induced structure

ecavallo:
Make it so that only properties are induced, not operations. This should be done for AbGroup also.

### DIFF
--- a/Cubical/Algebra/Group/GroupPath.agda
+++ b/Cubical/Algebra/Group/GroupPath.agda
@@ -52,32 +52,37 @@ GroupPath = ∫ 𝒮ᴰ-Group .UARel.ua
 
 -- TODO: Induced structure results are temporarily inconvenient while we transition between algebra
 -- representations
-module _ (G : Group ℓ) {A : Type ℓ} (m : A → A → A)
+module _ (G : Group ℓ) {A : Type ℓ}
+  (m : A → A → A)
+  (u : A)
+  (inverse : A → A)
   (e : ⟨ G ⟩ ≃ A)
   (p· : ∀ x y → e .fst (G .snd ._·_ x y) ≡ m (e .fst x) (e .fst y))
+  (pu : e .fst (G .snd .1g) ≡ u)
+  (pinv : ∀ x → e .fst (G .snd .inv x) ≡ inverse (e .fst x))
   where
 
   private
     module G = GroupStr (G .snd)
 
-    FamilyΣ : Σ[ B ∈ Type ℓ ] (B → B → B) → Type ℓ
-    FamilyΣ (B , n) =
-      Σ[ e ∈ B ]
-      Σ[ i ∈ (B → B) ]
-      IsGroup e n i
+    BaseΣ : Type (ℓ-suc ℓ)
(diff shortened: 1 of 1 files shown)


thomas-lamiaux:
I don't remember exactly what was said but I think you were saying that it would be pratical to have a raw ring structure ?
But there is https://github.com/agda/cubical/blob/master/Cubical/Algebra/CommRingSolver/RawRing.agda in the library, isn't what you were talking about ?

mortberg:
Yeah, but I don't know how useful it is... I'll implement the corresponding induced structure lemmas for rings today. If I don't run into any problems I guess we don't really need raw rings

mortberg:
@ecavallo  The CI fails. Seems to be something about cohomology... I guess @aljungstrom  was relying on the bad version of the lemma?

mortberg:
I'll investigate locally and hopefully make a PR fixing everything

aljungstrom:
The issue with this approach is of course that we don't always care about the unit and the inverse of the induced group (this is the case in The DirectProof file). I can fix it (and I probably should for various reasons) but doesn't it make more sense to have both versions of the induced group?

mortberg:
> The issue with this approach is of course that we don't always care about the unit and the inverse of the induced group (this is the case in The DirectProof file). I can fix it (and I probably should for various reasons) but doesn't it make more sense to have both versions of the induced group?

Yeah, having both version could make sense. But I'm surprised you don't care about the unit and inverse of the induced group... Don't they pop up at some later point in the proof? With the current approach they are most likely defined using useless transports which will be annoying to get rid of everywhere...

mortberg:
> They never show up actually. I am only using the induced stuff to be able to trace my term of interest step by step rather than in one go, so I am not really relying on much of the group structure on pi*. I just need to know that the maps are homomorphisms and that pi* is a group -- nothing more than that really.
> 
> In this particular case, I'll probably want to know what the other stuff in pi* is for future work, so I have no problem filling the holes if we decide to only use Evan's definition. But it still feels reasonable to keep both: using induced groups in order to split up longer proofs into smaller pieces is a really neat, and not having to specify additional data of course simplifies things.

Ok, fair enough! I'm convinced that we should have two versions of the induced structure, one where the user can specify as much of the raw structure as possible and one where they can specify the minimal amount

thomas-lamiaux:
Then a warning should be add so as people don't think it is reduced but equivalent version like for group morphism

mortberg:
> Then a warning should be add so as people don't think it is reduced but equivalent version like for group morphism

Well, the two induced groups are equivalent as the unit and inverse are always unique. But the operations are not definitionally the same. I'll add a comment about this

thomas-lamiaux:
This is true that they are equivalent. Do we have a word for two structures that are equivalent but have different equality or computation rules ? For instance Vec and FinVec ?

mortberg:
> This is true that they are equivalent. Do we have a word for two structures that are equivalent but have different equality or computation rules ? For instance Vec and FinVec ?

I can't think of a great name for this... Anyway,to make things even more confusing univalence implies that two such types are equal, but that of course doesn't mean that they are the same :-)

mortberg:
I've implemented @aljungstrom 's suggestion now on https://github.com/agda/cubical/pull/827, so I'll close this PR as the other will supersede it once I've figured out how to get the induced ring structure to work

PR #818, opened 2022-05-25
Principal ideal quotient fp

felixwellen:
Joint work with @MatthiasHu :
Construct finite representaion of quotients by principal ideals.
Depends on #810 -> Draft


Cubical/Algebra/CommAlgebra/Base.agda
```diff
@@ -1,4 +1,4 @@
-{-# OPTIONS --safe #-}
+{-# OPTIONS --safe --postfix-projections #-}
```
MatthiasHu, line 1: I think it was unintended to commit this?

Cubical/Algebra/CommAlgebra/Instances/Initial.agda
```diff
@@ -1,4 +1,4 @@
-{-# OPTIONS --safe #-}
+{-# OPTIONS --safe --postfix-projections #-}
```
MatthiasHu, line 1: (Here also.)

felixwellen:
@MatthiasHu thanks!

felixwellen:
The depency got replaced (by #816 ), and the replacement is merged. So this can be merged as well.

felixwellen:
The PR also contains a fruitless attempt to make things type check faster by using more copatterns. Since this is a rule anyway, I'll leave that as it is. I'll have a look if the PR checks with latest master and merge if yes.

PR #821, opened 2022-05-26
More copatterns in Cubical.Algebra

felixwellen:
Try to make things type-check faster by using copatterns everywhere.
However, nothing got faster so far...


thomas-lamiaux:
This file couldn't be type checked without the flag --loosy-experimentation.
Maybe here it would help ?
https://github.com/agda/cubical/blob/master/Cubical/Algebra/Polynomials/Multivariate/EquivCarac/Polyn-nPoly.agda

thomas-lamiaux:
@felixwellen I didn't understood, do you want to merge this after of before the renaming one ?

felixwellen:
Thanks for reminding - I think we decided to merge this now (yesterday, so I'll do that after checking that it type checks with newest master...

mortberg:
Looks good to me!

PR #824, opened 2022-05-27
Enforcing the new algebra naming conventions for Algebraic Structures

thomas-lamiaux:
I have started to enforce the [algebra naming convention ](https://github.com/agda/cubical/blob/master/Cubical/Algebra/NAMING.md) for algebraic structure. 
I will just modify the Base.agda and fix the dependencies.
However, I will not enforce notation like Invol in the properties files that might be done in a next PR.


thomas-lamiaux:
@felixwellen or @mortberg is there a reason for https://github.com/agda/cubical/blob/master/Cubical/Algebra/Monoid/Base.agda#L36 ?
Because IdL and IdR are beneath, and in practice one always use that and not identity.
So I don't see the point of having such constructors ?
Shouldn't we replace it by two constructors IdL and IdR ?

kangrongji:
I guess the reason is, `identity` is an h-proposition. One-side inverse is not unique in a non-commutative monoid.

mortberg:
> @felixwellen or @mortberg is there a reason for https://github.com/agda/cubical/blob/master/Cubical/Algebra/Monoid/Base.agda#L36 ? Because IdL and IdR are beneath, and in practice one always use that and not identity. So I don't see the point of having such constructors ? Shouldn't we replace it by two constructors IdL and IdR ?

I don't think there's any deep reason for this and it could as well be split into two fields

thomas-lamiaux:
> One-side unit is not unique under non-commutative multiplication.

I don't really understand what you mean ?
Because the way, one proved that it is a proposition is by using 
`isPropΣ : isProp A → ((x : A) → isProp (B x)) → isProp (Σ A B)`
and so by having as in hypothesis that it is a Set then by funExt it follows.

This would be case whether it separated or not.
Actually the current proof is done by proving them separately using 
`(λ semi → isPropΠ λ _ → isProp× (semi .is-set _ _) (semi .is-set _ _)))`

thomas-lamiaux:
@mortberg Then should I replace it by two fields ? (and the similar ones)
Because It seems to me as adding a useless construction in the basic files.

kangrongji:
> > One-side unit is not unique under non-commutative multiplication.
> 
> I don't really understand what you mean ? Because the way, one proved that it is a proposition is by using `isPropΣ : isProp A → ((x : A) → isProp (B x)) → isProp (Σ A B)` and so by having as in hypothesis that it is a Set then by funExt it follows.
> 
> This would be case whether it separated or not. Actually the current proof is done by proving them separately using `(λ semi → isPropΠ λ _ → isProp× (semi .is-set _ _) (semi .is-set _ _)))`

Oh... You're right, the ε is given as parameter.

mortberg:
> @mortberg Then should I replace it by two fields ? (and the similar ones) Because It seems to me as adding a useless construction in the basic files.

Sure, sounds good

thomas-lamiaux:
Should the following be deleted ? https://github.com/agda/cubical/blob/master/Cubical/Algebra/Group/Base.agda#L32

mortberg:
> Should the following be deleted ? https://github.com/agda/cubical/blob/master/Cubical/Algebra/Group/Base.agda#L32

Maybe not deleted, but maybe moved to abelian groups? Additive groups are usually abelian...

thomas-lamiaux:
Yes this what I meant, this shouldn't be in Group but AbGroup

thomas-lamiaux:
should be merge after the others

thomas-lamiaux:
I have introduced nice notation for AbGroup 
```
  open IsGroup isGroup public
    renaming
      ( ·Assoc      to +Assoc
      ; ·IdL        to +IdL
      ; ·IdR        to +IdR
      ; ·InvL       to +InvL
      ; ·InvR       to +InvR)
```
but then, it would also be nice to be able to do something like 
```
pres1       to pres0
presinv    to pres-
pres·        to pres+
```
but I don't really know how to that as right now the definition is just calling back the definition of group

mortberg:
> I have introduced nice notation for AbGroup
> 
> ```
>   open IsGroup isGroup public
>     renaming
>       ( ·Assoc      to +Assoc
>       ; ·IdL        to +IdL
>       ; ·IdR        to +IdR
>       ; ·InvL       to +InvL
>       ; ·InvR       to +InvR)
> ```
> 
> but then, it would also be nice to be able to do something like
> 
> ```
> pres1       to pres0
> presinv    to pres-
> pres·        to pres+
> ```
> 
> but I don't really know how to that as right now the definition is just calling back the definition of group

Won't we have the same problem for pretty much all the theory about groups? If lemmas are named using the multiplicative notation then they will look ugly for the additive notation. I don't know if it matters or what we can do.

thomas-lamiaux:
@felixwellen I am currently fixing the dependencies.
In you PR, why did you put stuff like AbGroup->Group or makeAbGroup in a module ?
Is that to prevent it to unfold ?

If there is really some specific trick that is known hat prevent unfolding  or thing like this, that really made working easier, maybe it should be mention in contributing ?
Like recommended practice to enhanced the code and computation, so people can be aware ?

thomas-lamiaux:
@mortberg Would it be possible to merge this before others conflict are added ? 
There is currently no PR that conflict it except mine but I rather fix the direct sum PR directly than fixing this one again

mortberg:
> @mortberg Would it be possible to merge this before others conflict are added ? There is currently no PR that conflict it except mine but I rather fix the direct sum PR directly than fixing this one again

Yes, I will merge as soon as the CI is finished

mortberg:
> @mortberg Would it be possible to merge this before others conflict are added ? There is currently no PR that conflict it except mine but I rather fix the direct sum PR directly than fixing this one again

Ignore my other message

Cubical/Algebra/CommAlgebra/Base.agda
```diff
@@ -84,57 +84,76 @@ module _ {R : CommRing ℓ} where
...
+                       r ⋆ (y · x) ≡⟨ sym (⋆AssocL _ _ _) ⟩
+                       (r ⋆ y) · x ≡⟨ ·Comm _ _ ⟩
+                       x · (r ⋆ y) ∎
+    makeIsCommAlgebra .IsCommAlgebra.·Comm = ·Comm
+
 
   module _ (S : CommRing ℓ') where
     open CommRingStr (snd S) renaming (1r to 1S)
     open CommRingStr (snd R) using () renaming (_·_ to _·R_; _+_ to _+R_; 1r to 1R)
+-- <<<<<<< HEAD
```
mortberg, line 134: This comment should be deleted

Cubical/Algebra/CommMonoid/GrothendieckGroup.agda
```diff
@@ -74,13 +72,13 @@ module _ (M : CommMonoid ℓ) where
 
             -- remove proof when MonoidSolver is done
```
mortberg, line 73: Don't we have a monoid solver that can be used now ( https://github.com/agda/cubical/tree/master/Cubical/Algebra/MonoidSolver)? @felixwellen

Cubical/Algebra/Group/Base.agda
```diff
@@ -1,8 +1,3 @@
-{-
```
mortberg, line 1: Why remove this header?

Cubical/Algebra/Group/Instances/Bool.agda
```diff
@@ -20,35 +20,37 @@ open import Cubical.Algebra.Semigroup
 open GroupStr
 open IsGroup
 open IsMonoid
-open IsSemigroup renaming (assoc to assoc')
+open IsSemigroup
 
 
 
 BoolGroup : Group₀
```
mortberg, line 27: Don't we have some operations for Bool that could be used here?

Cubical/Algebra/Group/Properties.agda
```diff
@@ -1,8 +1,3 @@
-{-
```
mortberg, line 1: Why remove comment?

Cubical/Algebra/Ring/Base.agda
```diff
@@ -100,36 +83,54 @@ Ring ℓ = TypeWithStr ℓ RingStr
 isSetRing : (R : Ring ℓ) → isSet ⟨ R ⟩
 isSetRing R = R .snd .RingStr.isRing .IsRing.·IsMonoid .IsMonoid.isSemigroup .IsSemigroup.is-set
 
+-- -- <<<<<<< HEAD
```
mortberg, line 86: Remove comment

Cubical/Algebra/Ring/Base.agda
```diff
@@ -140,8 +141,9 @@ module _ {R : Type ℓ} (0r 1r : R) (_+_ _·_ : R → R → R) (-_ : R → R)
   makeRing .snd .RingStr._·_ = _·_
   makeRing .snd .RingStr.-_ = -_
   makeRing .snd .RingStr.isRing =
-    makeIsRing is-setR +-assoc +-rid +-rinv +-comm
-                       ·-assoc ·-rid ·-lid ·-rdist-+ ·-ldist-+
+    makeIsRing is-setR +Assoc +IdR +InvR +Comm
+                       ·Assoc ·IdR ·IdL ·DistR+ ·DistL+
+-->>>>>>> master
```
mortberg, line 146: Please remove comment

Cubical/Algebra/Ring/BigOps.agda
```diff
@@ -138,7 +125,3 @@ module Product (R' : Ring ℓ) where
  ∏Ext = bigOpExt
  ∏0r = bigOpε
  ∏Last = bigOpLast
-
- -- only holds in CommRings!
```
mortberg, line 142: I think this comment can be good to keep in case someone is confused :-)

Cubical/Data/Int/Divisibility.agda
```diff
@@ -21,6 +21,7 @@ open import Cubical.Data.Nat.Divisibility
 open import Cubical.Data.Nat.Mod
 open import Cubical.Data.Int
   hiding   (_+_ ; _·_ ; _-_ ; -_ ; addEq ; ·Comm ; ·Assoc ; +Comm ; +Assoc)
+  renaming (·DistL+ to ·ℤDistL+)
```
mortberg, line 24: Is this really being used? Otherwise it might be better to hide

mortberg (changes requested):
Looks great! Massive improvement to the Algebra package!

I found a few small things that should be fixed before we can merge

Cubical/Algebra/Group/Base.agda
thomas-lamiaux: I don't see the point in a file name Group.Base + "adds some smart constructors" is not prise and nor helpful to not what is inside

thomas-lamiaux: It probably just refers to the make functions which are in every base file

Cubical/Algebra/Group/Properties.agda
thomas-lamiaux: It is a pleonasm. I don't see the point to add something already  in the name of the file.
But it can be keept too, it doesn't really matter to me

Cubical/Data/Int/Divisibility.agda
thomas-lamiaux: now hidden

thomas-lamiaux:
The bug that broke the last check is a bit unexpected. I think that I removed some useless import.
But then someone used them to add something in an other branch and then when merging it removed them even though they were now needed.

Cubical/Algebra/Group/Base.agda
thomas-lamiaux: > These constructors are extra smart, so it would be better to document them in the header instead of removing it
Then what would like to add ? Just saying some smart constructors is not really useful.
Should we add something in all Base files that have such constructors ?

thomas-lamiaux: Sounds good to me, I'll add that once the test is finished.

thomas-lamiaux:
@mortberg good to go

mortberg (approved):

mortberg:
No objections, so merged!

PR #827, opened 2022-05-30
More induced structures

mortberg:
Meant to replace https://github.com/agda/cubical/pull/817


Cubical/Algebra/Ring/Base.agda
```diff
@@ -240,6 +241,60 @@ uaRing {A = A} {B = B} = equivFun (RingPath A B)
...
+    module R = RingStr (R .snd)
+
+    BaseΣ : Type (ℓ-suc ℓ)
+    BaseΣ = Σ[ B ∈ Type ℓ ] B × B × (B → B → B) × (B → B → B) × (B → B)
+
+    FamilyΣ : BaseΣ → Type ℓ
+    FamilyΣ (B , u0 , u1 , a , m , i) = IsRing u0 u1 a m i
+
+    inducedΣ : FamilyΣ (A , 0a , 1a , add , mul , inv)
+    inducedΣ = {!!}
```
mortberg, line 271: @ecavallo I tried to define InducedRing, but Agda gets extremely slow when I try to fill this hole with the code below... Any clue why?

Cubical/Algebra/Matrix.agda
```diff
@@ -169,14 +188,13 @@ module _ (G' : AbGroup ℓ) where
   -- VecMatrix that is equal to the one on FinMatrix
   VecMatrixAbGroup : (m n : ℕ) → AbGroup ℓ
   VecMatrixAbGroup m n =
-    InducedAbGroup (FinMatrixAbGroup G' m n) addVecMatrix
-      FinMatrix≃VecMatrix (FinMatrix→VecMatrixHomAdd m n)
+    InducedAbGroup (FinMatrixAbGroup G' m n) addVecMatrix zeroVecMatrix negVecMatrix FinMatrix≃VecMatrix
```
mortberg, line 191: @aljungstrom This is the fix to the Matrix code with the new InducedAbGroup. Can you fix the DirectProof code? 

Alternatively we can have InducedAbGroupFromAbGroupHom which only requires preservation of +, but this seems bad as we also want to be able to pick what the 0 and - of the induced group are...

Cubical/Algebra/Ring/Base.agda
mortberg: What I've noticed is that there seems to be an exponential explosion when running `autoUARel BaseΣ` as  `BaseΣ`  gets bigger

thomas-lamiaux: Here I am not sure if it plays but you have not finished to define `FamilyΣ` but you are already using it. 
When I do that it often gets slow. I guess that agda must be trying to infer some constraints and then take time to see that you don't need the definition for it to work. 
You can try commenting the rest, fill the hole then uncomment ?
It can sometimes speed up the code drastically

mortberg: Yeah I know, it doesn't matter if I uncomment everything else in the file. I've localized the problem to autoUARel as I wrote above. It seems like the time it takes for the automation is exponential in the size of the input signature...

mortberg:
@thomas-lamiaux  : can you please check that this is what you need for your PR? 

For now I'm only doing Group, AbGroup and Ring. I can open an issue about doing this for other structures once we merge this

thomas-lamiaux:
> @thomas-lamiaux : can you please check that this is what you need for your PR?
That seems good to me
> For now I'm only doing Group, AbGroup and Ring. I can open an issue about doing this for other structures once we merge this
Or people can just add them when they need it ?

PR #829, opened 2022-06-01
Add non-dependent product special case to uarel automation

ecavallo:
Should help with #827. Adding a special case for structures that are built up with non-dependent products (as opposed to general sigmas) means we can avoid generating intermediate DUARels over very complicated base UARels when it isn't really necessary.


mortberg (approved):
Cool! Is it ready to be merged and tested on the other branch?

ecavallo:
yup

PR #836, opened 2022-06-04
Weak Equivalence between Categories

kangrongji:
This PR adds the notion of weak equivalence of category, i.e. functors that are both fully-faithful and essentially surjective. The main result is that, weak equivalence between univalent categories is equivalence, namely it admits explicit inverse functor.

P.S. I change the definition of isomorphisms in category `CatIso`, by separating it into a record predicate `isIso` and re-define `CatIso` as a Σ-type. This allows some definitions to be more directly formulated, and some arguments easier. Forgive me if this makes review more painful...

P.P.S. Instead of the brutal-force argument in HoTT book, I use a dependent version of isomorphism to formulate a more structural proof. Price is a new file `Cubical.Foundation.Equiv.Dependent`.


Cubical/Categories/Constructions/BinProduct.agda
```diff
@@ -62,3 +63,14 @@ module _ (C : Category ℓC ℓC')
     - define inverse to `assoc`, prove isomorphism
     - prove product is commutative up to isomorphism
 -}
+
+
+  -- The isomorphisms in product category
+
+  open isIso
+
+  iso× : {x y : C .ob}{z w : D .ob} → CatIso C x y → CatIso D z w → CatIso (C × D) (x , z) (y , w)
```
mortberg, line 72: Please rename to `CatIso×`

Cubical/Categories/Category/Base.agda
```diff
@@ -51,40 +52,63 @@ comp' = _∘_
 infixr 16 comp'
 syntax comp' C g f = g ∘⟨ C ⟩ f
 
+
 -- Isomorphisms and paths in categories
-record CatIso (C : Category ℓ ℓ') (x y : C .ob) : Type ℓ' where
-  constructor catiso
+
+record isIso (C : Category ℓ ℓ'){x y : C .ob}(f : C [ x , y ]) : Type ℓ' where
```
mortberg, line 58: Maybe we should have no-eta-equality for this? Might be good to avoid slow typechecking times

mortberg (changes requested):
Two small comments for things to consider changing. Apart from this it looks great! Sorry for the slow review. I'll take a look at your later PRs once this is merged.

This seems like a good refactoring of CatIso and some useful stuff about dependent isos/equivs

Cubical/Categories/Category/Base.agda
kangrongji: Oh, the extra judgemental equality is very useful, e.g. [symmetry of inverse](https://github.com/agda/cubical/blob/cc8812edc0d3ac57e60cfebe5dd42c49fcb773ff/Cubical/Categories/Morphism.agda#L129), and I think it is used in many places in the library.

kangrongji: The typechecking time is not a problem so far, do you think it is necessary to change it?

mortberg: Ah ok, if it's used several times and simplifies proofs but doesn't lead to any problems for typechecking times then we should not remove it

mortberg (approved):

PR #837, opened 2022-06-05
[ cosmetics ] Fix comments to address `--two-levels`

### LINKED ISSUE #771: Comments on types of primitives are wrong.
https://github.com/agda/cubical/blob/1b9c97a2140fe96fe636f4c66beedfd7b8096e8f/Cubical/Core/Primitives.agda#L150

Is this comment correct? It seems like `A` in `transp` can vary across universes. E.g.

```agda
forward : {ℓ : I → Level} (A : ∀ i → Type (ℓ i)) (r : I) (u : A r) → A i1
forward A r u = transp (λ i → A (i ∨ r)) r u
```

ice1000:
Fix #771

### DIFF
--- a/Cubical/Core/Primitives.agda
+++ b/Cubical/Core/Primitives.agda
@@ -30,7 +30,8 @@ import Agda.Builtin.Cubical.Glue
 -- import Agda.Builtin.Cubical.HCompU
 
 open import Agda.Primitive public
-  using    ( Level )
+  using    ( Level
+           ; SSet )
   renaming ( lzero to ℓ-zero
            ; lsuc  to ℓ-suc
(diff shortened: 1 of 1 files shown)


Saizan:
I have gotten some negative feedback on `SSet` as the name of the type. Maybe we should pick something something else

ice1000:
> I have gotten some negative feedback on `SSet` as the name of the type. Maybe we should pick something something else

Yes, I heard people saying that `SSet` means simplicial sets, not strict sets as we intended. Do you have some ideas?

ice1000:
Anyhow, I am going to merge this because at least it makes things better and we can rename `SSet` when we renamed it in the compiler.

PR #847, opened 2022-06-20
RP2

thomas-lamiaux:
Add RP2, in progress


thomas-lamiaux:
@aljungstrom I miss H4 RP2 to do my proof

thomas-lamiaux:
@mortberg 
Hi, just to tell you that the proof are basically done up to 

- @aljungstrom  add a full characterization of H^n (RP²), there is only 1-3 for the moment
- I add [the following lemma ](https://github.com/agda/cubical/pull/847/files#diff-594043243f8a19f0e56a80fe9d5d958a2d45c098e411924b3e8126119ea24560R422). 
Nothing very hard just annoying, I'll come back on that when Axel is done with what is missing

About working with the Z/k, I think this prove it stills works up to a few tricks.
Though it would be a bit more complicated in the general case as Axel used Bool for Z/2.
But this would only change some details.
The good news too, is that the scheme of the proof doesn't change and the additional properties can be proven relatively simply by level as before. Plus, having more then one object in the ideal quotienting doesn't change anything when working.

thomas-lamiaux:
@mortberg Finally done with that !

Cubical/Data/Int/IsEven.agda
```diff
@@ -0,0 +1,147 @@
...
+  using ()
+  renaming
+  (isEvenFalse  to isEvenℕFalse   ;
+   isEvenTrue   to isEvenℕTrue    ;
+   isOddFalse   to isOddℕFalse    ;
+   isOddTrue    to isOddℕTrue     ;
+   falseIsEven  to falseIsEvenℕ   ;
+   trueIsEven   to trueIsEvenℕ    ;
+   ¬IsEvenFalse to ¬IsEvenℕFalse  ;
+   ¬IsEvenTrue  to ¬IsEvenℕTrue   )
```
MatthiasHu, line 27: Hi, just a comment: I think instead of these renamings it might be better to write `import Cubical.Data.Nat.IsEven as ℕ`, without the `open`, to avoid name clashes.

thomas-lamiaux: Would this work ? 
From what I remember, importing a module as something doesn't hide the definition without N.X , right ? 
IBut it is still possible to call X directly and so it would conflict with the other definitions ?

MatthiasHu: Yes, it works, I just double-checked, see here: https://github.com/thomas-lamiaux/cubical/pull/1

It is important that you drop the "open". :-)

thomas-lamiaux: ha yeah, nice, I'll modify that before merging

thomas-lamiaux: resolved

mortberg (approved):

PR #848, opened 2022-06-23
Examples for modalities

MatthiasHu:
This PR provides a few simple examples of modalities (as defined in `Cubical.Modalities.Modality`). Note that there is already `HLevelTruncModality` in `Cubical.HITs.Truncation.Properties`.


MatthiasHu:
I wonder whether there is a more elegant definition for the open modality. But I can't find one, so I leave it like this for now.

felixwellen (approved):
Thanks!

felixwellen:
I don't know a better presentation of the open modality.

PR #852, opened 2022-06-30
Generalized an example for finitely presented algebra

MatthiasHu:
This PR generalizes a simple fact about finitely presented algebras. Instead of dividing out a single element in the base ring, we now divide out a finite list of elements.


felixwellen:
Looks all good -> merging.

PR #853, opened 2022-06-30
Self contained version of the new Brunerie number

mortberg:
This should probably not be merged, but I figured that it's easier to discuss it if we have a PR


mortberg:
@aljungstrom  @ecavallo  Can we gold the definition some more?

Cubical/Homotopy/Group/Pi4S3/BrunerieSelfcontained.agda
```diff
@@ -0,0 +1,111 @@
...
+encodeTruncS¹ : Ω ∥ S¹∙ ∥₁∙ .fst → ∥ ℤ ∥₀
+encodeTruncS¹ p = transport (λ i → codeTruncS¹ (p i) .fst) ∣ pos 0 ∣₀
+
+g8 : Ω² ∥ S²∙ ∥₂∙ .fst → Ω ∥ S¹∙ ∥₁∙ .fst
+g8 p i = encodeTruncS² (p i)
+
+g9 : Ω ∥ S¹∙ ∥₁∙ .fst → ∥ ℤ ∥₀
+g9 = encodeTruncS¹
+
+-- Use trick to eliminate away the truncation last
```
mortberg, line 64: @ecavallo  I wonder if this trick of yours makes any difference... What do you think?

mortberg: So g8/g9 could potentially land-in/start-from `Ω S¹`

PR #856, opened 2022-07-03
Abstract interface for QuotientAlgebra

MatthiasHu:
This PR provides an abstract interface to the `Cubical.Algebra.CommAlgebra.QuotientAlgebra` module as an attempt to reduce long type checking times.

`FPAlgebra` is the only module that uses `QuotientAlgebra` so far, and relatively small changes were necessary for it to use `QuotientAlgebraAbstract` instead. A witness `inducedHom∘quotientHom` had to be added to `QuotientAlgebra` for an equality that was a definitional equality before.

### DIFF
--- a/Cubical/Algebra/CommAlgebra/FPAlgebra.agda
+++ b/Cubical/Algebra/CommAlgebra/FPAlgebra.agda
@@ -163,12 +163,13 @@ module _ {R : CommRing ℓ} where
           (inducedHom A values relationsHold)
           f
           (sym (
-           {!f'                                   ≡⟨ sym (inv f') ⟩
+           f'                                    ≡⟨ sym (inv f') ⟩
            freeInducedHom A (evaluateAt A f')    ≡⟨ cong (freeInducedHom A)
                                                          (funExt hasCorrectValues) ⟩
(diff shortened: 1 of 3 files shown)


MatthiasHu:
Total checking time of `FPAlgebra.agda` dropped from 42,518ms to 14,423ms for me.

MatthiasHu:
Also, deriving the special case `R/⟨x⟩FP` from `R/⟨xs⟩FP` (see the very end of `FPAlgebra.agda`) was previously impossible (type checking would hang indefinitely).

MatthiasHu:
Note: We still can't remove the huge `abstract` block in `FPAlgebra.agda`. But perhaps we could if we also had an abstract interface for `FreeCommAlgebra`?

felixwellen:
I haven't looked at it too closely, but wouldn't it be better to just *replace* QuotientAlgebra with your abstract version? Or is that where you want to go anyway?
It makes sense for me to just keep/move to abstract interfaces/exports in all three cases.

MatthiasHu:
I made a new PR for that, since most changes made here are not needed any more: #857

felixwellen:
I this PR still needed or can we close it?

MatthiasHu:
Can be closed.

PR #857, opened 2022-07-04
abstractify QuotientAlgebra

MatthiasHu:
This PR declares almost all definitions in `Cubical.Algebra.CommAlgebra.QuotientAlgebra` as `abstract` in an attempt to reduce long type checking times.

`FPAlgebra` (the only module that uses `QuotientAlgebra` so far) had to be slightly adjusted too. A witness `inducedHom∘quotientHom` had to be added to `QuotientAlgebra` for an equality that was a definitional equality before.

This replaces #856.

### DIFF
--- a/Cubical/Algebra/CommAlgebra/FPAlgebra.agda
+++ b/Cubical/Algebra/CommAlgebra/FPAlgebra.agda
@@ -109,6 +109,11 @@ module _ {R : CommRing ℓ} where
         (values : FinVec ⟨ A ⟩ n)
         (relationsHold : (i : Fin m) → evPoly A (relation i) values ≡ 0a (snd A))
         where abstract
+        {-
+          We repeat the abstract keyword here because the contents of a child module are not
+          covered by the outer abstract keyword. This behaves like a single big abstract
+          block as long as the child module is anonymous.
+        -}
 
         private
           freeHom : CommAlgebraHom (Polynomials n) A
--- a/Cubical/Algebra/CommAlgebra/FPAlgebra.agda
+++ b/Cubical/Algebra/CommAlgebra/FPAlgebra.agda
@@ -149,13 +149,13 @@ module _ {R : CommRing ℓ} where
                          relationsHold
 
       unique :
-             {A : CommAlgebra R ℓ}
+             (A : CommAlgebra R ℓ)
              (values : FinVec ⟨ A ⟩ n)
              (relationsHold : (i : Fin m) → evPoly A (relation i) values ≡ 0a (snd A))
              (f : CommAlgebraHom FPAlgebra A)
              → ((i : Fin n) → fst f (generator i) ≡ values i)
              → inducedHom A values relationsHold ≡ f
-      unique {A = A} values relationsHold f hasCorrectValues =
+      unique A values relationsHold f hasCorrectValues =
         injectivePrecomp
           (Polynomials n)
           relationsIdeal
@@ -195,7 +195,7 @@ module _ {R : CommRing ℓ} where
           , (inducedHomOnGenerators A values relationsHold) )
         , λ {(f , mapsValues)
             → Σ≡Prop (λ _ → isPropΠ (λ _ → isSetCommAlgebra A _ _))
-                     (unique values relationsHold f mapsValues)}
+                     (unique A values relationsHold f mapsValues)}
 
       {- ∀ A : Comm-R-Algebra,
          ∀ J : Finitely-generated-Ideal,
@@ -246,7 +246,7 @@ module _ {R : CommRing ℓ} where
                 (funExt (inducedHomOnGenerators A (fst b) (snd b)))
       Iso.leftInv (FPHomIso {A}) =
         λ a → Σ≡Prop (λ f → isPropIsCommAlgebraHom {ℓ} {R} {ℓ} {ℓ} {FPAlgebra} {A} f)
-                 λ i → fst (unique {A}
+                 λ i → fst (unique A
                              (fst (evaluateAtFP {A} a))
                              (snd (evaluateAtFP a))
                              a
@@ -305,13 +305,13 @@ module Instances (R : CommRing ℓ) where
         inverse1 : fromA ∘a toA ≡ idAlgebraHom _
         inverse1 =
           fromA ∘a toA
-            ≡⟨ sym (unique _ _ _ _ _ (λ i → cong (fst fromA) (
+            ≡⟨ sym (unique _ _ _ _ _ _ (λ i → cong (fst fromA) (
                  fst toA (generator n emptyGen i)
                    ≡⟨ inducedHomOnGenerators _ _ _ _ _ _ ⟩
                  Construction.var i
                    ∎))) ⟩
           inducedHom n emptyGen B (generator _ _) (relationsHold _ _)
-            ≡⟨ unique _ _ _ _ _ (λ i → refl) ⟩
+            ≡⟨ unique _ _ _ _ _ _ (λ i → refl) ⟩
           idAlgebraHom _
             ∎
         inverse2 : toA ∘a fromA ≡ idAlgebraHom _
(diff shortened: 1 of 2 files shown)


MatthiasHu:
Just a note: I can't see how to "abstractify" `FreeCommAlgebra' in the same way, because it consists of 'FreeCommAlgebra.Base' and 'FreeCommAlgebra.Properties', and I think there is no way to let the latter module look through abstract definitions in the first module. :-/

mortberg:
Is it clear that making everything abstract like this is the right way to go? My experience is that constructions should not be opaque, but their properties can often be... Also, the "locking" trick that we use in some files for Z-cohomology could be helpful. It's an alternative to making things abstract

MatthiasHu:
I think to avoid the unfolding of algebra terms, it has to be the construction (`_/_`) that is abstract. I think it would not help to make only properties abstract?
I will look at the locking trick...

mortberg:
> I think to avoid the unfolding of algebra terms, it has to be the construction (`_/_`) that is abstract. I think it would not help to make only properties abstract? I will look at the locking trick...

Yeah, maybe a locked version of `_/_` is what you need. I'm not sure if there's some better place to look than https://agda.github.io/cubical/Cubical.Experiments.ZCohomologyOld.Properties.html#24355, but it's the first I found

MatthiasHu:
I experimented a bit with the locking trick. In principle it seems very very nice to provide an interface where the user of the module can decide themselves whether they want to see reducing or non-reducing behavior (by plugging in either `unlock` or some general `key : lockUnit`). But it seems to be a considerable amount of overhead compared to basically just adding a few `abstract`s as in the present PR.

For example, the definition of `quotientHom` currently reads:
```agda
  quotientHom : CommAlgebraHom A (_/_)
  fst quotientHom x = [ x ]
  IsAlgebraHom.pres0 (snd quotientHom) = refl
  ...
```
With the locking method it would be either something like this:
```agda
  quotientHomWithKey : (key : lockUnit) → CommAlgebraHom A (_/WithKey_ key)
  fst (quotientHomWithKey unlock) x = [ x ]
  IsAlgebraHom.pres0 (snd (quotientHomWithKey unlock)) = refl
  ...

  quotientHom : CommAlgebraHom A _/_
  quotientHom = quotientHomWithKey key
```
Or something like this:
```agda
  quotientHomUnlocked : CommAlgebraHom A (_/Unlocked_)
  fst quotientHomUnlocked x = [ x ]
  IsAlgebraHom.pres0 (snd quotientHomUnlocked) = refl
  ...

  quotientHom : CommAlgebraHom A (_/_)
  quotientHom = lock' key
    (λ k → CommAlgebraHom A (lock k _/Unlocked_))
    quotientHomUnlocked
```

In particular, I don't see how to avoid writing out all the type signatures two (or even three) times, in slightly modified ways. This feels bad to me as they can become relatively long, e.g.:
```agda
  injectivePrecomp : (B : CommAlgebra R ℓ) (f g : CommAlgebraHom (A / I) B)
                     → f ∘a (quotientHom A I) ≡ g ∘a (quotientHom A I)
                     → f ≡ g
```

felixwellen:
I'm certainly ok with using 'abstract' for now. Should I make a review?

MatthiasHu:
Was the question directed at me? :-)

felixwellen:
Yes. With a possibility for anders to intervene.

MatthiasHu:
Oh, ok. Yes, I would be pleased if you could review this PR.

felixwellen:
I can confirm a very nice type checking speed reduction in FPAlgebra! On my notebook, it used to take 80s and now it is down to 20s - which is mostly spend on deserialization, so I guess there is no immediate need to look for further speed improbements to FPAlgebra.

Cubical/Algebra/CommAlgebra/QuotientAlgebra.agda
```diff
@@ -186,21 +207,22 @@ module _ {R : CommRing ℓ} (A : CommAlgebra R ℓ) (I : IdealsIn A) where
...
-isZeroFromIdeal : {R : CommRing ℓ} {A : CommAlgebra R ℓ} {I : IdealsIn A}
-                  → (x : ⟨ A ⟩) → x ∈ (fst I) → fst (quotientHom A I) x ≡ CommAlgebraStr.0a (snd (A / I))
-isZeroFromIdeal {A = A} {I = I} x x∈I = eq/ x 0a (subst (λ y → y ∈ (fst I)) step x∈I )
-  where
-    open CommAlgebraStr (snd A)
-    step : x ≡ x - 0a
-    step = lemma (CommAlgebra→CommRing A) x
-    0' : ⟨ A / I ⟩
-    0' = fst (quotientHom A I) 0a
+abstract
```
felixwellen, line 219: Merely a cosmetic suggestion: When reading, it is a bit surprising to see just an 'abstract' all of a sudden - maybe you could write it in the same style as above, i.e. in an annoynmous module?

Cubical/Algebra/CommAlgebra/QuotientAlgebra.agda
```diff
@@ -186,21 +207,22 @@ module _ {R : CommRing ℓ} (A : CommAlgebra R ℓ) (I : IdealsIn A) where
                             (isPropIsCommIdeal (CommAlgebra→CommRing A))
                             refl ⟩
     _                  ≡⟨  CommRing.kernel≡I {R = CommAlgebra→CommRing A} I ⟩
-    I ∎
+    I                  ∎
 
 
 private
-  module _ (R : CommRing ℓ) where
+  module _ (R : CommRing ℓ) where abstract
```
felixwellen, line 214: Is this 'abstract' needed for something? 'lemma' should be behind an abstract barrier anyway, right?

Cubical/Algebra/CommAlgebra/FPAlgebra.agda
```diff
@@ -104,21 +104,16 @@ module _ {R : CommRing ℓ} where
...
-        → CommAlgebraHom FPAlgebra A
-      inducedHom A values relationsHold =
-        quotientInducedHom
-          (Polynomials n)
-          relationsIdeal
-          A
-          freeHom
-          isInKernel
-        where
+        where abstract
```
felixwellen, line 111: Is the 'abstract'-block inside an 'abstract' block needed due to some weirdness of how 'abstract' works, or is it needed for making the inside definitions abstract to the other abstract definitions?
I guess a comment would be good here.

felixwellen (changes requested):
Only a couple of minor comments/requests for clarifications. Otherwise this is good to merge.
I think we can work with abstract interfaces like this one and see how it goes.

Cubical/Algebra/CommAlgebra/QuotientAlgebra.agda
MatthiasHu: Good idea, I will do that.

Cubical/Algebra/CommAlgebra/QuotientAlgebra.agda
MatthiasHu: No, that is not how the `abstract` keyword works (unfortunately). [From the documentation](https://agda.readthedocs.io/en/v2.6.2.2/language/abstract-definitions.html): "The reach of the `abstract` keyword does not extend into modules". I just double-checked that this is even the case if the inner module is anonymous.

MatthiasHu: I moved the lemma, but left it marked abstract, because it seems more natural to me to have it as an abstract private definition than to have it as a non-abstract private definition. But this is only a slight preference, I can change it if you like it better the other way.

felixwellen: I have no strong opinion on the inner abstract - just do what you like. But now that you have the outer module, you could move the implicit arguments of 'isZeroFromIdeal' to the outer module and use the correct ones in the lemma. Might look a bit impure, but I think it would be easier to read, if there aren't two different 'R's... But I want to leave it up to you, just let me know if you want to merge it like it is or make a change.

MatthiasHu: Agreed. But now I finally saw how to avoid the lemma all together, which is I think even better. :-)

felixwellen (approved):
Also from mine - looks great!

PR #858, opened 2022-07-04
Useful functions in Algebra such as Op2 = A -> A -> A 

guilhermehas:
I created some functions that can be useful in the Cubical Library, such as Op1 and Op2.

### DIFF
--- a/Cubical/Algebra/AbGroup/Base.agda
+++ b/Cubical/Algebra/AbGroup/Base.agda
@@ -13,6 +13,7 @@ open import Cubical.Foundations.SIP
 open import Cubical.Data.Sigma
 open import Cubical.Data.Unit
 
+open import Cubical.Algebra.Core
 open import Cubical.Algebra.Monoid
 open import Cubical.Algebra.CommMonoid
 open import Cubical.Algebra.Group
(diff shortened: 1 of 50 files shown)


felixwellen:
Sorry for the long wait! 
Can you explain, what the purpose of your changes is?

guilhermehas:
This idea came from Agda Stdlib:
https://github.com/agda/agda-stdlib/blob/8ffc4cccf64ca1626f866ad0bc0f7d965e705d6b/src/Algebra/Core.agda#L18

It is to reduce the number of characters and make the code more succint.

mortberg:
> This idea came from Agda Stdlib: https://github.com/agda/agda-stdlib/blob/8ffc4cccf64ca1626f866ad0bc0f7d965e705d6b/src/Algebra/Core.agda#L18
> 
> It is to reduce the number of characters and make the code more succinct.

Is this used a lot in the standard library? It doesn't seem very useful to me and I definitely find `A → A → A` a lot more informative and readable than `Op₂ A`...

felixwellen:
I agree with Anders, I am also a lot faster at recognizing ```A → A → A``` and someone new to this part of the library would actually have to look up the op-notation to check if it is really just a function type. 
I don't know why the stdlib is using this notation - I always assumed it is used for some kind of abstraction somewhere, e.g. functions using ```Op₂``` as an input datum.

felixwellen:
Since this PR is only about introducing the op-notation, I'm going to close it now. If you think we misunderstood/missed something, feel free to continue discussing or reopen it.

felixwellen (commented):
Closed.

PR #860, opened 2022-07-05
Associativity of smash product

aljungstrom:
A direct construction of the associativity of smash product. No pentagon yet tho :'(


ecavallo:
:open_mouth:

Could it be easier to prove `Iso (⋀×3 A B C) (⋀×3 C B A)`? instead of `Iso (⋀×3 A B C) (⋀×3 C A B)`?

PR #864, opened 2022-07-18
Local rings

MatthiasHu:
This PR provides a definition of (commutative) local rings, some simple consequences and two alternative characterizations.

### DIFF
--- a/Cubical/Algebra/CommRing/LocalRing.agda
+++ b/Cubical/Algebra/CommRing/LocalRing.agda
@@ -5,32 +5,29 @@
 {-# OPTIONS --safe #-}
 module Cubical.Algebra.CommRing.LocalRing where
 
--- TODO: imports
 open import Cubical.Foundations.Prelude
-open import Cubical.Foundations.Function
-open import Cubical.Foundations.Structure
-open import Cubical.Foundations.Powerset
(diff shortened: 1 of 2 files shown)


Cubical/Algebra/CommRing/LocalRing.agda
```diff
@@ -0,0 +1,229 @@
+{-
+  Definition of local rings.
+-}
```
felixwellen, line 3: Maybe a comment like: "Note that local rings are more intricate constructively. See "A Course in Constructive Algebra" by Mines, Richman & Ruitenberg, p. 96 ff, where the two characterizations in this file are used as a definition of local ring, but without the requirement that 1 is different from 0."

felixwellen (changes requested):
Awesome! Only one documentation request...

Cubical/Algebra/CommRing/LocalRing.agda
MatthiasHu: Nice, thanks! I also added a link to [the nlab page](https://ncatlab.org/nlab/show/local+ring).

felixwellen (approved):
Thanks! Great stuff :-)

felixwellen (approved):
Thanks! Great stuff :-)

PR #865, opened 2022-07-19
Insertion into 'FinVec's

felixwellen:
'insert' takes a finite vector of lenght n and turns it into a finite vector of length n+1, with a given element inserted at a given index.  If element and index are fixed, this is an isomorphism, to the type of n+1-vectors with prescribed element at the fixed index.


felixwellen:
Obsoleted by work of @MatthiasHu .

PR #870, opened 2022-07-25
List operations

anuyts:
List operations.


Cubical/Data/FinData/Properties.agda
```diff
@@ -73,6 +81,25 @@ injSucFin : ∀ {n} {p q : Fin n} → suc p ≡ suc q → p ≡ q
 injSucFin {ℕsuc ℕzero} {zero} {zero} pf = refl
 injSucFin {ℕsuc (ℕsuc n)} pf = cong predFin pf
 
+injSucFinP : ∀ {n0 n1 : ℕ} {n : n0 ≡ n1} {p0 : Fin n0} {p1 : Fin n1}
```
felixwellen, line 84: Could you rename the path? E.g. ```{n m : ℕ} {p : n ≡ m}``` - I guess I'm not alone to think natural number, when I read ```n```...

felixwellen (changes requested):
Thanks for the contribution and I'm sorry for the long wait. It looks all good to me, I just have a small renaming request.

Cubical/Data/FinData/Properties.agda
anuyts: Well it is an `I`-indexed natural number :) But I'll rename.

felixwellen: Thanks!

felixwellen (approved):
Thans for the contribution!

PR #874, opened 2022-07-29
Naive injectivity implies isEmbedding even if only B is a set

MatthiasHu:
This PR removes an unnecessary assumption from `injEmbedding` in `Cubical.Functions.Embedding`. The naive definition of injectivity, `∀{w x} → f w ≡ f x → w ≡ x`, is already sufficient for `f` to be an embedding if `B` is a set, we don't need that `A` is a set.

I also deleted `retractableIntoSet→isEmbedding`, because it shows the same thing under the stronger assumption that `f` has a retract.

### DIFF
--- a/Cubical/Functions/Embedding.agda
+++ b/Cubical/Functions/Embedding.agda
@@ -150,14 +150,14 @@ isEmbedding≡hasPropFibers
 -- implies isEmbedding as long as B is an h-set.
 module _
   {f : A → B}
-  (iSB : isSet B)
+  (isSetB : isSet B)
   (inj : ∀{w x} → f w ≡ f x → w ≡ x)
   where
 
   injective→hasPropFibers : hasPropFibers f
   injective→hasPropFibers y (x , fx≡y) (x' , fx'≡y) =
     Σ≡Prop
-      (λ _ → iSB _ _)
+      (λ _ → isSetB _ _)
       (inj (fx≡y ∙ sym (fx'≡y)))
 
   injEmbedding : isEmbedding f
--- a/Cubical/Functions/Embedding.agda
+++ b/Cubical/Functions/Embedding.agda
@@ -55,24 +55,7 @@ isEmbedding→Inj
 isEmbedding→Inj {f = f} embb w x p
   = equiv-proof (embb w x) p .fst .fst
 
--- If A and B are h-sets, then injective functions between
--- them are embeddings.
---
--- Note: It doesn't appear to be possible to omit either of
--- the `isSet` hypotheses.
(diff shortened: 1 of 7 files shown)


MatthiasHu:
Wait a second, every function that admits a retract is an embedding, right? So `retractableIntoSet→isEmbedding` had an unnecessary `isSet` assumption too...

MatthiasHu:
No, I was wrong: for example, `Unit -> S^1` admits a retract but is not an embedding.

Cubical/Functions/Embedding.agda
```diff
@@ -163,6 +146,23 @@ isEmbedding≡hasPropFibers
            (λ _ → hasPropFibersIsProp _ _)
            (λ _ → isPropIsEmbedding _ _))
 
+-- We use the characterization as hasPropFibers to show that naive injectivity
+-- implies isEmbedding as long as B is an h-set.
+module _
+  {f : A → B}
+  (iSB : isSet B)
```
felixwellen, line 153: A more descriptive name like "isSetB" could help to make the proof below more readable.

Cubical/Functions/Embedding.agda
```diff
@@ -163,6 +146,23 @@ isEmbedding≡hasPropFibers
...
+
+  injective→hasPropFibers : hasPropFibers f
+  injective→hasPropFibers y (x , fx≡y) (x' , fx'≡y) =
+    Σ≡Prop
+      (λ _ → iSB _ _)
+      (inj (fx≡y ∙ sym (fx'≡y)))
+
+  injEmbedding : isEmbedding f
+  injEmbedding = hasPropFibers→isEmbedding injective→hasPropFibers
+
```
felixwellen, line 165: Mayb the stuff you deleted below, could still be a corollary here? I mean we could keep ```retractableIntoSet→isEmbedding``` but prove it using ```injEmbedding``` (and ```congRetract``` in a ```where```-block).

felixwellen (changes requested):
Again a nice simplification!
(Only some minor comments)

Cubical/Functions/Embedding.agda
MatthiasHu: Agreed.

Cubical/Functions/Embedding.agda
MatthiasHu: I will look into it.

MatthiasHu: Ok, I did it as you suggest, seems reasonable to me.

felixwellen (approved):

PR #875, opened 2022-07-29
Simplify discreteSetQuotients

MatthiasHu:
This PR removes two unnecessary assumptions from `discreteSetQuotients` in `Cubical.HITs.SetQuotients.Properties`. It now simply shows that the quotient `A / R` is discrete if `R` is an equivalence relation (not necessarily prop-valued) and `R a₀ a₁` is decidable.


felixwellen (approved):
Nice simplification!

PR #879, opened 2022-08-02
Cohomology Ring of S2 \/ S4

thomas-lamiaux:
PR to compute the cohomology ring of S2 \/ S4 which should be Z[X,Y]/<XY, X², Y²>

For CP2 / S2 \/ S4, this should give an example where the groups are the same but the ring are different.
This is example can be seen as better than Torus / S2 \/ S1 \/ S1 as : 
-  H*(T^2) is complicated and not yet formalized
- I don't know any nice form of H*(S2 \/ S1 \/ S1) 
- H*(CP2) and H*(S2 \/ S4) are nice


thomas-lamiaux:
@aljungstrom @mortberg On this proof, the function gets more complicated, as there is 2 variables, there is [3 trivial cases](https://github.com/agda/cubical/pull/879/files#diff-55f172e9e9b317a17b9e9b3d90476fcea449b07dbd2e9ea824266ee27485ac27R220) 
This annoying because, the [proof of the cup-product](https://github.com/agda/cubical/pull/879/files#diff-55f172e9e9b317a17b9e9b3d90476fcea449b07dbd2e9ea824266ee27485ac27R253) become hard to follow. 
Indeed, most cases are pointless, but we have to open the term (rec) for stuff to compute, to 0 \equiv 0.

Maybe add a partition on the direct sense ? But then it would be complicated to prove retraction. 
So there is probably nothing more to do without tactics, except proving had hoc lemma which wouldn't really change much

thomas-lamiaux:
Hi @mortberg and @aljungstrom, I have spend two days computing `S2 \/ S4` because it is a simpler example. 
So the ring is `Z[X,Y]/<XY, X², Y²>.`

Though I have a very weird issue. 
When I am working, I import the equivalence and [introduce notation ](https://github.com/agda/cubical/pull/879/files#diff-55f172e9e9b317a17b9e9b3d90476fcea449b07dbd2e9ea824266ee27485ac27R153)
`e₀ = invGroupIso H⁰-S²⋁S⁴≅ℤ, ϕ₀, ϕ₀str, ϕ₀⁻¹, ϕ₀⁻¹str, ϕ₀-sect, ϕ₀-retr`
so it is easier to work. 

When proving retraction, I have to prove `ϕ₄⁻¹ (ϕ₄ a) = a`, so I just want to call `ϕ₄-retr`. 
But weirdly he never manage to refine. Even weirder, it works for `ϕ₀`, and `ϕ₂`, and for `ϕ₄-sect` !
So I have checked, the RAM doesn't get saturated or anything, so it is type checking. 
So I have tried the following 
```
Foo : a -> ϕ₄⁻¹ (ϕ₄ a) = a, 
Foo = { leftInv }
```
And this won't refine !
So I have tried [adding the type explicitly](https://github.com/agda/cubical/pull/879/files#diff-55f172e9e9b317a17b9e9b3d90476fcea449b07dbd2e9ea824266ee27485ac27R174) which I wasn't doing before . 
Doesn't change a things, it looks like it is unfolding the phi function and retyping it to see if indeed they are of the form `A -> B // B -> A`. 
But even so, why would it work for the phi 0 and 2, the definition are basically the same. 
Maybe the recursive call increase the size but the RAM doesn't move at all so the term shouldn't be that big, et from the definition, I would have said it would increase by not much more than x4. So I shouldn't see the difference. 
So I don't get it

At the end, I finished by putting everything in a [ad hoc module](https://github.com/agda/cubical/pull/879/files#diff-55f172e9e9b317a17b9e9b3d90476fcea449b07dbd2e9ea824266ee27485ac27R169) and just pass it as a [parameter ](https://github.com/agda/cubical/pull/879/files#diff-55f172e9e9b317a17b9e9b3d90476fcea449b07dbd2e9ea824266ee27485ac27R509) as it seems that the issue is in leftInv. 
Work perfectly, but why ???

I mean, the file work this way, so it can be kept this way but this annoying as you loses the computation of phi_4 when doing that, hopefully I don't need them here but still

Cubical/ZCohomology/Groups/Sn.agda
```diff
@@ -306,23 +306,29 @@ Hⁿ-Sᵐ≅0 (suc n) (suc m) pf = suspensionAx-Sn n m
 
 -------------- A nice packaging for the Hⁿ-Sⁿ  ----------------
 
-code : (m n : ℕ) → Type ℓ-zero
-code zero     zero   = GroupIso (coHomGr 0 (S₊ 0)) (DirProd ℤGroup ℤGroup)
-code zero    (suc n) = GroupIso (coHomGr 0 (S₊ (suc n))) ℤGroup
-code (suc m)  zero   = GroupIso (coHomGr (suc m) (S₊ 0)) UnitGroup₀
-code (suc m) (suc n) with (discreteℕ m n)
+-- code : (m n : ℕ) → Type ℓ-zero
```
mortberg, line 309: Delete commented code?

Cubical/ZCohomology/CohomologyRings/S2wedgeS4.agda
```diff
@@ -0,0 +1,522 @@
...
+  ℤ[X,Y]/<XY,X²,Y²> = PolyCommRing-Quotient ℤCR <XY,X²,Y²>
+
+  ℤ[x,y]/<xy,x²,y²> : Type ℓ-zero
+  ℤ[x,y]/<xy,x²,y²> = fst ℤ[X,Y]/<XY,X²,Y²>
+
+  -- Import with notation
+  open IsGroupHom
+  open gradedRingProperties
+
+  open CommRingStr (snd ℤCR) using ()
```
mortberg, line 82: Didn't we conclude that there was a nicer way to do these big renamings?

Cubical/ZCohomology/CohomologyRings/S2wedgeS4.agda
```diff
@@ -0,0 +1,522 @@
...
+
+  e₂ = invGroupIso H²-S²⋁S⁴≅ℤ
+  ϕ₂ = fun (fst e₂)
+  ϕ₂str = snd e₂
+  ϕ₂⁻¹ = inv (fst e₂)
+  ϕ₂⁻¹str = snd (invGroupIso e₂)
+  ϕ₂-sect = rightInv (fst e₂)
+  ϕ₂-retr = leftInv (fst e₂)
+
+  module WhyWontYouWork
```
mortberg, line 169: This module name isn't great

mortberg: Alternatively document it

Cubical/ZCohomology/CohomologyRings/RP2.agda
```diff
@@ -0,0 +1,506 @@
...
+  ℤ[X]/<2X,X²> : CommRing ℓ-zero
+  ℤ[X]/<2X,X²> = PolyCommRing-Quotient ℤCR <2X,X²>
+
+  ℤ[x]/<2x,x²> : Type ℓ-zero
+  ℤ[x]/<2x,x²> = fst ℤ[X]/<2X,X²>
+
+  open IsGroupHom
+  open gradedRingProperties
+
+  open GroupStr (snd BoolGroup) using ()
```
mortberg, line 80: Big imports and renamings

Cubical/ZCohomology/CohomologyRings/RP2.agda
```diff
@@ -0,0 +1,506 @@
...
+  ℤ[x]→H*-RP²-pres+ : (x y : ℤ[x]) → ℤ[x]→H*-RP² (x +Pℤ y) ≡ ℤ[x]→H*-RP² x +H* ℤ[x]→H*-RP² y
+  ℤ[x]→H*-RP²-pres+ x y = refl
+
+  ϕ₀-gen : (n : ℕ) → (f : coHom n RP²) → ϕ₀ (pos 1) ⌣ f ≡ f
+  ϕ₀-gen n f = cong (λ X → X ⌣ f) ϕ₀-pres1 ∙ rUnit⌣ n f
+
+  open pres⌣
+
+  -- Nice packaging of the proof
+  pres·-base-case-int : (n : ℕ) → (a : ℤ) → (m : ℕ) → (b : ℤ) →
```
mortberg, line 265: Long lines in this proof

Cubical/ZCohomology/CohomologyRings/RP2.agda
```diff
@@ -0,0 +1,506 @@
...
+  ℤ[X]/<2X,X²>→H*R-RP² =
+    Quotient-FGideal-CommRing-Ring.inducedHom ℤ[X] (H*R RP²) ℤ[X]→H*-RP² <2X,X²> ℤ[x]→H*-RP²-cancelX
+
+  ℤ[x]/<2x,x²>→H*-RP² : ℤ[x]/<2x,x²> → H* RP²
+  ℤ[x]/<2x,x²>→H*-RP² = fst ℤ[X]/<2X,X²>→H*R-RP²
+
+  ℤ[x]/<2x,x²>→H*-RP²-pres0 : ℤ[x]/<2x,x²>→H*-RP² 0PℤI ≡ 0H*
+  ℤ[x]/<2x,x²>→H*-RP²-pres0 = refl
+
+  ℤ[x]/<2x,x²>→H*-RP²-pres+ : (x y : ℤ[x]/<2x,x²>) → ℤ[x]/<2x,x²>→H*-RP² ( x +PℤI y) ≡ ℤ[x]/<2x,x²>→H*-RP² x +H* ℤ[x]/<2x,x²>→H*-RP² y
```
mortberg, line 323: Long line

Cubical/ZCohomology/CohomologyRings/RP2.agda
```diff
@@ -0,0 +1,506 @@
...
+       base-neutral-eq k (is0 x) = cong [_] (cong (base {AGP = λ _ → snd ℤAG} (0 ∷ []))
+                                                  (cong ϕ₀⁻¹ (substG0 x) ∙ pres1 ϕ₀⁻¹str)
+                                            ∙ (base-neutral _))
+       base-neutral-eq k (is2 x) = cong [_] (cong (base (1 ∷ []))
+                                                  (cong (ψ₂⁻¹ ∘ ϕ₂⁻¹) (substG0 x)
+                                                  ∙ cong ψ₂⁻¹ (pres1 ϕ₂⁻¹str)) -- ψ₂⁻¹ pres1 by refl
+                                            ∙ base-neutral _)
+       base-neutral-eq k (else x) = refl
+
+       base-add-eq : (k : ℕ) → (a b : coHom k RP²) → (x : partℕ k) →
```
mortberg, line 383: Long lines

Cubical/ZCohomology/CohomologyRings/RP2.agda
```diff
@@ -0,0 +1,506 @@
...
+                                            ∙ cong (base (0 ∷ [])) (sym (pres· ϕ₀⁻¹str _ _) ∙ cong ϕ₀⁻¹ (substG+ _ _ _)))
+       base-add-eq k a b (is2 x) = Λ-pres+ (ϕ₂⁻¹ (substG x a)) (ϕ₂⁻¹ (substG x b))
+                                   ∙ cong [_] (cong (λ X → base {AGP = λ _ → snd ℤAG} (1 ∷ []) (ψ₂⁻¹ X))
+                                                    ((sym (pres· ϕ₂⁻¹str _ _) ∙ cong ϕ₂⁻¹ (substG+ _ _ _))))
+       base-add-eq k a b (else x) = +PℤIIdR _
+
+  H*-RP²→ℤ[x]/<2x,x²>-pres0 : H*-RP²→ℤ[x]/<2x,x²> 0H* ≡ 0PℤI
+  H*-RP²→ℤ[x]/<2x,x²>-pres0 = refl
+
+  H*-RP²→ℤ[x]/<2x,x²>-pres+ : (x y : H* RP²) → H*-RP²→ℤ[x]/<2x,x²> (x +H* y) ≡ (H*-RP²→ℤ[x]/<2x,x²> x) +PℤI (H*-RP²→ℤ[x]/<2x,x²> y)
```
mortberg, line 395: Long line

Cubical/ZCohomology/CohomologyRings/RP2.agda
```diff
@@ -0,0 +1,506 @@
...
+                            ∙ +PℤIdR _
+                            ∙ cong (base (1 ∷ [])) (sym (-DistL· _ _)))
+
+
+  e-retr-ψ₂ : (a : ℤ) → ((isEven a ≡ false) ⊎ (isEven a ≡ true)) → Λ (ψ₂ a) ≡ [ base (1 ∷ []) a ]
+  e-retr-ψ₂ a (inl x) = e-retr-ψ₂-false a x
+  e-retr-ψ₂ a (inr x) = e-retr-ψ₂-true a x
+
+
+  e-retr-base : (k :  ℕ) → (a : ℤ) → H*-RP²→ℤ[x]/<2x,x²> (ℤ[x]/<2x,x²>→H*-RP² [ base (k ∷ []) a ]) ≡ [ base (k ∷ []) a ]
```
mortberg, line 465: Long line

mortberg (changes requested):
A few quick comments

Cubical/ZCohomology/CohomologyRings/S2wedgeS4.agda
thomas-lamiaux: I don't remember, overall it is a bit long but I think it is clearer when reading the file

Cubical/ZCohomology/CohomologyRings/S2wedgeS4.agda
thomas-lamiaux: Well I was hoping to find a solution, but couldn't.
I changed the name and added a comment

mortberg (approved):

mortberg:
I fixed the conflicts. Will merge once the CI is finished

PR #880, opened 2022-08-05
Fin (suc n) ≡ Maybe (Fin n)

MatthiasHu:
This PR proves a small property of the `Fin` datatype in `Cubical.Data.FinData`:

```agda
finSuc≡Maybe : Fin (ℕ.suc n) ≡ Maybe (Fin n)
```

and also a pointed variant of that.


felixwellen (approved):
Looks good.

PR #881, opened 2022-08-05
Freely pointed types

MatthiasHu:
This PR exhibits the existing function `Maybe∙` as the "free pointing" operation, in the sense that there is an isomorphism:

```agda
freelyPointedIso : Iso (Maybe∙ A →∙ B) (A → ⟨ B ⟩)
```


felixwellen (approved):
Good!

PR #884, opened 2022-08-08
An Improvement for Cartesian Kan Operations

kangrongji:
Improve the definition of `coei→i` in ` Cubical.Foundations.CartesianKanOps`, so that it is definitional equality at the ends now. This is useful in some cases. Also an equality `coePath` between `coe` and the diagonal.


ecavallo (approved):
Nice, thanks!

PR #891, opened 2022-08-16
Added Unordered Pair

guilhermehas:
I created Unordered Pair, which means that `a , b = b , a`.


felixwellen:
Did you see the finite multisets: https://github.com/agda/cubical/blob/master/Cubical/HITs/FiniteMultiset/Base.agda ?
Unordered pairs are a special case and I think it would be better to define them as such, if you really have a use case specifically for pairs.

guilhermehas:
Yes, I know the relation. I will try to put in the properties. 
For example, in Agda Stdlib (https://github.com/agda/agda-stdlib/blob/master/src/Data/Vec/Recursive.agda), they defined Vec using Product. So maybe it is possible to define a FiniteMultiset using this Unordered Pair with the same idea.

Cubical/HITs/UnorderedProd/Base.agda
```diff
@@ -0,0 +1,41 @@
...
+  _,_ : (x y : A) → UnorderedProd A
+  swap : (x y : A) → (x , y) ≡ (y , x)
+  trunc : isSet (UnorderedProd A)
+
+module Elim {B : UnorderedProd A → Type ℓ′}
+  (_,*_ : (x y : A) → B (x , y))
+  (swap* : (x y : A) → PathP (λ i → B (swap x y i)) (x ,* y) (y ,* x))
+  (trunc* : (xs : UnorderedProd A) → isSet (B xs)) where
+
+  f : (xs : UnorderedProd A) → B xs
```
felixwellen, line 23: The module should be anonymous and 'f' should be called 'elim'.
Unfortunately, it is not better in FiniteMultiSets and ListedMultiSets...

Cubical/HITs/UnorderedProd/Base.agda
```diff
@@ -0,0 +1,41 @@
...
+  f (x , y) = x ,* y
+  f (swap x y i) = swap* x y i
+  f (trunc x y p q i j) =
+    isOfHLevel→isOfHLevelDep 2 trunc* (f x) (f y) (cong f p) (cong f q) (trunc x y p q) i j
+
+module ElimProp {B : UnorderedProd A → Type ℓ′}
+  (BProp : {xs : UnorderedProd A} → isProp (B xs))
+  (_,*_ : (x y : A) → B (x , y)) where
+
+  f : (xs : UnorderedProd A) → B xs
```
felixwellen, line 33: Same here.

Cubical/HITs/UnorderedProd/Base.agda
```diff
@@ -0,0 +1,41 @@
...
+  f = Elim.f _,*_ (λ x y → toPathP (BProp (transport (λ i → B (swap x y i)) (x ,* y)) (y ,* x)))
+    λ _ → isProp→isSet BProp
+
+module Rec {B : Type ℓ′} (BType : isSet B)
+  (_,*_ : (x y : A) → B) (swap* : (x y : A) → (x ,* y) ≡ (y ,* x)) where
+
+  f : UnorderedProd A → B
```
felixwellen, line 40: ... and here.

felixwellen (changes requested):
Only a renaming request...

felixwellen:
> Yes, I know the relation. I will try to put in the properties. For example, in Agda Stdlib (https://github.com/agda/agda-stdlib/blob/master/src/Data/Vec/Recursive.agda), they defined Vec using Product. So maybe it is possible to define a FiniteMultiset using this Unordered Pair with the same idea.

Maybe you can add a comment? Something like "the relation to ... should be ...".

Cubical/HITs/UnorderedProd/Base.agda
guilhermehas: I have already renamed it.

Cubical/HITs/UnorderedProd/Base.agda
guilhermehas: I have already renamed it.

Cubical/HITs/UnorderedProd/Base.agda
guilhermehas: I have already renamed it.

guilhermehas:
I was trying to define the recursive definition:
```agda
FMSet : (A : Type ℓ) (n : ℕ) → Type ℓ
FMSet A 0 = Unit*
FMSet A 1 = A
FMSet A (suc (suc n)) = {!!}
```

And I figured out that it is not possible to do, because both elements need to be of the same type.

> > Yes, I know the relation. I will try to put in the properties. For example, in Agda Stdlib (https://github.com/agda/agda-stdlib/blob/master/src/Data/Vec/Recursive.agda), they defined Vec using Product. So maybe it is possible to define a FiniteMultiset using this Unordered Pair with the same idea.
> 
> Maybe you can add a comment? Something like "the relation to ... should be ...".

Because of that, what I commented before was wrong.

guilhermehas:
I figured out how to do:
```agda
addN : UnorderedPair (A × ℕ) → ℕ
addN = rec isSetℕ (λ x y → proj₂ x + proj₂ y) λ x y → +-comm (proj₂ x) (proj₂ y)

data FMSet (A : Type ℓ) : ℕ → Type ℓ where
  one : A → FMSet A n
  join : (xs : UnorderedPair (A × ℕ)) → FMSet A (addN xs)
```

plt-amy:
Since this construction is limited to the sets, wouldn't it be cleaner to implement it by a set-coequalizer? Consider the coequalizer of `swap, id : A × A -> A × A`, for example

guilhermehas:
> Since this construction is limited to the sets, wouldn't it be cleaner to implement it by a set-coequalizer? Consider the coequalizer of `swap, id : A × A -> A × A`, for example

Thanks for your suggestion. I proved that both are isomorphic.

felixwellen:
As I just learned in the hallway, there is also an untruncated unordered pair notion, defined as maps from a set merely equivalent to Fin 2 to A. I guess for A a Set the Set-truncation of that notion should agree with your notion?

guilhermehas:
> As I just learned in the hallway, there is also an untruncated unordered pair notion, defined as maps from a set merely equivalent to Fin 2 to A. I guess for A a Set the Set-truncation of that notion should agree with your notion?

I didn't understand. Can you give a code example?

plt-amy:
Let `A` be a type (of arbitrary h-level). We define the type of unordered pairs with coordinates in `A` as the type
```agda
Σ[ X ∈ Type ] (∥ X ≃ Bool ∥ × X → A)
```
This construction is _at least_ a groupoid, but for A of h-level ≥ 3, it preserves h-levels. The idea is that, in the same way that a pair in A is the same as a function `Bool→A`, we can take away the ordering by allowing pairs to choose their own index types up to a mere equivalence; consumers of unordered pairs can't inspect the underlying type, and all they know is that it has two elements.

The pairing `{x,y}` is defined to be
```agda
Bool , inc id , λ { true → x; false → y }
```
This is equal to the pairing `{y,x}` because the type of identifications between unordered pairs is given by equivalences of their underlying types which commute with the projection function.

guilhermehas:
> ```agda
> Bool , inc id , λ { true → x; false → y }
> ```
> 
> This is equal to the pairing `{y,x}` because the type of identifications between unordered pairs is given by equivalences of their underlying types which commute with the projection function.

I am trying to do your implementation, but I am in trouble. What am I doing wrong?
```agda
UPairBool : Type ℓ → Type _
UPairBool A = Σ[ X ∈ Type ] (∥ X ≃ Bool ∥₁ × (X → A))

private
  uP₁ : UPairBool Bool
  uP₁ = Bool , ∣ idEquiv Bool ∣₁ , λ where false → false ; true → true

  uP₂ : UPairBool Bool
  uP₂ = Bool , ∣ notEquiv ∣₁ , λ where false → true ; true → false

  uP₁≡uP₂ : uP₁ ≡ uP₂
  uP₁≡uP₂ i = Bool , squash₁ ∣ idEquiv Bool ∣₁ ∣ notEquiv ∣₁ i  , {!!}
```

felixwellen:
> I am trying to do your implementation, but I am in trouble. What am I doing wrong?

What you are trying to prove is true. The first part of your term ```uP₁≡uP₂``` should not be constantly ```Bool```, but come from the ```not``` equivalence.

plt-amy:
Also, isn't that the set-truncation? The equivalence has to be _propositionally_-truncated.

felixwellen:
```∥_∥₁``` is propositional truncation.

**Comment by guilhermehas** [comment]
I am in trouble trying to prove this thing:

```agda
uP₁≡uP₂ : uP₁ ≡ uP₂
uP₁≡uP₂ = ΣPathP (notEq , ΣPathP (helper₁ , helper₂)) where
  helper₁ : PathP (λ z → ∥ notEq z ≃ Bool ∥₁) ∣ idEquiv Bool ∣₁ ∣ notEquiv ∣₁
  helper₁ i = {!squash₁ ∣ idEquiv Bool ∣₁ ∣ notEquiv ∣₁ i!}

  helper₂ : PathP (λ i → notEq i → Bool) (idfun Bool) not
  helper₂ i b = {!!}
```

felixwellen:
You can, of course, also postpone the proof and wrap up the current PR...

Cubical/HITs/FiniteMultiset/Recursive.agda
```diff
@@ -0,0 +1,21 @@
...
+open import Cubical.HITs.UnorderedPair
+
+private variable
+  ℓ : Level
+  A : Type ℓ
+
+pattern ,_ x = _,,_ _ x
+
++-commPair : UnorderedPair (A × ℕ) → ℕ
++-commPair = rec isSetℕ (λ (, x) (, y) → x + y) λ (, x) (, y) → +-comm x y
```
felixwellen, line 17: I think this would be easier to read if you use underscore and the usual notation for pairs, i.e. ```_ , x```.

felixwellen (changes requested):
I have a couple of comments and questions.

Cubical/Data/Sigma/Base.agda
```diff
@@ -25,6 +25,8 @@ A × B = Σ A (λ _ → B)
 
 infixr 5 _×_
 
+swap : ∀ {ℓ ℓ'} {A : Type ℓ} {B : Type ℓ'} → A × B → B × A
```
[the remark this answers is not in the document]
guilhermehas: I removed it.

Cubical/Data/Prod/Properties.agda
```diff
@@ -4,7 +4,7 @@ module Cubical.Data.Prod.Properties where
 open import Cubical.Core.Everything
 
 open import Cubical.Data.Prod.Base
-open import Cubical.Data.Sigma renaming (_×_ to _×Σ_) hiding (prodIso ; toProdIso ; curryIso)
+open import Cubical.Data.Sigma renaming (_×_ to _×Σ_) hiding (prodIso ; toProdIso ; curryIso; swap)
```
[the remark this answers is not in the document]
guilhermehas: I removed it.

Cubical/HITs/FiniteMultiset/Recursive.agda
```diff
@@ -0,0 +1,21 @@
+{-# OPTIONS --safe #-}
+module Cubical.HITs.FiniteMultiset.Recursive where
+
+open import Cubical.Foundations.Prelude
+
+open import Cubical.Data.Sigma using (_×_) renaming (_,_ to _,,_)
```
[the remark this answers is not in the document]
guilhermehas: I changed to the usual notation.

Cubical/HITs/FiniteMultiset/Recursive.agda
guilhermehas: I changed to the usual notation.

felixwellen:
Sorry for the long wait and thanks for the updates. I am a bit puzzled about the recursive FMSets - maybe it is better to save those for a PR where you prove something about them? All the other files look good to me.

guilhermehas:
I deleted the recursive file.

felixwellen (approved):
Thanks - will merge it now...

PR #892, opened 2022-08-16
List/permutation ≃ FreeComMonoid

marcinjangrzybowski:
- basic properties of permutations (defined as FinData.Fin n ≃ FinData.Fin m)
- _↔_ - equivalence relation of "being a permutation of" on List
- List/↔ ≃ FreeComMonoid (via isomorphism of List/↔ and FiniteMultiset)


marcinjangrzybowski:
- basic properties of permutations (defined as FinData.Fin n ≃ FinData.Fin m)
- _↔_ - equivalence relation of "being a permutation of" on List
- List/↔ ≃ FreeComMonoid (via isomorphism of List/↔ and FiniteMultiset)

Cubical/Data/Sum/Properties.agda
```diff
@@ -188,3 +189,8 @@ leftInv Σ⊎Iso (inr b , eb) = refl
 map-⊎ : (A → C) → (B → D) → A ⊎ B → C ⊎ D
 map-⊎ f _ (inl a) = inl (f a)
 map-⊎ _ g (inr b) = inr (g b)
+
+recMap : ∀ {ℓe} → {E : Type ℓe}
```
felixwellen, line 193: Maybe ```recMap≡``` or ```recMapComp≡```?

Cubical/Data/List/FinData.agda
```diff
@@ -34,12 +64,269 @@ lookup-tabulate : ∀ n → (^a : Fin n → A)
...
+               lookup (y ∷ x ∷ l) (swap0and1 k)
+lookup-swap x y l zero = refl
+lookup-swap x y l one = refl
+lookup-swap x y l (suc (suc k)) = refl
+
+
+
+infix 4  _↔_
+
+record _↔_ {ℓ} {A : Type ℓ} (x y : List A) : Type ℓ where
```
felixwellen, line 102: Just an idea: ```l l'``` instead of ```x y```.

Cubical/Data/List/FinData.agda
```diff
@@ -34,12 +64,269 @@ lookup-tabulate : ∀ n → (^a : Fin n → A)
...
+          ∀ k → lookup (xs ++ ys) k ≡
+           ⊎.rec (lookup xs) (lookup ys)
+             (FinSumChar.inv (length xs) (length ys)
+               (subst Fin (length++ xs ys) k))
+  lookup-FinSumChar {xs = []} {ys} _ = cong (lookup ys) (sym (transportRefl _))
+  lookup-FinSumChar {xs = x ∷ xs} {ys} zero =
+     cong (⊎.rec (lookup (x ∷ xs)) (lookup ys) ∘ (FinSumChar.inv _ _))
+       (transportFin-zero _)
+  lookup-FinSumChar {xs = x ∷ xs} {ys} (suc _) =
+     _ ≡⟨ lookup-FinSumChar {xs = xs} _ ⟩
```
felixwellen, line 205: ```≡⟨ ... ⟩``` is supposed to be used, if you want to write down the endpoints of the equalities. If you don't want to do that, just use the concatenation operator ```\.```. But I think here it would actually help to fill in those ```_```

Cubical/Data/List/FinData.agda
```diff
@@ -34,12 +64,269 @@ lookup-tabulate : ∀ n → (^a : Fin n → A)
...
+      h (inl _) = refl
+      h (inr _) = refl
+
+  cong↔++R : {xs ys : List A} → xs ↔ ys → ∀ l → xs ++ l ↔ ys ++ l
+  cong↔++R {xs = xs} {ys} (prm e p) _ =
+   let h = ⊎-equiv e (idEquiv _)
+   in prm (≡→Fin≃ _ ∙ₑ invEquiv (FinSumChar.Equiv _ _) ∙ₑ h ∙ₑ FinSumChar.Equiv _ _  ∙ₑ ≡→Fin≃ _)
+    λ _ →
+     let k' = FinSumChar.inv _ _ _
+     in _ ≡⟨ lookup-FinSumChar {xs = xs} _ ⟩
```
felixwellen, line 221: Same here

Cubical/Data/List/FinData.agda
```diff
@@ -34,12 +64,269 @@ lookup-tabulate : ∀ n → (^a : Fin n → A)
...
-  → lookup (map f xs) p0 ≡ f (lookup xs p1)
-lookup-map f (x ∷ xs) zero zero p = refl
-lookup-map f (x ∷ xs) zero (suc p1) p = ⊥.rec (znotsP p)
-lookup-map f (x ∷ xs) (suc p0) zero p = ⊥.rec (snotzP p)
-lookup-map f (x ∷ xs) (suc p0) (suc p1) p = lookup-map f xs p0 p1 (injSucFinP p)
+lookup-tabulateT : ∀ n → (^a : Fin n → A) → ∀ k
+    → lookup (tabulate n ^a) (subst⁻ Fin (length-tabulate n ^a) k) ≡ ^a k
+lookup-tabulateT n ^a k = (sym (transportRefl _)) ∙ funExt⁻ (fromPathP (lookup-tabulate n ^a)) k
+
+permute : ∀ {n} (l : List A) → Fin n ≃ Fin (length l) → List A
```
felixwellen, line 71: I guess it also makes sense here, to have a separate file about permutations.

Cubical/Data/List/FinData.agda
```diff
@@ -34,12 +64,269 @@ lookup-tabulate : ∀ n → (^a : Fin n → A)
...
+
+↔permute : ∀ {n} (l : List A) → (e : Fin n ≃ Fin (length l))
+                → l ↔ (permute l e)
+F≃ (↔permute l e) = invEquiv e ∙ₑ ≡→Fin≃ (sym (length-permute l e))
+l≡ (↔permute l e) k =
+        cong (lookup l) (sym (secEq e k))
+      ∙ sym (lookup-tabulateT _ (lookup l ∘ (fst e)) (invEq e k))
+
+
+isRefl↔ : BinaryRelation.isRefl (_↔_ {A = A})
```
felixwellen, line 118: Using relation-terminology is a bit confusing, since the type you are talking about should actually be a groupoid?
Did you think about that and figured it is still better this way? Then write down your conclusion in a comment.

Cubical/Data/List/FinData.agda
```diff
@@ -34,12 +64,269 @@ lookup-tabulate : ∀ n → (^a : Fin n → A)
...
+
+
+PunchHeadInOut/≡⁻ : ∀ (x : A) (l : List A) → ∀ k
+      → List→FMSet (x ∷ l) ≡
+         List→FMSet (permute (x ∷ l) (invEquiv (PunchHeadInOut≃ k)))
+PunchHeadInOut/≡⁻ x l zero = cong (List→FMSet ∘ (x ∷_))  (sym (tabulate-lookup l))
+PunchHeadInOut/≡⁻ x (x₁ ∷ l) one i =
+  comm x x₁ (List→FMSet ((tabulate-lookup l) (~ i))) i
+PunchHeadInOut/≡⁻ x (x₁ ∷ l) (suc (suc k)) =
+  _ ≡⟨ comm _ _ _ ⟩
```
felixwellen, line 268: Use concatenation operator or fill in terms.

Cubical/Data/List/FinData.agda
```diff
@@ -34,12 +64,269 @@ lookup-tabulate : ∀ n → (^a : Fin n → A)
...
+↔→FMSet≡ [] (_ ∷ _) = ⊥.rec ∘ ¬nil↔cons
+↔→FMSet≡ (_ ∷ _) [] = ⊥.rec ∘ ¬cons↔nil
+↔→FMSet≡ {A = A} (x ∷ xs) (y ∷ ys) r@(prm e _) =
+   let (e' , p') = unwindPermHead e
+       k' = equivFun e zero
+       xs' = permute xs (invEquiv e')
+       (prm _ p⁻) = isSym↔ _ _ r
+       pL = sym (length-tabulate _ (lookup (x ∷ xs) ∘ invEq (sucPerm e')))
+   in
+      _ ≡⟨ cong (x ∷fm_) (↔→FMSet≡ xs xs' (↔permute xs (invEquiv e'))) ⟩
```
felixwellen, line 286: Use concatenation operator or fill in terms.

Cubical/Data/List/FinData.agda
```diff
@@ -34,12 +64,269 @@ lookup-tabulate : ∀ n → (^a : Fin n → A)
...
+                    (PunchHeadInOut≃∙ₑ≡→Fin≃ pL k')) ⟩
+           _ ≡⟨ tabulate∘ (lookup (x ∷ xs) ∘ invEq (sucPerm e'))
+                           (invEq (PunchHeadInOut≃ k'))  ⟩
+           _ ≡⟨ cong (permute (x ∷ xs) ∘ invEquiv) (sym p') ⟩
+           _ ≡⟨ cong (tabulate _) (sym (funExt p⁻)) ⟩
+           _ ≡⟨ tabulate-lookup _ ⟩ (y ∷ ys) ∎
+          ) ⟩ _ ∎
+
+
+IsoList/↔FMSet : Iso (List/↔ A) (FMSet A)
```
felixwellen, line 299: Please add a comment, for example in Base.agda of FMSet, that this Iso is here.

felixwellen (changes requested):
Thanks for the nice contribution!
I only have some comments concerning code organization, naming and similar.

Cubical/Data/List/FinData.agda
marcinjangrzybowski: Are you referring to the fact that  _↔_ is not propositional but of hLevel >= 2 ? 
(set from permutation, and possibly more from path , depending on hLevel of A) 
For Sets and Grupoids, this gives hLevel of 2.

I saw that the BinaryRelation definition does not require a relation to be propositional (it even defines the property of relation being set-valued `isSetValued`).

I used it only to make signatures more readable. Also, it let me bundle all laws into BinaryRelation.isEquivRel.

currently, I am working on groupoid truncation using this permutation relation (but proving hexagon identity is more challenging)

felixwellen: Yes, that is what I was referring to. 
Just write a small comment "_↔_ is not expected to be a proposition, using Relation terminology for readability and bundling" or something like that.

felixwellen: > currently, I am working on groupoid truncation using this permutation relation (but proving hexagon identity is more challenging)

Not sure I understand. Sounds interesting - if you like to discuss it, just make a draft PR and I'll have a look.

mzeuner:
This looks cool and I'm also interested to see how this relates to groupoid truncation.

Did you see that we have another relation on lists that gives you finite multisets in the quotient: https://github.com/agda/cubical/blob/master/Cubical/Relation/ZigZag/Applications/MultiSet.agda

This relation arises automatically from the multiset structure on lists and the functional relation to association lists, but when you unfold it essentially relates two lists that contain the same elements with the same multiplicity. Do you think it would be hard to prove that the two relations are logically equivalent?

Cubical/Data/List/FinData.agda
mzeuner: You can always take the propositional truncation of your relation and make it prop-valued, but when you take the set-quotient you end up with the same thing either way. 

That's a little lemma called [truncRelIso](https://github.com/agda/cubical/blob/master/Cubical/HITs/SetQuotients/Properties.agda#L282) and I always took it as justification for quotienting by non-prop-valued relations. 

As long as your relation is an equivalence relation in the sense above, it gives you a nice characterization of the quotients path type: [isEquivRel→TruncIso](https://github.com/agda/cubical/blob/master/Cubical/HITs/SetQuotients/Properties.agda#L295)

felixwellen: Ah, I thought that's true, but I didn't know about the lemma. I guess we use the terminology somewhere else in the same way and withdraw my request for a comment.

marcinjangrzybowski:
> Do you think it would be hard to prove that the two relations are logically equivalent?

I do not expect it to be hard since the other relation brings `Discreate A` to the table, 
with this `_↔_` relations become decidable. 

I will include, proof, thanks for the idea!

marcinjangrzybowski:
I am in process of cleaning up and spliting into smaller pieces version of this PR for both sets and groupoids, (I gave overview of it to @felixwellen during last AIM) , it is mainly making use of machinery for permutations, PR #1065 is first in a series which will hopefully bring me there.

PR #895, opened 2022-08-16
More on images

felixwellen:
This PR shows, that a map is an embedding iff the restriction to its image is an equivalence.
Also adds a bundled version of the composition of embeddings (not used, but that's what I would have found sooner).

### DIFF
--- a/Cubical/Functions/Image.agda
+++ b/Cubical/Functions/Image.agda
@@ -46,7 +46,8 @@ module _ (f : A → B) where
              → ∣ x , Σ≡Prop isPropIsInImage fx≡y ∣₁)
            y∈im
 
-
+  imageFactorization : fst imageInclusion ∘ restrictToImage ≡ f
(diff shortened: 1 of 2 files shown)


Cubical/Functions/Image.agda
```diff
@@ -73,3 +74,9 @@ module _ (f : A ↪ B) where
                  (equivToIso (restrictionHasSameFibers y))
                  (isEmbedding→hasPropFibers (snd f) (fst y)) ) ,
       (isSurjectionImageRestriction (fst f)))
+
+{-
+  This is the extension to a 'iff', which is also a general modal fact.
```
mortberg, line 79:
```suggestion
  This is the extension to an 'iff', which is also a general modal fact.
```

mortberg (approved):
Looks good except for grammar in a comment. I'll just commit my suggestion and merge

PR #897, opened 2022-08-16
A small lemma on equivalences

felixwellen:
Just one proof, that a retract of an equivalence is itself an equivalence.


felixwellen:
I added the symmetric statement. Because I messed up and that was what I actually needed - but I guess its better to have both anyway.

PR #899, opened 2022-08-16
generalize levels for some AlgebraHom operations

MatthiasHu:
This PR just generalizes the type signatures of operations like `AlgebraHom≡` or `_∘≃a_` to algebras of different levels.

EDIT: Also, use declared variables as much as possible.


felixwellen (changes requested):
Did you try to use generalized variables (for R, A, B, C, D) - that should be a lot simpler?
If yes, then it would be good to have a comment explaining why it doesn't work.

MatthiasHu:
This is better, thank you for pointing it out.

In `Algebra.Base`, it is not so easy to do consistently. Does it look ok the way it is now?

felixwellen:
Yes, looks good to me.

felixwellen (approved):

felixwellen:
Now it just needs to pass ci...

PR #902, opened 2022-08-19
The Internal n-Cubes

kangrongji:
This PR contains a mutual recursive definition of n-cubes, their boundaries and n-cubes with fixed boundary, as family of types parametrized over `ℕ`. They're strictly isomorphic to external ones as, for example, `(i j : I) → A`, `(i j : I) → Partial (i ∨ ~ i ∨ j ∨ ~ j) A` and `(i j : I) → A [ (i ∨ ~ i ∨ j ∨ ~ j) ↦ u i j ]`. Unless one fixes a canonical natural number `n` as dimension, these isomorphisms cannot be described internally, so for more general results one must resort to macros or 2LTT.

My aim is to prove that a type has h-level `n` if and only if n-boundaries can always be filled into n-cubes, generalizing the special cases about `isSet'` and `isGroupoid'` already existing in library.


Cubical/Foundations/Cubes/Base.agda
```diff
@@ -0,0 +1,155 @@
...
+
+makeCube : {n : ℕ}{A : Type ℓ}{a₀ a₁ : Cube n A} → a₀ ≡ a₁ → Cube (suc n) A
+makeCube {n = 0} a₋ = _ , a₋
+makeCube {n = suc n} a₋ = _ , λ i → a₋ i .snd
+
+-- A cube is just a path of cubes of one-lower-dimension.
+-- Unfortunately the following function cannot begin at 0,
+-- because Agda doesn't support pattern matching on ℕ towards pre-types.
+pathCube : (n : ℕ) → (I → Cube (suc n) A) → Cube (suc (suc n)) A
+pathCube n p = _ , λ i → p i .snd
```
ecavallo, line 86: Let's add a reminder here that we can fix this in 2.6.3 when I -> A becomes fibrant.

Cubical/Foundations/Cubes/Base.agda
```diff
@@ -0,0 +1,155 @@
...
+constCube {n = 0} a = _ , λ i → a
+constCube {n = suc n} (∂ , cube) = _ , λ i → cube
+
+retConst : {n : ℕ}{A : Type ℓ} → (cube : Cube n A) → ∂₀ (constCube {n = n} cube) ≡ cube
+retConst {n = 0} _ = refl
+retConst {n = suc n} _ = refl
+
+setConst : {n : ℕ}{A : Type ℓ} → (cube : Cube (suc n) A) → constCube (∂₀ cube) ≡ cube
+setConst {n = 0} (_ , p) i = _ , λ j → p (i ∧ j)
+setConst {n = suc n} (_ , p) i  = _ , λ j → p (i ∧ j)
```
ecavallo, line 123: Is `set` short for section?

Cubical/Foundations/Cubes/Base.agda
kangrongji: done

ecavallo:
Looks good to me!

kangrongji:
Ehhh.... How could I write 2.9.3?

PR #903, opened 2022-08-19
Cubes in Path Type

kangrongji:
This PR contains a few simple cases of cubical subtype, and shows:

- Cubes with a pair of fixed constant opposite faces is equivalent to cubes in the path type;

- Every cubes can be deformed to have (a pair of) fixed constant opposite faces, and this procedure gives an equivalence.


kangrongji:
Update: I've made some simplifications of the arguments and cut about 100 LOC. Hope it would make things clearer and easier to understand.

ecavallo (approved):
LGTM

Issue #833, opened 2022-06-01
Remove Syntax folder


mortberg:
I think it might be good to remove the Syntax folder and revert https://github.com/agda/cubical/pull/595 . The problem being that it causes problems when displaying goals

felixwellen:
Done by PR #907

PR #907, opened 2022-08-22
Revert "Add syntax typeclass for types with carriers."

felixwellen:
This implements #833


mortberg:
Thanks! It's nice that nothing depended on this

PR #912, opened 2022-08-23
Alternative definition of the torus

ncfavier:
An alternative definition of the torus whose surface is `p ∙ q ≡ q ∙ p` rather than `Square p p q q`, with a proof that the two definitions are equivalent.

The forward and inverse maps make use of a single 3-dimensional "tube" that connects `p ∙ q ≡ q ∙ p` to `Square p p q q` (`compPath→Square-faces`), so that the homotopies follow from the fact that `hcomp` behaves like an equivalence (`hcomp-equivFillerSub`).

But this isn't quite enough to make the proof go through; many thanks to @plt-amy and @kangrongji for explaining [what was wrong](https://github.com/agda/agda/issues/6017#issuecomment-1212409372) and [how to fix it](https://github.com/kangrongji/cubical-experiments/blob/main/Torus.agda), respectively.

The core of the issue is that there are two ways to compose two squares along a chosen "first" direction: *pointwise*, where we abstract over the second dimension and compose along the first with `_∙_`; or directly, using `comp` to compose the two `PathP`s in the varying path space `compPath-filler p q i j ≡ compPath-filler p q i j`. These two composites are equal, by uniqueness of composites, because they are both lids for the same cube (see `∙v≡∙v'`); but they don't *compute* to the same thing.

In `from-to (square' i j)`, the left boundary is the dependent function `from-to` applied to the composite `line1' ∙ line2'`, which turns out to be the "direct" composite `(λ j → refl) ∙v' (λ j → refl)`, but the cube given by `hcomp-equivFiller` has `(λ j → refl) ∙v (λ j → refl)` (in other words, `λ j → refl`) as its boundary, so we need to fix up the boundary with `∙v≡∙v'`.

cc. @mortberg

#### Bureaucracy

- Is there a good reason for calling `_∙₂_` "vertical" composition? I think this is based on the way we draw squares, which is not really set in stone. On the other hand, since `_∙₂_` composes `PathP`s along their *second* dimension, it really corresponds to *horizontal* composition in a 2-category. This is why I named the operation above `∙v`: it composes `PathP`s along their *first* dimension.
- Should we also have `_∙h_` (alias for `_∙₂_`), `_∙h'_` and `∙h≡∙h'`? The proofs would essentially be the same up to `flipSquare`, so maybe it's not worth cluttering the library?
- I'm not sure where the composition stuff should go, and more generally the relationship between `Cubical.Foundations.{Prelude,GroupoidLaws,Path}` seems kind of blurry. I put it in `Path` for now, the only constraint is that `∙v≡∙v'` depends on `comp-unique` in `GroupoidLaws` (and also that I'd like composition-related thing to be as close together as possible).
- I renamed the old `lid-unique` to `hlid-unique` and introduced heterogeneous counterparts `comp-unique` and `lid-unique`, so this is a breaking change (but nothing uses `lid-unique` in the library).
- I renamed `comm→PathP` to `compPath→Square` for homogeneity.
- Generally, names: please feel free to suggest.
- This interacts with https://github.com/agda/cubical/pull/906 as it introduces more `inS`/`outS`.

### DIFF
--- a/Cubical/Foundations/GroupoidLaws.agda
+++ b/Cubical/Foundations/GroupoidLaws.agda
@@ -209,22 +209,40 @@ cong-∙ : ∀ {B : Type ℓ} (f : A → B) (p : x ≡ y) (q : y ≡ z)
          → cong f (p ∙ q) ≡ (cong f p) ∙ (cong f q)
 cong-∙ f p q = cong-∙∙ f refl p q
 
-hcomp-unique : ∀ {ℓ} {A : Type ℓ} {φ} → (u : I → Partial φ A) → (u0 : A [ φ ↦ u i0 ]) →
-               (h2 : ∀ i → A [ (φ ∨ ~ i) ↦ (\ { (φ = i1) → u i 1=1; (i = i0) → outS u0}) ])
-               → (hcomp u (outS u0) ≡ outS (h2 i1)) [ φ ↦ (\ { (φ = i1) → (\ i → u i1 1=1)}) ]
+hcomp-unique : ∀ {ℓ} {A : Type ℓ} {φ}
+             → (u : I → Partial φ A) → (u0 : A [ φ ↦ u i0 ])
+             → (h2 : ∀ i → A [ (φ ∨ ~ i) ↦ (\ { (φ = i1) → u i 1=1; (i = i0) → outS u0}) ])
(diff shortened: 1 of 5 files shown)


ecavallo (approved):
Looks good to me! I think your bureaucratic changes are all reasonable. Feel free to add the horizontal composition synonym / lemmas in another PR if you want, otherwise someone can add them if they ever find they need them. Thanks!

PR #914, opened 2022-08-28
Refactor polynomials

felixwellen:
This is an effort to push the algebra part of the library somewhat more towards a common style.
In particular, so far, (while reading) I am refactoring the various definitions of polynomials a bit, in the following way:
* More bundled use of CommRings, e.g. ```A : CommRing``` instead of ```Acr@(A , Astr) : CommRing``` - I think this is a lot more readable in conjunction with
* Use of instance search for CommRing-Structures, e.g. ```x + y``` instead of ```_+_ Astr x y```
* Name recursors and eliminators in the usual way, e.g. ```elimProp``` instead of ```Ind-Prop.f```  

Let me know what you think!


thomas-lamiaux:
could this PR be merged after all the cohomology ones ? 
Otherwise it going to create a lot of stuff to fix

felixwellen:
Sure - can you have a look here:
https://github.com/agda/cubical/pull/914/files#diff-10188ff3797498a9126f5c681c626d1393821b30b54631a297879880a90fe8b1L96-L106
Maybe this is something you want to use in ongoing work.

thomas-lamiaux:
I had seen the last comment. 
More precisely, there is one last PR by Axel that his going to use polynomials, but then it should be all.

felixwellen:
I'll do this from scratch some time - should be easier than updating this one... -> closing

PR #915, opened 2022-08-29
Cup product of klein bottle (Z coeffs)

aljungstrom:
A small PR containing a lemma for @thomas-lamiaux about the cup product vanishing on H*(K^2)


mortberg:
Thanks!

Issue #781, opened 2022-05-04
Show the definitions of polynomials are equivalent


felixwellen:
There are three definitions of polynomials (one is called FreeCommAlgebra) and it should be shown, that all of them are equivalent. Possibly the best strategy is to show that all fullfill the universal property of the algebra of polynomials over a commutative ring k, which is:
```
Hom(k[X_1,...,X_n], A) = (Fin n -> A)
```
where ```Hom``` denotes the type of homomorphisms of k-Algebras.

thomas-lamiaux:
I don't have the time right now but if someone wants to do it, I have the following idea :

There is actually four potential definition of the polynomials in the library :

1. The HIT one located in Polynomials
2. The truncated list one located in Polynomials
3. The freely generated one located ?
4. The almost null sequences one located inside Univariate.Base. 
     No one ever developed a structure on the definition.
     However it should be easy as it is what is done in #798 in the complicated case where is dependency.
    
So about structure, it would be good to move 1 and 4, complete 4 + find new names.
Possible do the same as in Int, keeping one as the preferred one.

To prove the equivalence, one could prove the universal property but actually in this case, it probably easier to give syntactic translation. 
Indeed there is already a proof of 1 <-> 2. 2 <-> 4 is proved on types, so it suffices to define the structure and prove it on the operations. 1 <-> 3 should be easy as it is very similar structure.
Thus given us the equivalence without too much more work.

thomas-lamiaux:
update :
#798 is going to add full CommRing structure around 4 that was missing and prove the equivalence.
Hence in the library, there is going to be 4 <-> 1 <-> 2 so it suffices to add 1 <-> to finish the job

felixwellen:
I'll do 1 <-> 3 by:
* Constructing algebra structure on the list-polynomials
* Show the map from the universal property of the polynomial HiT in Algebra.CommAlgebra.FreeCommAlgebra is surjective
* ... show it is an Embedding, by constructing a one-sided inverse (showing left invertibility with the universal property)

felixwellen:
Or maybe I don't - just realized I have to show, that the inverse to the induced morphism is an algebra-hom...

felixwellen:
This is done.

PR #917, opened 2022-08-30
Universal property of list-based polynomials

felixwellen:
The aim of this PR, is to show that the UnivariatePolyList - polynomials have the correct universal property.
This is a missing piece, connecting all definitions of polynomials in the library (see #781).


mortberg:
Don't we already have that the list polynomials are equivalent to some of the other polynomials? Then this could just be transported over?

felixwellen:
As far as I know, only the CommAlgebra.FreeCommAlgebra has the universal property. And the aim of #781 is to show that the definitions in Algebra.Polynomials are equivalent to FreeCommAlgebra. @thomas-lamiaux suggested, that it is probably easier to construct directly an equivalence of types, which I also tried, but ultimately realized I need the parts of the universal property I formalized now here anyway.

felixwellen:
I grepped for 'universal' in the algebra part of the library and didn't find a universal property for any of the definitions of polynomials (except the FreeCommAlgebra). Let me know, if you think there is one I missed.

felixwellen:
Done, but depends on #916 -> still draft

felixwellen:
Thought this might be something for @MatthiasHu , if you agree, please assign the PR to yourself.

Cubical/Algebra/CommAlgebra/UnivariatePolyList.agda
```diff
@@ -2,29 +2,218 @@
...
+      _ = snd (CommRing→Ring R)
 
   ListPolyCommAlgebra : CommAlgebra R ℓ
   ListPolyCommAlgebra =
     Iso.inv (CommAlgChar.CommAlgIso R)
             (ListPoly ,
              constantPolynomialHom R)
+
+  {- Universal Property -}
+  module _ (A : Algebra (CommRing→Ring R) ℓ') where
```
mzeuner, line 40: It is a bit strange to me that we're in `Algebra.CommAlgebra` but actually consider not-necessarily commutative algebras...

Cubical/Algebra/Polynomials/UnivariateList/UniversalProperty.agda
```diff
@@ -0,0 +1,8 @@
+{-# OPTIONS --safe #-}
+module Cubical.Algebra.Polynomials.UnivariateList.UniversalProperty where
```
mzeuner, line 2: Is there a reason why you didn't want to prove the universal property here? That would resolve the issue about having a statement about Algebras in CommAlgebras.

Cubical/Algebra/CommAlgebra/UnivariatePolyList.agda
```diff
@@ -2,29 +2,218 @@
...
+              fst f (r ⋆ 1a + X · p)           ≡⟨ pres+ (r ⋆ 1a) (X · p) ⟩
+              fst f (r ⋆ 1a) + fst f (X · p)   ≡[ i ]⟨ pres⋆ r 1a i + pres· X p i ⟩
+              r ⋆ fst f 1a + fst f X · fst f p ≡[ i ]⟨ r ⋆ pres1 i + fX≡x i · fst f p ⟩
+              r ⋆ 1a + x · fst f p             ≡[ i ]⟨ r ⋆ 1a + (x · IH i) ⟩
+              r ⋆ 1a + x · inducedMap p        ≡⟨⟩
+              inducedMap (r ∷ p) ∎)
+            (is-set _ _)
+          where
+            useSolver : (r : ⟨ R ⟩) → r ≡ (r · 1r) + 0r
+            useSolver = solve R
```
mzeuner, line 219: Would it be hard to state the universal property as "there is an equivalence between A and the R-homs from R[X] to A"?
I feel like that would look nice if you write it down explicitly.

Cubical/Algebra/Algebra/Properties.agda
```diff
@@ -46,6 +46,13 @@ module AlgebraTheory (R : Ring ℓ) (A : Algebra R ℓ') where
                        (0r ⋆ x) + (0r ⋆ x) ∎
     in RingTheory.+Idempotency→0 (Algebra→Ring A) (0r ⋆ x) idempotent-+
 
+  0a-absorbs : (r : ⟨ R ⟩) → r ⋆ 0a ≡ 0a
```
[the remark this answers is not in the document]
felixwellen: Don't know, why I wrote this. According to the naming conventions, it should be ```·AnnihilR```.
I'll refactor it.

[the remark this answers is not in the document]
felixwellen: Or to be completely precise: ```⋆AnnihilR```

Cubical/Algebra/CommAlgebra/UnivariatePolyList.agda
```diff
@@ -2,29 +2,218 @@
...
+  {- Universal Property -}
+  module _ (A : Algebra (CommRing→Ring R) ℓ') where
+    open AlgebraStr ⦃...⦄ using (_⋆_; 0a; 1a; ⋆IdL; ⋆DistL+; ⋆DistR+; ⋆AssocL; ⋆AssocR; ⋆Assoc)
+    private instance
+      _ = snd A
+      _ = snd (Algebra→Ring A)
+      _ = snd (CommAlgebra→Algebra ListPolyCommAlgebra)
+
+    module _ (x : ⟨ A ⟩) where
+      open AlgebraTheory using (0-actsNullifying; 0a-absorbs)
```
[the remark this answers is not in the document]
felixwellen: I'll rewrite that to ```⋆AnnihilL```

Cubical/Algebra/Algebra/Properties.agda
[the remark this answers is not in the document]
felixwellen: Done

Cubical/Algebra/CommAlgebra/UnivariatePolyList.agda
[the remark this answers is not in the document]
felixwellen: Done

Cubical/Algebra/CommAlgebra/UnivariatePolyList.agda
felixwellen: Yes, I feel with you ;-)
But: We are proving the universal property of an CommAlgebra, which happens to quantify over Algebras. An analogous example where this is completely okay, would be the universal property of the abelianisation of a group. There it feels completely natural, that the ump quantifies over something more general. I think the ump of the CommAlgebra of Polynomials should also naturally quantify over something more general (e.g. Algebras), the situation is just more complex.

Cubical/Algebra/Polynomials/UnivariateList/UniversalProperty.agda
felixwellen: I think my reasoning was, that it would be good to be consistent about where things are. So 'Polynomials' could just contain a lot of pointers to Instances of CommRings and CommAlgebras and their Properties *and* theorems on the relations between different implementations of polynomials. But we are actually far away from that now, so I could move the ump as well.

Cubical/Algebra/CommAlgebra/UnivariatePolyList.agda
felixwellen: Yes, good idea. It shouldn't be hard, I'll try to that later...
Btw.: Have you thought about a general definition of UMP for algebraic things, that would allow us to change between the different forms and have some simple consequences? I would have some use for that, but haven't tried to write something. I only tried to use category theory (from the library), but that is not general enough (because of universe levels).

mzeuner: That would be cool. What do the textbooks say, do they ever give a general definition?
Maybe it helps to think about typical consequences, like the only homomorphism from the object to itself is the identity etc...

felixwellen: Yes, that's exactly something I would have liked to use a couple of times. That already seems to need functoriality, though, meaning we would have to redefine that with varying universe levels.

Cubical/Algebra/Polynomials/UnivariateList/UniversalProperty.agda
mzeuner: Fair enough, I hadn't looked at the `Polynomials/` too much but that seems to be more or less the standard, so let's try to keep it that way.

Cubical/Algebra/CommAlgebra/UnivariatePolyList.agda
felixwellen: Made an issue for that: #943

Cubical/Algebra/Polynomials/UnivariateList/UniversalProperty.agda
felixwellen: A notable exception seems to be the UnivariateList polynomials, which have their ring structure defined in Polynomials.UniveriateList.Properties. I'll make an issue with the suggestion to move that.

felixwellen: Done, #944

Cubical/Algebra/CommAlgebra/UnivariatePolyList.agda
felixwellen: Reformulated this universal property like suggested.

felixwellen:
@mzeuner : Do you think it can be merged?

mzeuner:
Yeah, I think I can live with the `Algebra` vs. `CommAlgebra` issue regarding the universal property, let's merge it.

PR #919, opened 2022-08-31
X is a regular element in the ring of list-based polynomials

felixwellen:
This means that multiplication with X is injective.
Depends on #917


Cubical/Algebra/Polynomials/UnivariateList/Properties.agda
```diff
@@ -639,6 +639,38 @@ module PolyModTheory (R' : CommRing ℓ) where
...
 
+---------------------------------------------------------------------------
+-- The indeterminate is a regular element, i.e. multiplication with it is injective
+
+  prod-X-injective : (p : Poly R') → prod-Xn 1 p ≡ 0P → p ≡ 0P
+  prod-X-injective p Xp≡0 = p≡0
+    where
+      shift = shiftFun (Poly→PolyFun p)
+
+      0r∷p≡shift : Poly→PolyFun (0r ∷ p) ≡ shift
```
MatthiasHu, line 650: If `shiftFun` is exported, then perhaps this lemma (read as a property of `shiftFun`) should also be exported. Do you agree?

MatthiasHu (commented):
I like it very much!
One suggestion below.

Cubical/Algebra/Polynomials/UnivariateList/Properties.agda
felixwellen: Yes, I agree...

felixwellen: Done

MatthiasHu (approved):

PR #921, opened 2022-09-01
[ re agda/#1625 ] Rename --experimental-lossy-unification -> --lossy-unification

jespercockx:
See https://github.com/agda/agda/issues/1625#issuecomment-1091699806


felixwellen:
part of #948 now -> closing

PR #922, opened 2022-09-01
Typevariate and list-based polynomials are isomorphic

felixwellen:
This PR shows that there is an isomorphism of algebras between the typevariate polynomials (with index type Unit) defined as a HIT and the list-based polynomials. 
Depends on #917


felixwellen:
#917 merged -> ready for review.

felixwellen:
...once the conflicts are resolved.

Cubical/Algebra/CommAlgebra/UnivariatePolyList.agda
```diff
@@ -216,10 +219,30 @@ module _ (R : CommRing ℓ) where
...
     fst indcuedHomEquivalence f = fst f X
     fst (fst (equiv-proof (snd indcuedHomEquivalence) x)) = inducedHom x
     snd (fst (equiv-proof (snd indcuedHomEquivalence) x)) = inducedMapGenerator x
     snd (equiv-proof (snd indcuedHomEquivalence) x) (g , gX≡x) =
       Σ≡Prop (λ _ → isSetAlgebra A _ _) (sym (inducedHomUnique x g gX≡x))
+
+
+  {- A corollary, which is useful for constructing isomorphisms to
+     algebras with the same universal property -}
+  isIdByUMP : (f : CommAlgebraHom ListPolyCommAlgebra ListPolyCommAlgebra)
```
MatthiasHu, line 234: I think if we have `isIdByUMP` here as we do in `FreeCommAlgebra/Properties.agda`, then we should also factor it over an `equalByUMP` as done there. (But proved using `inducedHomUnique` in this case.)

Cubical/Algebra/Polynomials/TypevariateHIT/EquivUnivariateListPoly.agda
```diff
@@ -0,0 +1,73 @@
...
+private variable
+  ℓ : Level
+
+module _ {R : CommRing ℓ} where
+  open Theory renaming (inducedHom to inducedHomHIT)
+  open CommRingStr ⦃...⦄
+  private instance
+    _ = snd R
+    X-HIT = Construction.var {R = R} {I = Unit} tt
+    X-List = generatorList R
```
MatthiasHu, line 32: These two should not be instances, or should they?

Cubical/Algebra/Polynomials/TypevariateHIT/EquivUnivariateListPoly.agda
```diff
@@ -0,0 +1,73 @@
...
+    idList = AlgebraHoms.idAlgebraHom (CommAlgebra→Algebra (ListPolyCommAlgebra R))
+    idHIT = AlgebraHoms.idAlgebraHom (CommAlgebra→Algebra (R [ Unit ]))
+
+    toFrom : (x : ⟨ ListPolyCommAlgebra R ⟩) → fst to (fst from x) ≡ x
+    toFrom = isIdByUMP-List R (to ∘a from) (cong (fst to) fromPresX)
+
+    fromTo : (x : ⟨ R [ Unit ] ⟩) → fst from (fst to x) ≡ x
+    fromTo = isIdByUMP-HIT (from ∘a to) λ {tt → fromPresX}
+
+    asIso : Iso ⟨ R [ Unit ] ⟩  ⟨ ListPolyCommAlgebra R ⟩
```
MatthiasHu, line 61: I thought as a rule of thumb, if you construct an equivalence by constructing an iso, then you should export the iso too, because later you might want to use your knowledge about the `inv` part too. Is this true?

MatthiasHu (changes requested):
Nice. Two comments about consistent style / reusability.

Cubical/Algebra/CommAlgebra/UnivariatePolyList.agda
felixwellen: Done

Cubical/Algebra/Polynomials/TypevariateHIT/EquivUnivariateListPoly.agda
felixwellen: Yes, they shouldn't, thanks!

Cubical/Algebra/Polynomials/TypevariateHIT/EquivUnivariateListPoly.agda
felixwellen: Yes.

MatthiasHu (approved):

PR #923, opened 2022-09-01
remove duplicate of Pushout.elimProp

MatthiasHu:
This PR removes `PushoutToProp` from `Cubical.HITs.Pushout.Properties`, which was just a duplicate of `elimProp`, with the same definition.


felixwellen (approved):
Thanks!

PR #924, opened 2022-09-02
Move and split FPAlgebra

felixwellen:
Create a folder for fp algebras. One could also go further and split up the instances.


MatthiasHu (approved):
I like it!

PR #928, opened 2022-09-12
H*(Kleinbottle,Z/2)

aljungstrom:
WIP


thomas-lamiaux:
Shouldn't this def be in algebra ?
https://github.com/agda/cubical/pull/928/files#diff-2ac1bdc3baa3997bd843c09403b37cfa86fbb20580d09320cc88bbc58ad940e9R205

mortberg:
> Shouldn't this def be in algebra ? https://github.com/agda/cubical/pull/928/files#diff-2ac1bdc3baa3997bd843c09403b37cfa86fbb20580d09320cc88bbc58ad940e9R205

I can't see which proof you're referring to. Please comment on the actual line in the code instead

mortberg (approved):
Looks good to me after a quick skim through

PR #931, opened 2022-09-13
Use commAlgebraFromCommRing in the equivalence between algebras and over-rings

MatthiasHu:
This small PR uses the `commAlgebraFromCommRing` construction more consequently, in code that was written before it. We also single out a property of the construction (`commAlgebraFromCommRing→CommRing`) from the surrounding proof.


mzeuner:
Nice, I actually moved the prove of `APath` outside of the `where` block as well but didn't use `commAlgebraFromCommRing` in my latest [PR](https://github.com/agda/cubical/pull/929/files#diff-98faa58f8c9d8a08689393b86fc64666ae13b667f58c78870def606175a3e21e).
Do you think it makes any difference in the computational behavior?

MatthiasHu:
Ah, I forgot I still have to figure out what went wrong here. Trying to do that soon.

felixwellen:
ping @MatthiasHu

felixwellen:
This is now part of #1007 -> closing.

PR #939, opened 2022-09-30
Allow nullification to take a family of types as input rather than just one type

awswan:
The current version of nullification in the library only took a single type S as input. However, the definition in [RSS] is defined for a family of types. Many of the examples of nullification in [RSS] and in my papers really require the more general definition with a family of types, so I suggest changing the definition in the library.

In the first commit I changed the definition of nullification following the notation in `Localization/`, which is already defined for families of maps. In the second commit I updated the files in the library that refer to nullification to use the more general definition, which turned out to just be truncation. It type checks successfully for me on agda 2.6.2.2.

[RSS] Rijke, Shulman, Spitters, *Modalities in homotopy type theory*


ecavallo (approved):
Looks good to me, thanks!

PR #940, opened 2022-10-03
R is not zigzag complete if R<- and R-> are PERs. 

phadej:


ecavallo:
Ouch, thanks pointing this out. For context for others, we claimed this was true in our representation independence paper (https://doi.org/10.1145/3434293, just above Definition 5.3). I think it was me who wrote this, and I got it from [Krishnaswami and Dreyer](https://drops.dagstuhl.de/opus/volltexte/2013/4212/) (just below Definition 1). Should have double-checked! Maybe they meant something different... anyway, fortunately nothing depends on it.

It's probably not super useful to have in the library, but it will be good to have a record of the mistake. Could you add a comment noting the source of the false claim? Then I'll merge.

jonsterling:
Interesting. the claim of krishnaswami and dreyer however looks indeed like this one here, so it seems to be an error in their paper.

phadej:
Added a note in the source as well (it's in commit message too).

PR #945, opened 2022-10-23
Define profunctors and representability of profunctors in 2 ways

maxsnew:
This defines profunctors and what it means for a functor to represent a profunctor. It gives two definitions, analogous to definitions of adjunctions by hom-iso and universal morphism: `Representable` defines it as a functor with a hom isomorphism `Hom[ F , Id ] =~ R` and `Representable'` defines it as a function on objects `F_0` with a universal morphism and induction principle (very similar to a presentation of a positive type by constructors, induction principle and beta-eta). I provided mappings between these two notions but haven't proven that it's an equivalence yet because I wasn't quite sure how to do it. My hope is that this profunctor approach can be used to flesh out the category theory part of the library, as the simplest proofs of many theorems like RAPL are structured as isomorphisms of profunctors.

This is my first time contributing to the cubical library so I'm not sure I structured it in an idiomatic way. Advice on this would be appreciated.

Related discussion in https://github.com/agda/cubical/pull/873.

### DIFF
--- a/Cubical/Categories/Profunctor/Base.agda
+++ b/Cubical/Categories/Profunctor/Base.agda
@@ -1,0 +1,285 @@
+{-# OPTIONS --safe #-}
+module Cubical.Categories.Profunctor.Base where
+
+open import Cubical.Foundations.Prelude hiding (Path)
+open import Cubical.Foundations.Structure
+
+open import Cubical.Data.Graph.Base
+open import Cubical.Data.Graph.Path
+
+open import Cubical.Categories.Category
+open import Cubical.Categories.Functor
+open import Cubical.Categories.Instances.Functors
+open import Cubical.Categories.NaturalTransformation
+open import Cubical.Categories.Constructions.BinProduct
+open import Cubical.Categories.Instances.Sets
+open import Cubical.Categories.Functors.HomFunctor
+
+-- The simplest case is when the level of the sets in the profunctors
+-- is the same as the level of the hom sets of categories. Roughly
+-- this means the categories involved are "locally small".
+module Profunctors (ℓ ℓ' : Level) where
+  Cat : Type (ℓ-max (ℓ-suc ℓ) (ℓ-suc ℓ'))
+  Cat = Cubical.Categories.Category.Category ℓ ℓ'
+  PROF : (C D : Cat) → Category (ℓ-max ℓ (ℓ-suc ℓ')) (ℓ-max ℓ ℓ')
+  PROF C D = FUNCTOR ((C ^op) × D) (SET ℓ')
+
+  record Profunctor (C D : Cat) : Type (ℓ-max ℓ (ℓ-suc ℓ')) where
+    constructor fromFunc
+    field
+      asFunc : Category.ob (PROF C D)
+
+    private
+      R = asFunc
+
+    module C = Category C
+    module D = Category D
+    _⋆C_ = C._⋆_
+    _⋆D_ = D._⋆_
+
+    Het[_,_] : C.ob → D.ob → Type ℓ'
+    Het[ c , d ] = ⟨ asFunc ⟅ c , d ⟆ ⟩
+
+    _⋆L⟨_⟩⋆R_ : ∀ {c c' d' d} → (f : C.Hom[ c , c' ])(r : Het[ c' , d' ])(g : D.Hom[ d' , d ]) → Het[ c , d ]
+    f ⋆L⟨ r ⟩⋆R g = Functor.F-hom R (f , g) r
+
+    ⋆L⟨⟩⋆RAssoc : ∀ {c c' c'' d'' d' d} → (f : C.Hom[ c , c' ])(f' : C.Hom[ c' , c'' ])(r : Het[ c'' , d'' ])(g' : D.Hom[ d'' , d' ])(g : D.Hom[ d' , d ])
+                → (f ⋆C f') ⋆L⟨ r ⟩⋆R (g' ⋆D g) ≡ f ⋆L⟨ f' ⋆L⟨ r ⟩⋆R g' ⟩⋆R g
+    ⋆L⟨⟩⋆RAssoc f f' r g' g i = Functor.F-seq R (f' , g') (f , g) i r
+
+    ⋆L⟨⟩⋆RId : ∀ {c d} → (r : Het[ c , d ])
+             → C.id ⋆L⟨ r ⟩⋆R D.id ≡ r
+    ⋆L⟨⟩⋆RId r i = Functor.F-id R i r
+
+
+    _⋆L_ : {c c' : C.ob} {d : D.ob}
+              → C.Hom[ c , c' ]
+              → Het[ c' , d ]
+              → Het[ c  , d ]
+    _⋆L_ f r = f ⋆L⟨ r ⟩⋆R D.id
+    infixr 9 _⋆L_
+
+    ⋆LId : ∀ {c d} → (r : Het[ c , d ]) → C.id ⋆L r ≡ r
+    ⋆LId = ⋆L⟨⟩⋆RId
+
+    ⋆LAssoc : ∀ {c c' c'' d} → (f : C.Hom[ c , c' ])(f' : C.Hom[ c' , c'' ])(r : Het[ c'' , d ])
+            → (f ⋆C f' ⋆L r) ≡ (f ⋆L f' ⋆L r)
+    ⋆LAssoc f f' r =
+      ((f ⋆C f') ⋆L⟨ r ⟩⋆R D.id) ≡⟨ (λ i → (f ⋆C f') ⋆L⟨ r ⟩⋆R sym (D.⋆IdL D.id) i) ⟩
+      ((f ⋆C f') ⋆L⟨ r ⟩⋆R (D.id ⋆D D.id)) ≡⟨ ⋆L⟨⟩⋆RAssoc f f' r D.id D.id ⟩
+      f ⋆L f' ⋆L r ∎
+            
+
+    _⋆R_ : {c : C.ob} {d' d : D.ob}
+         → Het[ c , d' ]
+         → D [ d' , d ]
+         → Het[ c , d ]
+    _⋆R_ r g = C.id ⋆L⟨ r ⟩⋆R g
+    infixl 9 _⋆R_
+
+    ⋆RId : ∀ {c d} → (r : Het[ c , d ]) → r ⋆R D.id ≡ r
+    ⋆RId = ⋆L⟨⟩⋆RId
+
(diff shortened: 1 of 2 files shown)


felixwellen:
Hi Max!
We have the (rarely followed) convention, to explain what happens in a file at the beginning of the file. I think it would make sense to add your description of this PR directly as a comment, since it is certainly about concept not everyone is familiar with. Here is an example of such a comment:

https://github.com/agda/cubical/blob/master/Cubical/Foundations/Equiv/PathSplit.agda

You might want to check out the hints about whitespaces here:

https://github.com/agda/cubical/blob/master/CONTRIBUTING.md

Maybe you already found this file:

https://github.com/agda/cubical/blob/master/Cubical/Algebra/NAMING.md

On a first glance, your code looks like you are already aware of this - I'm not too familiar with the category theory part of the library and will wait a bit to give @mortberg a chance to look at your PR in more detail.

plt-amy:
(FWIW, PRs from new contributors need approval for every commit. I've approved your latest one)

Cubical/Categories/Profunctor/Base.agda
```diff
@@ -0,0 +1,307 @@
...
+      -- aka the elimination rule/pattern matching
+      induction      : ∀ {c : C.ob} {d : D.ob} → R.Het[ c , d ] → D.Hom[ F₀ c , d ]
+      -- aka the β-reduction principle
+      computation    : ∀ {c : C.ob} {d : D.ob} → (r : R.Het[ c , d ])
+                     → (unit ⋆R induction r) ≡ r
+      -- aka the η-expansion principle
+      extensionality : ∀ {c d} (f : D.Hom[ F₀ c , d ]) → f ≡ induction (unit ⋆R f)
+
+    weak-extensionality : ∀ {c} → D.id ≡ induction (unit {c = c})
+    weak-extensionality = D.id ≡⟨ extensionality D.id ⟩ induction (unit ⋆R D.id) ≡⟨ (λ i → induction (R.⋆RId unit i)) ⟩ induction unit ∎
```
felixwellen, line 201: This is a bit long, please add line breaks.

Cubical/Categories/Profunctor/Base.agda
```diff
@@ -0,0 +1,307 @@
...
+
+    naturality : ∀ {c : C.ob}{d d' : D.ob}(r : R.Het[ c , d' ]) (k : D.Hom[ d' , d ])
+               → (induction r ⋆D k) ≡ induction (r ⋆R k)
+    naturality r k =
+      induction r ⋆D k ≡⟨ extensionality (induction r ⋆D k) ⟩
+      induction (unit ⋆R induction r ⋆D k) ≡⟨ (λ i → induction (R.⋆RAssoc unit (induction r) k i)) ⟩
+      induction ((unit ⋆R induction r) ⋆R k) ≡⟨ (λ i → induction (computation r i ⋆R k)) ⟩
+      induction (r ⋆R k) ∎
```
felixwellen, line 209: align

Cubical/Categories/Profunctor/Base.agda
```diff
@@ -0,0 +1,307 @@
...
+      ; F-id =
+             induction (C.id ⋆L unit) ≡⟨ (λ i → induction (R.⋆LId unit i)) ⟩
+             induction unit ≡⟨ sym weak-extensionality ⟩
+             D.id ∎
+      ; F-seq = λ f g →
+        induction ((f ⋆C g) ⋆L unit)  ≡⟨ (λ i → induction (R.⋆LAssoc f g unit i)) ⟩
+        induction (f ⋆L (g ⋆L unit))  ≡⟨ (λ i → induction (f ⋆L sym (computation (g ⋆L unit)) i)) ⟩
+        induction (f ⋆L (unit ⋆R (induction (g ⋆L unit))))  ≡⟨ (λ i → induction (sym (R.⋆L⋆RAssoc f unit (induction (g ⋆L unit))) i)) ⟩
+        induction ((f ⋆L unit) ⋆R (induction (g ⋆L unit)))  ≡⟨ sym (naturality (f ⋆L unit) (induction (g ⋆L unit))) ⟩
+        induction (f ⋆L unit) ⋆D induction (g ⋆L unit) ∎
```
felixwellen, line 225: align / move equality proofs

Cubical/Categories/Profunctor/Base.agda
```diff
@@ -0,0 +1,307 @@
...
+              f ⋆L⟨ unit ⋆R IP ⟩⋆R g ∎
+
+    F-represents-R : F Represents R
+    F-represents-R = record { trans = induction⁻¹ ; nIso = λ x → isiso induction (λ i r → induction⁻¹-induction≡id r i) λ i IP → induction-induction⁻¹≡id IP i }
+      where induction-induction⁻¹≡id : ∀ {c d}(IP : D.Hom[ F₀ c , d ]) → induction (unduction IP) ≡ IP
+            induction-induction⁻¹≡id IP =
+              induction (unit ⋆R IP) ≡⟨ sym (naturality unit IP) ⟩
+              induction unit ⋆D IP ≡⟨ (λ i → sym weak-extensionality i ⋆D IP) ⟩
+              D.id ⋆D IP               ≡⟨ D.⋆IdL _ ⟩
+              IP ∎
```
felixwellen, line 256: align

Cubical/Categories/Profunctor/Base.agda
```diff
@@ -0,0 +1,307 @@
...
+            r ∎
+
+          extensionality : ∀ {c d} (f : D.Hom[ F.F-ob c , d ]) → f ≡ induction (unduction D.id ⋆R f)
+          extensionality f =
+            f ≡⟨ sym (λ i → isIso.ret (NatIso.nIso F-repr-R _) i f) ⟩
+            induction (unduction f) ≡⟨ (λ i → induction (unduction (sym (D.⋆IdL f) i))) ⟩
+            induction (unduction (D.id ⋆D f)) ≡⟨ (λ i → induction (unduction (sym (D.⋆IdL D.id) i ⋆D f))) ⟩
+            induction (unduction ((D.id ⋆D D.id) ⋆D f)) ≡⟨ (λ i → induction (unduction ((sym F.F-id i ⋆D D.id) ⋆D f))) ⟩
+            induction (unduction ((F.F-hom C.id ⋆D D.id) ⋆D f)) ≡⟨ (λ i → induction (unduction-homo.homomorphism C.id D.id f i)) ⟩
+            induction (C.id ⋆L⟨ unduction D.id ⟩⋆R f) ∎
```
felixwellen, line 303: align

Cubical/Categories/Profunctor/Base.agda
```diff
@@ -0,0 +1,307 @@
...
+open import Cubical.Categories.Instances.Functors
+open import Cubical.Categories.NaturalTransformation
+open import Cubical.Categories.Constructions.BinProduct
+open import Cubical.Categories.Instances.Sets
+open import Cubical.Categories.Functors.HomFunctor
+
+-- The simplest case is when the level of the sets in the profunctors
+-- is the same as the level of the hom sets of categories. Roughly
+-- this means the categories involved are "locally small".
+module Profunctors (ℓ ℓ' : Level) where
```
felixwellen, line 44: I think it would make sense to use an anonymous module here.

felixwellen (changes requested):
Thanks for the contribution!
Looks good to me - I only have some comments/suggestions on formatting and things like that.
There is one additional issue: In the case of putting something into a folder, like Categories/Profunctor/Base.agda, we always have a Categories/Profunctor.agda, which just imports the defaults (Profunctor/Base in this case) publicly. Please add a file like that (there should be lots of examples).

Cubical/Categories/Profunctor/Base.agda
```diff
@@ -0,0 +1,307 @@
+{-
+
+  Definition of profunctors (https://ncatlab.org/nlab/show/profunctor)
+  and some basic facts about them.
+
+  We define a profunctor C -/-> D as a functor C^o x D -> Set. We pick
+  the universe levels such that the hom sets of C and D match Set,
```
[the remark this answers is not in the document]
maxsnew: As mentioned on the nlab, different people follow different conventions. This convention has the benefit of matching how you apply it to arguments. I.e., if R : C -/-> D then the set is R.Het[ c , d ] whereas with the opposite convention it would be R.Het[ d , c ] . The alternative convention has the benefit that the functor `Cat -> Prof` is doesn't involve an op or co, but I don't think that that will matter so much in this library that is for the moment quite concrete.

Cubical/Categories/Profunctor/Base.agda
```diff
@@ -0,0 +1,307 @@
...
+
+    _⋆L_ : {c c' : C.ob} {d : D.ob}
+              → C.Hom[ c , c' ]
+              → Het[ c' , d ]
+              → Het[ c  , d ]
+    _⋆L_ f r = f ⋆L⟨ r ⟩⋆R D.id
+    infixr 9 _⋆L_
+
+    ⋆LId : ∀ {c d} → (r : Het[ c , d ]) → C.id ⋆L r ≡ r
+    ⋆LId = ⋆L⟨⟩⋆RId
```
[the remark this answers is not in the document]
maxsnew: I think based on reading that file, it's not necessary because there is no analogous ⋆LIdR.

Cubical/Categories/Profunctor/Base.agda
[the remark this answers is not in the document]
maxsnew: Mike Shulman on Zulip made the following suggestion: we could define both `C ⊶ D` and `D ⊷ C` to mean profunctors where the open side is contravariant (to suggest an "op"). This has the downside that the unicode characters here don't have names in Agda-mode, but I think it makes it nicely unambiguous and doesn't globally fix an orientation.

The only naming issue is what we call the category of all such profunctors, maybe just PROF⊶ C D and  PROF ⊷ C D?

Cubical/Categories/Profunctor/Base.agda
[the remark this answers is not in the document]
felixwellen: Yes, agreed.

Cubical/Categories/Profunctor/Base.agda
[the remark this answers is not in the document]
felixwellen: I see, thanks for explaining!
Just in case this is not clear: I was just pointing something out I was confused about - I'm fine with accepting the code as is.
I like the suggestion though and I don't mind using characters without  a name in the agda-mode.

maxsnew:
Ok I updated it with the notation I mentioned and added the `Profunctor.agda` file. Let me know if there's anything else

felixwellen:
No - looks all good to me! -> Merging...

PR #946, opened 2022-10-26
Minor adjustment to universe level parameters and a correction

awswan:
I made a couple of minor adjustments to universe level parameters to make the corresponding functions more general. While I was doing this I noticed the statement of `idemNull` was wrong and also corrected that.


ecavallo (approved):
LGTM!

PR #947, opened 2022-10-30
Added Braided, Symmetric and Closed Monoidal Categories

jakeKonrad:
These seem like useful definitions to have. All of them are taken from ncatlab.org. Please let me know if I got anything wrong or if anything could been done better or if these definitions just aren't useful in the library. Thanks!

### DIFF
--- a/Cubical/Categories/Monoidal/Braided.agda
+++ b/Cubical/Categories/Monoidal/Braided.agda
@@ -15,30 +15,28 @@ module _ {ℓ ℓ' : Level} (C : Category ℓ ℓ') where
   open Category C
   open Functor
 
-  -- private record to define the functors that describe the braiding
-  -- and nice names to define the actual braiding
   private
-    record BraidedStr : Type (ℓ-max ℓ ℓ') where
-      field
-        monstr : MonoidalStr C
+    -- swaps the arguments of a bifunctor
+    Swap : Functor (C × C) C → Functor (C × C) C
+    F-ob (Swap F)  (x , y) = F .F-ob (y , x)
+    F-hom (Swap F) (f , g) = F .F-hom (g , f)
+    F-id  (Swap F)         = F .F-id
+    F-seq (Swap F) f g     = F .F-seq (snd f , fst f) (snd g , fst g)
 
-      open MonoidalStr monstr public
+  record BraidedStr : Type (ℓ-max ℓ ℓ') where
+    field
+      monstr : MonoidalStr C
 
+    open MonoidalStr monstr public
+
+    -- private names to make the braiding look nice
+    private
       [x⊗y] : Functor (C × C) C
       [x⊗y] = ─⊗─
 
       -- just ─⊗─ but swaps the arguments
       [y⊗x] : Functor (C × C) C
-      F-ob  [y⊗x] (x , y) = y ⊗ x
-      F-hom [y⊗x] (f , g) = g ⊗ₕ f
-      F-id  [y⊗x]         = ─⊗─ .F-id
-      F-seq [y⊗x] f g     = ─⊗─ .F-seq (snd f , fst f) (snd g , fst g)
-
-  record BraidedMonStr : Type (ℓ-max ℓ ℓ') where
-    field
-      braidedstr : BraidedStr
-
-    open BraidedStr braidedstr public
+      [y⊗x] = Swap ─⊗─
 
     field
       -- the braiding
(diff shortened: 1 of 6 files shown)


Cubical/Categories/Monoidal/Closed.agda
```diff
@@ -0,0 +1,60 @@
...
+open import Cubical.Categories.Adjoint
+
+open UnitCounit
+
+module _ {ℓ ℓ' : Level} (C : Category ℓ ℓ') where
+  open Category C
+  open Functor
+
+  -- the closed structure
+  record ClosedStr : Type (ℓ-max ℓ ℓ') where
```
MatthiasHu, line 23: Perhaps "closed structure" / `ClosedStr` is not the optimal name for just a bifunctor `[─,─]`, since there is also this notion:
https://ncatlab.org/nlab/show/closed+category

What about `InternalHomStr` or `BifunctorStr`?

Cubical/Categories/Monoidal/Braided.agda
```diff
@@ -0,0 +1,73 @@
...
+open import Cubical.Foundations.Prelude
+open import Cubical.Categories.Monoidal.Base
+
+module _ {ℓ ℓ' : Level} (C : Category ℓ ℓ') where
+  open Category C
+  open Functor
+
+  private
+    -- swaps the arguments of a bifunctor
+    Swap : Functor (C × C) C → Functor (C × C) C
```
MatthiasHu, line 20: I think instead of this special-purpose function it would be better to compose `─⊗─` with a certain swapping functor of general type `Functor (C × D) (D × C)`, to be defined in `Cubical.Categories.Constructions.BinProduct`. This would incidentally also be the first step to resolve a todo note there: "- prove product is commutative up to isomorphism".

MatthiasHu (commented):
Nice work. Here are three comments.

MatthiasHu:
> All of them are taken from ncatlab.org.

It would be good to state this at the beginning of each file, with a link to the respective nalb article.

MatthiasHu:
I think in the definition of closed monoidal category you actually deviated quite a bit from [the nlab definition](https://ncatlab.org/nlab/show/closed+monoidal+category#symmetric_closed_monoidal_category): Instead of just requiring a right adjoint for all the functors b ⊗─, separately for each b, you require a functor [─,─] defined on C ^op × C and then a certain adjunction for every b. (One could also say that the nlab is cheating there, because they then act as if they had a functor [─,─] in two arguments and don't really explain why.)

The two definitions are equivalent, but this is not entirely trivial. One has to show that the individual functors [ b ,─] from the nlab definition canonically assemble into a functor in two arguments [─,─]. This should be a special case of a theorem talking about a functor into a functor category F : C --> [ D , E ] such that for all c, the functor F(c) has an adjoint, right?

Perhaps it is indeed better to keep the definition as you wrote it, I can't really judge this now.

I expect that the equivalence of the two definitions (or a "smart constructor" that expects just the data mentioned on the nlab) would be good to have when formalizing examples of closed monoidal categories.

jakeKonrad:
I think I addressed everything you said and learned a lot along the way thanks!

MatthiasHu (changes requested):
Nice, I am glad this worked out so well! :-)

Just one comment change request below.

MatthiasHu:
It would be great if the quite big proofs of `fillsOut-Bifunctor` and `F-Curry` could be made more readable or more modular, but I don't see how to do this right now.

@felixwellen pointed out to me that recently there was a pull request (#945) about Profunctors, see `Cubical.Categories.Profunctor.Base`, and the contents of that are also closely related to adjunctions. Not sure if there is a connection.

jakeKonrad:
Do you have a good reference for the relation between profunctors and adjunctions? 

Mac Lane proves the theorem using the notion of conjugate natural transformations, but I didn't find that terminology anywhere else so I didn't use it. But maybe it would make the proof more modular? If so, then there is also the category of adjunctions between categories C and D defined using conjugate natural transformations. Do you think this is a useful direction to travel down?

MatthiasHu (approved):

MatthiasHu:
> Mac Lane proves the theorem using the notion of conjugate natural transformations

I think that conjugate natural transformations can be very useful, but this is a topic for another pull request.

MatthiasHu:
> Do you have a good reference for the relation between profunctors and adjunctions?

Here is a bit about it in the nlab:
https://ncatlab.org/nlab/show/adjoint+functor#InTermsOfCographsHeteromorphisms

MatthiasHu:
I merged this PR without checking that it still passes CI after the rebase. It did not pass, so I had to revert. Sorry to all for the noise!

MatthiasHu:
@jakeKonrad I can unfortunately not re-open this PR, since it was merged. Could you please open a new PR, merge the current master branch (or rebase onto it) and then try to fix the issues that arise?

PR #958, opened 2022-11-23
workflow_dispatch for ci-nix

MatthiasHu:
This should allow us to manually trigger the ci-nix workflow.


MatthiasHu:
Now I am not so sure any more if I am doing this right. I can't find a button to actually start the workflow anywhere in this PR.

PR #959, opened 2022-11-24
Added agda-2.6.3 in flakes

guilhermehas:
Cubical Agda in Nix is compiled using Agda 2.6.3.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,5 +1,25 @@
 {
   "nodes": {
+    "agda": {
+      "inputs": {
+        "flake-utils": "flake-utils",
+        "nixpkgs": "nixpkgs"
+      },
+      "locked": {
+        "lastModified": 1669235596,
+        "narHash": "sha256-u4yZPqadMfTnE98c5FKV7yMY3cTehXqjNVZmXPXfkYo=",
+        "owner": "agda",
+        "repo": "agda",
+        "rev": "19960e93da5bd775a795a4a3f7c301aceeca645d",
+        "type": "github"
+      },
+      "original": {
+        "owner": "agda",
+        "ref": "release-2.6.3",
+        "repo": "agda",
+        "type": "github"
+      }
+    },
     "flake-compat": {
       "flake": false,
       "locked": {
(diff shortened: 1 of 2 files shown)


felixwellen:
Nice!
I don't understand why the old ci-ubuntu job doesn't run - do you know?

felixwellen:
More questions: 
* How the version of agda picked? Right now, we should use the branch release-2.6.3 for testing and we should switch eventually to the tag v2.6.3 which doesn't exist yet.
* How do nix flakes compose? If I have a project, that uses flake A and flake B, which both have agda defined as outputs, which agda will be available in the environment I get in the end?

guilhermehas:
> Nice! I don't understand why the old ci-ubuntu job doesn't run - do you know?
No, I don't know

> * How the version of agda picked? Right now, we should use the branch release-2.6.3 for testing and we should switch eventually to the tag v2.6.3 which doesn't exist yet.
The version of file is in this Line:
https://github.com/agda/cubical/blob/7f947d86ad872f70bdb63aed852745eefa97d847/flake.nix#L10
For your example, the line will be:
```nix
inputs.agda.url = "github:agda/agda/?ref=v2.6.3";
```

> * How do nix flakes compose? If I have a project, that uses flake A and flake B, which both have agda defined as outputs, which agda will be available in the environment I get in the end?
It is composed using overlay:
For example:
```nix
overlays = [ overlayAgda2 overlayAgda3 ]
pkgs = import nixpkgs { inherit overlays}
```
Nixpkgs have Agda1, overlayAgda2 has Agda2 and overlayAgda3 has Agda3.
I think that in the end, Agda3 will be used.

@MatthiasHu maybe can help if I explained anything wrong.

felixwellen:
Thanks for explaining and for pointing me to the agda-version - I just overlooked that line in the diff...

MatthiasHu:
> Nice! I don't understand why the old ci-ubuntu job doesn't run - do you know?

That is because the ci-ubuntu workflow is configured to run on pull requests into master, but this is not a pull request into master. :-)

MatthiasHu:
I was a bit confused why we have two versions of nixpkgs in `flake.lock` now, `nixpkgs` and `nixpkgs_2`. I think this is because the Agda flake (in the version specified in our flake.flake) specifies (in its flake.lock) a different version of nixpkgs than our flake.lock does. I am not sure if this is the way it is supposed to be when you compose flakes, but I also don't see a good way to avoid it.

MatthiasHu:
I think I understand now that we could declare that our nixpkgs input `follows` Agda's nixpkgs -- or the other way round. But it does not seem to be necessary: ["[...] it is generally not useful to eliminate transitive nixpkgs flake inputs [...]"](https://nixos.org/manual/nix/unstable/command-ref/new-cli/nix3-flake.html#flake-inputs). We could still do it for flake-utils, but probably it is not important.

PR #961, opened 2022-11-24
Nixpkgs of Agda using following another nixpkgs

guilhermehas:
Instead of using two nixpkgs in flake.lock, it is using just one.


guilhermehas:
@MatthiasHu

MatthiasHu:
LGTM

PR #964, opened 2022-11-27
remove isTruncatedFun (duplicate of isOfHLevelFun)

rwbarton:
This was not used anywhere.
Even in the same file, isOfHLevelFun is used instead.


rwbarton:
It might be nice to rename `isOfHLevelFun` to `isTruncatedFun` instead; but having both names seems confusing to me.

felixwellen:
I agree. With both points. If we rename to `isTruncFun` we should however rename all the other `isOfHLevel`-things...

PR #966, opened 2022-11-30
Affine schemes paper

mzeuner:
A summary file for our paper on formalizing affine schemes in Cubical Agda.
(Preprint on arXiv coming soon)


mzeuner:
Should we wait for the arXiv-link before merging though?

PR #976, opened 2023-01-27
[ re agda/#6368 ] Add blanket instance for TypeWithStr

jespercockx:
I would like to merge https://github.com/agda/agda/pull/6368, but this requires a (very small) change to the cubical library. I would say it is an improvement, since it means less instances have to be declared by hand.


jespercockx:
I'm confused, there's a CI error about a name that's not in scope but it's not even in the part that I modified.

mortberg:
I have no clue what the CI problem is, but the PR looks good to me!

mortberg (approved):

ecavallo:
We removed the `has-⟨⟩` typeclass from the cubical library a while back: https://github.com/agda/cubical/pull/907

ecavallo:
Thanks, looks good now!

PR #981, opened 2023-02-02
Simpler bool

### LINKED ISSUE #962: Data.Bool.and and Data.Bool.or are pattern matching on second argument
Is there a reason to do that? If not, I'll be happy to make a PR defining

```agda
true  or _ = true
false or x = x
```

etc. which makes proofs about `or` and `and` shorter.

But if there is a good reason, we should add a comment why they are implemented in the way they are.

phadej:
Resolve #962


ecavallo (approved):
Looks good to me, thanks!

PR #994, opened 2023-03-03
Remove cong′, duplicate of congS from Prelude

ecavallo:
Doesn't look like it's used anywhere either.


ecavallo:
There were some, I just didn't see them because cong′ looks like cong' in my font...

felixwellen:
Checks on current master -> merging.

PR #997, opened 2023-03-16
Remove commutativity assumption to `bigOpSplit++`

jpoiret:
`bigOpSplit++` only needs associativity.


jpoiret:
The test failure seems to be related to a version mismatch between Cabal and GHC (Cabal is too old for GHC 9.6). The change is minimal so I don't expect any problems with this.

jpoiret:
I'm focusing on fixing CI for now so I need my main branch for something else, I'll resend a PR later with another branch on my side :)

Issue #998, opened 2023-03-24
Swapping the argument order in equational reasoning


maxsnew:
In porting a category solver from the 1lab (https://github.com/maxsnew/multi-poly-cats/tree/main/Cubical/Tactics/CategorySolver), I ran into an issue: the solvers can't work in equational reasoning proofs because they never have the right hand side of the goal inferred. For example this proof:
```
  ex : ∀ {A B C}(f : 𝓒 [ A , B ])(g : 𝓒 [ B , C ])(h : 𝓒 [ A , C ])
    → (f ⋆ (id ⋆ g)) ≡ h
    → f ⋆ g ≡ h ⋆ id
  ex f g h p =
    f ⋆ g ≡⟨ solveCat! 𝓒 ⟩
    (f ⋆ (id ⋆ g)) ≡⟨ p ⟩
    h ≡⟨ solveCat! 𝓒 ⟩
    h ⋆ id ∎
```
Doesn't succeed with the current definition of equational reasoning, in each case the call to `solveCat!` fails because the rhs is an unsolved metavariable.

Reed Mullanix showed me that the issue can be solved by redefining the equational-reasoning syntax to swap the order the arguments are passed:
```
≡⟨⟩-syntax : ∀ {ℓ} {A : Type ℓ} (x : A) {y z} → y ≡ z → x ≡ y → x ≡ z
≡⟨⟩-syntax x q p = p ∙ q

infixr 2 ≡⟨⟩-syntax

syntax ≡⟨⟩-syntax x q p = x ≡⟨ p ⟩ q
```
Agda's type checker works from left to right but the equational reasoning macro is right-nested, so swapping the arguments like this gives better type inference and can even lead to performance improvements. With this change the equational reasoning proof above works.

This change has already been done in the agda-stdlib three years ago (https://github.com/agda/agda-stdlib/pull/1045) and is used in the 1lab as well. It seems quite critical for allowing the use of solvers in practice.

felixwellen:
Wow - this is great! I ran into this problem once, but didn't know it has an easy fix like this. Does it have drawbacks? Does the use of ```syntax``` mess up the goals?

MatthewDaggitt:
The issues the stdlib has had with it: i) very verbose to rename the syntax, ii) much more non-obvious what to do when trying to import explicitly using `using`.

maxsnew:
What do you mean by it being verbose? Don't we just need to change a couple of definitions to be macros?

felixwellen:
[...]

PR #999, opened 2023-03-25
swap the argument order in equational reasoning

maxsnew:
Addressing https://github.com/agda/cubical/issues/998 .

Currently not building because the current solvers have some hacks in them related to equational reasoning that I haven't fixed because they might not be necessary after this change. It would be nice to have an example of how this helps with a solver. I have a category solver in a different repo that I'm currently using to verify that this helps.

In the stdlib they also changed other *-reasoning syntax. I'm not familiar enough with the stdlib to know where all of those would be. Additionally I haven't change `≡⟨⟩⟨⟩-syntax` or `_≡⟨_⟩≡⟨_⟩_` yet but they should be similarly adapted as well.

### DIFF
--- a/Cubical/Foundations/Prelude.agda
+++ b/Cubical/Foundations/Prelude.agda
@@ -28,7 +28,7 @@ open import Cubical.Core.Primitives public
 infixr 30 _∙_
 infixr 30 _∙₂_
 infix  3 _∎
-infixr 2 _≡⟨_⟩_ _≡⟨⟩_
+infixr 2 step-≡ _≡⟨⟩_
 infixr 2.5 _≡⟨_⟩≡⟨_⟩_
 infixl 4 _≡$_ _≡$S_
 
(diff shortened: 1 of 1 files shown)


felixwellen:
Here it is, just in case it doesn't show up automatically (not checked yet, but I guess it is good to merge into this PR):

https://github.com/maxsnew/cubical/pull/1

felixwellen:
Type checks on my machine and I actually could just substitute the worst ringsolver-macro with another one, because the RHS is available now in equational reasoning.

felixwellen:
As far as I can tell, that should be ready for merging - or do you want to change anything?

maxsnew:
Looks good to me. Main thing I would want to add is a test that demonstrate it works with equational reasoning

felixwellen:
Have you looked at the CommRingSolver/Examples?
That shows it works in simple cases und you can check it won't work in more complicated cases.
But this is due to a shortcoming of how deBruijn-indices are handled in the CommRingSolver.
I think it would be good to rewrite the latter using more existing code for the reflection interface, which is not neccessarily in the cubical-library. So I would say that is out of scope here...

maxsnew:
Oh I guess I missed those. LGTM then

felixwellen:
Okay -> merging.

PR #1000, opened 2023-03-26
Refactor CI actions

cmcmA20:
- split the job into three phases: cabal/ghc, agda, library
- cache all the intermediate binaries
- bump ghc and cabal version

Old action failing to populate caches on unsuccessful library tests was quite infuriating.
Example usage: https://github.com/cmcmA20/cubical-mini/actions/runs/4527022375
Documentation deploy doesn't work as intended but I'm not sure if I haven't botched setting it up in my repo.

@andreasabel FYI

### DIFF
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -32,82 +32,53 @@ on:
 ##
 ########################################################################
 
+########################################################################
+## SETTINGS
+##
+## AGDA_BRANCH picks the branch of Agda to use to build the library.
+## It doesn't really track the branch, so you have to drop caches to
+## get a new branch version if it changes. This will be fixed in the
+## next PR.
+########################################################################
+
 env:
-  GHC_VERSION: 8.6.5
-  CABAL_VERSION: 3.2.0.0
+  AGDA_BRANCH: v2.6.3
+  GHC_VERSION: 9.2.5
+  CABAL_VERSION: 3.6.2.0
   CABAL_INSTALL: cabal install --overwrite-policy=always --ghc-options='-O1 +RTS -M6G -RTS'
+  CACHE_PATHS: |
+    ~/.cabal/packages
+    ~/.cabal/store
+    ~/.cabal/bin
 
 jobs:
   test-cubical:
     runs-on: ubuntu-latest
     steps:
-
-      ########################################################################
-      ## SETTINGS
-      ##
-      ## AGDA_BRANCH picks the branch of Agda to use to build the library.
-      ########################################################################
-
-      - name: Initialise variables
-        run: |
-          # Pick Agda version
-          echo "AGDA_BRANCH=v2.6.3" >> $GITHUB_ENV
-
-
-      ########################################################################
-      ## CHECKOUT
-      ########################################################################
-
-      # By default github actions do not pull the repo
-      - name: Checkout cubical
-        uses: actions/checkout@v3
-
-      ########################################################################
-      ## CACHING
-      ########################################################################
+      - name: Install cabal
+        uses: haskell/actions/setup@v2
+        with:
+          ghc-version: ${{ env.GHC_VERSION }}
+          cabal-version: ${{ env.CABAL_VERSION }}
+          cabal-update: true
 
       # This caching step allows us to save a lot of building time by only
-      # downloading ghc and cabal, rebuilding Agda, and re-checking unchanged
-      # library files if absolutely necessary i.e. if we change either the
-      # version of Agda, ghc, or cabal that we want to use for the build.
-      - name: Cache cabal packages
-        uses: actions/cache@v3
-        id: cache-cabal
+      # rebuilding Agda, and re-checking unchanged library files if
+      # absolutely necessary i.e. if we change either the version of Agda,
+      # ghc, or cabal that we want to use for the build.
+      - name: Restore external dependencies cache
+        uses: actions/cache/restore@v3
+        id: cache-external-restore
         with:
-          path: |
-            ~/.cabal/packages
-            ~/.cabal/store
-            ~/.cabal/bin
+          path: ${{ env.CACHE_PATHS }}
           key: ${{ runner.os }}-${{ env.GHC_VERSION }}-${{ env.CABAL_VERSION }}-${{ env.AGDA_BRANCH }}
-
-      - name: Cache library build
-        uses: actions/cache@v3
-        id: cache-library
-        with:
-          path: ./_build
-          key: library-${{ runner.os }}-${{ env.GHC_VERSION }}-${{ env.CABAL_VERSION }}-${{ env.AGDA_BRANCH }}-${{ hashFiles('Cubical/**') }}
-          restore-keys: |
-            library-${{ runner.os }}-${{ env.GHC_VERSION }}-${{ env.CABAL_VERSION }}-${{ env.AGDA_BRANCH }}-
 
       ########################################################################
       ## INSTALLATION STEPS
       ########################################################################
 
-      - name: Install cabal
-        if: steps.cache-cabal.outputs.cache-hit != 'true'
-        uses: haskell/actions/setup@v2
-        with:
-          ghc-version: ${{ env.GHC_VERSION }}
-          cabal-version: ${{ env.CABAL_VERSION }}
-
-      - name: Put cabal programs in PATH
-        run: echo "~/.cabal/bin" >> $GITHUB_PATH
-
-      - name: Cabal update
-        run: cabal update
-
       - name: Download and install Agda from github
-        if: steps.cache-cabal.outputs.cache-hit != 'true'
+        if: steps.cache-external-restore.outputs.cache-hit != 'true'
         run: |
           git clone https://github.com/agda/agda --branch ${{ env.AGDA_BRANCH }} --depth=1
           cd agda
@@ -118,16 +89,44 @@ jobs:
           rm -rf agda
 
       - name: Download and install fix-whitespace
-        if: steps.cache-cabal.outputs.cache-hit != 'true'
+        if: steps.cache-external-restore.outputs.cache-hit != 'true'
         run: |
           git clone https://github.com/agda/fix-whitespace --depth=1
           cd fix-whitespace
           ${{ env.CABAL_INSTALL }} fix-whitespace.cabal
           cd ..
 
+      - name: Save external dependencies cache
+        if: steps.cache-external-restore.outputs.cache-hit != 'true'
+        uses: actions/cache/save@v3
+        id: cache-external-save
+        with:
+          path: ${{ env.CACHE_PATHS }}
+          key: ${{ steps.cache-external-restore.outputs.cache-primary-key }}
+
+      ########################################################################
+      ## CHECKOUT
+      ########################################################################
+
+      # By default github actions do not pull the repo
+      - name: Checkout cubical
+        uses: actions/checkout@v3
+
       ########################################################################
       ## TESTING
       ########################################################################
+
+      - name: Restore library cache
+        uses: actions/cache/restore@v3
+        id: cache-library-restore
+        with:
+          path: ./_build
+          key: library-${{ runner.os }}-${{ env.GHC_VERSION }}-${{ env.CABAL_VERSION }}-${{ env.AGDA_BRANCH }}-${{ hashFiles('Cubical/**', 'cubical.agda-lib', 'Everythings.hs') }}
+          restore-keys: |
+            library-${{ runner.os }}-${{ env.GHC_VERSION }}-${{ env.CABAL_VERSION }}-${{ env.AGDA_BRANCH }}-
+
+      - name: Put cabal programs in PATH
+        run: echo "~/.cabal/bin" >> $GITHUB_PATH
 
       - name: Test cubical
         run: |
@@ -135,6 +134,14 @@ jobs:
             AGDA_EXEC='~/.cabal/bin/agda -WnoUnsupportedIndexedMatch -W error' \
             FIX_WHITESPACE='~/.cabal/bin/fix-whitespace' \
             EVERYTHINGS='cabal run Everythings'
+
+      - name: Save library cache
+        if: steps.cache-library-restore.outputs.cache-hit != 'true'
+        uses: actions/cache/save@v3
+        id: cache-library-save
+        with:
+          path: ./_build
+          key: ${{ steps.cache-library-restore.outputs.cache-primary-key }}
 
       ########################################################################
       ## HTML GENERATION


andreasabel:
@cmcmA20 : Thanks for the PR (with the lucky number 1000!).

I looked at the code and it seems like you want to cache 3 things:
1. GHC and Cabal installation and Cabal packages.
2. Agda binary and dependencies of Agda.
3. The cubical library build.

1+2 share the same set of paths (`~/.ghcup` and `~/.cabal`), 3 uses a different path (`_build`).
Since 3 was already separate from 1+2, I only comment on the first 2 caches.

The key structure in terms of program versions you suggest is (leaving out `os` because it is fixed to `Linux`):
1. ghc-cabal
2. ghc-cabal-agda

Why do you separate steps 1 + 2?  You want to save the GHC installation in terms Agda fails to build?  Shouldn't you assume that all your external dependencies (GHC/Cabal/Agda) build fine and you only worry about building the cubical library in the CI?

Note that the cache use of your new workflow is huge: short of 2.5 GB.  With the total cache limit of 10GB, you could only accommodate master and 3 PRs.  Anything beyond this would bump caches and make new commits to PRs much slower.
Screenshot 2023-03-27 at 10 25 35
- One improvement thus would to fuse 1+2 into a single cache saving action.
- Whether caching the GHC installation and the cabal package index makes sense is a bit of a question per se.  Restoring the cache also amounts to downloading hundreds of MB and unpacking them.  Ideally installing GHC with ghcup should not do much more than that.  And downloading the cabal package index with `cabal update` shouldn't do much more than that.  And these are fetched from external locations where you do not have to pay with cache space.
- But caching GHC/Cabal was already part of the original CI.  Was this working correctly or did your changes (like also caching `~/.ghcup`) fix it?

Your original problem seemed to have been this one:
> Old action failing to populate caches on unsuccessful library tests was quite infuriating.

I agree this problem needs addressing.  
Maybe it can be fixed more conservatively by just adding a `actions/cache/save` step for the Agda installation before trying to build the cubical library?

cmcmA20:
>Shouldn't you assume that all your external dependencies (GHC/Cabal/Agda) build fine and you only worry about building the cubical library in the CI?

You're absolutely right! Regarding cache size, I was not aware of such a strict quota (10GB), my fault.

>But caching GHC/Cabal was already part of the original CI. Was this working correctly or did your changes (like also caching ~/.ghcup) fix it?

If you set a different cabal/ghc version (e.g. 3.6.2.0 + 9.2.5), you can observe non-idempotence of current workflow, it works correctly only if there was a cache miss. ghcup is the culprit, it fails to find its' config (because it's not cached) and defaults to a wrong ghc version. Seems like it's using cabal version to automatically infer ghc version.

>My suggestion would be to make a new PR with minimal changes to the current CI that achieve your goal (caching Agda before trying building the cubical library) and see if this already does the job (the bumps to GHC and Cabal versions should also be included, of course).

Thank you for guidance, Andreas, will ping you when it's done!

andreasabel:
> If you set a different cabal/ghc version (e.g. 3.6.2.0 + 9.2.5), you can observe non-idempotence of current workflow, it works correctly only if there was a cache miss. ghcup is the culprit, it fails to find its' config (because it's not cached) and defaults to a wrong ghc version. Seems like it's using cabal version to automatically infer ghc version.

I was suspecting something like that.  Thanks for fixing this!

andreasabel (approved):
Looks great!

cmcmA20:
Thank you! One last fix though: `cache` -> `cache/restore` in library section. I see it works fine but better safe than sorry.

.github/workflows/ci-ubuntu.yml
```diff
@@ -32,82 +32,58 @@ on:
...
 
-      - name: Put cabal programs in PATH
-        run: echo "~/.cabal/bin" >> $GITHUB_PATH
-
       - name: Cabal update
         run: cabal update
 
       - name: Download and install Agda from github
-        if: steps.cache-cabal.outputs.cache-hit != 'true'
+        if: steps.cache-external-restore.outputs.cache-hit != 'true'
```
andreasabel, line 86: Just a comment (not on your change, but on this CI design):
Since Agda is only rebuilt if the cache key changes, you could not have a movable `AGDA_BRANCH` identifier like `master`.
To facilitate the latter, one would first have to resolve `ADGA_BRANCH` to a SHA and then use this as cache key (component).
```sh
SHA=$(git rev-parse ${{ env.AGDA_BRANCH }})
```
Of course, this has the drawback that one has to checkout the Agda repo _always_ (and before cache restoration) just to do this resolution.

.github/workflows/ci-ubuntu.yml
```diff
@@ -32,82 +32,58 @@ on:
...
+########################################################################
+## SETTINGS
+##
+## AGDA_BRANCH picks the branch of Agda to use to build the library.
+########################################################################
+
 env:
-  GHC_VERSION: 8.6.5
-  CABAL_VERSION: 3.2.0.0
+  AGDA_BRANCH: v2.6.3
```
andreasabel, line 42: Suggestion: Add a comment that movable identifiers like `master` would not work here because they do not make adequate cache keys.
I.e., if I put `master` here, this CI would restore the Agda it last built from the cache rather than building Agda anew when `master` changes.

andreasabel (approved):
Suggestion: comment that `AGDA_BRANCH` cannot safely be a movable identifier such as `master`.

.github/workflows/ci-ubuntu.yml
cmcmA20: Maybe we should use GraphQL action to query the latest hash on `AGDA_BRANCH`?

andreasabel: If you want released versions of Agda, there is now https://github.com/wenkokke/setup-agda as the official Agda installer (should be moved to the `agda` organization at some point).

Re GraphQL: You mean https://github.com/octokit/graphql-action ?

andreasabel:
@ecavallo @felixwellen @mortberg : Who is in charge in CI here?  If you feel in charge, please assign yourself to this PR and handle the merge.  I'd say this PR is ready, but I don't want to overstep the owners...
@cmcmA20 : If you do not think this PR is ready, please communicate this (e.g. by stating it here or converting the PR to a draft).

cmcmA20:
@andreasabel I think this is only marginally better than a previous one, so let's wait for maintainers to say their word about possible directions regarding CI architecture. Until then, I'll experiment with Wen's setup/agda.

jpoiret:
Hi, I was also independently working on a similar fix on my end, so let me share my thoughts:
- There's no reason we should cache GHC and Cabal, they can be installed quite fast using the haskell setup action. However, they need to always be set-up, and not just on cache miss as it's being done right now, since they aren't in the cache either. This would reduce cache use even more as well.
- Given cubical's development model, I'm not sure the `AGDA_BRANCH` ref moving around should be an immediate problem (and it's not a regression with current behavior so it shouldn't be a blocker). But, if cubical wants to test against master at some point, the agda-setup action won't help either (from what I understand). This could be done in a future PR.

Regarding the documentation not building, you'd need to set "Read and write permissions" for workflows under Settings → Actions. One other solution that would work out of the box would be to add the write permission to the action in the yaml file, but I'm not a fan of this either since it grants the rights to the whole job. Some refactoring into multiple jobs could probably be done as well, or use the new Beta for deploying pages directly via actions.

My commits were a bit more conservative, see https://github.com/agda/cubical/compare/master...jpoiret:cubical:ci-clean-up
The goal is to just fix the failing CI builds for now, the rest could come later.

.github/workflows/ci-ubuntu.yml
cmcmA20: Yes I meant this action, but I've missed the elephant: GitHub also has a classic REST API interface.

cmcmA20:
@jpoiret: Hi! Big thanks for the doc tip.

>There's no reason we should cache GHC and Cabal, they can be installed quite fast

You'r right, tests show that the difference is negligible so it's a total waste of cache space. However, the CI flow can be decomposed into these jobs: `(ghc and cabal) <+> (Agda <|> fix-whitespace) <+> test library <+> docs`. And then the third stage nontrivially depending on the binaries from the first one really bites. What would be a good way to handle this?

>But, if cubical wants to test against master at some point, the agda-setup action won't help either

I hadn't much time to test this but skimming through the README gave me the same impression. Anyway I've just implemented this feature.

andreasabel:
If you are thinking about more revolutionary changes of this CI, please also consider using binary installs of `fix-whitespace`  (rather than building from source).  Here is an action I wrote to this end: https://github.com/andreasabel/fix-whitespace-action

andreasabel:
@cmcmA20 wrote:
> Documentation deploy doesn't work as intended

According to my own experiments, deployment only works if the CI is run on `master`, because one usually does not want update documentation while WIP.

Yesterday I set up a web page deployment using the new fashion (in beta stage), using `actions/upload-pages-artifact@v1` and `actions/deploy-pages@v2`: https://github.com/teach-plt/www/blob/79b4094c422fd8d89c0bcdb83304d55947cfdf46/.github/workflows/deploy.yml#L75-L105
This requires to change Settings/Pages to "using Github Actions" in the repo, though.

andreasabel:
@cmcmA20 
> @andreasabel I think this is only marginally better than a previous one, so let's wait for maintainers to say their word about possible directions regarding CI architecture.

I thought the cache improvement (status as of 3 days ago) would be an incremental improvement worth merging, until a more thorough overhaul is in place.  (Atm, though, the workflow is failing on `octokit/request-action`.)

cmcmA20:
@andreasabel
>workflow is failing

It's exactly what is says on the tin, "v.2.6.3" is not a valid branch name, though it is a valid _tag_ name, sorry for overlooking that, I've pushed a fix. This raises a question: do we want to leave it as it is (the original action config was actually misleading about the nature of these things) or do we want to support both branches and tags?

For now I consider this change ready for merge but your doc deploy actions would be a great addition, so all the CI tweaks are grouped under the same PR. Sorry for dragging this, I was under the impression that doing more than was originally planned is fine. If you have specific preferences about PR decomposition, please let me know, so we can split it into a few fine-grained ones.

andreasabel:
> It's exactly what is says on the tin, "v.2.6.3" is not a valid branch name, though it is a valid _tag_ name, sorry for overlooking that, I've pushed a fix. This raises a question: do we want to leave it as it is (the original action config was actually misleading about the nature of these things) or do we want to support both branches and tags?

I think so.  Tags are fixed targets, branches are movable targets.  There is no guarantee that `release-2.6.3` will forever match `v2.6.3`.  In fact it even does not so atm.  (And they might deviate further in the future if we push to `release-2.6.3` e.g. to prepare 2.6.3.1, who knows?)

> For now I consider this change ready for merge but your fix-whitspace and doc deploy actions would be a great addition, so all the CI tweaks are grouped under the same PR. Sorry for dragging this, I was under the impression that doing more than was originally planned is fine. If you have specific preferences about PR decomposition, please let me know, so we can split it into a few fine-grained ones.

Well, in terms of PR strategy, I think one should _separate concerns_.  
- The first issue was about improving caching.  A fixed point was reached when caching the Agda build before trying to compile the cubical library was working.  At this point there is a meaningful commit which achieves one thing, which can be merged.
- The next thing is about resolving `AGDA_BRANCH` to a SHA rather than using it verbally.  You started to work on this (which is great), but now we are again in the middle of something and cannot merge.  Also, this is somewhat orthogonal to the cache improvement, it is more about getting the right version of Agda, which is another improvement.
- Then, there will be an improvement of `fix-whitespace` etc.

All these are separate concerns and should go into separate commits, so then in case on of these points actually wasn't a strict improvement, it can be reverted easily without affecting the others.

If you agree, then let's work like this in small steps, each of which solves one topic and can be reviewed easily.   (Sorry, I didn't communicate this clearly before.)

cmcmA20:
Thank you for the explanation! I will revert the changes to the first stage where we got the caching right, so we can reap the benefits. Everything else will go into a new folow-up PR.

.github/workflows/ci-ubuntu.yml
```diff
@@ -32,82 +32,53 @@ on:
...
-      ########################################################################
+      - name: Install cabal
+        uses: haskell/actions/setup@v2
+        with:
+          ghc-version: ${{ env.GHC_VERSION }}
+          cabal-version: ${{ env.CABAL_VERSION }}
+          cabal-update: true
 
       # This caching step allows us to save a lot of building time by only
       # downloading ghc and cabal, rebuilding Agda, and re-checking unchanged
```
jpoiret, line 66: Minor nitpick, it would help to remove the "downloading ghc and cabal" part. Not a blocker though IMO.

jpoiret (approved):
LGTM, apart from the minor nitpick that shouldn't be a blocker!

.github/workflows/ci-ubuntu.yml
```diff
@@ -118,38 +89,75 @@ jobs:
...
+          path: ./_build
+          key: library-${{ runner.os }}-${{ env.GHC_VERSION }}-${{ env.CABAL_VERSION }}-${{ env.AGDA_BRANCH }}-${{ hashFiles('Cubical/**') }}
+          restore-keys: |
+            library-${{ runner.os }}-${{ env.GHC_VERSION }}-${{ env.CABAL_VERSION }}-${{ env.AGDA_BRANCH }}-
+
+      - name: Put cabal programs in PATH
+        run: echo "~/.cabal/bin" >> $GITHUB_PATH
+
       - name: Test cubical
+        if: steps.cache-library-restore.outputs.cache-hit != 'true'
```
andreasabel, line 132: This is a bit of an unconventional setup, I'd say, to skip the main step of a job.
Currently only changes under `Cubical/` will trigger this step, judging from the cache key.
What about changes in `cubical.agda-lib`, or `Everythings.hs` in the root directory?
What about new files added to the root?

Could you say a bit more about scenarios where it would be the right thing to skip the main step?

andreasabel (changes requested):
I was about to merge this, but then I noticed the last CI run hadn't actually checked the cubical library (skipped step)...

.github/workflows/ci-ubuntu.yml
cmcmA20: I've miseed this case because I've been mainly working on files under `Cubical/` and considered everything else more or less fixed, my apologies!

Skipping any step is desirable if it would be a recomputation of already known data. If it's correctly implemented of course :-)

cmcmA20: @andreasabel
Adding new files in the root directory will incur additional costs as you'll need to also modifiy the CI config. Not caching will waste 50 minutes of compute on every re-run. Human time may be more valuable than machine time in this case, but changing project structure requires human attention anyway.
What's your decision on this?

andreasabel: @cmcmA20: I was hoping for a more concrete answer to my question, like enumerating some concrete scenarios, so that one can better judge what is the right approach here:

> Could you say a bit more about scenarios when it would be the right thing to skip the main step?

cmcmA20: @andreasabel:
Forgive my opacity, I think skipping the main step on cache hit is natural and right thing to do. Here I assume that building project on CI is equivalent to building it locally with `make`. If both Agda and project files are unchanged, and you haven't explicitly dropped the cache (like running `make clean`), rebuilding/redeploying anything should be considered a bug in the workflow or the build system.

Maybe I don't see something obvious, please tell me where's the error in reasoning.

andreasabel: Thanks for your reply.  I am not convinced yet:

You currently have in the cache key `hashFiles('Cubical/**', 'cubical.agda-lib', 'Everythings.hs')`.
So if someone changes the makefile and none of these other files and opens a PR, CI would skip the step one wants to test (provided caches have not been deleted).

I think If you insist that the main step be skipped if nothing changed, you have to define the _nothing_ correctly!  Also, it add some subtlety, and needs some extra maintenance (who is going to do this?).
So, I would advise against it.

>  I think skipping the main step on cache hit is natural and right thing to do.

I would prefer to also hear a concrete scenario.  
The scenario I can think of is when a PR just changes a `.md` file like `CONTRIBUTING`, or the `.gitignore` file.  But aren't these cases rare enough to ignore them?
Do you have other scenarios?

cmcmA20: Thank you, Andreas, for uncovering the weak points, I really appreciate it.
This design provides modest benefits, but the effort to get it right seems unreasonably high.

>CONTRIBUTING, or the .gitignore file. But aren't these cases rare enough to ignore them? Do you have other scenarios?
I had no scenarios in mind besides documentation changes.

.github/workflows/ci-ubuntu.yml
```diff
@@ -118,24 +89,60 @@ jobs:
...
+
       - name: Test cubical
         run: |
           make test \
             AGDA_EXEC='~/.cabal/bin/agda -WnoUnsupportedIndexedMatch -W error' \
             FIX_WHITESPACE='~/.cabal/bin/fix-whitespace' \
             EVERYTHINGS='cabal run Everythings'
 
+      - name: Save library cache
+        if: steps.cache-library-restore.outputs.cache-hit != 'true'
```
andreasabel, line 139: Since we have `restore-keys`, it makes sense to save the cache also if it was restored from a fall-back key.  It will be saved under the primary key then and refreshes the cache to be up to date.
https://github.com/haskell/actions/blob/87b3442d9877a4ab61f04a4e3d0da2e14be5d51e/setup/README.md?plain=1#L169-L171
```yaml
        if: ${{ !steps.cache.outputs.cache-hit
          || steps.cache.outputs.cache-primary-key != steps.cache.outputs.cache-matched-key }}
```
What do you think?

andreasabel (approved):
Thanks for the good work!

I think this can be merged.  Optionally, refine the "save library cache" conditional (see my suggestion) before mergin.

.github/workflows/ci-ubuntu.yml
cmcmA20: >When key exactly matches an existing cache, it's called a cache hit, and the action restores the cached files to the path directory.
When key doesn't match an existing cache, it's called a cache miss, and a new cache is automatically created if the job completes successfully.

Looks like only exact match count as cache hit, partial match on `restore-keys` will try to save the cache anyway. Hold on, I'll test this and return with a definite answer.

cmcmA20: https://github.com/actions/cache/blob/main/restore/README.md#outputs
>`cache-hit` will be set to true only when cache hit occurs for the exact key match. For a partial key match via restore-keys or a cache miss, it will be set to false.

andreasabel:
Thanks for this journey, where we all learned new things.

PR #1001, opened 2023-03-31
Simplify CartesianKanOps/FunExtEquiv using "interpolation" on I

pi3js2:
[note: in the latest, the `erp` operator has been renamed to `interpolateI` and is defined in `Cubical.Core.Interpolate`]

Here I added an "interpolation" operator on the interval, in Cubical.Foundations.Erp:

```agda
erp : I → I → I → I
erp t i j = (~ t ∧ i) ∨ (t ∧ j) ∨ (i ∧ j)
```

I have found the `erp` operator to be very useful generally, so I think it is worth exposing in the library.

Using `erp`, I simplified CartesianKanOps and FunExtEquiv.

I am not really sure whether the changes to CartesianKanOps and FunExtEquiv are worth doing. The new versions seem simpler, but I don't have any real examples where they improve performance or make proofs easier, and I guess this might break some user code.

I have also added `funExtNonDep` for paths in non-dependent function types (but where both domain and codomain depend on the interval) which can be even simpler than the new `funExtDep`. Maybe there is a better name than `funExtNonDep`?


kangrongji:
Your `erp` is the first "not so trivial" map between de Morgan cubes  `I³ → I`, and I think it's the only one in this 3-dimensional case, up to certain renaming or involution. It's already used in many parts of the library to construct higher degenerate cubes like [here](https://github.com/agda/cubical/blob/f3cf89decfb494365147e65f549057d1525225ec/Cubical/Data/Int/MoreInts/QuoInt/Properties.agda#L79) or [here](https://github.com/agda/cubical/blob/f3cf89decfb494365147e65f549057d1525225ec/Cubical/Homotopy/BlakersMassey.agda#L281). I'm not sure if give it a name would make things more clear...

phijor:
I just had a look, this seems very useful! It's not too hard to show that `coei→j` as defined in terms of `erp` is an equivalence (https://github.com/phijor/cubical/commit/cd210450eaa77142d7405ff5f007e9ba9482eb73). Since everything computes nicely, this turns into proofs of `isEquiv (transport p)` and `transpEquiv : ∀ i → p i ≃ B` (cf. https://github.com/agda/cubical/pull/1069) for free.

Regarding the name: Why not call it `interp`? This is in line with the pattern of primitive operations ending in _p_, like in `transp` and `comp`.

mortberg:
Indeed, calling it `interp` or `interpolate` sounds good to me. Otherwise I think this could be merged? @ecavallo Any opinions?

phijor:
> It could just go in CartesianKanOps, even private. Or maybe it doesn't even need a name, like @kangrongji said. It is easy to define yourself or even (in simple cases) use directly when you want it.
> 
> I created Erp.agda:
> 
>     * to advertise to people who don't know about this thing, like me not very long ago...
> 
>     * to provide a place to demonstrate the nice equations it satisfies. I have more equations to add, by the way.

I think it very valuable to have these things written down somewhere, and this library seems like a good place. Consider that cubical type theory is a niche thing that is (for me at least) hard to learn, so any resource that helps understand it better is appreciated. Even (or in particular) if it concerns "trivialities".

Personally, I also find the ad-hoc interval expressions hard to read; there's not enough context to see at a glance what their _intent_ is. I think having descriptive names, such as `squeeze`, `spread`, `coe` and `interpolate` helps readability of low-level proofs a lot.

> About the name, I just like short names, and I like to do nested erps.
> 🤷 I am happy to rename it in my personal files.

Agree, short names are nice when it's unambiguous what they mean. Agda supports `renaming` on import, so

```agda
import Cubical.Foundations.Interpolate renaming (interpolate to erp) public
```

is something I'd put in my personal prelude. For library use, I'd keep a more descriptive name though and use an alias whenever it becomes bothersome.

> By the way, @kangrongji, if you know about any other interesting "not so trivial" maps, e.g. in higher dimension, I would love to hear about them. 😄

I haven't quite figured it out yet, but are there 3D-analouges of `erp` and `eqI`? The equivalence I built in https://github.com/phijor/cubical/commit/cd210450eaa77142d7405ff5f007e9ba9482eb73 doesn't reduce to `idIsEquiv` on `(i = i0 ∧ j = i0)` or `(i = i1 ∧ j = i1)`, but I think it could with slightly better interpolation.

ecavallo:
> Indeed, calling it `interp` or `interpolate` sounds good to me. Otherwise I think this could be merged? @ecavallo Any opinions?

Looks good to me.

kangrongji:
That's cool. What about using `intp`? not easy to pronounce though.

kangrongji:
Btw, a bit trivial but worth to mention, this operation can interpolate between any `u v : I^n -> I` by writing `(~ t ∧ u) ∨ (t ∧ v) ∨ (u ∧ v)`. More generally apply to each coordinates it can interpolate any maps $I^m -> I^n$.

Cubical/Foundations/CartesianKanOps.agda
```diff
@@ -5,6 +5,7 @@ module Cubical.Foundations.CartesianKanOps where
 open import Cubical.Foundations.Prelude
 open import Cubical.Foundations.GroupoidLaws
 open import Cubical.Foundations.Transport
```
phijor, line 7: I think with your changes, these imports are not needed anymore:

```suggestion
```

mortberg: I guess I can commit this and we see if the CI is happy

mortberg:
> That's cool. What about using `intp`? not easy to pronounce though.

This might be a reasonable compromise? What I don't like about `interp` is that my brain will complete it to `interpret`, not `interpolate`...

kangrongji:
Other suggestion: I think `mid` or `midst` might be more informative.

phijor:
I just remembered this PR... It looks like last time, we were all busy bike-shedding names for `erp`. I'd like to have `funExtNonDep` and friends in the library though. Could this be rebased and merged?

mortberg (approved):
Looks good to me!

mortberg:
Oops, CI unhappy... Please fix

Cubical/Foundations/CartesianKanOps.agda
```diff
@@ -3,8 +3,7 @@
 module Cubical.Foundations.CartesianKanOps where
 
 open import Cubical.Foundations.Prelude
-open import Cubical.Foundations.GroupoidLaws
-open import Cubical.Foundations.Transport
+open import Cubical.Foundations.Erp
```
phijor, line 6: CI is failing because it cannot find the removed module, delete the import:

```suggestion
```

Cubical/Foundations/CartesianKanOps.agda
```diff
@@ -45,20 +44,30 @@ coei0→0 A a = refl
...
+  -- and more...
+  interpolateI : I → I → I → I
+  interpolateI t i j = (~ t ∧ i) ∨ (t ∧ j) ∨ (i ∧ j)
+
+  -- "Equality" on the interval, chosen for the next definition
+  -- (coei→j), because interpolateI k i j is constant in k on
+  -- eqI i j. Note that eqI i i is not i1 but i ∨ ~ i.
+  equalI : I → I → I
+  equalI i j = (i ∧ j) ∨ (~ i ∧ ~ j)
+
```
phijor, line 63: Why is this marked private? Isn't the point that `interpolateI` is a helpful combinator?

In any case, `interpolateI` is definitely the best name so far.

phijor (changes requested):

Cubical/Foundations/CartesianKanOps.agda
```diff
@@ -3,8 +3,7 @@
 module Cubical.Foundations.CartesianKanOps where
 
 open import Cubical.Foundations.Prelude
```
phijor, line 5: Another observation: This module has no dependencies on the prelude, except for `refl`:

```suggestion
open import Cubical.Core.Everything

private
  refl : ∀ {ℓ} {A : Type ℓ} {x : A} → x ≡ x
  refl {x = x} _ = x
```

So in theory, it could go into `Cubical.Core.*` or be re-exported from `Cubical.Foundations.Prelude`. In any case, it is now possible to use the Kan operations inside of `Cubical.Foundations.*` without introducing dependency cycles. This would perhaps clean up a few ad-hoc uses of `coe0→i` and friends, such as in [isEquivTransp](https://agda.github.io/cubical/Cubical.Foundations.Equiv.Base.html#isEquivTransp).

Cubical/Foundations/CartesianKanOps.agda
phijor: I have no authority here, but... I'd make it public. It seems like a useful one-liner to reuse and there's worse definitions that are public. I'd hate to redefine myself :smiley:

Cubical/Foundations/CartesianKanOps.agda
```diff
@@ -2,9 +2,12 @@
 {-# OPTIONS --safe #-}
 module Cubical.Foundations.CartesianKanOps where
 
-open import Cubical.Foundations.Prelude
-open import Cubical.Foundations.GroupoidLaws
-open import Cubical.Foundations.Transport
+open import Cubical.Core.Everything
+open import Cubical.Core.Interpolate
```
mortberg, line 6: Why import both Everything and the separate file? Wouldn't it make more sense to put interpolate in Prelude and then you don't need to make a local definition of refl?

mortberg (changes requested):
I don't think this should go in the Core. The idea of the Core is to expose builtin nicer, but the rest of the library should rely on what is in the Prelude.

mortberg (approved):
Looks good to me, merging if CI is happy

mortberg:
The CI was finally happy so I merged. Thanks for the contribution!

PR #1007, opened 2023-06-05
Localization algebra

jpoiret:
Hi,

This is an attempt to define the localization of an algebra at an arbitrary multiplicatively closed subset of it, compared with the current localization at a multiplicatively closed subset of the base ring. I've had many attempts, trying to use CT and the fact that algebras over a base ring are equivalent to rings under a base ring, but the definitional behavior was really hard to predict and things just didn't compute well. I chose instead to just extend the existing construction in the most straightforward way. I'm still trying to formulate a recursion principle using the universal property, so that we're able to prove things about the localization without resorting to its actual construction.

This depends on a rebased version of #931.

LMKWYT

### DIFF
--- a/Cubical/Algebra/CommAlgebra/LocalisationAlgebra.agda
+++ b/Cubical/Algebra/CommAlgebra/LocalisationAlgebra.agda
@@ -23,6 +23,7 @@ import      Cubical.HITs.PropositionalTruncation as PropTrunc
 
 open import Cubical.Algebra.Algebra
 open import Cubical.Algebra.CommAlgebra
+open import Cubical.Algebra.CommAlgebra.Subalgebra
 open import Cubical.Algebra.CommRing as CommRing hiding (_ˣ;module Units)
 open import Cubical.Algebra.CommRing.Localisation using (isMultClosedSubset)
 open import Cubical.Algebra.Ring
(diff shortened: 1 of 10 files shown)


Cubical/Algebra/CommAlgebra/LocalisationAlgebra.agda
```diff
@@ -0,0 +1,254 @@
...
+extending the ring construction mechanically to using the fact the category of
+algebras is a coslice category and exactly re-using the localization of rings.
+Here, we chose the former, because after trying out the latter for some time, it
+turned out that its computational behavior wasn't as nice as expected, and it
+made Agda struggle a lot.
+-}
+
+{-# OPTIONS --safe #-}
+
+open import Cubical.Core.Everything
```
felixwellen, line 13: Probably not needed.

Cubical/Algebra/CommAlgebra/LocalisationAlgebra.agda
```diff
@@ -0,0 +1,254 @@
...
+
+    _⁻¹ : (a : ⟨ A ⟩) → ⦃ a ∈ _ˣ ⦄ → ⟨ A ⟩
+    _⁻¹ a ⦃ a-inv ⦄ = fst a-inv
+
+  open Units
+
+  open isMultClosedSubset SMultClosedSubset
+  private Aᵣ = CommAlgebra→CommRing A
+
+  private module S⁻¹ADef where
```
felixwellen, line 60: Maybe a more descriptive name.

felixwellen: Maybe "AlgebraStructureOnLocalization"

Cubical/Algebra/CommAlgebra/LocalisationAlgebra.agda
```diff
@@ -0,0 +1,254 @@
...
+  S⁻¹AAsCommAlgebra : CommAlgebra R ℓAlg
+  S⁻¹AAsCommAlgebra = let open S⁻¹ADef
+    in commAlgebraFromCommRing {R = R} S⁻¹Aᵣ _⋆_ ·Assoc⋆ ⋆DistR+ ⋆DistL+ ⋆IdL ⋆AssocL
+
+  S⁻¹AAsCommAlgebra→CommRing : CommAlgebra→CommRing S⁻¹AAsCommAlgebra ≡ RLoc.S⁻¹RAsCommRing
+  S⁻¹AAsCommAlgebra→CommRing = let open S⁻¹ADef
+    in commAlgebraFromCommRing→CommRing {R = R} S⁻¹Aᵣ _⋆_ ·Assoc⋆ ⋆DistR+
+                                        ⋆DistL+ ⋆IdL ⋆AssocL
+
+  module S⁻¹AUnversalProp where
```
felixwellen, line 133: Typo "Universal"

felixwellen (changes requested):
Discussed in person.

felixwellen:
Checking the library took 47min in the CI, we should investigate.

felixwellen:
Both master and this branch take 27min on my machine, so it should be fine.

Cubical/Algebra/CommAlgebra/LocalisationAlgebra.agda
jpoiret: I've removed all of those unnecessary imports.

felixwellen (changes requested):
LGTM - just a request for comment.

PR #1008, opened 2023-06-05
Categorical bits and pieces

jpoiret:
Hi,

While attempting a categorical formulation of #1007 that I abandoned in the end, I collected a lot of simple results and modifications.

Noteworthy changes are 2491f5fd4a58ce99f6fdf93c4fd45e08e7b95950, a change to the definition of being an equivalence, which previously was not an hProp. I fixed all uses of it in the library to be compatible with the new definition. I also separated triangle identities from the definition of an adjunction in c55988cfb8a9c66526068fb134bf2547c5c93ef2, so that I could more easily add adjoint equivalences in c9cb16baa0ba94a140b4706a4d418bbe4fe31a21.

From what I can tell this is independent to #988, and shouldn't interfere with it.

LMKWYT


maxsnew:
This is independent of what's in our current PR and should only have minor conflicts with our followup. Thanks for the heads up it shouldn't be a big issue for us to work around.

Cubical/Categories/Presheaf/Constructions.agda
```diff
@@ -0,0 +1,25 @@
...
+module Cubical.Categories.Presheaf.Constructions where
+
+open Functor
+
+module Single
+  {ℓC ℓ'C ℓSet : Level}
+  {C : Category ℓC ℓ'C}
+  where
+
+  constantPresheaf : hSet ℓSet → Presheaf C ℓSet
```
maxsnew, line 19: There's already https://github.com/agda/cubical/blob/master/Cubical/Categories/Functors/Constant.agda

jpoiret: Oops, didn't think of looking that way, thanks!

jpoiret:
I just removed the stuff about subpresheaves and representable presheaves, as there's going to be something better with #988 anyways (I have to make time to send some changes to it).

felixwellen:
Can you rebase?

felixwellen (approved):
Thanks for the contribution!
I commented only on one small detail and will merge anyway if you don't have the time to fix it. 
One question though: Are there any candidates we should consider as a definition of ```isEquivalence```, which are naturally a proposition?

felixwellen:
Thanks! Maybe you can put what you know into a comment at ```isEquivalence```, I guess it is natural to wonder about that when reading the definition.

felixwellen:
With "what you know" I mean that we don't know a notion which works analogous to "isEquiv" for not necessarily univalent categories.

felixwellen:
Thanks!
I'll merge.

PR #1011, opened 2023-06-12
Add more results for inductively defined equality type

JonasHoefer:
#1005 wants to remove the swan identity type because `Agda.Builtin.Equality` now has the same computational behavior. This pull request extends the `Cubical.Data.Equality` module, which contains some basic results for this equality type. The main changes are the following.
1. A proof of univalence which is behaves like "Book HoTT", that is, `pr₁ (idToEquiv p) x = transport id p x` holds, where `transport` is defined in terms of pattern matching. Thus, the function which is an equivalence is the same one as in the book (see Lemma 2.10.1 and Axiom 2.10.3).
2. The interface for `S¹` is extended with the eliminator and its corresponding β and η rules.
3. A more general result (similar to the one [here](https://www.cse.chalmers.se/~nad/listings/equality/Equality.Path.Isomorphisms.html)) is proven, which simplifies lifting HITs to the inductively defined equality type.


Cubical/Data/Equality.agda
```diff
@@ -1,327 +1,23 @@
...
--- Functions for going between the various definitions. This could
--- also be achieved by making lines in the universe and transporting
--- back and forth along them.
-
-fiberPathToFiber : {f : A → B} {y : B} → fiberPath f y → fiber f y
-fiberPathToFiber (x , p) = (x , pathToEq p)
-
-fiberToFiberPath : {f : A → B} {y : B} → fiber f y → fiberPath f y
-fiberToFiberPath (x , p) = (x , eqToPath p)
+- Basic therory for _≡_ (J, ap, transport, ...)
```
felixwellen, line 7: Typo

Cubical/Data/Equality/Conversion.agda
```diff
@@ -0,0 +1,314 @@
...
+        → isContrPath A → isContr B
+helper1 f g h (x , p) =
+  (f x , λ y → pathToEq (λ i → hcomp (λ j → λ { (i = i0) → f x
+                                              ; (i = i1) → h y j })
+                                     (f (p (g y) i))))
+
+helper2 : {A B : Type ℓ} (f : A → B) (g : B → A) (h : ∀ y → Path _ (g (f y)) y)
+        → isContr B → isContrPath A
+helper2 {A = A} f g h (x , p) = (g x , λ y → eqToPath (ap g (p (f y)) ∙ pathToEq (h y)))
+
```
felixwellen, line 163: These definitions should not be public.

Cubical/Data/Equality/FunctionExtensionality.agda
```diff
@@ -0,0 +1,53 @@
+{- Function Extensionality in terms of _≡_
+
+- Function Extensionality stated in terms of the inductively defined equality type
```
felixwellen, line 3: Merge these two comments?

Cubical/Data/Equality/PropositionalTruncation.agda
```diff
@@ -0,0 +1,33 @@
+{- Propositional truncation using ≡ with paths under the hood -}
```
felixwellen, line 1: Maybe "Propositional truncation for the inductively defined equality type _≡_, with paths under the hood"?

Cubical/Data/Equality/Base.agda
```diff
@@ -0,0 +1,114 @@
+{-
+
+This module converts between the path types and the inductively
```
felixwellen, line 3: Isn't that in "Conversion" now?

PR #1012, opened 2023-06-17
Lindenbaum-Tarski algebra

croos90:
Lindenbaum-Tarski algebra of propostional logic.

- Definition of provability
- Properties of propositional logic
- Biprovability relation
- Definition of Lindenbaum-Tarski algebra
- Lindenbaum-Tarski algebra is complemented distributive lattice, hence Boolean

### DIFF
--- a/Cubical/Algebra/LindenbaumTarski.agda
+++ b/Cubical/Algebra/LindenbaumTarski.agda
@@ -1,4 +1,5 @@
-{-# OPTIONS --cubical --safe #-}
+{-# OPTIONS --safe #-}
+
 module Cubical.LindenbaumTarski where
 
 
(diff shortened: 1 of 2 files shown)


Cubical/Algebra/LindenbaumTarski.agda
```diff
@@ -0,0 +1,391 @@
+{-# OPTIONS --safe #-}
```
mortberg, line 1: Maybe add a short comment in the top of the file saying what it contains?

mortberg (approved):
Apart from a small comment I think this looks good. 

One could maybe split this into multiple files and move some stuff to other parts, like the definition of propositional formulas and the proof system... On the other hand, the logical system is somewhat tailored for the construction of the LT algebra.  What does @ecavallo  and @felixwellen think?

felixwellen:
I think it's fine like it is now (agreeing that a comment explaining the file contents would be great). 

We can think about splitting off the logic-parts in our future refactoring that finally puts the logic related things in the library at a place where they are not completely unexpected ;-)

Cubical/Algebra/LindenbaumTarski.agda
croos90: Valid point!

Cubical/Algebra/LindenbaumTarski.agda
```diff
@@ -0,0 +1,403 @@
+{-
+
+This file contains:
+
+  - Classical propositional logic and provability
+  - Properties of classical propositional logic
+  - Equivalence relation on formulas in terms of provability
+  - Definition of Lindenbaum-Tarski algebra
+  - Proof the Lindenbaum-Tarski algebra is a Boolean algebra
```
mortberg, line 9:
```suggestion
  - Proof that the Lindenbaum-Tarski algebra is a Boolean algebra
```

Cubical/Algebra/LindenbaumTarski.agda
mortberg: Great!

mortberg (approved):
Merging as soon CI is finished checking it

Issue #1018, opened 2023-06-30
Note licence exceptions


felixwellen:
We already have or will soon have files which are essentially copied from the 1lab. Since the 1lab uses the AGPL 3 licence and we use an MIT licence, we should note the exceptions in our licence file. Maybe like this:
"There are a couple of files, which are licenced under more restrictive licences. Those files are marked with the corresponding SPDX identifier." 
I do not know if this is how it should be done, if you know, please speak up.
See the discussion here: #988.

felixwellen:
I have not found and not heard of anything against this way of dealing with it, so I'll implement it.

felixwellen:
The solution was that the authors of the corresponding parts of the 1lab agreed to relicense under some BSD license. The license exception is noted now.

PR #1021, opened 2023-07-05
#1018: Note licence exceptions

felixwellen:


jpoiret:
(IANAL) Thinking about it a bit more, I don't think you can license the whole project MIT if it includes GPL'd code that is then used by the MIT code, since it is infective.

Also, [1] might be a good resource, it's an FSFE initiative with clear standards.

[1] https://reuse.software/

felixwellen:
What do you mean with 'infective'? Are you saying if I use something under the GPL in my source code, then my source code also needs to adopt the GPL principles? 
Also, note, that with this change, the project wouldn't be licensed under the MIT license anymore. Instead, it would have a license, which says some files are MIT, some something else. I wouldn't call that 'licensed under MIT'.

jpoiret:
> What do you mean with 'infective'? Are you saying if I use something under the GPL in my source code, then my source code also needs to adopt the GPL principles? Also, note, that with this change, the project wouldn't be licensed under the MIT license anymore. Instead, it would have a license, which says some files are MIT, some something else. I wouldn't call that 'licensed under MIT'.
(IANAL) While you can license some files under a license and some others using another one, you still have to respect the contents of both licenses, but here, the GPL explicitely covers modified versions of the original program, including things that are "based" on it.

```quote
To “modify” a work means to copy from or adapt all or part of the work in a fashion requiring copyright permission, other than the making of an exact copy. The resulting work is called a “modified version” of the earlier work or a work “based on” the earlier work.

A “covered work” means either the unmodified Program or a work based on the Program.
```
Then, the provision of section 5 applies, by which
```quote
c) You must license the entire work, as a whole, under this License to anyone who comes into possession of a copy. This License will therefore apply, along with any applicable section 7 additional terms, to the whole of the work, and all its parts, regardless of how they are packaged. This License gives no permission to license the work in any other way, but it does not invalidate such permission if you have separately received it.
```

You could say this is the whole point of the GPL, ie. to force things that use GPL'd programs to themselves be free software. The MIT license doesn't force this, it is not "infective".

felixwellen:
Adapted the text to using BSD-licensed files, which we can actually do.

PR #1026, opened 2023-07-17
Add opcartesian fibrations

jpoiret:
Hi everyone,

This adds reindexing of displayed categories, the basic definitions of opcartesian fibrations, as well as discrete opfibrations, some nice properties of opcartesian morphisms and the construction of the substitution functor. I stopped short of proving the pseudo-functorial laws of the latter mostly because I didn't care too much.

As opposed to the 1lab which reasons with displayed morphisms using regular paths and transports, I thought we could try having reasoning combinators for the `PathP` version, so I went with that and added the two abominations that are `eq1 ≡[ p ∙ q ]∙ eq2` and `eq1 ≡[ p ⋆ q ]⋆ eq2` which are respectively displayed versions of vertical and horizontal compositions of paths. This lets us input mostly "regular" proofs, with the caveat that Agda can't always infer p and q, and when it can it might hinder performance, so I chose to leave them explicit in the syntax. It's a little bit of an experiment but I found that the verbosity of proofs is not too bad considering you don't really need to read `p` and `q`. Also, you often end up with equalities over the wrong equality, but because the base category has homsets you can always rectify using `≡[]-rectify`.

WDYT?


Cubical/Categories/Displayed/Cartesian.agda
```diff
@@ -0,0 +1,308 @@
...
+            (cleavage σ x .snd .fst ⋆ᴰ ϕ
+               ≡[ B.⋆IdR σ ∙ sym (B.⋆IdL σ) ]
+             f ⋆ᴰ cleavage σ y .snd .fst)
+        stepf = substituteArrow f .fst
+        stepg :
+          Σ[ ϕ ∈ verticalCategory Cᴰ b [ cleavage σ y .fst , cleavage σ z .fst ] ]
+            (cleavage σ y .snd .fst ⋆ᴰ ϕ
+               ≡[ B.⋆IdR σ ∙ sym (B.⋆IdL σ) ]
+             g ⋆ᴰ cleavage σ z .snd .fst)
+        stepg = substituteArrow g .fst
```
ecavallo, line 308: any reason for writing out the type signatures here?

Cubical/Categories/Displayed/Cartesian.agda
```diff
@@ -0,0 +1,308 @@
...
+        , ΣPathPProp
+            {B = λ _ f → isOpcartesian f}
+            {u = _ , isOpcartesianDiscreteOpfibration _}
+            {v = snd (opcleavage f aᴰ)}
+            isPropIsOpcartesian (pair-paths .snd))
+        i
+      where
+        pair-paths : _
+        pair-paths = PathPΣ (unique-lift f aᴰ .snd (map-snd fst (opcleavage f aᴰ)))
+
```
ecavallo, line 131: How about
```suggestion
    uniqueOpcleavageDiscreteOpfibration .snd opcleavage =
      implicitFunExt $ implicitFunExt $ funExt λ f → funExt λ aᴰ →
      ΣPathP $
      map-snd (ΣPathPProp isPropIsOpcartesian) $
      PathPΣ $
      unique-lift f aᴰ .snd (map-snd fst (opcleavage f aᴰ))
```
to avoid filling in the implicits?

Cubical/Categories/Displayed/Reasoning.agda
```diff
@@ -0,0 +1,108 @@
...
+    → fᴰ ≡[ p ] gᴰ
+    → gᴰ ≡[ q ] hᴰ → fᴰ ≡[ p ∙ q ] hᴰ
+  ≡[]∙ {fᴰ = fᴰ} {hᴰ = hᴰ} p q eq1 eq2 =
+    subst
+      (λ p → PathP (λ i → p i) fᴰ hᴰ)
+      (sym $ congFunct Hom[_][ _ , _ ] p q)
+      (compPathP eq1 eq2)
+
+  infixr 30 ≡[]∙
+  syntax ≡[]∙ p q eq1 eq2 = eq1 ≡[ p ∙ q ]∙ eq2
```
ecavallo, line 60: My two difficulties with this syntax are that

-  it looks so similar to the `≡⟨ ⟩` equational reasoning syntax---I keep looking at the paths inside the brackets first, when it's the ones on the sides that are the "important" part. 
- one can misread the `p ∙ q` as an ordinary path composition (although the outside then doesn't make sense)
 
What do you think of something like `eq1 [ p ]∙[ q ] eq2` ?

ecavallo (commented):
Looks good to me on the whole---just have a few minor comments and an opinion about the new notation.

Cubical/Categories/Displayed/Cartesian.agda
jpoiret: None :) I just removed them

Cubical/Categories/Displayed/Reasoning.agda
jpoiret: Seems better, yes, esp. regarding the second point. I also did the same with `[ p ]⋆[ q ]`!

Cubical/Categories/Additive/Instances/Terminal.agda
```diff
@@ -24,20 +24,17 @@ open AdditiveCategory
...
 
-  homProp : (x y : Category.ob {ℓ = ℓ} terminalCategory) → isProp (terminalCategory [ x , y ])
+  homProp : (x y : Category.ob {ℓ = ℓ} TerminalCategory) → isProp (TerminalCategory [ x , y ])
   homProp x y = isOfHLevelUnit* 2 x y
 
-  homContr : (x y : Category.ob {ℓ = ℓ} terminalCategory) → isContr (terminalCategory [ x , y ])
+  homContr : (x y : Category.ob {ℓ = ℓ} TerminalCategory) → isContr (TerminalCategory [ x , y ])
   homContr x y = isProp→isContrPath (isOfHLevelUnit* 1) x y
 
   terminalPreAdd : PreaddCategory ℓ ℓ
```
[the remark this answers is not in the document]
jpoiret: You're right, although I didn't want this PR to go too much out of scope.

PR #1027, opened 2023-07-25
Move Rationals around

timorl:
Now they are in `Data` and structured more like the various integers. In particular, `QuoQ` is now "the preferred version of the rationals in the library" – I picked it since it was the only one used in other places.

This is intended to be part of #404 – if this gets accepted then the only remaining work there will be proving the equivalence of `HITQ` and `QuoQ`. I'm already working on that, the main hurdle being that they are defined using different integers, so I first want to prove that arithmetic operations on the integers are equal.

### DIFF
--- a/Cubical/Algebra/CommRing/Instances/Rationals.agda
+++ b/Cubical/Algebra/CommRing/Instances/Rationals.agda
@@ -4,12 +4,12 @@
 
 -}
 {-# OPTIONS --safe #-}
-module Cubical.Algebra.CommRing.Instances.QuoQ where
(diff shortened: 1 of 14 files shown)


timorl:
Hm, can I request a review here, say from @ecavallo ?

felixwellen:
Don't know if anyone is working right now (I am also on vacation). I am back next week and might have time for reviewing (don't know what yet).

mortberg:
This seems reasonable to me. What do you think @felixwellen @ecavallo  ?

PR #1028, opened 2023-07-25
+Int≡+

timorl:
This is more of a warmup, since I actually want `·Int≡·` (useful to prove `QuoQ≡HITQ` I think?). But this might also prove useful to prove that, so no effort wasted.

I'm aware of some problems with this, but I'm not sure how to fix them, so I'll be grateful for any comments.
1. The import looks ridiculous and I'm pretty sure it's not conformant with the contribution guidelines. However, I wasn't able to figure out how to write it in any other way and have all this still work (`as` imports still cause conflicts I think?).
2. I'm not sure whether the changes to `Iso` are ok or if there is a better way of doing a similar thing. Similarly,  exporting `isoIntℤ` might be controversial?
3. I think at least some of the reasoning should be `private`, but I'm not sure what the best way of achieving this is – `private` is kinda obvious, but I think I've also seen some unnamed modules? What should I do with this?
4. The big comment in line 298 – is this something worth pursuing? Within this PR?
5. In general I have a feeling there might be a much simpler way of doing all this...


Cubical/Data/Int/MoreInts/QuoInt/Properties.agda
```diff
@@ -6,13 +6,15 @@ open import Cubical.Foundations.Prelude
...
 open import Cubical.Foundations.Equiv
+open import Cubical.Foundations.Isomorphism
 open import Cubical.Relation.Nullary
 
 open import Cubical.Data.Nat as ℕ using (ℕ; zero; suc)
 open import Cubical.Data.Bool as Bool using (Bool; not; notnot)
 open import Cubical.Data.Empty
 open import Cubical.Data.Unit renaming (Unit to ⊤)
 
+open import Cubical.Data.Int renaming (_+_ to _+Int_; _·_ to _·Int_; ℤ to Int; sucℤ to sucInt; predℤ to predInt; sucℤ+ to sucInt+; predℤ+ to predInt+) hiding (neg; pos; abs; isSetℤ; -_; _+'_; +'≡+)
```
jpoiret, line 18: You probably forgot to remove the `open` when importing the module qualified, hence the conflicts. I'd try a `import Cubical.Data.Int as Int`.

Cubical/Foundations/Isomorphism.agda
```diff
@@ -201,6 +201,11 @@ codomainIso : ∀ {ℓ ℓ' ℓ''} {A : Type ℓ} {B : Type ℓ'} {C : Type ℓ'
            → Iso (A → B) (A → C)
 codomainIso z = codomainIsoDep λ _ → z
 
+endoIso : Iso A B → Iso (A → A) (B → B)
+endoIso is = compIso (domIso is) (codomainIso is)
+
+opIso : Iso A B → Iso (A → A → A) (B → B → B)
```
jpoiret, line 207: The name is very unclear for me, maybe at least `binaryOpIso`?

Cubical/Data/Int/MoreInts/QuoInt/Properties.agda
```diff
@@ -233,3 +235,82 @@ sign-pos-suc-· m (signed s (suc n)) = refl
 
 ·-injʳ : ∀ m n o → m · pos (suc o) ≡ n · pos (suc o) → m ≡ n
 ·-injʳ m n o p = ·-injˡ o m n (·-comm (pos (suc o)) m ∙ p ∙ ·-comm n (pos (suc o)))
+
+opIsoIntℤ : Iso (Int → Int → Int) (ℤ → ℤ → ℤ)
+opIsoIntℤ = (opIso isoIntℤ)
```
jpoiret, line 241: Extra pair of parentheses, also this is so short that this can probably be inlined for a bit less cognitive indirection.

Cubical/Data/Int/MoreInts/QuoInt/Properties.agda
```diff
@@ -233,3 +235,82 @@ sign-pos-suc-· m (signed s (suc n)) = refl
...
+predℤ→Int (pos zero) = refl
+predℤ→Int (pos (suc n)) = refl
+predℤ→Int (neg zero) = refl
+predℤ→Int (neg (suc n)) = refl
+predℤ→Int (posneg i) = refl
+
+ℤ→Int+Int≡+ : ∀ (n m : ℤ) → (ℤ→Int n) +Int (ℤ→Int m) ≡ ℤ→Int (n + m)
+ℤ→Int+Int≡+ (pos zero) m =
+  (ℤ→Int (pos zero)) +Int (ℤ→Int m)
+  ≡⟨ sym (pos0+ (ℤ→Int m)) ⟩
```
jpoiret, line 263: I'd suggest indenting the reasoning lines with a different number of spaces to improve readability.

Cubical/Data/Int/MoreInts/QuoInt/Properties.agda
```diff
@@ -233,3 +235,82 @@ sign-pos-suc-· m (signed s (suc n)) = refl
...
+  negsuc (suc n) +Int ℤ→Int m
+  ≡⟨ sym (predInt+ (negsuc n) (ℤ→Int m)) ⟩
+  predInt ((negsuc n) +Int ℤ→Int m)
+  ≡⟨ cong predInt (ℤ→Int+Int≡+ (neg (suc n)) m) ⟩
+  predInt (ℤ→Int ((neg (suc n)) + m))
+  ≡⟨ predℤ→Int ((neg (suc n)) + m)⟩
+  ℤ→Int (predℤ ((neg (suc n)) + m))
+  ≡⟨ cong ℤ→Int (predℤ-+ˡ (neg (suc n)) m) ⟩
+  ℤ→Int ((neg (suc (suc n))) + m) ∎
+  -- uh, this looks like an extremely generic thing that could be done when defining any function from ℤ, thus only requiring defining it on `pos n` and `neg (suc n)` as one would kinda expect for integers; why don't we do that everywhere?`
```
jpoiret, line 299: I agree that the `elim` given in `Base` is quite tautological and this could probably be added alongside it (or even replace it if it's not used anywhere else).

jpoiret (commented):
I left some generic style comments. Regarding whether to put some of the reasoning in private, I don't really think it's an issue here

Cubical/Data/Int/MoreInts/QuoInt/Properties.agda
timorl: Ha, I was sure I tried that combination, but perhaps I wanted to have some things imported renamed/without the prefix. I still don't know how to do that, but perhaps it's better not to anyway. <_<"

Cubical/Foundations/Isomorphism.agda
timorl: Sure, changed.

Cubical/Data/Int/MoreInts/QuoInt/Properties.agda
timorl: Yeah, I'm not sure why I did that anyway, inlined.

Cubical/Data/Int/MoreInts/QuoInt/Properties.agda
timorl: Hm, I changed the formatting into something else entirely, because it felt nicer. Lemme know what you think and if you want a specific different one then tell me what it should be.

Cubical/Data/Int/MoreInts/QuoInt/Properties.agda
timorl: It wasn't used, so I replaced it, and also used the new version to define the only complicated part of the PR, I think this is more readable now.

timorl:
I also privated a lot of the PR after all, I think they are pretty irrelevant.

Cubical/Data/Int/MoreInts/QuoInt/Properties.agda
```diff
@@ -237,81 +237,82 @@ sign-pos-suc-· m (signed s (suc n)) = refl
...
+  predℤ→Int (neg zero) = refl
+  predℤ→Int (neg (suc n)) = refl
+  predℤ→Int (posneg i) = refl
+
+  ℤ→Int+Int≡+ : ∀ (n m : ℤ) → (ℤ→Int n) Int.+ (ℤ→Int m) ≡ ℤ→Int (n + m)
+  ℤ→Int+Int≡+ n m = (ℤelim (λ n → ∀ (m : ℤ) → (ℤ→Int n) Int.+ (ℤ→Int m) ≡ ℤ→Int (n + m)) posℤ→Int+Int≡+ negsucℤ→Int+Int≡+ n) m
+    where
+    posℤ→Int+Int≡+ : ∀ (n : ℕ) (m : ℤ) → (ℤ→Int (pos n)) Int.+ (ℤ→Int m) ≡ ℤ→Int ((pos n) + m)
+    posℤ→Int+Int≡+ zero m =
+      (ℤ→Int (pos zero)) Int.+ (ℤ→Int m) ≡⟨
```
jpoiret, line 263: That's too funky of a notation wrt. the rest of the library. I feel better with seperate lines for `≡` steps, with differing indentation. Also, it's not consistent with +'≡+

jpoiret (commented):
Rest looks good imo.

Cubical/Data/Int/MoreInts/QuoInt/Properties.agda
timorl: Better now? Or is this not what you had in mind?

I also fixed the other series of equalities, I completely forgot it was there before. ^^"

Cubical/Data/Int/MoreInts/QuoInt/Properties.agda
```diff
@@ -233,3 +236,84 @@ sign-pos-suc-· m (signed s (suc n)) = refl
...
+
+  predℤ→Int : ∀ (n : ℤ) → Int.predℤ (ℤ→Int n) ≡ ℤ→Int (predℤ n)
+  predℤ→Int (pos zero) = refl
+  predℤ→Int (pos (suc n)) = refl
+  predℤ→Int (neg zero) = refl
+  predℤ→Int (neg (suc n)) = refl
+  predℤ→Int (posneg i) = refl
+
+  ℤ→Int+Int≡+ : ∀ (n m : ℤ) → (ℤ→Int n) Int.+ (ℤ→Int m) ≡ ℤ→Int (n + m)
+  ℤ→Int+Int≡+ n m = (ℤelim (λ n → ∀ (m : ℤ) → (ℤ→Int n) Int.+ (ℤ→Int m) ≡ ℤ→Int (n + m)) posℤ→Int+Int≡+ negsucℤ→Int+Int≡+ n) m
```
felixwellen, line 259: I think a name containing something like 'IsHom+' would be better.
And then this fact should be exported as well (maybe even as an homomorphism of abelian groups - that might fit in better somewhere else though).

Cubical/Data/Int/MoreInts/QuoInt/Properties.agda
```diff
@@ -233,3 +236,84 @@ sign-pos-suc-· m (signed s (suc n)) = refl
...
+
+  op≡Intℤ : (Int.ℤ → Int.ℤ → Int.ℤ) ≡ (ℤ → ℤ → ℤ)
+  op≡Intℤ = isoToPath (binaryOpIso isoIntℤ)
+
+  +Int≡+' : (λ i → op≡Intℤ i) [ Int._+_ ≡ _+'_ ]
+  +Int≡+' = transport-filler op≡Intℤ Int._+_
+
+
++Int≡+ : (λ i → (op≡Intℤ ∙ refl) i) [ Int._+_ ≡ _+_ ]
++Int≡+ = compPathP +Int≡+' +'≡+
```
felixwellen, line 319: Can you explain why you want to export this particular form?

felixwellen (changes requested):
Looks good to me as well - I have only some naming/exporting questions.

Cubical/Data/Int/MoreInts/QuoInt/Properties.agda
timorl: My idea was that reasoning on equality is the simplest reasoning there is, so that's likely what other modules would want to use. In particular, if I want to show that two definitions of rationals using different definitions of integers as their base, result in actually equal types (which is the direct motivation for this work) then this form lets me directly show that specific parts of the definition are equal as integers.

Cubical/Data/Int/MoreInts/QuoInt/Properties.agda
felixwellen: Just to be clear: The equality you proved is exactly the statement that the map ```ℤ→Int``` is a homomorphism with respect to ```+```. As you suspect, the performance of a homomorphism proof extracted from an equality of groups is not going to be great, so we should export this one separately.
Regarding conventions, it would be better to turn the equality around. In general, we try to write equations like this in a "evaluating/simplifying/normalizing direction" (meaning ```f(x+y)=f(x)+f(y)``` for a homomorphism or ```x*(y+z)=x*y+x*z``` for distributivity). This is not a well defined term and there is no reason to do it this way -- it is just about having a convention people can remember.

Cubical/Data/Int/MoreInts/QuoInt/Properties.agda
felixwellen: Ok, enough reason to export that.

Cubical/Data/Int/MoreInts/QuoInt/Properties.agda
timorl: Right, the thing I wrote about "iso-" was just silly, sorry. Anyway, exported, renamed, and flipped.

I also rebased this, but in the meantime another thing got merged so that was a tad pointless.

felixwellen:
Thanks! Looks all good to me now -> will merge if CI is happy.

PR #1031, opened 2023-08-22
Sites, sheaves and sheafification as a QIT

MatthiasHu:
This PR provides a definition of a coverage on a category (turning it into a site) and formulates the sheaf condition (amalgamation property) for presheaves on such a site. Furthermore, the sheafification of a presheaf on a site is constructed by means of a quotient inductive type (QIT), and the appropriate universal property is shown.


Cubical/Categories/Site/Sheaf.agda
```diff
@@ -0,0 +1,125 @@
...
+      ((x ≡ y) , str (P ⟅ _ ⟆) _ _)
+
+  amalgamationProperty :
+    hProp (ℓ-max (ℓ-max (ℓ-max (ℓ-max ℓ ℓ') ℓcov) ℓpat) ℓP)
+  amalgamationProperty =
+    ∀[ c ] ∀[ cov ∶ ⟨ covers c ⟩ ]
+    amalgamationPropertyForCover P (str (covers c) cov)
+
+  isSheaf : Type (ℓ-max (ℓ-max (ℓ-max (ℓ-max ℓ ℓ') ℓcov) ℓpat) ℓP)
+  isSheaf = ⟨ amalgamationProperty ⟩
```
dwarn, line 102: Why have two names for this? It's unclear why `isSheaf` should be a raw type, when just above, `isSeparated` is an hProp. I would suggest renaming `amalgamationProperty` to `isSheaf`, and removing the abbrevation for the underlying type.

Cubical/Categories/Site/Sheafification/Base.agda
```diff
@@ -0,0 +1,126 @@
...
+      (f : Hom[ d , c ]) →
+      (x : ⟨ P ⟅ c ⟆ ⟩) →
+      η⟦⟧ ((P ⟪ f ⟫) x) ≡ restrict f (η⟦⟧ x)
+
+    sep :
+      {c : ob} →
+      (cover : ⟨ covers c ⟩) →
+      let cov = str (covers c) cover in
+      (x y : ⟨F⟅ c ⟆⟩) →
+      (i~j :
```
dwarn, line 71: Is this mean to be called `x~y`?

Cubical/Categories/Site/Sheafification/Base.agda
```diff
@@ -0,0 +1,126 @@
...
+      let
+      -- Is there any way to make this definition now and reuse it later?
+      F : Presheaf C _
+      F = record
+        { F-ob = λ c → ⟨F⟅ c ⟆⟩ , trunc
+        ; F-hom = restrict
+        ; F-id = funExt restrictId
+        ; F-seq = λ f g → funExt (restrictRestrict f g)
+        }
+      in
```
dwarn, line 86: Why not define `F` mutually with `F⟅ _ ⟆`?

Cubical/Categories/Site/Sheafification/Base.agda
```diff
@@ -0,0 +1,126 @@
...
+      {c : ob} →
+      (cover : ⟨ covers c ⟩) →
+      let cov = str (covers c) cover in
```
dwarn, line 89: Perhaps it would be nice to have some syntax for this construction. Something like `∀[x in Fam] P x ≔ ∀ (i : ⟨ Fam ⟩) → P (str Fam i)` where `Fam : TypeWithStr ℓ λ I → I → A` and `P : A → Type`

Cubical/Categories/Site/Sheafification/UniversalProperty.agda
```diff
@@ -0,0 +1,183 @@
...
+open import Cubical.Categories.Site.Sheafification.ElimProp
+
+
+module UniversalProperty
+  {ℓ ℓ' ℓcov ℓpat : Level}
+  {C : Category ℓ ℓ'}
+  (J : Coverage C ℓcov ℓpat)
+  where
+
+  -- We assume 'P' to have this level to ensure that 'F' has the same level as 'P'.
```
dwarn, line 35: Ultimately it would of course be nice to know that sheafification has the expected universal property also when it raises universe levels, and it seems the only obstacle to this is that you have to talk about natural transformations between presheaves of different levels? Perhaps this comment could be elaborate to make clear that this is mathematically an unnecessary restriction.

Cubical/Categories/Site/Sheafification/Base.agda
```diff
@@ -0,0 +1,126 @@
...
+      (fam : CompatibleFamily F cov) →
+      (patch : ⟨ cov ⟩) →
+      restrict (patchArr cov patch) (amalgamate cover fam) ≡ fst fam patch
+
+  F : Presheaf C (ℓ-max (ℓ-max (ℓ-max (ℓ-max ℓ ℓ') ℓcov) ℓpat) ℓP)
```
dwarn, line 109: Should this not have a more descriptive name, like `sheafification`?

Cubical/Categories/Site/Sheaf.agda
MatthiasHu: Agreed, thanks.

Cubical/Categories/Site/Sheafification/Base.agda
MatthiasHu: Yes, thanks!

Cubical/Categories/Site/Sheafification/Base.agda
MatthiasHu: I have trouble getting that to work... Might try again later.

Cubical/Categories/Site/Sheafification/Base.agda
MatthiasHu: The exposed name is `Sheafification.F`. Do you think it should still be called `sheafification` inside the named module `Sheafification`?

Cubical/Categories/Site/Sheafification/Base.agda
MatthiasHu: Here is why it seems impossible to define the presheaf `F` and the QIT `⟨F⟅_⟆⟩` mutually recursively.

In the type signatures of some of the constructors of `⟨F⟅_⟆⟩` we use `F` and we need it to unfold. So `F` needs to come before `⟨F⟅_⟆⟩` in the mutual block (I gather), for example like this:
```agda
  data ⟨F⟅_⟆⟩ : ob → Type ...

  F : Presheaf C ...
  F = ...

  data ⟨F⟅_⟆⟩ where
    ...
```

But the definition of `F` uses both `⟨F⟅_⟆⟩` itself as well as some of its constructors (`trunc`, `restrict`, `restrictId` and `restrictRestrict`), and I don't see a way to make the types of these constructors known ahead of time.

In other words: We would need to interrupt the list of constructors of the QIT in order to state the definition of `F`. Agda does not seem to support that. (As opposed to records.)

MatthiasHu: I am happy for any other ideas how to avoid the code duplication!

mzeuner (commented):
Great stuff, looking forward to use it. I only had some stylistic comments.

Cubical/Categories/Site/Sheaf.agda
```diff
@@ -0,0 +1,120 @@
...
+  module _
+    {c : ob}
+    {ℓ'' : Level}
+    (cov : Cover C ℓ'' c)
+    where
+
+    FamilyOnCover : Type (ℓ-max ℓP ℓ'')
+    FamilyOnCover = (i : ⟨ cov ⟩) → ⟨ P ⟅ patchObj cov i ⟆ ⟩
+
+    isCompatibleFamily : FamilyOnCover → hProp (ℓ-max (ℓ-max (ℓ-max ℓ ℓ') ℓP) ℓ'')
```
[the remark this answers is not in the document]
MatthiasHu: Ok, I will try to follow the `isPropIs...` convention then.

Cubical/Categories/Site/Sieve.agda
```diff
@@ -0,0 +1,105 @@
...
+
+-- Definition of sieves on an object
+-- and the sieve generated by a cover.
+
+open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.Function
+open import Cubical.Foundations.HLevels
+open import Cubical.Foundations.Structure
+
+open import Cubical.Functions.Logic using (∀[]-syntax; ∀[∶]-syntax; _⇒_; _⇔_)
```
mzeuner, line 12: There is some hPropy stuff in this file as well. Maybe do the same as in `Site.Sheaf` for consistency's sake ;)

MatthiasHu: Yes, indeed, I just didn't finish that file yet! :smile:

Cubical/Categories/Site/Sheafification/Base.agda
MatthiasHu: I think I actually went for a named module precisely to give users the option to open it with arguments (either just `J` or both `J` and `P`). So let's go with `sheafification` and `isSheafSheafification`, right?

MatthiasHu: Side remark: I can think of the following notations for sheafification in the literature, which seem not very helpful either.

- $a$ (for "associated sheaf functor") [MacLane, Moerdijk: Sheaves in Geometry an Logic]
- $i^*$ (because it is the inverse image part of the geometric morphism from sheaves to presheaves)

MatthiasHu: > it would probably be best to make `F⟅_⟆` private.

That also makes all the constructors of the HIT private, unfortunately. At least I can't find a way to mark only the name `⟨F⟅_⟆⟩` as private.

So I guess I should rename it to `⟨sheafification⟅_⟆⟩`.

Cubical/Categories/Site/Sieve.agda
MatthiasHu: Done.

Cubical/Categories/Site/Sieve.agda
```diff
@@ -0,0 +1,105 @@
...
+  {ℓ ℓ' : Level}
+  (C : Category ℓ ℓ')
+  where
+
+  open Category C
+
+  record Sieve (ℓsie : Level) (c : ob) : Type (ℓ-max ℓ (ℓ-max ℓ' (ℓ-suc ℓsie))) where
+    no-eta-equality
+    field
+      passes : {d : ob} → Hom[ d , c ] → hProp ℓsie
```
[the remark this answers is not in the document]
MatthiasHu: Done.

Cubical/Categories/Site/Sheaf.agda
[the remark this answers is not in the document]
MatthiasHu: I made these changes. No `hProp`s, no `Cubical.Functions.Logic` any more in this PR.

Is this how you meant it?

Cubical/Categories/Site/Sheafification/Base.agda
MatthiasHu: I replaced all `F`s now (that stood for a sheafification). Good like that?

Cubical/Categories/Site/Sheafification/Base.agda
MatthiasHu: I think this is out of scope for this PR.

In principle I would like to have a file dedicated to "families" of things (with a bundled index type). But I wouldn't know where to put it in the library. Or how to find places in the library where it could already be used. Or how to handle potential variants like "a family indexed by an h-set" nicely.

Cubical/Categories/Site/Sheafification/UniversalProperty.agda
MatthiasHu: Done.

MatthiasHu:
@mzeuner I think I handeled everything now!

Cubical/Categories/Site/Sheaf.agda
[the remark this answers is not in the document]
mzeuner: Yes, exactly, cheers.

Cubical/Categories/Site/Sheafification/Base.agda
mzeuner: Yeah, looks good to me, but it is of course a purely stylistic matter. There are instances of both in the library, named modules with short names, like `f` for a certain function, and named modules with terms that have essentially the same name as the module (intended to opened). I don't want to advocate too heavily for one of the options, but if everyone is happy now, we can leave things as they are, I would say.

mzeuner:
I think the PR is ready to be merged @mortberg  @felixwellen

felixwellen:
Nice!
Please rebase onto/merge master...

MatthiasHu:
Note to my future self: it would probably have been smarter to split this PR into two (1. sites, 2. sheafification). :-)

felixwellen:
Looks good to me (just skimming - two people already looked at it).

PR #1033, opened 2023-08-25
simplify isSet<SomeAlgebraicStructure>

MatthiasHu:
This PR simplifies, or at least harmonizes, the definitions of the following lemmas:
```
isSetGroup
isSetRing
isSetDistLattice
isSetCommAlgebra
isSetLeftModule
isSetField
isSetAlgebra
isSetLattice
isSetCommRing
isSetAbGroup
isSetOrderedCommMonoid
isSetCommMonoid
```

Each of them is a simple `is-set` after opening the respective `<Something>Str`. In the case of `Algebra` and `OrderedCommMonoid`, the `is-set` field was impossible to use since it was exported twice from the respective records; I fixed that by hiding one copy (as it was already done in `Ring`, for example).

### DIFF
--- a/Cubical/Algebra/AbGroup/Base.agda
+++ b/Cubical/Algebra/AbGroup/Base.agda
@@ -139,7 +139,7 @@ module _ ((G , abgroupstr _ _ _ GisGroup) : AbGroup ℓ) where
   AbGroup→CommMonoid .snd .CommMonoidStr.isCommMonoid .IsCommMonoid.·Comm = IsAbGroup.+Comm GisGroup
 
 isSetAbGroup : (A : AbGroup ℓ) → isSet ⟨ A ⟩
-isSetAbGroup A = isSetGroup (AbGroup→Group A)
+isSetAbGroup A = is-set (str A)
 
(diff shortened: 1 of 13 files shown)


Cubical/Algebra/AbGroup/Base.agda
```diff
@@ -139,7 +139,7 @@ module _ ((G , abgroupstr _ _ _ GisGroup) : AbGroup ℓ) where
   AbGroup→CommMonoid .snd .CommMonoidStr.isCommMonoid .IsCommMonoid.·Comm = IsAbGroup.+Comm GisGroup
 
 isSetAbGroup : (A : AbGroup ℓ) → isSet ⟨ A ⟩
-isSetAbGroup A = isSetGroup (AbGroup→Group A)
+isSetAbGroup A = is-set (str A)
```
MatthiasHu, line 142: This is the single case where the `open ...Str ...` method was impractical, since the file already opens multiple records with `is-set` fields.

felixwellen:
Thanks for the nice refactoring!

PR #1035, opened 2023-08-25
elimProp for the FreeCommAlgebra HIT

MatthiasHu:
This PR replaces the huge repetitive `proveEq` lemma in `FreeCommAlgebra.Properties` by a simpler, more general and somewhat shorter (but also repetitive) `elimProp` eliminator.

### DIFF
--- a/Cubical/Algebra/CommAlgebra/FreeCommAlgebra/Properties.agda
+++ b/Cubical/Algebra/CommAlgebra/FreeCommAlgebra/Properties.agda
@@ -48,8 +48,8 @@ module Theory {R : CommRing ℓ} {I : Type ℓ'} where
          using (0r; 1r)
          renaming (_·_ to _·r_; _+_ to _+r_; ·Comm to ·r-comm; ·IdR to ·r-rid)
 
-  open Construction using (var; const)
   module C = Construction
+  open C using (var; const)
 
   {-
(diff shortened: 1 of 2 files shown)


Cubical/Algebra/CommAlgebra/FreeCommAlgebra/Properties.agda
```diff
@@ -46,10 +48,110 @@ module Theory {R : CommRing ℓ} {I : Type ℓ'} where
...
+      mkPathP (C.+-rinv x)
+        (on+ (elimProp x) (on- (elimProp x)))
+        onConst
+        i
+    elimProp (C.+-comm x y i) =
+      mkPathP (C.+-comm x y)
+        (on+ (elimProp x) (elimProp y))
+        (on+ (elimProp y) (elimProp x))
+        i
+    elimProp (C.·-assoc x y z i) =
```
MatthiasHu, line 108: I would be interested if all these path constructor cases can be made any shorter than this.

felixwellen: No good ideas here. Did you try dropping the 'i' on both sides?

felixwellen:
Thanks! I wanted to have/do this for a long time...

PR #1038, opened 2023-08-29
Avoid unnecessary `syntax` declarations

MatthiasHu:
This PR replaces [`syntax` declarations](https://agda.readthedocs.io/en/latest/language/syntax-declarations.html) which

- do not bind variables and
- do not change the order of arguments in such a way that some argument depends on some later argument

by ordinary infix (or mixfix) operator declarations.

The motivation for avoiding syntax declarations is that Agdas behavior around them can be confusing, for example:
- "jumping to the definition" of a declared syntax does not jump to the syntax declaration but to the definition of the name it stands for.
- When you type an operator that is actually a declared syntax, like `_⇒_` (from `NaturalTransformation.Base`), in a hole and ask for its type (or try to "refine", or "give" the hole), you get a "Not in scope" error. (If I understand correctly, `_⇒_` does indeed not exist as a name. But it is easy to expect it to exist.)
- If you want to import a declared syntax by `open ... using (...)`, you have to write the name it refers to, even though you then never use that name.

### DIFF
--- a/Cubical/Algebra/Algebra/Properties.agda
+++ b/Cubical/Algebra/Algebra/Properties.agda
@@ -85,7 +85,8 @@ module AlgebraHoms where
   compAlgebraHom f g .fst = g .fst ∘ f .fst
   compAlgebraHom f g .snd = compIsAlgebraHom (g .snd) (f .snd)
 
-  syntax compAlgebraHom f g = g ∘a f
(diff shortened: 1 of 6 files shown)


felixwellen:
For the slim chance that there is such an advantage, we can still revert -> merging.

PR #1039, opened 2023-08-30
·Int≡·

timorl:
Natural continuation of #1028 . It follows the same pattern (post-review ofc) and exists for the same reason, so I suspect it shouldn't be controversial.


felixwellen:
Yes, that's good to have and uncontroversial.

PR #1041, opened 2023-09-04
Generalize types of inspect idiom

dolio:
This generalizes some types to allow doing an `inspect` of a dependent function. I've found myself wanting this from time to time, and I haven't noticed the generalization having any adverse effects (I think the fact that `f` is an argument nails down potential ambiguities about `B`).


mortberg (approved):
Seems like a reasonable change that I can merge directly. Thanks for the contribution!

Issue #598, opened 2021-08-23
Remove the `makeX` functions but keep the `makeIsX`


mortberg:
We could remove for example `makeGroup` as groups should be constructed using copatterns anyway. It might still be nice to keep the `makeIsGroup` and it should hopefully not be a problem as the copatterns in the group def will block the laws from getting unfolded unless necessary

barrettj12:
I disagree, since I think `makeX` gives much more readable definitions. E.g. looking at the current definition of the group ℤ in `Cubical.Algebra.Group.Instances.Int`:
```agda
ℤ : Group₀
fst ℤ = ℤType
1g (snd ℤ) = 0
_·_ (snd ℤ) = _+ℤ_
inv (snd ℤ) = _-ℤ_ 0
isGroup (snd ℤ) = isGroupℤ
  where
  abstract
    isGroupℤ : IsGroup (pos 0) _+ℤ_ (_-ℤ_ (pos 0))
    isGroupℤ = makeIsGroup isSetℤ +Assoc (λ _ → refl) (+Comm 0)
                           (λ x → +Comm x (pos 0 -ℤ x) ∙ minusPlus x 0)
                           (λ x → minusPlus x 0)
```
vs using `makeGroup`:
```agda
ℤ = makeGroup
  0
  _+ℤ_
  (_-ℤ_ 0)
  isSetℤ
  +Assoc
  (λ _ → refl)
  (+Comm 0)
  (λ x → +Comm x (pos 0 -ℤ x) ∙ minusPlus x 0)
  (λ x → minusPlus x 0)
```
You could even annotate it:
```agda
ℤ = makeGroup
  0          -- identity
  _+ℤ_       -- group operation
  (_-ℤ_ 0)   -- inverses
  ...
```

barrettj12:
On the other hand, I just tried doing this with the `GroupStr` constructor `groupstr`:
```agda
ℤ : Group₀
ℤ = ℤType , groupstr
      0
      _+ℤ_
      (_-ℤ_ 0)
      (makeIsGroup 
        isSetℤ
        +Assoc
        (λ _ → refl)
        (+Comm 0)
        (λ x → +Comm x (pos 0 -ℤ x) ∙ minusPlus x 0)
        (λ x → minusPlus x 0)
      )
```
which looks very similar. Probably not all of these are necessary to keep.

[...]

PR #1042, opened 2023-09-10
Introduce Semiring and refactor finite sums

felixwellen:
So far, we only have CommSemiring. I don't have a Semiring that is not commutative, but I want to move the definition of sum from Ring to a place where it can also be used for Nat.

### DIFF
--- a/Cubical/Algebra/CommSemiring/Base.agda
+++ b/Cubical/Algebra/CommSemiring/Base.agda
@@ -62,29 +62,3 @@ makeIsCommSemiring {_+_ = _+_} {_·_ = _·_}
                                (x · z) + (y · z) ∎ )
             ( λ x → ·Comm x _ ∙ AnnihilL x) AnnihilL
         IsCommSemiring.·Comm x = ·Comm
-
-makeCommSemiring : (R : Type ℓ)
-                 (0r 1r : R) (_+_ _·_ : R → R → R)
-                 (is-setR : isSet R)
-                 (+Assoc : (x y z : R) → x + (y + z) ≡ (x + y) + z)
-                 (+IdR : (x : R) → x + 0r ≡ x)
-                 (+Comm : (x y : R) → x + y ≡ y + x)
-                 (·Assoc : (x y z : R) → x · (y · z) ≡ (x · y) · z)
-                 (·IdR : (x : R) → x · 1r ≡ x)
-                 (·DistR+ : (x y z : R) → x · (y + z) ≡ (x · y) + (x · z))
-                 (AnnihilL : (x : R) → 0r · x ≡ 0r)
-                 (·Comm : (x y : R) → x · y ≡ y · x)
-               → CommSemiring ℓ
-makeCommSemiring
-  R
-  0r 1r _+_ _·_
-  is-setR +Assoc +IdR +Comm ·Assoc ·IdR ·DistR+ AnnihilL ·Comm
-  = (R , str)
-  where module CS = CommSemiringStr
-        str : CommSemiringStr R
-        CS.0r str = 0r
-        CS.1r str = 1r
-        CS._+_ str = _+_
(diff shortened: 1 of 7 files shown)


Cubical/Algebra/CommSemiring/Base.agda
```diff
@@ -48,3 +37,54 @@ record CommSemiringStr (A : Type ℓ) : Type (ℓ-suc ℓ) where
...
+            is-setR +Assoc +IdR +Comm ·Assoc
+            ·IdR (λ x → ·Comm _ x ∙ (·IdR x))
+            ·DistR+ (λ x y z → (x + y) · z       ≡⟨ ·Comm (x + y) z ⟩
+                               z · (x + y)       ≡⟨ ·DistR+ z x y ⟩
+                               (z · x) + (z · y) ≡[ i ]⟨ (·Comm z x i + ·Comm z y i) ⟩
+                               (x · z) + (y · z) ∎ )
+            ( λ x → ·Comm x _ ∙ AnnihilL x) AnnihilL
+        IsCommSemiring.·Comm x = ·Comm
+
+makeCommSemiring : (R : Type ℓ)
```
mzeuner, line 66: There has been a motion to remove `make...` and only keep the `makeIs...` for all the algebraic structures:
https://github.com/agda/cubical/issues/598

I guess the discussion has remained inconclusive, but maybe something to keep in mind...

mzeuner (commented):
Only had a tiny remark, which is related to an old issue https://github.com/agda/cubical/issues/598.

Since that issue was never closed, I think we can leave it to your conscience, whether you want to keep `makeCommSemiRing` or whether you want define `UpperNat.asCommSemiRing` using copatterns.

Otherwise it should be ready to be merged.

Cubical/Algebra/CommSemiring/Base.agda
felixwellen: I think Anders' arguments do not really apply in this case, since there is no difference in readability and I defined ```makeCommSemiring``` a copattern - apart from using ```(R , str)``` - but I think we wan't that one to unfold. Or does it make a difference to use a function defined with a copattern instead of a direct copattern? 
On the other hand, it is really just boilerplate code and the interesting stuff is in ```makeIsCommSemiring``` so I will have a look how things look like, when I remove it.

felixwellen: Looks somewhat better - thanks!

Cubical/Algebra/CommSemiring/Instances/UpperNat.agda
```diff
@@ -186,16 +186,17 @@ module ConstructionBounded where
...
-  IsCommSemiring.+IsCommMonoid (CommSemiringStr.isCommSemiring (snd asCommSemiring)) =
-    OrderedCommMonoidStr.isCommMonoid (snd ℕ↑-+b)
-  IsCommSemiring.·IsCommMonoid (CommSemiringStr.isCommSemiring (snd asCommSemiring)) =
-    OrderedCommMonoidStr.isCommMonoid (snd ℕ↑-·b)
-  IsCommSemiring.·LDist+ (CommSemiringStr.isCommSemiring (snd asCommSemiring)) =
-    λ x y z → Σ≡Prop (λ s → PropCompletion.isPropIsBounded ℓ-zero ℕ≤+ s)
-                     (ConstructionUnbounded.+LDist· (fst x) (fst y) (fst z))
-  IsCommSemiring.AnnihilL (CommSemiringStr.isCommSemiring (snd asCommSemiring)) = AnnihilL
+  asCommSemiring =
+    makeCommSemiring ℕ↑b 0↑ (OrderedCommMonoidStr.ε (snd ℕ↑-·b))
```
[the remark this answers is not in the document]
felixwellen: See above

PR #1043, opened 2023-09-10
NatSolver should be able to deal with 'suc's

felixwellen:
This PR fixes an issue in the reflection interface of the NatSolver. Now any 'suc x' is translated correctly to '1 + x' in the Nat-Expression language which is used for solving. I also removed some unused code.


felixwellen:
Since it is only a small bug fix, including a test, I'll just merge that.

PR #1045, opened 2023-09-11
Remove commutativity assumption to bigOpSplit++

jpoiret:
This is only needed for bigOpSplit.


felixwellen:
Sorry - I just produces a conflict, since I moved those bigops from Rings to Semiring. I guess it would make sense to just move products in the same way.

jpoiret:
That should solve the conflict, let's wait for CI

felixwellen (approved):
Thanks!

Issue #1046, opened 2023-09-12
Change the Constructor Name of Sequential Colimits


kangrongji:
When using [sequential colimits from the library](https://github.com/agda/cubical/blob/master/Cubical/HITs/SequentialColimit/Base.agda), the constructor `inl` can be hard to align with `push` during pattern patching, etc. So it doesn't look good. Renaming it to `incl` and maybe changing the field name `space` to `obj` would improve alignment and give a more category-theory-inspired feel. If you all agree, I could make a PR very soon.

felixwellen:
Not a user of sequential colimits (yet), but this looks like an improvement to me.
What do you think about the name ```Lim→```? 
I consistently forger the traditional math notation it resembles over the years, but maybe that's just me. It is also a lot shorter than ```SeqColim```.

kangrongji:
Yeah... I think `Lim→` looks a bit weird. There is no hurt to use `SeqColim` instead. I could do all these changes.

PR #1047, opened 2023-09-12
Renaming to Sequential Coimits

kangrongji:
See #1046 .

I also changed some names in `SequentialColimit/Properties.agda` for it being more informative. These names are only used in this one file. Also add some spaces to make them look better.

@felixwellen

### DIFF
--- a/Cubical/HITs/James/Inductive/PushoutFormula.agda
+++ b/Cubical/HITs/James/Inductive/PushoutFormula.agda
@@ -26,7 +26,7 @@ open import Cubical.Data.Sigma
 open import Cubical.HITs.Wedge
 open import Cubical.HITs.Pushout
 open import Cubical.HITs.Pushout.PushoutProduct
-open import Cubical.HITs.SequentialColimit renaming (incl to incl')
+import Cubical.HITs.SequentialColimit as SColim
 open import Cubical.HITs.James.Inductive.Base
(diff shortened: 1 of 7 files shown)


Cubical/HITs/James/Inductive/PushoutFormula.agda
```diff
@@ -26,7 +26,7 @@ open import Cubical.Data.Sigma
 open import Cubical.HITs.Wedge
 open import Cubical.HITs.Pushout
 open import Cubical.HITs.Pushout.PushoutProduct
-open import Cubical.HITs.SequentialColimit
+open import Cubical.HITs.SequentialColimit renaming (incl to incl')
```
felixwellen, line 29: I guess this is due to a name collision. I would prefer a name which clearly connects ```incl``` to ```SeqColim```, so it is clear on reading where it comes from.
Sometimes it is good to ```import``` something as, say ```SColim``` to later write ```SColim.incl```, but I don't know if it is a good idea here.

felixwellen (changes requested):
Thanks! Just one comment so far. Please keep it as a draft as long as you are still adding commits.

Cubical/HITs/James/Inductive/PushoutFormula.agda
kangrongji: Make sense. This is the only place where conflicting happens, I'll follow you here.

kangrongji:
It's done now @felixwellen

felixwellen (approved):
Looks all good!

PR #1048, opened 2023-09-12
Add workflow with Agda master.

jpoiret:
Hi, this is a first try at adding a CI workflow using Agda master, without rebuilding it ourselves. It simply fetches the nightly release and uses that instead.

This seems to be working fine on my fork, see https://github.com/jpoiret/cubical/actions/runs/6158237986/job/16710596838 (and it has already caught something).

WDYT?

### DIFF
--- a/.github/workflows/ci-master.yml
+++ b/.github/workflows/ci-master.yml
@@ -1,0 +1,143 @@
+## Adapted from the agda/agda-stdlib workflow
+
+name: Ubuntu build
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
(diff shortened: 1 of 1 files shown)


felixwellen:
> (and it has already caught something)

Which is good to know:

```
/home/runner/work/cubical/cubical/Cubical/Tactics/Reflection.agda:92,5-19
warning: -W[no]UserWarning
DEPRECATED: Use `withReduceDefs` instead of `dontReduceDefs`
when scope checking dontReduceDefs
```

felixwellen:
Also closing, like #1051

Issue #1015, opened 2023-06-20
Citation.cff


anuyts:
This repository needs a [citation.cff](https://citation-file-format.github.io/) file so people can easily cite it in papers.

See also [this stackexchange answer](https://academia.stackexchange.com/a/172780/144474).

anuyts:
Here's a handy generator: <https://citation-file-format.github.io/cff-initializer-javascript/#/>.

felixwellen:
Maybe we can just adapt that one:

https://github.com/agda/agda-stdlib/blob/master/CITATION.cff

Opinions please @mortberg @ecavallo @mzeuner @aljungstrom (incomplete list).

ecavallo:
Following the stdlib is fine with me :+1:

mortberg:
Me too!

PR #1049, opened 2023-09-12
#1015: CITATION.cff

felixwellen:
Add a CITATION.cff and update instructions for RELEASE


felixwellen:
This one is only bureaucratic and the essential change was decided in #1015, so I'll just merge once CI is happy.

PR #1056, opened 2023-09-15
Additions to the Powerset module 

### LINKED ISSUE #1052: Additions to the powerset module
I believe the following additions could be made to the `Powerset` module : 

1. [x] A `∈ₚ` operator that produces an `hProp` for `x ∈ₚ A` rather than a general type
2. [x] Binary unions and intersections 
3. [x] Indexed unions and intersections 
4. [x] A separate type for "inhabited subsets" 
5. [x] The complement of a subset

While these are easy to define I found consistently needing these and believe others could benefit from it too. 

I'd be happy to drop in a PR.

rahulc29:
Fixes #1052 
Do note that two definitions had to be rewritten to prevent circular dependencies. 
I've highlighted them in the comments and marked them as `private`.

### DIFF
--- a/Cubical/Foundations/Powerset.agda
+++ b/Cubical/Foundations/Powerset.agda
@@ -18,6 +18,12 @@ open import Cubical.Foundations.Function
 open import Cubical.Foundations.Univalence using (hPropExt)
 
 open import Cubical.Data.Sigma
+open import Cubical.Data.Unit
+open import Cubical.Data.Empty
+
+open import Cubical.HITs.PropositionalTruncation.Base
+
+open import Cubical.Relation.Nullary using (¬_)
 
 private
   variable
@@ -41,6 +47,26 @@ A ⊆ B = ∀ x → x ∈ A → x ∈ B
 ∈-isProp : (A : ℙ X) (x : X) → isProp (x ∈ A)
 ∈-isProp A = snd ∘ A
 
+infix 5 _∈ₚ_
+
+-- x ∈ₚ A differs from x ∈ A
+-- in the sense that it produces an hProp
+_∈ₚ_ : {X : Type ℓ} → X → ℙ X → hProp ℓ
+x ∈ₚ A = x ∈ A , ∈-isProp A x
+
+-- "not in" relations
+
+infix 5 _∉_
+_∉_ : {X : Type ℓ} → X → ℙ X → Type ℓ
+x ∉ A = ¬ x ∈ A
+
+∉-isProp : (A : ℙ X) (x : X) → isProp (x ∉ A)
+∉-isProp A x = isPropΠ λ _ → isProp⊥
+
+infix 5 _∉ₚ_
+_∉ₚ_ : {X : Type ℓ} → X → ℙ X → hProp ℓ
+x ∉ₚ A = x ∉ A , ∉-isProp A x
+
 ⊆-isProp : (A B : ℙ X) → isProp (A ⊆ B)
 ⊆-isProp A B = isPropΠ2 (λ x _ → ∈-isProp B x)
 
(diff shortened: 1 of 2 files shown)


mortberg:
Could some reorganization be done to avoid duplication?

Another thing is that I generally don't like operations that produce hProp's. It's often better to have something produce a Type and then have the proof that it is an h-prop separately. The reason being that many times one can do things purely for Type's without any need to pass around the h-prop proofs. A more technical issue is that Agda sometimes get confused and implicit arguments tend to work worse as it cannot infer the h-prop proofs... I made a comment about this at some point (https://github.com/agda/cubical/blob/master/Cubical/Functions/Logic.agda#L6) and would be quite happy to delete/rewrite the Logic file... Maybe this could be done as part of reorganization so that you can avoid duplication in this PR?

rahulc29:
The main aspect is that `Cubical.Functions.Embeddings` needs some proofs regarding powersets such as `Embedding→Subset` and the like. It might be possible to isolate the proofs involving powersets in the `Embeddings` module? The `Embeddings` module has many very important lemmata that (in my opinion) should be refactored into logically separate modules. In particular, it is easy to see a characterisation into the following
1. Elementary properties of embeddings
2. Embeddings wrt isomorphisms and equivalences
3. "Embedding-into-X→X" style proofs
4. Miscellaneous (composition, transitivity, universe embeddings)
I have not included the fibration identity principle and the embedding identity principle since they are already in their separate modules and refactoring them would be rather easy.

I agree that `hProp`s are not necessarily convenient to work with. My primary motivation for relying on `hProp`s was to not have to duplicate the hProp logic given in `Cubical.Functions.Logic`. 
I believe it might be a better idea to rewrite the `Logic` module anyway to just keep the actual type and the `isProp` proofs separate.

mortberg:
> The main aspect is that `Cubical.Functions.Embeddings` needs some proofs regarding powersets such as `Embedding→Subset` and the like. It might be possible to isolate the proofs involving powersets in the `Embeddings` module? The `Embeddings` module has many very important lemmata that (in my opinion) should be refactored into logically separate modules. In particular, it is easy to see a characterisation into the following
> 
>     1. Elementary properties of embeddings
> 
>     2. Embeddings wrt isomorphisms and equivalences
> 
>     3. "Embedding-into-X→X" style proofs
> 
>     4. Miscellaneous (composition, transitivity, universe embeddings)
>        I have not included the fibration identity principle and the embedding identity principle since they are already in their separate modules and refactoring them would be rather easy.
> 
> 
> I agree that `hProp`s are not necessarily convenient to work with. My primary motivation for relying on `hProp`s was to not have to duplicate the hProp logic given in `Cubical.Functions.Logic`. I believe it might be a better idea to rewrite the `Logic` module anyway to just keep the actual type and the `isProp` proofs separate.

I'm not against splitting Functions.Embeddings into multiple files in a separate directory and then have a file exporting everything if this helps with dependencies. I'm also all for rewriting Functions.Logic. Feel free to have a go at it in this PR!

rahulc29:
Sure, I'll have a go at it! :D

Cubical/Foundations/Powerset.agda
```diff
@@ -111,7 +111,7 @@ private data _⊎_ (A B : Type ℓ) : Type ℓ where
   inr : B → A ⊎ B
```
anshwad10: I don't think it's a good idea to have two versions of a datatype, especially since they are not definitionally equal. It would be better to split Cubical.Data.Sum into Cubical.Data.Sum.Base and .Properties, import sum from .Base, and have .Properties import from Embeddings so there is no circular dependency and no duplication. (And I don't think making it private fixes it: How does a user of this library give an element of A U B without being able to access the constructors of the datatype underlying it?)

PR #1062, opened 2023-09-24
Dependent Version of Bi-Invertible Equivalences

kangrongji:
Some facts about dependent bi-invertible maps. They are useful to deal with HITs with bi-invertible maps constructors, for example, the `BiInvInt`.


mortberg (approved):
Looks good to me. @ecavallo Should we merge?

PR #1063, opened 2023-09-30
more connectivity lemmas

rwbarton:
From https://github.com/CMU-HoTT/serre-finiteness/blob/cellular/Cellular/Connectivity.agda


mortberg (approved):
Looks good to me!

Issue #657, opened 2021-12-02
Schemes


FredericPaugam:
If schemes are not implemented yet, i wanted to recommend you the reading of Demazure-Gabriel ``groupes algébriques'', because their presentation of schemes should be adapted to a cubical agda implementation (without choice). For example, they define explicitely the open subset $V(I)$ as a concrete subsheaf of $Spec(A)$, the sheaf of sets associated to $A$ by Yoneda for the standard open topology. It is not representable in general, but you may show it is a sheaf and clearly, it is covered by standard opens $D(f)$ corresponding to generators of $I$. Schemes are then sheaves on the Grothendieck standard open site covered by some $Spec(A)$s,  and they inherit a Zariski topology. The structural sheaf on arbitrary opens (without points) comes from functions with values in the affine line. I saw you had the classical opens in the cubical library, so i hope this may be useful. Projective space and Grasmaniann are defined similarly by their points in EGAI. Remark that the etale topology is also very close (just take etale algebras over rational opens instead to define sheaves). A similar approach works for wide classes of rigid analytic and complex analytic spaces (using affinoid Banach algebras and their rational algebras). However, in the analytic situation, you need ind-objets and completed tensor products, which probably makes the implementation more complicated.

mzeuner:
Thanks for your recommendation. There is indeed an ongoing effort to add schemes or at least affine schemes to the library.
The idea is to not only work without choice but follow an approach that is entirely constructive. The main inspiration being the following paper:
https://link.springer.com/article/10.1007/s10472-009-9160-7

As a consequence we can not describe the opens of `Spec(A)` as sets of prime ideals...
Skimming through the first chapter of Demazure-Gabriel, I can't see how to make that work constructively,
but maybe I'm missing something?

In any case, I would love to have a classical definition of the Zariski spectrum and the structure sheaf and prove it equivalent
to the constructive definition (using LEM). So it's definitely worthwhile to keep this approach in mind.

FredericPaugam:
[...]

PR #1068, opened 2023-10-23
ℤ-Functors

mzeuner:
This PR introduces the category of ℤ-functors, i.e. functors from commutative rings to sets (of a given size), i.e. presheaves on `CommRing ^op`. We show/introduce two things

1. The Yoneda embedding has something that looks like a left adjoint, although not quite, because we run into a mismatch of universe levels.
2. To each ℤ-functor, we can associate the lattice of compact open subfunctors that are classified by the Zariski lattice. This can be used to define the notion of an affine cover.

There are still a bunch of TODOs in `Categories.Instances.ZFunctors` that will hopefully be proved in later PRs. 

Once we have sites (https://github.com/agda/cubical/pull/1031), we can define the Zariski topology on ℤ-functors and define quasi-compact, quasi-separated schemes as the Zariski sheaves that have an affine cover in the sense of this PR. See also the discussion in https://github.com/agda/cubical/issues/657.

The ultimate goal is to do all of that for finitely presented algebras, where we can avoid the size issues. This would give us a predicative and constructive account of schemes of finite presentation (over some base `Spec(A)`).


Cubical/Algebra/DistLattice/Properties.agda
```diff
@@ -0,0 +1,55 @@
...
+
+  idCompDistLatticeHom : {L M : DistLattice ℓ} (f : DistLatticeHom L M)
+                    → compDistLatticeHom _ _ _ f (idDistLatticeHom M) ≡ f
+  idCompDistLatticeHom = idCompLatticeHom
+
+  compAssocDistLatticeHom : {L M N U : DistLattice ℓ}
+                         (f : DistLatticeHom L M) (g : DistLatticeHom M N) (h : DistLatticeHom N U)
+                       → compDistLatticeHom _ _ _ (compDistLatticeHom _ _ _ f g) h
+                       ≡ compDistLatticeHom _ _ _ f (compDistLatticeHom _ _ _ g h)
+  compAssocDistLatticeHom = compAssocLatticeHom
```
mortberg, line 55: Please package this up as a category instance

Cubical/Categories/Instances/ZFunctors.agda
```diff
@@ -0,0 +1,412 @@
...
+    (λ _ _ _ → makeNatTransPath (funExt₂
+                 (λ A _ → ZariskiLattice A .snd .DistLatticeStr.∧lAssoc _ _ _)))
+    (λ _ → makeNatTransPath (funExt₂ (λ A _ → ZariskiLattice A .snd .DistLatticeStr.∧lRid _)))
+    (λ _ _ → makeNatTransPath (funExt₂ (λ A _ → ZariskiLattice A .snd .DistLatticeStr.∧lComm _ _)))
+    (λ _ _ → makeNatTransPath (funExt₂ -- don't know why ∧lAbsorb∨l doesn't work
+               (λ A _ → ZariskiLattice A .snd .DistLatticeStr.absorb _ _ .snd)))
+    (λ _ _ _ → makeNatTransPath (funExt₂ -- same here
+                 (λ A _ → ZariskiLattice A .snd .DistLatticeStr.∧l-dist-∨l _ _ _ .fst)))
+
+  -- TODO: (contravariant) action on morphisms
```
mortberg, line 394: Please fix this as it should be easy, then we can merge

mortberg (changes requested):
Good stuff, a few comments that could be fixed before we merge and then you can do the other TODOs as separate PRs

Cubical/Algebra/DistLattice/Properties.agda
mzeuner: Done

Cubical/Categories/Instances/ZFunctors.agda
mzeuner: Done

mortberg (approved):
Nice! Will merge once CI is done

PR #1071, opened 2023-10-28
Rational order

LuuBluum:
Defines an ordering over the rationals and proves several key properties. Also refactors the rationals to use the integers instead of QuoInt; the original definition are relabeled as QuoQ and put into MoreRationals.

At some point, ideally under a different PR, QuoQ should be removed and all proofs involving it replaced with proofs against the variant using the default version of Int rather than QuoInt.


Cubical/Data/Int/Properties.agda
```diff
@@ -86,6 +98,288 @@ predSuc (pos n)          = refl
 predSuc (negsuc zero)    = refl
 predSuc (negsuc (suc n)) = refl
 
+sucDistMin : ∀ m n → sucℤ (min m n) ≡ min (sucℤ m) (sucℤ n)
```
mortberg, line 101: Wow, a lot of the proofs in this file are quite wild. Big pattern-matches where almost every case is `refl`.. Can one set things up so that this goes a bit smoother or is this necessary? Some tactics to solve the trivial cases automatically would be nice...

Cubical/Data/Int/Properties.agda
```diff
@@ -451,6 +827,311 @@ inj-+z {z} {l} {n} p = inj-z+ {z = z} (+Comm z l ∙ p ∙ +Comm n z)
 n+z≡z→n≡0 : ∀ n z → n + z ≡ z → n ≡ 0
 n+z≡z→n≡0 n z p = inj-z+ {z = z} {l = n} {n = 0} (+Comm z n ∙ p)
 
+pos+posLposMin : ∀ x y → min (pos (x +ℕ y)) (pos x) ≡ pos x
```
mortberg, line 830: In the naming of these, should Min/Max come first?

Cubical/Data/Int/Properties.agda
```diff
@@ -451,6 +827,311 @@ inj-+z {z} {l} {n} p = inj-z+ {z = z} (+Comm z l ∙ p ∙ +Comm n z)
 n+z≡z→n≡0 : ∀ n z → n + z ≡ z → n ≡ 0
```
mortberg, line 827: Not your fault, but it's bad to have lemma names named after arbitrary choices of variable names. Can you fix?

Cubical/Data/Rationals/MoreRationals/QuoQ/Properties.agda
```diff
@@ -0,0 +1,248 @@
...
+-- useful functions for defining operations on ℚ
+
+onCommonDenom :
+  (g : ℤ × ℕ₊₁ → ℤ × ℕ₊₁ → ℤ)
+  (g-eql : ∀ ((a , b) (c , d) (e , f) : ℤ × ℕ₊₁) (p : a ℤ.· ℕ₊₁→ℤ d ≡ c ℤ.· ℕ₊₁→ℤ b)
+           → ℕ₊₁→ℤ d ℤ.· (g (a , b) (e , f)) ≡ ℕ₊₁→ℤ b ℤ.· (g (c , d) (e , f)))
+  (g-eqr : ∀ ((a , b) (c , d) (e , f) : ℤ × ℕ₊₁) (p : c ℤ.· ℕ₊₁→ℤ f ≡ e ℤ.· ℕ₊₁→ℤ d)
+           → (g (a , b) (c , d)) ℤ.· ℕ₊₁→ℤ f ≡ (g (a , b) (e , f)) ℤ.· ℕ₊₁→ℤ d)
+  → ℚ → ℚ → ℚ
+onCommonDenom g g-eql g-eqr = SetQuotient.rec2 isSetℚ
```
mortberg, line 39: Wasn't there an issue with SetQuotient.rec2 being defined by rec instead of double pattern-matching? Maybe this get easier with these definition if that is fixed?

Cubical/Data/Rationals/MoreRationals/QuoQ/Properties.agda
```diff
@@ -0,0 +1,248 @@
...
+  (λ { (a , b) (c , d) (e , f) p →
+    ℕ₊₁→ℤ d ℤ.· (a ℤ.· ℕ₊₁→ℤ f ℤ.+ e ℤ.· ℕ₊₁→ℤ b)
+      ≡⟨ sym (ℤ.·-distribˡ (ℕ₊₁→ℤ d) (a ℤ.· ℕ₊₁→ℤ f) (e ℤ.· ℕ₊₁→ℤ b)) ⟩
+    ℕ₊₁→ℤ d ℤ.· (a ℤ.· ℕ₊₁→ℤ f) ℤ.+ ℕ₊₁→ℤ d ℤ.· (e ℤ.· ℕ₊₁→ℤ b)
+      ≡[ i ]⟨ lem₁ a (ℕ₊₁→ℤ d) c (ℕ₊₁→ℤ b) (ℕ₊₁→ℤ f) p i ℤ.+ lem₂ (ℕ₊₁→ℤ d) e (ℕ₊₁→ℤ b) i ⟩
+    ℕ₊₁→ℤ b ℤ.· (c ℤ.· ℕ₊₁→ℤ f) ℤ.+ ℕ₊₁→ℤ b ℤ.· (e ℤ.· ℕ₊₁→ℤ d)
+      ≡⟨ ℤ.·-distribˡ (ℕ₊₁→ℤ b) (c ℤ.· ℕ₊₁→ℤ f) (e ℤ.· ℕ₊₁→ℤ d) ⟩
+    ℕ₊₁→ℤ b ℤ.· (c ℤ.· ℕ₊₁→ℤ f ℤ.+ e ℤ.· ℕ₊₁→ℤ d) ∎ })
+
++-comm : ∀ x y → x + y ≡ y + x
```
mortberg, line 123: With naming conventions, shouldn't this be `+Comm`, and similarly for the rest of the names in the file?

mortberg (changes requested):
Thanks, lots of good stuff in this PR! 

A general question, can't the ring solver be used to prove a lot of the equations for Z that are needed in the proofs?

mortberg:
Can someone who knows how the CI really works help with fixing:

```
agda: Heap exhausted;
agda: Current maximum heap size is 3758096384 bytes (3584 MB).
agda: Use `+RTS -M<size>' to increase it.
make: *** [GNUmakefile:53: check] Error 251
Error: Process completed with exit code 2.
```

Maybe @felixwellen or @mzeuner ?

LuuBluum:
The heap issue seems to be happening on master as well, so it's a broader issue with the library.

Cubical/Data/Rationals/Order.agda
```diff
@@ -0,0 +1,626 @@
...
+                        (⊎.rec (⊥.rec ∘ m≮n) (<Weaken≤ n m))
+                        (isConnected< m n m≢n)
+
+0<+ : ∀ m n → 0 < m ℚ.+ n → (0 < m) ⊎ (0 < n)
+0<+ m n 0<m+n with <Dec 0 m | <Dec 0 n
+... | no 0≮m | no 0≮n = ⊥.rec (≤→≯ (m ℚ.+ n) 0 (≤Monotone+ m 0 n 0 (≮→≥ 0 m 0≮m) (≮→≥ 0 n 0≮n)) 0<m+n)
+... | no _    | yes 0<n = inr 0<n
+... | yes 0<m | _ = inl 0<m
+
+0<1 : 0 < 1
```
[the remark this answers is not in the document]
LuuBluum: It's one of those last properties required for something to be an ordered field as per the HoTT book's definition. I don't think it's actually strictly necessary, so I have no issue with removing it.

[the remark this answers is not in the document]
mortberg: With the definition of `<` that we're discussing on Discord this will be completely trivial, so please remove it

Cubical/Data/Int/Properties.agda
LuuBluum: I don't think I necessarily tried for all of these (sucDist and predDist namely), but I recall that attempting to consolidate the cases would run into issues. The problem is that min and max themselves are defined by pattern-matching, so they'll only actually compute once you apply the appropriate argument case.

Which means a lot of pattern matching for proofs about either.

Cubical/Data/Rationals/MoreRationals/QuoQ/Properties.agda
LuuBluum: Yes (which is why it's changed in Rationals.Base).

In reality this entire file isn't necessary other than the fact that some of the identities (proving equivalence with SigmaQ, namely) are hard to map rationals with Z instead of QuoInt as their integer definition. Ideally this whole file can be ripped out and the couple of places that still refer to it instead refer to Rationals instead.

Cubical/Data/Rationals/Order.agda
```diff
@@ -0,0 +1,626 @@
...
+_≥_ : ℚ → ℚ → Type₀
+m ≥ n = n ≤ m
+
+_>_ : ℚ → ℚ → Type₀
+m > n = n < m
+
+_#_ : ℚ → ℚ → Type₀
+m # n = (m < n) ⊎ (n < m)
+
+data Trichotomy (m n : ℚ) : Type₀ where
```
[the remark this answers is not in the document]
LuuBluum: We do, but the issue here is that it only works for [ a , b ] [ c , d ] cases. Upstreaming it would require a gnarly pattern match like the definition of < or <=/

Cubical/Data/Rationals/MoreRationals/QuoQ/Properties.agda
mortberg: Oops yes, I missed that this was in a `MoreRationals` file... So let's not bother about changing it. Ideally the files in `MoreX` should not leak too much into the rest of the library

Cubical/Data/Int/Properties.agda
mortberg: Hmm, I wonder if the best is to define max/min by deep pattern-matching like this? Wouldn't it be better to just use < or <= together with if-then-else? (Assuming that the comparison operations are bool valued...)

mortberg: Alternatively, define them with slightly less deep pattern-matching are use max/min for nat instead?

LuuBluum: I think part of the problem here is that they're defined by themselves rather than in terms of the equivalent operator on the nats. Given that I want to do a broader refactoring of Nat and Int (per our discussion), I'd rather that be its own PR instead of included in this one.

mortberg: Yeah, that makes sense. Please make a note of this and fix it in a future PR :-)

felixwellen:
Please rebase/merge master

Cubical/Data/Int/Properties.agda
LuuBluum: I'm going to redo most of these in the later PR where I will revise all the Nat/Int properties to defer to their algebraic structure.

Cubical/Data/Rationals/MoreRationals/QuoQ/Properties.agda
LuuBluum: I can see about fixing rec2 to see if things compute a bit easier, and if not I can define this with pattern-matching as well.

LuuBluum: Threw in the change to rec2.

Cubical/HITs/SetQuotients/Properties.agda
```diff
@@ -119,14 +119,32 @@ rec set f feq (squash/ x y p q i j) = set (g x) (g y) (cong g p) (cong g q) i j
...
+    toPath a b rab = elimProp (λ _ → set _ _) λ c → feql a b c rab
+
+    fun : A / R → B / S → C
+    fun [ a ] y = fun₀ a y
+    fun (eq/ a b r i) y = toPath a b r y i
+    fun (squash/ x y p q i j) z = isSet→SquareP (λ _ _ → set)
+      (λ _ → fun x z)
+      (λ _ → fun y z)
+      (λ i → fun (p i) z)
+      (λ i → fun (q i) z) j i
```
mortberg, line 147: Wouldn't something like this be simpler:

```
rec2 : isSet C
  → (f : A → B → C)
  → (∀ a b c → R a b → f a c ≡ f b c)
  → (∀ a b c → S b c → f a b ≡ f a c)
  → A / R → B / S → C
rec2 setC f feql feqr [ a ] [ b ] = f a b
rec2 setC f feql feqr [ a ] (eq/ x y z i) = feqr a x y z i
rec2 setC f feql feqr [ a ] (squash/ x y p q i j) =
  setC _ _ (cong (rec2 setC f feql feqr [ a ]) p) (cong (rec2 setC f feql feqr [ a ]) q) i j
rec2 setC f feql feqr (eq/ x y z i) b = {!!}
rec2 setC f feql feqr (squash/ x y p q i j) b =
  setC _ _ (cong (λ x → rec2 setC f feql feqr x b) p) (cong (λ x → rec2 setC f feql feqr x b) q) i j
```

I don't have time to figure out the remaining hole, but it should follow from elimProp I think

LuuBluum: I prefer the original proof, because the logic used here can be used to convert any call to rec2 to a pattern-matching approach. It's also, at least in my opinion, a lot easier to follow when it doesn't involve a bunch of recursive calls to itself.

Cubical/Data/Int/Properties.agda
LuuBluum: When this is refactored per our discussion, this lemma won't exist anymore.

mortberg: Great!

Cubical/HITs/SetQuotients/Properties.agda
mortberg: Ok, as long as it computes properly on `[ a ] [ b ]` I guess it doesn't really matter how the rest is done

mortberg (approved):

mortberg:
Thanks for this! Looking forward to future PRs cleaning things even more

MatthiasHu:
@LuuBluum Could you explain the motivation for 1bc03c771f5c9143b73811c3187fc3c614c2d8a9 ? The previous definition of `rec2` was shorter and seems more readable and elegant to me.

LuuBluum:
> @LuuBluum Could you explain the motivation for [1bc03c7](https://github.com/agda/cubical/commit/1bc03c771f5c9143b73811c3187fc3c614c2d8a9) ? The previous definition of `rec2` was shorter and seems more readable and elegant to me.

Pattern matching. Defining `rec2` recursively (as it was before) results in significant slowdown on usage due to needing to carry over the proof of transitivity over the quotient relation every time you invoke it, particularly with things that don't care about that proof at all.

For context, under the previous definition of `rec2`, if I were to define the operations on the rational numbers that way, you would never be able to use them because they could never compute. It was just prohibitively slow. This was the only way to actually define the rationals as a set quotient so that it could actually be used.

PR #1073, opened 2023-10-30
CI: set heap size to 6G

felixwellen:


felixwellen:
There was no maximum set as far as I can tell. Now the suggested heap size is 6G instead of 3G - not clear to me why that should help, but who knows whats going on.

felixwellen:
Ok, so the first commit just *decreased* the maximum heap size and it worked. The last commit increased the suggested heap size - I also don't see why that should fix the problem. Maybe it is just randomly working?

felixwellen:
Anyway - I'll just merge it..

felixwellen:
Anyway - I'll just merge it..

PR #1075, opened 2023-10-30
CI: bump GHC to recommended version (9.4.7)

andreasabel:


felixwellen:
I looks a lot like this commit pushed the checking time for the library up from around 35min to 50min. Is there a reason for that?

felixwellen:
Might be just noise in the github-builttimes...

andreasabel:
Looking at recent runs, I see build times between 3 and 110 minutes: https://github.com/agda/cubical/actions?query=branch%3Amaster+event%3Apush
What seems to matter most here is how much of the cache can be reused.  

A priori, I do not see a reason why bumping the GHC version would cause a regression in Agda type-checking times, but of course there could be a GHC regression...  One would need some reproducer...

felixwellen:
Thanks for the input - I guess I should do some profiling locally, when I have some time...

Issue #1022, opened 2023-07-08
Slightly more generalizable universes


Trebor-Huang:
https://github.com/agda/cubical/blob/132a2a3197b490c571356f0399a2a6fbfab40f2a/Cubical/Foundations/Transport.agda#L170C1-L170C1

This could be slightly improved with `∀ {ℓ ℓ'} {A : Type ℓ} {B : Type ℓ'}`. It's too trivial so I won't submit a PR here. Also we might need to test whether it breaks level inference downstream.

mortberg:
Feel free to make a PR if nothing breaks. I like short PRs that make things better :-)

Trebor-Huang:
I take that it is already fixed by the PR?

PR #1076, opened 2023-10-31
Generalized Universe Levels in 'substInPaths'

kesleta:
Made `substInPaths` slightly more general by allowing A and B to be Types with different levels. Suggested by #1022.


mortberg (approved):
Thanks for fixing this! Will merge once the CI is finished

PR #1079, opened 2023-11-12
Bump `cubical-utils` to GHC 9.8

andreasabel:
- Bump cubical-utils to GHC 9.8
- Everythings.hs: explicit imports
- Add `cabal.project` so that `cubical-utils` also works as a submodule


andreasabel:
This is a pure maintenance change that I want in Agda 2.6.4.1, so I take the liberty to merge this now...

PR #1083, opened 2023-12-04
Release for agda 2.6.4.1

felixwellen:
The suggested heap size also seems to set a maximum. I don't why we have a suggested heap size anyway, so I just removed that, since make failed with agda 2.6.4.1, compiled with ghc 9.8.1.


felixwellen:
I think we should stick to removing the line ```RTS_OPTIONS=+RTS -H6G -RTS```, since it causes problems with newer ghc versions (than we are using in the CI - that's why it didn't show here).
Can someone (@phijor @jpoiret ?) check the flake and do ```nix flake update``` and commit the result?

phijor:
I've updated the flake inputs locally; I'll get back to you with the result once it's finished building.

*NB:* I'm not a contributor to the repo, so I cannot directly push to this branch. I can put the commit in a fork of mine though.

felixwellen:
Thanks!

felixwellen:
> _NB:_ I'm not a contributor to the repo, so I cannot directly push to this branch. I can put the commit in a fork of mine though.

Yes, just do that and I'll drag it over here.

phijor:
I updated the flake inputs, everything builds successfully. Changes should be available as a commit [here](https://github.com/agda/cubical/commit/21314e6a9cefda56573f9c7394d5855cfe04086d) (plain-text patch [here](https://github.com/agda/cubical/commit/21314e6a9cefda56573f9c7394d5855cfe04086d.patch)).

felixwellen:
Looks all good to me now!

felixwellen:
@mortberg @mzeuner @ecavallo are you okay with a release or should I wait for anything with merging this? If you have the time, maybe it is good to test it on your machine, since there was some trouble with the make file - I noticed some weird behaviour around RTS options with newer GHC versions.

felixwellen:
I actually missed one instance of "Experiments"-exclusion in the makefile - maybe this went by the CI because of caching...

felixwellen:
It is fixed, looks all good to me now (I checked 977 files in 25 min, so there is also no tc-time blow up).

felixwellen:
I assume no complaints means I can go ahead with merging this and doing all the release stuff (agda 2.6.4.2 is around the corner..).

felixwellen:
On one of my machines, there is *somehow* a too low default maximum for the heap. Setting a high maximum (I chose 32G) solves that and I don't see any harm in that, so I'll leave that in as it is now.

PR #1085, opened 2023-12-05
add proof that Sigma preserves null types

awswan:
I couldn't find a proof that null types are preserved by Sigma, so here's one, added to Nullification.Properties with some similar proofs.


mortberg:
Looks good to me so merging. Thanks for the contribution!

PR #1087, opened 2024-01-12
Remove redundant third constructor to replacement

ecavallo:
The last constructor in the current inductive-recursive definition of replacement is unnecessary. Thanks go to @dwarn and @plt-amy who independently pointed this out to me some time ago.


mortberg:
Nice!

Issue #1095, opened 2024-02-01
Algebraic geometry


mzeuner:
By now, the library has quite a bit of stuff on (constructive) algebraic geometry, but it is spread out.
The stuff based on the lattice theoretic approach of [Coquand, Lombardy and Schuster](https://link.springer.com/article/10.1007/s10472-009-9160-7) is in `Cubical.Algebra.ZariskiLattice`, whereas the functor of points approach is developed in `Cubical.Categories.Instances.ZFunctors`.

I would suggest to make a folder `Cubical.AlgebraicGeometry` that itself contains to folders for the two approaches currently (partly) formalized and move everything algebraic geometry related there.

Does that sound like a reasonable idea?

mortberg:
Yes, that would be great

mortberg:
Has this been fixed now?

PR #1086, opened 2023-12-06
Functorial qcqs-schemes

mzeuner:
This PR builds on (https://github.com/agda/cubical/pull/1082).
It proves that the Zariski coverage on `CommRing^op` is subcanonical, i.e. that representables are Zariski sheaves.
Quasi-compact,  quasi-separated schemes are defined as Zariski sheaves that have a finite affine cover by compact open subfunctors. Subcanonicity is then used to show that affine schemes (i.e. representables) are qcqs-schemes.

PR #1082, opened 2023-12-01
Zariski coverage on CommRing^op

mzeuner:
Finishing the draft PR (https://github.com/agda/cubical/pull/1081) with @MatthiasHu.

PR #1096, opened 2024-02-01
Open subschemes

mzeuner:
This builds on (https://github.com/agda/cubical/pull/1082) and (https://github.com/agda/cubical/pull/1086),
but even with those two merged it is still a long one, my apologies...

The PR contains the following:
- definition of standard (compact) open subfunctors of an affine scheme and proof that those are affine
- proof that any compact open subfunctor of an affine scheme is a qcqs-scheme

The file `Cubical.Categories.Instances.ZFunctors` has gotten quite big and type-checking it is starting to get slow.
Ultimately, I think it should be split up into several files in a separate (functorial) `AlgebraicGeometry` directory,
see this issue: https://github.com/agda/cubical/issues/1095
I'm postponing this for a separate PR though


MatthiasHu:
@mzeuner Could you try to resolve the merge conflicts here or would you like me to do it?

mzeuner:
> @mzeuner Could you try to resolve the merge conflicts here or would you like me to do it?

This was harder than I expected, but now it should work...

MatthiasHu:
In the future, it might be good to move the definition of support maps and related lemmas (currently in `ZariskiLattice.UniversalProperty`) to a separate file, since it does not depend on the Zariski lattice. But perhaps not in this PR. (Just discussed with @mzeuner .)

felixwellen:
Is this PR intentionally open?

MatthiasHu:
> Is this PR intentionally open?

Yes, @mzeuner and I discussed it a bit today and I am working on reviewing it.

felixwellen:
> > Is this PR intentionally open?
> 
> Yes, @mzeuner and I discussed it a bit today and I am working on reviewing it.

Good - then I'll leave you to it.

Cubical/Algebra/DistLattice/BigOps.agda
```diff
@@ -107,6 +107,15 @@ module Join (L' : DistLattice ℓ) where
  ≤-⋁Ext = ≤-bigOpExt
 
 
+module JoinMap {L : DistLattice ℓ} {L' : DistLattice ℓ'} (φ : DistLatticeHom L L') where
+  private module L = Join L
+  private module L' = Join L'
+  open BigOpMap (LatticeHom→JoinSemilatticeHom φ)
+
+  pres⋁ : {n : ℕ} (U : FinVec ⟨ L ⟩ n) → φ .fst (L.⋁ U) ≡ L'.⋁ (φ .fst ∘ U)
+  pres⋁ = presBigOp
```
MatthiasHu, line 116: It seems weird to have this lemma only for `Join` and not for `Meet`. But this was already the case for other definitions in this file (and I find it pretty hard to see which ones)...

I guess this is part of the general question how to handle the duality between join and meet nicely.

Cubical/Algebra/ZariskiLattice/Properties.agda
```diff
@@ -0,0 +1,221 @@
...
+      where
+      fgHelper : Σ[ α ∈ FinVec (fst R) 1 ] 1r ^ n ≡ linearCombination R α (replicateFinVec 1 f)
+               → f ∈ₚ R ˣ
+      fgHelper (α , 1ⁿ≡α₀f+0) = α zero , path
+        where
+        path : f · α zero ≡ 1r
+        path = f · α zero      ≡⟨ solve! R ⟩
+               α zero · f + 0r ≡⟨ sym 1ⁿ≡α₀f+0 ⟩
+               1r ^ n          ≡⟨ 1ⁿ≡1 n ⟩
+               1r ∎
```
MatthiasHu, line 80: About half of this proof is proving that if `1 ∈ √⟨f⟩` then `f` is a unit. It would be good to have that in a reusable form. Perhaps as two lemmas (in appropriate files): (1) if `1 ∈ √I` then `1 ∈ I`, (2) if `1 ∈ ⟨f⟩` then `f` is a unit.

mzeuner: Yeah agreed, I'll have a look at it later.

mzeuner: Is something like this what you had in mind :
https://github.com/agda/cubical/pull/1096/commits/ddc4fa53198b8eb3c3bd8b0cb55c1680648d3e93

MatthiasHu: Yes, very nice!

Cubical/Algebra/DistLattice/BigOps.agda
mzeuner: I added the analogous case for meets of this one: https://github.com/agda/cubical/pull/1096/commits/ef4ae22842f8d581b2bc80f048ba05387624ed60

I did treat meets and joins symmetrically, but I never ended up using big meets, so a bunch of lemmas are only done for big joins. After all, they are what you need to define a "cover", so arguably the more useful one...

felixwellen:
ZFunctors should definitely be split up - no need to do it in this PR though.
Otherwise it looks good to me and it still checks on current master -> merging.

PR #1105, opened 2024-02-21
Define free wild categories

felixwellen:
I define something which might be a free wild category - I don't know. It is intended to be used for a solver for wild categories.


felixwellen:
obsoleted

PR #1106, opened 2024-02-25
Functor Comprehension

stschaef:
Adding a module called `Comprehension`, and associated dependencies, that provides a method for constructing functors without providing the full functorial structure up front.

That is, say you wish to define a functor `F :  C → D` via some universal property `P`. Instead of manually constructing the action on objects, morphisms, and proving functoriality, one can instead just give an action on objects and prove the functoriality of the property `P`.

This is a pretty big PR. I have tried to integrate most of our changes into the appropriate spots so far, and I'll be marking it as a draft until I address these 

- Continue renaming and integrating our changes. For instance, the modules that currently end in `More` 
- Considering moving `Comprehension.agda` back under the `Profunctor` directory.
- Some dangling TODOs and naming concerns

PR #1108, opened 2024-03-01
Displayed Category Constructions

stschaef:
Meant to be the first thrust of #1106 code as we separate that large hunk of code into modular and manageable PRs

Here we add the following things to reason about displayed categories

- displayed full subcategories
- displayed preorders
- displayed terminal category
- displaying a category D over a category C
- displayed opposite category and opposite functor
- introduction principles to define functors into certain classes of displayed categories. Such as the displayed total category and when a displayed category has prop hom sets
- the left adjoint to reindexing


felixwellen:
@ecavallo @mortberg does anyone have any idea why the CI is not running? Might be because @stschaef  is first-time contributor, but usually there appears some button to run the CI...

mortberg:
> @ecavallo @mortberg does anyone have any idea why the CI is not running? Might be because @stschaef is first-time contributor, but usually there appears some button to run the CI...

Indeed, very strange. I see no button either...

ecavallo:
Same here, I don't know the problem.

stschaef:
Could it be that I'm merging in a branch with a name that isn't `master`

Cubical/Categories/Displayed/AdjointToReindex.agda
```diff
@@ -0,0 +1,173 @@
...
+          (pf i₁) ((F-seq g h ◁ (λ i₂ → seq' D (pg i₂) (ph i₂))) i₁)))
+        i ,
+      Cᴰ .⋆Assocᴰ fᴰ gᴰ hᴰ i
+  isSetHomᴰ ∃F {d}{d'}{f}{u}{v} =
+    isSetΣ (C .isSetHom)
+      λ g → isSet×
+        (isProp→isSet (isPropHomP _ _ _ _))
+        (Cᴰ .isSetHomᴰ)
+
+-- Examples of ∃F instantiated
```
stschaef, line 112: I thought I'd at least try to demonstrate to y'all that this construction does indeed generalize what we'd discussed in #1106 

These examples can probably be removed though

Cubical/Categories/Displayed/AdjointToReindex.agda
```diff
@@ -0,0 +1,173 @@
...
+    isPropHomP : ∀ {A B A' B'} (p : A ≡ A') (q : B ≡ B') →
+                (f : D [ A , B ]) → (f' : D [ A' , B' ]) →
+                isProp (hom-pathP p q f f')
+    isPropHomP pdom pcod f f'' x y =
+      isoFunInjective (PathPIsoPath _ _ _) x y fromPathPeq
+      where
+      fromPathPeq : fromPathP x ≡ fromPathP y
+      fromPathPeq = D .isSetHom _ _ (fromPathP x) (fromPathP y)
+
+  ∃F : Categoryᴰ D (ℓ-max (ℓ-max ℓC ℓD) ℓDᴰ) (ℓ-max (ℓ-max ℓC' ℓD') ℓDᴰ')
```
stschaef, line 58: It isn't clear to me if this general construction would be convenient to use in lieu of the usual construction for `∫Cᴰ` or the ad-hoc way we built `∫Cᴰsr`. I have a feeling there may end up being too many transports around to use this easily. 

However, if nothing else it should be useful to have this more general construction. As one could show that the usual `∫Cᴰ` and `∫Cᴰ'` as defined here are equivalent, and then you can port over proofs about each of these constructions across the equivalence

ecavallo:
Now the button to run CI appeared. Not sure what changed, but I clicked it.

stschaef:
> Now the button to run CI appeared. Not sure what changed, but I clicked it.
I pushed another commit to fix a naming typo. Seems like that did the trick

Cubical/Categories/Displayed/Preorder.agda
```diff
@@ -0,0 +1,56 @@
+{-# OPTIONS --safe #-}
+module Cubical.Categories.Displayed.Preorder where
```
maxsnew, line 2: Since this is analogous to `FullSubcategory` we should put in under `Constructions` too. Also for naming, if we like `FullSubcategory` then this could just be `Subcategory`

Cubical/Categories/Displayed/AdjointToReindex.agda
```diff
@@ -0,0 +1,173 @@
+{-# OPTIONS --safe #-}
+--
+module Cubical.Categories.Displayed.AdjointToReindex where
```
maxsnew, line 3: Adjoint isn't specific enough. It's the left adjoint. (I'm not sure if the right adjoint always exists or if it requires a Conduche condition.)

This could also be called dependent sum, though I think that's confusing because to me a dependent sum is the special case where F is a projection.

Cubical/Categories/Displayed/AdjointToReindex.agda
```diff
@@ -0,0 +1,173 @@
...
+    Σ[ g ∈ C [ c , c' ] ] hom-pathP p p' (F-hom g) f × Cᴰ.Hom[ g ][ x , x' ]
+  ∃F .idᴰ {d} {c , p , x} =
+    (C .id) ,
+    (F-id ◁ (cong (λ v → D .id {v}) p)) ,
+    Cᴰ .idᴰ
+  ∃F ._⋆ᴰ_ (g , p , gᴰ) (h , q , hᴰ) =
+      g ⋆⟨ C ⟩ h ,
+      F-seq g h ◁ (λ i → p i ⋆⟨ D ⟩ q i) ,
+      (Cᴰ._⋆ᴰ_ gᴰ hᴰ)
+  ⋆IdLᴰ ∃F {d}{d'}{g}{c , pc , xc}{c' , pc' , xc'} (f , p , x) =
```
maxsnew, line 70: These proofs can probably be simplified using https://github.com/agda/cubical/blob/master/Cubical/Categories/Displayed/Reasoning.agda

Cubical/Categories/Displayed/Constructions/FullSubcategory.agda
```diff
@@ -0,0 +1,65 @@
+-- | Full subcategories (not necessarily prop) viewed as displayed categories.
+{-# OPTIONS --safe #-}
+module Cubical.Categories.Displayed.Constructions.FullSubcategory where
```
maxsnew, line 3: If we don't like "Full subcategory" we could call this `FullyFaithful` because every displayed cat where `Fst` is fully faithful is equivalent to one of this form. We should also consider removing `Cubical.Categories.Constructions.FullSubcategory` and replacing it by this. But there is some univalent stuff in there now that would have to be ported

Cubical/Categories/Displayed/Functor.agda
```diff
@@ -157,3 +172,56 @@ module _ {C : Category ℓC ℓC'} {D : Category ℓD ℓD'}
...
+
+  Fst : Functor (∫C Cᴰ) C
+  Fst .F-ob = fst
+  Fst .F-hom = fst
+  Fst .F-id = refl
+  Fst .F-seq =
+    λ f g → cong {x = f ⋆⟨ ∫C Cᴰ ⟩ g} fst refl
+
+  -- Functor into the total category
+  module TotalCategory {D : Category ℓD ℓD'}
```
maxsnew, line 189: Should this be a named module? The analogs of this, e.g., `Data.Sum.elim` are not in named modules, you just tend to have to import the module `as Sum` to use them without name clashes.

Cubical/Categories/Displayed/Functor.agda
```diff
@@ -157,3 +172,56 @@ module _ {C : Category ℓC ℓC'} {D : Category ℓD ℓD'}
   ∫F .F-hom (_ , fᴰ) = _ , Fᴰ.F-homᴰ fᴰ
   ∫F .F-id = ΣPathP (_ , Fᴰ.F-idᴰ)
   ∫F .F-seq _ _ = ΣPathP (_ , (Fᴰ.F-seqᴰ _ _))
+
+-- Projections out of the total category
```
maxsnew, line 176: This should be in a different file. I think we should move `∫C` to `Displayed.TotalCategory` or `Displayed.TotalCategory.Base` and this can be in that or `Displayed.TotalCategory.Properties`

Cubical/Categories/Displayed/Functor.agda
```diff
@@ -157,3 +172,56 @@ module _ {C : Category ℓC ℓC'} {D : Category ℓD ℓD'}
...
+           (Fᴰ : Functorᴰ F (Unitᴰ D) Cᴰ)
+           where
+    intro : Functor D (∫C Cᴰ)
+    intro .F-ob d = F ⟅ d ⟆ , Fᴰ .F-obᴰ _
+    intro .F-hom f = (F ⟪ f ⟫) , (Fᴰ .F-homᴰ _)
+    intro .F-id = ΣPathP (F .F-id , Fᴰ .F-idᴰ)
+    intro .F-seq f g = ΣPathP (F .F-seq f g , Fᴰ .F-seqᴰ _ _)
+
+-- Projection out of the displayed total category
+module _ {C : Category ℓC ℓC'}
```
maxsnew, line 200: Same comment here as above, should be in a different file.

Cubical/Categories/Displayed/Preorder.agda
```diff
@@ -0,0 +1,56 @@
...
+      isProp→PathP ((λ i → Pᴰ .isPropHomᴰ {f = ((C .⋆Assoc _ _ _) i)})) _ _
+    ; isSetHomᴰ = isProp→isSet (Pᴰ .isPropHomᴰ)
+    }
+
+  hasPropHomsPreorderᴰ : hasPropHoms Preorderᴰ→Catᴰ
+  hasPropHomsPreorderᴰ _ _ _ = Pᴰ .isPropHomᴰ
+
+  open Functor
+
+  Preorderᴰ→FstFaithful : isFaithful (Fst {Cᴰ = Preorderᴰ→Catᴰ})
```
maxsnew, line 53: maybe just `FstFaithful`?

maxsnew (commented):
Mostly starting some discussions about naming/library structure.

Cubical/Categories/Displayed/AdjointToReindex.agda
jpoiret: Yes, having a right adjoint is exactly being Conduché.

Cubical/Categories/Displayed/AdjointToReindex.agda
maxsnew: Should `∫Cᴰsr` and `∫Cᴰsl` be in this PR? I don't see them

Cubical/Categories/Displayed/Preorder.agda
maxsnew: In Section 9.8 of the HoTT book this is called a "notion of structure" over C, likely following https://ncatlab.org/nlab/show/stuff,+structure,+property terminology since the resulting projection is faithful.

jpoiret: Yeah, that's a great idea! Structure over C instead of `Preorder` and property over C for `FullSubcategory`?

maxsnew: By this logic, we should change Category^D to be `Stuff`...

stschaef: I can see the argument for renaming `FullSubcategory > PropertyOverC` and `Preoder > StructureOverC`, but this feels maybe out of scope for this PR. Shouldn't those changes also come with analogous changes to the non-displayed variants?

Cubical/Categories/Displayed/TotalCategory/Base.agda
```diff
@@ -0,0 +1,90 @@
+{-# OPTIONS --safe #-}
+module Cubical.Categories.Displayed.TotalCategory.Base where
```
maxsnew, line 2: Maybe a radical suggestion but I think we should put this in `Categories.Constructions.TotalCategory` (not under `Displayed`) because it constructs a `Category`, not a displayed category. Then `DisplayedTotalCategory` would go in `Categories.Displayed.Constructions.TotalCategory`.

Cubical/Categories/Displayed/Constructions/StructureOverC.agda
```diff
@@ -0,0 +1,58 @@
...
+open import Cubical.Categories.Displayed.Base
+open import Cubical.Categories.Displayed.Functor
+open import Cubical.Categories.Displayed.HLevels
+open import Cubical.Categories.Displayed.TotalCategory.Base
+
+private
+  variable
+    ℓC ℓC' ℓCᴰ ℓCᴰ' ℓD ℓD' ℓDᴰ ℓDᴰ' : Level
+
+record StructureOverC (C : Category ℓC ℓC') ℓCᴰ ℓCᴰ' :
```
maxsnew, line 21: For naming, I think `StructureOver` is nicer (and same for `PropertyOver`). Then for example you would write `Algebra M : StructureOver Set` for `M : Monad Set`.

Cubical/Categories/Displayed/TotalCategory/Base.agda
```diff
@@ -0,0 +1,90 @@
...
+    intro .F-ob d = F ⟅ d ⟆ , Fᴰ .F-obᴰ _
+    intro .F-hom f = (F ⟪ f ⟫) , (Fᴰ .F-homᴰ _)
+    intro .F-id = ΣPathP (F .F-id , Fᴰ .F-idᴰ)
+    intro .F-seq f g = ΣPathP (F .F-seq f g , Fᴰ .F-seqᴰ _ _)
+
+-- Display a category D over C
+module _ (C : Category ℓC ℓC') (D : Category ℓD ℓD') where
+  open Category
+  open Categoryᴰ
+  weaken : Categoryᴰ C ℓD ℓD'
```
maxsnew, line 80: See comment about `weakenᴰ`.

Cubical/Categories/Displayed/Constructions/PropertyOverC.agda
```diff
@@ -0,0 +1,51 @@
...
+open Categoryᴰ
+open Functorᴰ
+
+module _ (C : Category ℓC ℓC') (P : Category.ob C → Type ℓP) where
+  private
+    module C = Category C
+  open Category
+  open Functor
+
+  PropertyOverC : Categoryᴰ C ℓP ℓ-zero
```
maxsnew, line 31: See comment about `StructureOverC`

maxsnew (changes requested):
Just a couple more naming/moving things around. Then it should be good to go IMO

maxsnew (approved):
LGTM. @jpoiret @felixwellen and/or @ecavallo this is ready for review

Cubical/Categories/Displayed/LeftAdjointToReindex.agda
```diff
@@ -0,0 +1,107 @@
+{-# OPTIONS --safe #-}
+--
+module Cubical.Categories.Displayed.LeftAdjointToReindex where
```
maxsnew, line 3: This should be `Cubical.Categories.Displayed.Constructions.LeftAdjointToReindex`

stschaef:
Whitespace is fixed, and the left adjoint has been moved. Should be ready for review

maxsnew:
@felixwellen anything we can do to move this along?

Cubical/Categories/Displayed/HLevels.agda
```diff
@@ -0,0 +1,75 @@
+{-# OPTIONS --safe #-}
```
felixwellen, line 1: Maybe a comment that this file is about ```HLevels``` of ```Hom```s would be good here.

Cubical/Categories/Displayed/Constructions/StructureOver.agda
```diff
@@ -0,0 +1,58 @@
...
+      isProp→PathP ((λ i → Pᴰ .isPropHomᴰ {f = ((C .⋆Assoc _ _ _) i)})) _ _
+    ; isSetHomᴰ = isProp→isSet (Pᴰ .isPropHomᴰ)
+    }
+
+  hasPropHomsStructureOver : hasPropHoms StructureOver→Catᴰ
+  hasPropHomsStructureOver _ _ _ = Pᴰ .isPropHomᴰ
+
+  open Functor
+
+  FstFaithful : isFaithful (Fst {Cᴰ = StructureOver→Catᴰ})
```
felixwellen, line 55: Name...

Cubical/Categories/Displayed/Constructions/LeftAdjointToReindex.agda
```diff
@@ -0,0 +1,106 @@
...
+    isPropHomP : ∀ {A B A' B'} (p : A ≡ A') (q : B ≡ B') →
+                (f : D [ A , B ]) → (f' : D [ A' , B' ]) →
+                isProp (hom-pathP p q f f')
+    isPropHomP pdom pcod f f'' x y =
+      isoFunInjective (PathPIsoPath _ _ _) x y fromPathPeq
+      where
+      fromPathPeq : fromPathP x ≡ fromPathP y
+      fromPathPeq = D .isSetHom _ _ (fromPathP x) (fromPathP y)
+
+  ∃F : Categoryᴰ D (ℓ-max (ℓ-max ℓC ℓD) ℓDᴰ) (ℓ-max (ℓ-max ℓC' ℓD') ℓDᴰ')
```
felixwellen, line 54: Maybe this would be a good place for a comment which says that ```∃F``` is adjoint to reindexing (and just to be sure to say what reindexing means).

felixwellen (changes requested):
Looks all good to me - I just have a couple of documentation/naming comments.

Cubical/Categories/Constructions/TotalCategory.agda
```diff
@@ -0,0 +1,74 @@
...
+  ∫F .F-hom (_ , fᴰ) = _ , Fᴰ.F-homᴰ fᴰ
+  ∫F .F-id = ΣPathP (_ , Fᴰ.F-idᴰ)
+  ∫F .F-seq _ _ = ΣPathP (_ , (Fᴰ.F-seqᴰ _ _))
+
+-- Projections out of the total category
+module _ {C : Category ℓC ℓC'} {Cᴰ : Categoryᴰ C ℓCᴰ ℓCᴰ'} where
+  open Functor
+  open Functorᴰ
+
+  Fst : Functor (∫C Cᴰ) C
```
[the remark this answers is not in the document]
stschaef: Sounds good. We were trying to keep consistent with `Categories.Constructions.BinProduct`, which has the first projection functor out of the product category as `Fst` as well. 
We should probably then change the instance in `BinProduct` too then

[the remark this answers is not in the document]
stschaef: Although I don't immediately see any good candidate for `SOMETHING`. Maybe something like `fstProjection` or `π₁F`?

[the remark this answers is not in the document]
maxsnew: I think we should either leave it as `Fst` to match BinProduct or change both of them to `fst` to match Data.Sigma and then use them as qualifies imports when you need to disambiguate

Cubical/Categories/Displayed/Constructions/StructureOver.agda
maxsnew: isFaithfulFst?

Cubical/Categories/Displayed/Constructions/LeftAdjointToReindex.agda
stschaef: Does this sound alright?
```
  -- Reindexing a dislayed category Dᴰ over D along a functor F : C → D
  -- gives a displayed category over C
  -- Fiberwise pullback the objects over D along F to display them over C
  --
  --    reindex Dᴰ F                Dᴰ
  --         _                      _
  --         |                      |
  --         v           F          v
  --         C -------------------> D
  --
  -- Which may be read as a 2-functor from displayed categories over D to
  -- displayed categories over C. This operation has a left 2-adjoint, which
  -- we call ∃F, that maps displays over C to displays over D
  --
  --         Cᴰ                  ∃F Cᴰ F
  --         _                      _
  --         |                      |
  --         v           F          v
  --         C -------------------> D
  --
```

Cubical/Categories/Displayed/HLevels.agda
stschaef: Added

Cubical/Categories/Displayed/Constructions/TotalCategory.agda
```diff
@@ -0,0 +1,69 @@
...
+open import Cubical.Categories.Displayed.Base
+open import Cubical.Categories.Displayed.Functor
+open import Cubical.Categories.Displayed.Instances.Terminal
+open import Cubical.Categories.Constructions.TotalCategory hiding (intro)
+
+private
+  variable
+    ℓC ℓC' ℓD ℓD' ℓE ℓE' ℓCᴰ ℓCᴰ' ℓDᴰ ℓDᴰ' ℓEᴰ ℓEᴰ' : Level
+
+-- Displayed total category, i.e. Σ for displayed categories
```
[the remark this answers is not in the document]
stschaef: Added this as well
```
-- Displayed total category, i.e. Σ for displayed categories
--
-- The ordinary total category (∫C Cᴰ) has as objects
-- pairs (x, xᴰ) where x is an object of C and xᴰ is an object of Cᴰ over x.
--
-- Whereas if we had a category Dᴰ displayed over (∫C Cᴰ),
-- and a category Cᴰ displayed over C, then the
-- displayed total category (∫Cᴰ Cᴰ Dᴰ) likewise has pairs
-- as displayed objects.
--
-- In the displayed total category, we have objects (xᴰ, yᴰ) displayed
-- over x, where x is an object of C, xᴰ an object in Cᴰ displayed over x,
-- and yᴰ is an object of Dᴰ over (x, xᴰ).
```

Cubical/Categories/Displayed/Constructions/LeftAdjointToReindex.agda
maxsnew: typo: `dislayed`

Cubical/Categories/Displayed/Constructions/TotalCategory.agda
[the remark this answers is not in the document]
felixwellen: Yes that's great!

Cubical/Categories/Displayed/Constructions/LeftAdjointToReindex.agda
felixwellen: Very good!

Cubical/Categories/Constructions/TotalCategory.agda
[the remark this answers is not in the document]
felixwellen: I see. I think the solution with qualified imports is better, guessing that it is very unlikely that you want to have both available, ```fst``` from ```Data.Sigma``` and this one. For example in this file, it makes a lot of sense to me to qualify ```Data.Sigma```.  Feel free to disagree and go with some other name, my main problem with ```Fst``` is that it starts in upper case for disambiguation (maybe we should fix a convention for disambiguation which is already in use, like appending apostrophies to a name). Another option might be to use ```π₁```.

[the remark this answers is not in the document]
stschaef: Qualified imports could work here, but they cause more problems than the seem to solve. Choosing the name `fst` for both here and `BinProduct` requires renamings in both `Data.Sigma` and `Prelude` because Sigma types are defined as primitives. 

That is, anytime we simultaneously import any of `Prelude`, `Data.Sigma`, `Constructions.TotalCategory`, or `Constructions.BinProduct` we need to add several qualified imports. This is every import site for `BinProduct`. These qualified imports unnecessarily complicates imports in files that are unrelated to this PR. It seems like `Fst` is preferable to make this PR a more modular addition to the repo. 

I think it makes the most sense to keep the naming of this functor aligned with the projection in `BinProduct`. So as an alternative, we can go with `π₁` in each of these spots and correspondingly adapt any code that imports `BinProduct`

Cubical/Categories/Displayed/Constructions/StructureOver.agda
stschaef: added

Cubical/Categories/Constructions/TotalCategory.agda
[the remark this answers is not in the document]
maxsnew: I don't think what you're describing is actually a problem, Steven, it's normal to have to use a lot of qualified imports IMO.

But I agree that this should really be combined with fixing BinProduct as well in a separate PR. So I recommend not changing the names for now and then making an issue where we can bikeshed this more.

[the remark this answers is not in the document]
felixwellen: Yes - agreed :-)
I'll merge it.

PR #1122, opened 2024-04-18
Lift experiment

mzeuner:
Is it worth a PR?

PR #1123, opened 2024-04-24
Properties  of the lifting functor

mzeuner:
This PR proves that lifting functor from sets in a universe to sets in the successor universe is fully faithful and preserves (small) limits. It is a less opaque rewrite of the accidental https://github.com/agda/cubical/pull/1122.


mortberg (approved):
Looks good to me

PR #1125, opened 2024-05-10
Topological modalities

awswan:
I needed some results about topological modalities for a project, and I think it makes sense to add these to the library in case anyone else wants to use them. I'm not 100% sure about the naming, so feel free to change if necessary.  There are a couple of naming clashes that come from the literature, so there's not  much I can do about them: 1. Separated is a general definition that applies to any modality, whereas in the library (and many places in the literature) it is specific to the double negation modality; 2. So far Injective in the library is only referring to functions, I think, but I also use it for injective objects, as in https://ncatlab.org/nlab/show/injective+object .

### DIFF
--- a/Cubical/HITs/Nullification/Topological.agda
+++ b/Cubical/HITs/Nullification/Topological.agda
@@ -1,6 +1,7 @@
 {-# OPTIONS --safe #-}
 open import Cubical.Foundations.Prelude
 open import Cubical.Foundations.Equiv
+open import Cubical.Foundations.Equiv.Fiberwise
 open import Cubical.Foundations.Equiv.Properties
 open import Cubical.Foundations.Function
 open import Cubical.Foundations.HLevels
@@ -8,6 +9,7 @@ open import Cubical.Foundations.Univalence
 
 open import Cubical.Data.Sigma.Properties
 open import Cubical.Data.Unit
+open import Cubical.Data.Nat using (ℕ; zero; suc)
 
 open import Cubical.HITs.Nullification.Base
 open import Cubical.HITs.Nullification.Properties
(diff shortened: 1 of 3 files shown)


Cubical/HITs/Nullification/Topological.agda
```diff
@@ -0,0 +1,110 @@
...
+  equivPreservesIsNull (invEquiv e) (isNullEquiv (snd X) (snd Y))
+  where
+    e : (X ≡ Y) ≃ ((fst X) ≃ (fst Y))
+    e =
+      X ≡ Y ≃⟨ invEquiv (Σ≡PropEquiv λ _ → isPropIsNull) ⟩
+      (fst X) ≡ (fst Y) ≃⟨ univalence ⟩
+      (fst X) ≃ (fst Y) ■
+
+module _ (isPropS : (α : A) → isProp (S α)) where
+  {- Recall that a type Z is injective when can extend any map S α → Z to
```
felixwellen, line 57: Typo, "when we can extend"

Cubical/HITs/Nullification/Topological.agda
```diff
@@ -0,0 +1,110 @@
...
+  ℓ ℓ' : Level
+
+{-
+  We use the formulation of lexness that the universe of null types is
+  null itself.
+
+  We choose the level of the universe of null types to be as low as
+  possible such that it is null when the generators are propositions.
+-}
+nullTypes : Type (ℓ-max (ℓ-suc (ℓ-max ℓ ℓs)) ℓα )
```
felixwellen, line 40: "NullType" would be more in line with the naming conventions.

felixwellen (changes requested):
Just a typo and a renaming suggestion, otherwise ready to merge. Thanks for contributing!

awswan:
Thanks! I've made both those changes.

PR #1130, opened 2024-05-29
No-eta for CommAlgebras

felixwellen:
Just an experiment to see if it helps with tc speed when CommAlgbras have no eta law.


felixwellen:
tc speed seems to be only minimally affected: Cubical.AlgebraicGeometry.ZariskiLattice.StructureSheafPullback check a bit more than a second faster and I didn't see anything with a bigger change.

felixwellen:
It might make more of a difference to have no-eta for Algebras (because e.g. type checking CommAlgebra-homomorphism compositions should actually be about Algebras being equal...)

felixwellen:
obsoleted by #1145

PR #1132, opened 2024-06-01
Add `Π-contractDom`

ncfavier:
Analogous to [`Σ-contractFst`](https://github.com/agda/cubical/blob/60f18987bb1819a15fccc325343ef7b469bb2eeb/Cubical/Data/Sigma/Properties.agda#L318-L328). Not sure where this should live; can't go in `Cubical.Foundations.Function` because of cyclic imports.


mortberg (approved):
Looks good to me!

PR #1134, opened 2024-06-18
Simplify proof that cong ⟨_⟩ is injective on groups

ecavallo:
We can use `declareRecordIsoΣ` for this.


felixwellen:
Thanks @ecavallo !
I'll merge it....

PR #1137, opened 2024-06-19
Algebra structures without eta

felixwellen:
I deactivated the eta for the records IsCommAlgebra and AlgebraStr and fixed the problems it causes. This does not seem to have tc speed impact, but I think it is still better because type checking will fail in a situation where two Algebra's are normalized for comparison (instead of taking forever).


felixwellen:
depends on  #1130

felixwellen:
outdated experiment -> closing

PR #1138, opened 2024-06-21
Another minor universe level generalisation

awswan:
Another minor generalisation in universe levels, this one in Unit.Properties. I seem to send a lot of these, but they are all things that actually came up while I was using the library.


awswan:
Less sure about this after noticing it breaks stuff, but it is following the principle of maximum universe generality.

awswan:
Sorry for sending two commits to fix the same thing after sending the pull request. I'm satisfied now and I'll stop touching it.

mortberg:
Looks good to me! Thanks for catching these small things

PR #1139, opened 2024-06-24
Add --guardedness globally and remove Codata.Everything

mortberg:
Any objections to this? As far as I understand --guardedness is safe and doesn't interfere with anything, but the Codata folder needs it. 

This should allow us to simplify the build process in a future PR as we just need to generate the Everything files without any special treatment for some folders


mortberg:
Ps: me looking into this was prompted by https://github.com/agda/cubical/pull/1129 which would benefit from global `--guardedness`

Issue #1147, opened 2024-08-10
Formalizing Devalapurkar & Haine


Trebor-Huang:
I'm making some headway in formalizing the paper

- Devalapurkar and Haine, [On the James and Hilton–Milnor splittings, & the metastable EHP sequence](https://people.math.rochester.edu/faculty/doug/otherpapers/DEVALAPURKAR-HAINE.pdf).

And I'd like to contribute this to the cubical library. But before that I want to confirm I'm not repeating work, and that these are suitable for the library. Here's a list of things to prove:

- Main results:
  - Fundamental James Splitting: ΣΩΣX ≃ ΣX ∨ (X ∧ ΣΩΣX).
  - Fundamental Hilton–Milnor Splitting: Ω(X ∨ Y) ≃ ΩX × ΩY × ΩΣ(ΩX ∧ ΩY).
  - These both have infinitary versions, but the proof in Devalapurkar and Haine reduces them to presheaf toposes, which are hypercomplete. I wonder if it is provable in HoTT without additional assumptions.
  - Metastable EHP: I think this is out of reach.
- Funny intermediary results:
  - The pushout of X ← X ∨ Y → X is contractible (immediate from pasting lemma).
  - Σ(X ⋀ Y) ≃ X join Y (done).
  - the fiber of X ∨ Y → X × Y is Σ(ΩX ∧ ΩY) (done except pasting lemma).
  - If A is the fiber of B → C, and ΩB → ΩC has a section, then ΩB is equivalent to ΩA × ΩC. (I tried to prove it directly but it seems to require a large coherence result between 2-paths.)
- Lemmas that I expect to be in the library, but I can't find (if they actually exist feel free to link):
  - Suspension is equivalent to pushout of 1 ← X → 1.
  - Pushout of 1 ← 1 → X is X; Pushout of 1 ← X = X is 1.
  - The pasting lemma for pushouts. Seems like it is already in #1133, so maybe we can cherrypick?

PR #1151, opened 2024-08-31
Pi4S3 Paper

aljungstrom:
This PR adds the formalisation behind 
_Formalising and computing the fourth homotopy group of the 3-sphere in Cubical Agda_ by me and @mortberg

I've had this stuff lying around for months but I forgot to make a PR... The primary addition is the `Cubical/Papers/Pi4S3-JournalVersion.agda` which extends `Cubical/Papers/Pi4S3.agda`.

There is some new maths but nothing that should be interfering. Most of it is in the new file `Cubical/HITs/Sn/Multiplication.agda`. It contains the 

There's also the equivalence between suspensions of smash products and joins (in  `Cubical/HITs/Sn/SmashProduct/Base.agda`.) which I saw was mentioned in Issue #1147 . I hope this doesn't cause any problems for @Trebor-Huang (please tell me in that case and we can probably change it back to your version).


Trebor-Huang:
I think I proved it by pasting a lot of pushout diagrams. If there is a cubical flavored proof that's certainly better, though I might have relied on some judgmental equalities when further pasting the thing with other squares. It shouldn't be too much trouble to adjust though since squares are very modular.

Trebor-Huang:
Somehow agda took a lot longer checking one particular judgemental equality when I swapped the proof out. Should I be concerned?

aljungstrom:
@Trebor-Huang Hmm.. Maybe Agda is too keen on unfolding this definition or something? Could you point me to where in your code this happened?

PR #1120, opened 2024-04-02
Base change, sliced adjoints

marcinjangrzybowski:

PR #1152, opened 2024-09-02
Univalent Category of SETOIDs ,  Setoids are not LCCC

marcinjangrzybowski:
### Cubical/Relation/Binary/Setoid.agda
* Setoid as Pair of hSet and propositional equivalence relation
### Cubical/Categories/Instances/Setoids.agda
* Univalent Category of Setoids 
* Adjoint quadruple of functors from/to SET : Quot ⊣ IdRel ⊣ Forget ⊣ FullRel (crude, basic cohesion?)
* `¬BaseChange⊣SetoidΠ` - Base change functor does not have right adjoint (so SETOID cannot be LCCC, at least not in the literal sense) implementation of `Setoids are not an LCCC` by Thorsten Altenkirch and Nicolai Kraus (https://www.cs.nott.ac.uk/~psznk/docs/setoids.pdf)


marcinjangrzybowski:
Depends on https://github.com/agda/cubical/pull/1120

maxsnew:
A couple of things before a more detailed review: 
1. I don't think we should commit to the terminology "Setoid" for this, as often a Setoid is defined as coming with a `Type`-valued relation rather than a `Prop`-valued relation. I would suggest we just call these "equivalence relations" or even "thin groupoid"
2. The link https://people.cs.nott.ac.uk/psznk/docs/setoids.pdf doesn't work for me, I get a 404

marcinjangrzybowski:
1. Updated the link to working one: https://nicolaikraus.github.io/docs/setoids.pdf
2.I will change terminology.

Cubical/Categories/Category/Base.agda
```diff
@@ -154,10 +156,29 @@ _⋆_ (ΣPropCat C P) = _⋆_ C
...
               → isIso C f → isIso (ΣPropCat C P) {x , p} {y , q} f
 inv (isIsoΣPropCat p q f isIsoF) = isIsoF .inv
 sec (isIsoΣPropCat p q f isIsoF) = isIsoF .sec
 ret (isIsoΣPropCat p q f isIsoF) = isIsoF .ret
+
+ΣℙCat : (C : Category ℓ ℓ') (P : ℙ (ob C)) → Category ℓ ℓ'
+ΣℙCat = ΣPropCat
+
+isSmall : (C : Category ℓ ℓ') → Type ℓ
+isSmall C = isSet (C .ob)
```
anshwad10, line 171: I don't think `isSmall` is a good name; These categories are conventionally called 'strict' categories, and calling them 'small' gives the impression that this is about universe level, which it's not.

marcinjangrzybowski: how about `isStrict` ? I do not have anything more descriptive in mind

anshwad10: yes, that's also [what the 1lab calls it](https://1lab.dev/Cat.Strict.html).

PR #1153, opened 2024-09-06
Rework displayed category reasoning.

jpoiret:
This small PR improves the reasoning system for displayed categories, both in terms of ergonomy and performance. Proofs that were previously costly in terms of memory and time are now near instantaneous, and the proof terms are much more focused and readable.

We don't have many displayed category equality proofs right now, and I think this was probably due to how hard they were to use.


maxsnew:
This seems to have conflicts with the just merged https://github.com/agda/cubical/pull/1149 can you resolve them?

PR #1154, opened 2024-09-07
Basic Order Theory

LuuBluum:
Introduces basic order theory (basics of lattices, downsets, upsets, poset closures; defines galois connections as well). Also cleans up a couple of order-related files for naturals and rationals and adds some useful lemmas to some algebra structures for use later. Renames and cleans up the various orderings as well. Also proves that embeddings form a complete lattice. Defines preimages as well.


LuuBluum:
Note that while this is not immediately related to #1181, it does help set up some additional foundational stuff for orderings in general, such as basic definitions of lattices.

felixwellen (approved):
Thanks for the contribution! And sorry for the long waiting time.
Looks all good to me.

PR #1158, opened 2024-09-24
WellFounded: Well-founded relations are irreflexive.

shlevy:


Cubical/Induction/WellFounded.agda
```diff
@@ -3,6 +3,7 @@
 module Cubical.Induction.WellFounded where
 
 open import Cubical.Foundations.Prelude
+open import Cubical.Relation.Nullary
```
anshwad10, line 6: this import is unnecessary; you don't even use `isIrrefl` later in the file.

anshwad10: oh right it was needed for negation

PR #1163, opened 2024-10-28
Categories of propositions and sieves

anuyts:


Cubical/Categories/Instances/SievesAsFunctors.agda
```diff
@@ -0,0 +1,37 @@
+{-# OPTIONS --safe #-}
+
+module Cubical.Categories.Instances.SievesAsFunctors where
+
+-- There are also sieves in Cubical.Site.Sieve
```
felixwellen, line 5: I think it would be good to also have a pointer there.
Maybe it would also be good to mention that the definitions are equivalent.

felixwellen (changes requested):
Just a minor documentation request...

anuyts:
After a brief discussion with @maxsnew , I think it's better to define PROP as the total category of a displayed category over SET, so maybe we shouldn't go with this pull request at all.

PR #1172, opened 2024-11-25
Indexed W-types: hlevel without univalence.

anuyts:
Use univalence sparingly, in anticipation of https://github.com/agda/agda/issues/3750.


felixwellen:
I guess less dependencies are always good...

PR #1173, opened 2024-11-27
fix incorrectly named isIsoToIso, add IsoToIsIso

maxsnew:
I had misnamed this function, oops


felixwellen:
I'm a bit confused that there is `IsoToIsIso` but also `isoToPath`. I'm not sure if we ever came up for a good rule for the case when you have something which should start in lower case, but starts with a word that is usually in uppercase. I prefer using uppercase then, so I agree and just merge.

PR #1174, opened 2024-11-28
A few minor utility functions

awswan:
I needed a few minor utility functions while I'm using the library, and it made sense to add them here, so here they are. I think these ones should be noncontroversial and I tried to match the notation of the file I was adding to, but let me know if there are any problems.


felixwellen:
Thanks!

PR #1180, opened 2025-01-15
Formalize Theorem 7.2.2 in the HoTT Book

anshwad10:


Cubical/Relation/Binary/Base.agda
```diff
@@ -205,6 +205,25 @@ module BinaryRelation {ℓ ℓ' : Level} {A : Type ℓ} (R : Rel A A ℓ') where
         q : isContr (relSinglAt a)
         q = isOfHLevelRespectEquiv 0 (t , totalEquiv _ _ f λ x → invEquiv (u a x) .snd)
                                    (isContrSingl a)
+  -- Theorem 7.2.2 in the HoTT Book
```
felixwellen, line 208: Could this be moved to `Cubical.Relation.Binary.Properties`? I think that would be a better place if there is no good reason to have it in `Base`.

felixwellen:
Thanks!

PR #1184, opened 2025-02-09
Move `factorial` from `Cubical.Data.Fin.LehmerCode` to `Cubical.Data.Nat.Properties`

anshwad10:


Cubical/Data/Fin/LehmerCode.agda
```diff
@@ -22,6 +22,7 @@ open import Cubical.Relation.Nullary
 open import Cubical.Data.Unit as ⊤
 open import Cubical.Data.Empty as ⊥
 open import Cubical.Data.Nat
+open import Cubical.Data.Nat.Properties using (factorial) public
```
felixwellen, line 25: It would be better to make modules use factorial more directly, i.e. import it from Data.Nat.Properties.

felixwellen: Try to remove the `public` to see if it is still used.

PR #1185, opened 2025-02-12
Fix missed rename of StrictPoset to Quoset

LuuBluum:
When renaming StrictPoset to Quoset, missed a function's name during the rename. This fixes the missed rename.


mortberg:
Thanks! Good catch

PR #1188, opened 2025-02-16
Add Rezk Completion by HIT

anshwad10:


anshwad10:
Just noticed that the HIT can be replaced with a [GroupoidQuotient](https://agda.github.io/cubical/Cubical.HITs.GroupoidQuotients.html). This would have saved me from writing so many elimination principles if I had found it earlier

PR #1195, opened 2025-02-22
Strict monoidal categories are strict categories #1194

anshwad10:


anshwad10:
I did not realize renaming the branch would close the PR

Issue #753, opened 2022-03-29
Strict monoidal categories lack coherence laws


anuyts:
In Cubical.Categories.Monoidal.Base, there is a type `MonoidalStr` of non-strict monoidal structures and a type `StrictMonStr` of strict ones. One would expect the following results:
* Strict structures give rise to non-strict ones
* For univalent categories, this is an isomorphism.

However, I think the first result doesn't even hold (unless your object type is a set) as the strict structure lacks coherence laws.

anshwad10:
This can be closed since #1196 was merged

PR #1196, opened 2025-02-22
Strict monoidal categories are strict categories

anshwad10:


anshwad10:
#753

ecavallo:
The change looks good, but I don't know what the intent behind this definition was in the first place (since it is barely used). There are two possibilities here:

1. A `StrictMonCategory` is a monoidal {strict category}. This is the interpretation taken by this PR.
1. A `StrictMonCategory` is a `MonoidalCategory` where the coherence isomorphisms are identities. This the interpretation suggested by @anuyts in #753.

Opinions from @barrettj12 or @anuyts? If there are no objections I will just merge.

anuyts:
I think this is wrong before and after. The unit and associativity laws should concern the entire functors, not just their object part. The current formulation is only correct on [thin/posetal categories](https://ncatlab.org/nlab/show/thin+category).

ecavallo:
As it stands the PR does assume the type of objects is an h-set, as part of the definition of `IsMonoid`. This is what I meant by strict category in my option "1".

anshwad10:
Yeah,  I had forgotten to axioms for the morphisms too (so I added it now). I think strict monoidal categories should have underlying strict categories, which is why I required the type of objects to be a set instead of assuming more coherence laws.

anuyts:
Ok, that makes sense.

PR #1103, opened 2024-02-15
move precategories to separate folder and rename to wild categories

felixwellen:
refactor (split into files)
rename
add instance for Types

PR #1200, opened 2025-02-28
Missed rename of Precat to WildCat

anshwad10:


anshwad10:
@felixwellen missed renaming `SymmetricMonoidalPrecat` to `SymmetricMonoidalWildCat` in #1103.

Cubical/Papers/SmashProducts.agda
```diff
@@ -41,10 +41,10 @@ open WCat2 using (PointedCat)
 open WCat3 using (isMonoidalWildCat)
 
 -- Definition 4 (Symmetric Monoidal wild categories)
-open WCat3 using (SymmetricMonoidalPrecat)
+open WCat3 using (isSymmetricWildCat)
 
 --- 2.2 Smash Products
--- Definition 5 (Smash products -- note: defined as a coproduct in the
```
felixwellen, line 47: We should leave Axel's formulation as it is - note that he *uses* a pushout to construct a coproduct in pointed types.

anshwad10: but calling it a coproduct doesn't make sense - the smash product is a coproduct neither in the (wild) category of types nor in pointed types. I am pretty sure calling it a coproduct was a typo/mistake

felixwellen: Ah, ok - I didn't read carefully enough. But since it is a collection of results from a paper (and it could be that he wanted to point out that it is a definition where a coproduct is used), let's better check with @aljungstrom.

PR #1205, opened 2025-05-09
Prep for Agda 2.8.0: remove some spurious `private`

andreasabel:
Adress new/stricter warnings of Agda 2.8.0:
- `private` where it has no effect
- `open` in `abstract` block


andreasabel:
Merged this as I need it for the release of Agda-2.8.0.

PR #1212, opened 2025-05-28
Define multicategories

anshwad10:
This is a generalization of a multicategory like https://gist.github.com/plt-amy/174cf949e4cad7a24c1c73d0cdc6c35b However, I didn't require `isArity` to be a proposition so that the finite ordinals also qualify as arities. This is a generalization of both non-symmetric and symmetric multicategories, and it also allows multicategories with infinite-arity morphisms. I am currently not sure of the relation to Tom Leinster's generalized multicategories.


anshwad10:
I'll add more when the next version of cubical is supported and the error is fixed

anshwad10:
> I am currently not sure of the relation to Tom Leinster's generalized multicategories.

I think Tom Leinster's multicategories are actually more general than these: Given some `S : ArityType` there is a (wild) monad `M` on `Type` defined by `M X = Σ[ I ∈ Arity ] ⟨ I ⟩ → X`; `isArity-⊤` provides the unit, `isArity-Σ` gives the join, and `isArity-idl` `isArity-idr` `isArity-assoc` give the monad laws. An `S`-multicategory in this sense is precisely a `M`-multicategory in their sense
So  I'm going to try again with a more general definition parameterized by an arbritrary wild monad

PR #1231, opened 2025-07-22
Spectrum maps and fiber spectrum

mzhang28:
More spectrum stuff based on Floris' work.


anshwad10:
> But here is the idea: You generalize from pointed types to the arrow category - then pointed types are included as maps 1 -> A. Squares between those maps are pointed maps. If you now switch from maps to type families, you naturally arrive at a definition of pointed map with definitionally associative composition.

Interesting; I think if you normalized this, the definition of pointed map you get would be similar to the [Yoneda identity type](https://unimath.github.io/agda-unimath/foundation.yoneda-identity-types.html), because the type family corresponding to an element x : 1 -> A is precisely the identity type x =_ : A -> Type so for a function to preserve the basepoint means (z : A) -> pt A = z -> pt B = f z which kinda looks like the yoneda identity type (z : A) → (z ＝ x) → (z ＝ y) and their use cases are similar too (making associativity definitional)

PR #1219, opened 2025-06-23
Intervals in posets and the covering relation

anshwad10:
#1218

Issue #1218, opened 2025-06-23
Formalize 'Combinatorics of higher-categorical diagrams' by Amar Hadzihasanovic


anshwad10:
I think this is perfect for formalizing in cubical for a few reasons:

- From the introduction:
> Another point of departure with most earlier approaches[4] is that molecules are defined _inductively_, as produced by a number of “constructors”.
- Also from the introduction:
> For anything that concerns regular directed complexes, or any of the surrounding combinatorial structures, I attempted to be detailed and careful in excess of the expected standard for research-level mathematics.
- This is why univalence and inductive types will be helpful here:
> Comment 3.3.3 — We will prove many properties of molecules by induction on their construction. Since we only consider properties that are invariant under isomorphism, in these proofs we will not explicitly consider closure under isomorphisms.
(When defining the predicate 'isMolecule' we don't need to give cases for closure under isomorphisms due to univalence making it automatic)

I am going to formalize it over a few PRs, starting with oriented graded posets.

PR #1232, opened 2025-07-23
Define the category of posets

anshwad10:
This is part of my project to formalize [Hadzihasanovic's work](https://arxiv.org/abs/2404.07273) (#1218). The PR #1219 had a lot of different things and it was getting to big, so I'm splitting it.

### DIFF
--- a/Cubical/Categories/Instances/Posets.agda
+++ b/Cubical/Categories/Instances/Posets.agda
@@ -1,0 +1,25 @@
+{-# OPTIONS --safe #-}
+module Cubical.Categories.Instances.Posets where
+
+open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.Function as Function
+open import Cubical.Foundations.HLevels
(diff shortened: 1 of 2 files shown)


maxsnew:
Should this be called `POSET` or `PosetCategory` or `PosetsCategory`? I think the library is inconsistent about this now, i.e., we have `SET` and we have `PresheafCategory` but I don't think anything follows the structure of `PosetsCategory`.

anshwad10:
I was following the naming convention of [`CommRingsCategory`](https://agda.github.io/cubical/Cubical.Categories.Instances.CommRings.html#1142), [`LatticesCategory`](https://agda.github.io/cubical/Cubical.Categories.Instances.Lattices.html#268) etc. [`PosetCategory`](https://agda.github.io/cubical/Cubical.Categories.Instances.Poset.html#316) is already used for the construction of a thin category from a poset.

PR #1235, opened 2025-07-23
Define the downward and upward closure of a subset

anshwad10:
This is part of my project to formalize [Hadzihasanovic's work](https://arxiv.org/abs/2404.07273) (#1218). The PR #1219 had a lot of different things and it was getting to big, so I'm splitting it.


anshwad10:
I also added an eliminator and recursor for `∃`. I found it useful, and it was pretty easy to define using `uncurry` and the eliminator for propositional truncation.

maxsnew:
LGTM

Issue #259, opened 2020-03-29
Why is pathToEquiv/lineToEquiv in Agda.Builtin.Glue?


m-yac:
Seems like it better belongs in `Cubical.Foundations.Transport`, keeping it in a Builtin file makes it seem more mysterious than it is.

...Though maybe it's more mysterious than I think: Is the construction of `pathToEquiv` in `Agda.Builtin.Glue` any better than `isoToEquiv (pathToIso p)` in `Cubical.Foundations.Transport`? My understanding is that the key property `pathToEquiv` has is that `invEq (pathToEquiv)` is definitionally equal to `transport⁻ p`, but both have this property. Is the fiber contraction of `pathToEquiv` superior to `isoToEquiv (pathToIso p)` in some way?

mortberg:
@Saizan Is pathToEquiv/lineToEquiv used internally these days? I guess you might not use it anymore after implementing the direct definition for transport in hcomp U? If so then it could be moved out of Agda.Builtin.Glue.

I don't remember if `pathToEquiv` is better than the other definition... It looks kind of scary, but it might still be a bit smaller than `isoToEquiv (pathToIso p)`

Saizan:
No, it's not used internally anymore, we could move it.

I would guess that the inverse you get out of the equiv is simpler than what you get from `isoToEquiv` but I have not checked.

mortberg:
@Saizan What should we do about this issue?

Saizan:
Put the `pathToEquiv` code into this library somewhere?

mortberg:
Actually we rename it to `lineToEquiv`

https://github.com/agda/cubical/blob/master/Cubical/Core/Glue.agda#L40 

and this is then only used in one place in the whole library

https://github.com/agda/cubical/blob/master/Cubical/Foundations/Univalence.agda#L195

So we should probably just inline it there?


Note that there's of course a much simpler proof of this: https://github.com/agda/cubical/blob/master/Cubical/Foundations/Univalence.agda#L215

Do you think it would be worse to use that proof instead?

Saizan:
It would probably be worse to use `eqweqmap`, while `pathToIso` might be very close.

It would be fine to inline `lineToEquiv` into `pathToEquiv`.

mortberg:
Yeah, let's inline `lineToEquiv` in `pathToEquiv` then!

[...]

PR #1237, opened 2025-07-27
Improvements to Univalence.agda

anshwad10:
I had a few gripes with this file. First of all, I noticed that `ua→` and `ua→⁻` used transports unnecessarily, so I rewrote them and made it simpler, and while I was at it I also proved that `ua→` is an equivalence. For this I needed the lemma `ua-unglueIso` and I also added `commSqIsEq'` to `Equiv.agda` which proves the other triangle identity for equivalences. I also rewrote `uaInvEquiv` and `uaCompEquiv` to use a direct proof instead of `EquivJ`. I also noticed that almost all the definitions in the module took `{A B : Type ℓ}` as implicit arguments, so I made them into private variables.
Also, I noticed the comment `-- TODO: there should be a direct proof of this that doesn't use equivToIso` about `invEquiv` so I might fix that in this PR as well later.

### DIFF
--- a/Cubical/Foundations/Equiv.agda
+++ b/Cubical/Foundations/Equiv.agda
@@ -98,6 +98,19 @@ module _ {f : A → B} (equivF : isEquiv f) where
         })
       (commSqIsEq a i j)
 
+  commPathIsEq' : ∀ a → retIsEq (invIsEq a) ≡ cong invIsEq (secIsEq a)
+  commPathIsEq' a i j = hcomp (λ where
+      k (i = i0) → retIsEq (invIsEq (secIsEq a k)) j
+      k (i = i1) → invIsEq (secIsEq (secIsEq a k) j)
+      k (j = i0) → invIsEq (f (invIsEq (secIsEq a k)))
+      k (j = i1) → invIsEq (secIsEq a k)
+    ) (hcomp (λ where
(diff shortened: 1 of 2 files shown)


mortberg:
Sounds like very good improvements to me. Let me know when it's ready for review and I'll take a closer look

ecavallo:
(Mentioned this on Discord, repeating here for the record) We added a definition of `pathToEquiv` to the library with the intent of removing it from Agda.Builtin.Cubical: https://github.com/agda/cubical/issues/259. Is there anything stopping us from doing so now? (Which definition we should use here is a different question that I have not formed an opinion about.)

anshwad10:
The current definition of `pathToEquiv` in the library is not that good, because `equivTransp` That's why I though it would be better to use the builtin one.
 So I suppose I can inline that definition, or we can use `isoToEquiv pathToIso`. `isoToEquiv pathToIso` isn't perfectly optimized, because `isoToEquiv` has to mess up one of the units.
 I think I should use the definition currently in the builtin, because it also proves that isos have propositional fibres in the process which might be useful on its own.

anshwad10:
Once we have identity systems defined (#1254) I'll also prove that equivalences are an identity system and derive the computation rule for `EquivJ`

PR #1241, opened 2025-07-30
fixup: make generate-everything.sh predicative

ncfavier:
Due to the order in which shells do things, `find > foo` will find `foo`. This means that `generate-everything.sh` puts everything on the table including the table, resulting in a dependency cycle.

This change fixes that by excluding `Cubical/Everything.agda` from "everything". Also sorts the modules alphabetically and makes `make listing` display "Checking" messages so we know what it's doing.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -37,9 +37,8 @@ timings: clean
 
 .PHONY : listings
 listings: $(wildcard Cubical/**/*.agda)
-	rm -f Cubical/Everything.agda
 	./generate-everything.sh > Cubical/Everything.agda
-	$(AGDA) Cubical/Everything.agda -i. -isrc --html -v0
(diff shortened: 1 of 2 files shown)


foxyseta:
> Due to the order in which shells do things, `find > foo` will find foo.

Oops, you're right. Sorry about that.

felixwellen:
Haven't figured out yet where I can change that.

ncfavier:
There's supposed to be a ⚙️ next to it if you've got the rights (which I'm guessing you don't). Maybe @mortberg does?

ncfavier:
If not it might be an organisation-level thing, so maybe @andreasabel

felixwellen:
Yes, I don't seem to have the rights.

PR #1244, opened 2025-08-01
Pour `Categories.Commutativity` into `Categories.Morphism`

### LINKED ISSUE #1199: Categories.Commutativity only has one lemma
Why does the library have an entire file dedicated to a simple lemma? It would make sense to either add more theory of commutative squares or just move `compSq` to Categories.Morphism, where a few other lemmas (`invFlipSquare`) related to commutative squares also are.

anshwad10:
fixes #1199
I accidentally closed #1243 while trying to resolve a conflict


anshwad10:
I noticed that Agda was unable to infer the implicit argument `C` of `compSq`, so I'm making it explicit.

Issue #1181, opened 2025-01-15
Formalize Ordered CommRings


anshwad10:
Add a new module `Cubical.Algebra.OrderedCommRing` and create instances for the rationals and dyadic rationals

PR #1246, opened 2025-08-10
Definition of `OrderedCommRing`

LorenzoMolena:
The definition is adapted from ordered fields present in the HoTT Book (defn 11.2.7).
This also adds a pseudolattice type with structure. The property was already encoded in `Poset.Properties`, but the record was missing.


anshwad10:
Thanks! (this is because #1181 i presume?)

mortberg:
Looks good to me! Merging

Issue #1240, opened 2025-07-29
Category and Displayed Category Performance aka Should we no-eta-equality Category?


maxsnew:
We recently ran into major performance issues in github.com/maxsnew/cubical-categorical-logic and in the process of solving them, I forked `cubical` and tried out two breaking changes:
1. Add `no-eta-equality` to `Category`
2. Make some equalities `opaque`. Specifically `makeNatTransPath` and everything in `Displayed.Reasoning`.

These both led to improvements in ergonomics and performance. In `cubical-categorical-logic` we use `--lossy-unification` almost everywhere because inference of category parameters is abysmal without it. With `no-eta-equality`, these become unnecessary. Things that take a minute or longer to typecheck instead typecheck instantaneously.

I have less direct evidence that making equalities opaque sped things up but there's very little downside and it can certainly reduce some proof goals from blowing up.

Of course `no-eta-equality` is a breaking change as it breaks some definitional equalities of categories, especially involving `op` (like `C ^op ^op = C`). From my experience updating `cubical-categorical-logic` I think it is not that hard at all to add in explicit coercions or eta-expand to work around them. This shouldn't be surprising as we know that 1lab.dev works fine without it.

I can clean this up and make a PR but this is a big breaking change so I want to make sure that there are no objections on principle before I go about changing the library @ecavallo @jpoiret @felixwellen . I don't have any systematic performance measurements but I can include them when I make the PR.

jpoiret:
No objections here either. I'd be interested to know *exactly* what patterns make `eta-equality` explode in our faces if anyone knows.

PR #1247, opened 2025-08-12
Category no eta equality

maxsnew:
Implementing https://github.com/agda/cubical/issues/1240 . Haven't been able to test locally because checking `Cubical.Categories.Instances.CommAlgebras` seems to hang


maxsnew:
The comment I made is out of date. This PR is actually ready to merge. I had meant to do some performance numbers before we merge. Rebuilding the entire library from scratch the difference was negligible (~0.9% improvement) but I was able to remove one `--lossy-unification`.

PR #1249, opened 2025-08-13
standard bijections of Natural numbers

Freek98:
Added proofs showing that the type of natural numbers is isomorphic to it's own sum and product.

- Showed that if we have a function $f : \mathbb N \to \mathbb N$, which is increasing and starts at 0. We can then partition $\mathbb N$ into the pieces $[f k , f(k+1) )$. 
- Applying this fun fact to the double function, derived that $\mathbb N \cong \mathbb N + \mathbb N$
- Applying this fun fact to the function ($n \mapsto \frac{n (n+1)}{2} $), derived that $\mathbb N \cong \mathbb N \times \mathbb N$. 
- There are also some small lemmas on the order of natural numbers and increasing functions.

### DIFF
--- a/Cubical/Data/Nat/Bijections/Product.agda
+++ b/Cubical/Data/Nat/Bijections/Product.agda
@@ -6,12 +6,12 @@ open import Cubical.Data.Nat
 open import Cubical.Data.Sigma
 open import Cubical.Data.Nat.Bijections.Triangle
 
-Triangle⊂ℕ≅ℕ×ℕ : Iso Triangle⊂ℕ×ℕ (ℕ × ℕ)
-Iso.fun Triangle⊂ℕ≅ℕ×ℕ (_ , k , m , _) = m , k
-Iso.inv Triangle⊂ℕ≅ℕ×ℕ (m , k)         = m + k , k , m , refl
-Iso.rightInv Triangle⊂ℕ≅ℕ×ℕ _ = refl
-Iso.leftInv  Triangle⊂ℕ≅ℕ×ℕ (n , k , m , p) = J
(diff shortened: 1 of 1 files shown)


Cubical/Data/Nat/Bijections/Triangle.agda
```diff
@@ -0,0 +1,52 @@
...
+triangle (suc n) = n + suc (triangle n)
+
+strictIncTriangle : isStrictlyIncreasing triangle
+strictIncTriangle = sucIncreasing→StrictlyIncreasing triangle triangleN<triangleSN where
+  triangleN<triangleSN : (n : ℕ) → triangle n < triangle (suc n)
+  triangleN<triangleSN n = n , refl
+
+private
+  1+k+t=k+t+1 : (n : ℕ) → (t : ℕ ) → suc (n + t) ≡ n + suc t
+  1+k+t=k+t+1 n t = solveℕ!
```
anshwad10, line 26: The solver here is also overkill, considering that this is exactly the type of `+-suc`

Cubical/Data/Nat/Bijections/Sum.agda
```diff
@@ -0,0 +1,74 @@
...
+
+open import Cubical.Tactics.NatSolver
+open import Cubical.Data.Nat.Bijections.IncreasingFunction
+
+double : ℕ → ℕ
+double n = n + n
+
+private
+  2Sn=2n+2 : {n : ℕ} → double (suc n) ≡ double n + 2
+  2Sn=2n+2 = solveℕ!
```
anshwad10, line 24: This is also an instance of `+-comm` or `+-suc`, the solver is overkill

anshwad10:
Well, I had some nitpciks, but seems im too late because its already merged

Issue #1124, opened 2024-04-29
arithmetic operations of Cubical.Data.Rationals is super slow


onestruggler:
The the stdlib Data.Rational is fast. The follwing code compute instanrtly.

``` agda
{-# OPTIONS --safe --cubical --two-level #-}
module Scalar where
open import Data.Rational as ℚ hiding (_+_ ; _*_)
open import Data.Integer as ℤ hiding (_+_ ; _*_)

data ℚ[ω] : Set where
  _ω³+_ω²+_ω+_ : ℚ -> ℚ -> ℚ -> ℚ -> ℚ[ω]

infixl 6 _+_
infixl 7 _*_
infix 7 _s*_

_+_ :  ℚ[ω] ->  ℚ[ω] ->  ℚ[ω]
(a ω³+ b ω²+ c ω+ d) + (a' ω³+ b' ω²+ c' ω+ d') = (a ℚ.+ a') ω³+ (b ℚ.+ b') ω²+ (c ℚ.+ c') ω+ (d ℚ.+ d') 


_*_ :  ℚ[ω] ->  ℚ[ω] ->  ℚ[ω]
(a ω³+ b ω²+ c ω+ d) * (a' ω³+ b' ω²+ c' ω+ d') = a'' ω³+ b'' ω²+ c'' ω+ d'' where
  a'' = a ℚ.* d'  ℚ.+  b ℚ.* c'  ℚ.+  c ℚ.* b'  ℚ.+  d ℚ.* a'
  b'' = b ℚ.* d'  ℚ.+  c ℚ.* c'  ℚ.+  d ℚ.* b' ℚ.- a ℚ.* a'
  c'' = c ℚ.* d'  ℚ.+  d ℚ.* c' ℚ.- a ℚ.* b' ℚ.- b ℚ.* a'
  d'' = d ℚ.* d'  ℚ.-  a ℚ.* c'  ℚ.-  b ℚ.* b'  ℚ.-  c ℚ.* a'

_s*_ : ℚ -> ℚ[ω] -> ℚ[ω]
q s* (a ω³+ b ω²+ c ω+ d) = (q ℚ.* a) ω³+ (q ℚ.* b) ω²+ (q ℚ.* c) ω+ (q ℚ.* d)

Q0 = 0ℚ
P0 = (Q0 ω³+ Q0 ω²+ Q0 ω+ Q0) -- 0

[...]

PR #1258, opened 2025-09-05
More efficient implementation of /, % and GCD

LorenzoMolena:
The definition of `gcd` that uses well-founded induction and `%` from `Cubical.Data.Fin` has some performance issues.
I also noticed that `quotient_/_` and `remainder_/_` take quite some time to normalize, so I added alternative implementations, called `quotient'_/_` and `remainder'_/_`, using the built-in helper functions from `Agda.Builtin.Nat`. Lemmas about these helper functions are adapted from the standard library.

I only proved the necessary properties (`≡remainder'+quotient'` and `mod'<`) required to define the efficient version of GCD.
I have already tested whether the performance issues come solely from this, but it seems there are other factors as well, particularly the use of well-founded induction over the naturals.

In this PR, the new proposed implementation becomes the default one, but the previous implementation can still be found at the end of the `GCD` file, with a prime (`'`) symbol in the name.
The extensional equality between the two implementations is easily proved by the fact that `GCD m n` is a proposition.

This improvement enables fast fraction reduction and, combined with PR #1245, addresses efficiency for rational arithmetic (#1124)

@felixwellen, please let me know if this new implementation is appropriate.


Cubical/Data/Nat/GCD.agda
```diff
@@ -66,46 +69,94 @@ oneGCD m = symGCD (divsGCD (∣-oneˡ m))
...
+    g∣1+n            = w .fst .fst
+    g∣m%1+n          = w .fst .snd
+    d∣1+n→d∣m%1+n→d∣g = w .snd
+  in
+    (∣%→∣ g∣m%1+n g∣1+n , g∣1+n) ,
+    λ d' (d'∣m , d'∣1+n) → d∣1+n→d∣m%1+n→d∣g d' (d'∣1+n , (∣→∣% d'∣m d'∣1+n))
+
+-- putting it all together using an auxiliary variable to pass the termination checking
+
+euclid-helper< : (m n f : ℕ) → (n < m) → (m ≤ f) → GCD m n
```
felixwellen, line 127: private?

felixwellen:
Thanks - merging.

PR #1260, opened 2025-09-18
Add commutative ring and field instances for `Data.Rationals`

broughjt:
This PR adds commutative ring and field instances for the `Data.Rationals`
version of the rational numbers.

The existing commutative ring and field instances, which worked with the
`MoreRationals.QuoQ` version defined using the quotient integers, have been
moved into new modules (`QuoQRationals`). New instances for `Data.Rationals` have
been added in their place.

### DIFF
--- a/Cubical/Algebra/CommRing/Instances/QuoQRationals.agda
+++ b/Cubical/Algebra/CommRing/Instances/QuoQRationals.agda
@@ -1,0 +1,31 @@
+{-
+
+ℚ (the QuoQ version) is a Commutative Ring
+
+-}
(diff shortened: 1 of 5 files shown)


mortberg:
@LorenzoMolena What do you think about this? Anything that could be improved wrt efficiency or should I merge?

mortberg:
> and circling back to curently discussed PR, I do not have any notes, nor conflicts wiht our code, :) just wanted to chime in to give context about work wich is coming soon.

Ok, good! Then I guess this could be merged unless someone has some comments about it?

PR #1261, opened 2025-09-22
Nat and Int instances of some order structures 

LorenzoMolena:
This PR adds some missing instances of order structures for `Nat` and `Int`.
These instances are derived for simpler orders from more complex ones. For example, I defined `ℤ≤Poset = Toset→Poset ℤ≤Toset`, rather than manually constructing all the fields.

I am not sure if this is the preferred approach. If not, I also have a version ready where all instances are explicitly constructed.

Please let me know what you think.


mortberg (approved):
Looks good to me! What does @felixwellen think?

PR #1262, opened 2025-09-25
Add boolean builtins in `Nat`

LorenzoMolena:
This PR refactors the definitions of `min`, `max`, `discreteℕ`, `<-split`, and `_≟_`
to use boolean builtins, improving the performance of their computation. 
For `min` and `max`, this improvement applies unconditionally, while for the other
functions it depends on whether proof terms are ignored 
(e.g., when defining functions by cases via `_≟_`).

The main changes are:
- `Nat.Base`: imports the previously missing boolean builtins.
- `Nat.Properties` and `Nat.Order`: update the implementations of the functions listed above.
- `Prelude`: adds aliases for `isContrSingl ... .fst` and `isContrSinglP ... .fst`
  to support with-abstraction without losing information about the orginal terms
  (a workaround for the "with-abstraction equality" available in standard Agda but missing in Cubical).
- `DirectSum.Equiv-DSHIT-DSFun`, `Instances.NatMax`, and `Truncation.Properties`:
  adjusted for compatibility with the updated definitions.

Some notes:
- Only the properties of the boolean builtins necessary for the new implementations are proven.
- The previous implementation of `_≟_` remains in the file under the name `_≟'_`.
- Some comments and names may still need revision to align with the library’s conventions.

Please let me know if this structure and approach are appropriate.

### DIFF
--- a/Cubical/Foundations/Prelude.agda
+++ b/Cubical/Foundations/Prelude.agda
@@ -504,6 +504,8 @@ isContrSinglP A a .snd (x , p) i =
 infixl 0 _UsingEq
 infixl 0 _i0:>_UsingEqP
 
+-- Similar to `inspect`, but more convenient when `a` is not a function
+-- application, or when the applied function is not relevant
 _UsingEq : (a : A) → singl a
 a UsingEq = isContrSingl a .fst
 


felixwellen:
You can keep `UsingEq`, but make a comment there that it is similar to `inspect`.

PR #1268, opened 2025-10-24
Instances for `Pseudolattice` and `OrderedCommRing`

LorenzoMolena:
This PR adds instances for `Nat` and `Int` as `Pseudolattice`, and for `Int` as `OrderedCommRing`.
It also introduces a helper function constructing a `Pseudolattice` from a `Poset`, which is used here to define these instances.


Cubical/Relation/Binary/Order/Pseudolattice/Instances/Nat.agda
```diff
@@ -0,0 +1,37 @@
...
+open import Cubical.Data.Nat
+open import Cubical.Data.Nat.Order renaming (_≤_ to _≤ℕ_)
+
+open import Cubical.Relation.Binary.Order.Poset.Instances.Nat
+open import Cubical.Relation.Binary.Order.Pseudolattice
+
+open PseudolatticeStr
+
+private
+  minGLB : ∀ {m n x} → x ≤ℕ m → x ≤ℕ n → x ≤ℕ min m n
```
anshwad10: This seems like it would be useful in other places, I reccomend you make it `public` and move it to somewhere like `Data.Nat.Order` or `Data.Nat.Properties`

Cubical/Relation/Binary/Order/Pseudolattice/Base.agda
```diff
@@ -57,3 +60,33 @@ makeIsPseudolattice {_≤_ = _≤_} is-setL is-prop-valued is-refl is-trans is-a
     PS : IsPseudolattice _≤_
     PS .IsPseudolattice.isPoset = isposet is-setL is-prop-valued is-refl is-trans is-antisym
     PS .IsPseudolattice.isPseudolattice = is-meet-semipseudolattice , is-join-semipseudolattice
+
+module _
+  (P : Poset ℓ ℓ') (_∧_ _∨_ : ⟨ P ⟩ → ⟨ P ⟩ → ⟨ P ⟩) where
+  open PosetStr (str P) renaming (_≤_ to infix 8 _≤_)
+  module _
+    (π₁ : ∀ {a b x} → x ≤ a ∧ b → x ≤ a)
```
anshwad10: Wouldn't it be simpler to ask that `π₁ : ∀ {a b} → a ∧ b ≤ a` and use `≤-trans` to reconstruct it (and likewise for `π₂`, `ι₁`, and `ι₂`) ? That seems to be more convenient, considering that that is how you prove it in the `Nat` and `Int` instances anyways

PR #1182, opened 2025-01-22
Book cauchy reals

marcinjangrzybowski:
This is an implementation of real numbers as defined in Chapter 11.3: Cauchy Reals of the HoTT Book.
(Note: results 11.3.11 and 11.3.50 are currently missing.)

Modules:

- `Cubical/HITs/CauchyReals/Base.agda`
- `Cubical/HITs/CauchyReals/Closeness.agda`
- `Cubical/HITs/CauchyReals/Continuous.agda`
- `Cubical/HITs/CauchyReals/Inverse.agda`
- `Cubical/HITs/CauchyReals/Lems.agda`
- `Cubical/HITs/CauchyReals/Lipschitz.agda`
- `Cubical/HITs/CauchyReals/Multiplication.agda`
- `Cubical/HITs/CauchyReals/Order.agda`

These modules closely follow the book. However, I had to adjust the order of definitions due to the issue discussed here: [HoTT/book#899](https://github.com/HoTT/book/issues/899).

Additionally, the following modules:

- `Cubical/HITs/CauchyReals/Sequence.agda`
- `Cubical/HITs/CauchyReals/Derivative.agda`

are attempts to benchmark the definitions and evaluate their behavior. In `Derivative.agda`, I've managed to derive the Power rule (for natural exponents). In `Sequence.agda`, I defined `exp` (currently only for ℚ₊), allowing the definition of `e`.

To implement these properly (assuming we want to continue without countable choice or LEM), I plan to consult some books on constructive analysis. The HoTT Book references "Errett Bishop. Foundations of Constructive Analysis," but I welcome other suggestions.

While the development is not yet up to library standards, everything currently type checks. I will likely split this into smaller PRs. There is one remaining definition, `invℝ`, hidden behind `abstract`. However, I have a promising idea based on previous optimizations to expose it (or at least manage its evaluation more controllably by guarding it with `lockUnit`).

There is some redundancy, as I am still experimenting with different versions of definitions.

### TODO:
- Implement square root
- Implement chain rule
- Complete results 11.3.11 and 11.3.50 from the HoTT Book
- Extend `exp` to all reals and provide a characterization
- Properly generalize convergence tests and add a few more
- Define logarithm
- Parameterize definitions by implementation of Rationals
- Figure out a more elegant way to employ `CommRingSolver` (temporary solution in `Cubical/HITs/CauchyReals/Lems.agda`)

PR #1270, opened 2025-11-11
Fix to the definitions of `min` and `max`

LorenzoMolena:
While working on a future PR, I noticed that when performing a "with-abstraction with `UsingEq`" on `n <ᵇ m`, each case was not normalizing to the expected term. 
This issue is resolved by adding `UsingEq` into the definitions of `min` and `max`. Interestingly, all proofs that relied on "with-abstraction *withot* `UsingEq`" still works unchanged. 

I added a comment about this limitation above the definition of `UsingEq`, but I'm not entirely sure how clear it is, so I also referenced its use in `min` and `max` over ℕ. 
Please let me know if there is a clearer way to explain this behavior.

### DIFF
--- a/Cubical/Data/Nat/Properties.agda
+++ b/Cubical/Data/Nat/Properties.agda
@@ -23,9 +23,9 @@ private
 min : ℕ → ℕ → ℕ
 min zero m = zero
 min (suc n) zero = zero
-min (suc n) (suc m) with n <ᵇ m
-... | false = suc m
-... | true  = suc n
+min (suc n) (suc m) with n <ᵇ m UsingEq
+... | false , _ = suc m
+... | true  , _ = suc n
 
 minSuc : min (suc n) (suc m) ≡ suc (min n m)
 minSuc {zero} {zero} = refl
@@ -44,9 +44,9 @@ minComm (suc n) (suc m) = minSuc ∙∙ cong suc (minComm n m) ∙∙ sym minSuc
 max : ℕ → ℕ → ℕ
 max zero m = m
 max (suc n) zero = suc n
-max (suc n) (suc m) with n <ᵇ m
-... | false = suc n
-... | true  = suc m
+max (suc n) (suc m) with n <ᵇ m UsingEq
+... | false , _ = suc n
+... | true  , _ = suc m
 
 maxSuc : max (suc n) (suc m) ≡ suc (max n m)
 maxSuc {zero} {zero} = refl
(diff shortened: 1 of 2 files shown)


mortberg:
Why not just define min and max by recursion without any `with` or calls to `<`?

mortberg:
So having `min (suc n) (suc m) = min n m` and similarly for max

anshwad10:
Also, I think this is precisely the advantage of using `inspect` over `UsingEq`, because when using `inspect` we abstract on the value of `n <ᵇ m` and the path separately

Cubical/Data/Nat/Properties.agda
```diff
@@ -23,9 +23,9 @@ private
 min : ℕ → ℕ → ℕ
 min zero m = zero
 min (suc n) zero = zero
-min (suc n) (suc m) with n <ᵇ m
-... | false = suc m
-... | true  = suc n
+min (suc n) (suc m) with n <ᵇ m UsingEq
```
anshwad10: Like `discrete-ℕ` in the previous PR, but I think this case-split also seems to be redundant: I'm pretty sure that
```agda
min : ℕ → ℕ → ℕ
min n m with n <ᵇ m UsingEq
... | false = m
... | true  = n
```
would have exactly teh same behaviour because `<ᵇ` would compute away when its argument is zero. Perhaps change this too? (likewise for `max`)

anshwad10:
> Ah right, I had forgotten about that. I wonder how much one should worry about efficiency for functions on unary nat though... If we want things to be fast we should at least have a binary representation?

I don't know if that would help; It wouldn't be better in terms of storage at least, because [Agda compiles Nat to built in arbitrary-precision binary integers](https://agda.readthedocs.io/en/latest/tools/compilers.html#builtin-natural-numbers), and idt it would speed up arithmetic because the builtin operations are O(1) while binary arithmetic is O(logN)

marcinjangrzybowski:
@mortberg this is part of work to enable fast operations on Integers -> Rationals -> and finaly beeing able to quickly calculatate things like Nth digit of `e` in https://github.com/agda/cubical/pull/1182 :)
@LorenzoMolena identified bootlenecks, and we have very promising experiments.
We discussed those plans with @felixwellen , but are happy to meet @mortberg to get the feedback on our plan.
With @LorenzoMolena , we will be doing code sprint hopefully integrating this work, 2 weeks from now during AIM.

PR #1272, opened 2025-11-14
Properties of `Pseudolattice`

LorenzoMolena:
Update to the `Pseudolattice` module. 
It introduces the definition of equivalences between pseudolattices, following the pattern used in the other order structures, and adds a basic `Properties` module containing the results that seem most generally useful.


LorenzoMolena:
@mortberg I fixed the conflicts

Issue #1236, opened 2025-07-26
The powerset is not universe polymorphic enough


anshwad10:
Currently, for `A : Type l`, `ℙ A` is defined as `A -> hProp l`, however I frequently find myself needing to use `A -> hProp l'` where `l'` is greater than the universe level of `A`. I could use `Embeddings` instead which are maximally universe polymorphic, however I run into a problem: `Subset→Embedding` use `ℙ`, so I can't use that without compromising the universe polymorphism. Also, this causes more problems in the rest of the library: for example, subgroups and ideals are not maximally universe polymorphic because they use `ℙ`, and neither is `ΣPropCat` (though that is redundant because of `FullSubcategory` and should be removed anyways). I see two solutions to this:
- Define an alternate version `ℙ'` and use that whenever the extra universe polymorphism is needed
- Modify the definition of `ℙ` to take another universe level argument; This would be a breaking change and would require editing a lot of the library

Thoughts

maxsnew:
Personally, I would update it to be more universe polymorphic. I consider the current design to be an anti-pattern as it forces tedious lifting on library users to recover universe polymorphism.

In the categories part of the library, we only have universe polymorphic definitions of presheaves/presheaf categories: https://github.com/agda/cubical/blob/master/Cubical/Categories/Presheaf/Base.agda#L15

anshwad10:
> Hopefully Agda can infer it in many places

I'm not so sure.. For example `ℙ A l` and `ℙ A l-zero` would have the same universe level of `l-suc l`, so agda would probably not be able to infer what is meant just based on the type. Likely making it implicit would cause a similar situation to what happened with `Lift`

PR #1273, opened 2025-11-15
Make `Lift`'s level argument explicit

anshwad10:
See the discussion at [#1236](https://github.com/agda/cubical/issues/1236#issuecomment-3531207216)


anshwad10:
I'm also making `Unit*`, `⊥*` and `Bool*` take their level arguments explicitly for the same reason

maxsnew:
@anshwad10 can you write a summary of what is changed in this PR to make it easier to review? It is quite large so it would be good to know that you changed arguments to be explicit specifically in XYZ files and everything else is just updated to fit the new interface. Since you made the change yourself it is much easier for you to write up than it is for a reviewer to figure out.

PR #1274, opened 2025-11-16
Resolve naming inconsistency between Iso and Category.isIso

### LINKED ISSUE #1251: Inconsistency between Iso and Category.isIso
The [paths in `Iso` are called `rightInv` and `leftInv`](https://github.com/agda/cubical/blob/629503aeff501437ab03a225d7b5eee23a245a6c/Cubical/Foundations/Isomorphism.agda#L33) while [paths in `isIso` for `Category` are called `sec` and `ret`](https://github.com/agda/cubical/blob/629503aeff501437ab03a225d7b5eee23a245a6c/Cubical/Categories/Category/Base.agda#L65). I guess the most sensible refactor would be to change the name of the fields in `isIso` to match `Iso` since more code uses `Iso`.

msyds:
Closes #1251.

This is a very big rename. It would've been much easier to rename `isIso`'s fields to match `Iso`'s rather than the other way around, but I strongly believe that would leave us with an [even worse inconsistency](https://github.com/agda/cubical/issues/1251#issuecomment-3538697148).

### DIFF
--- a/Cubical/Categories/Equivalence/Properties.agda
+++ b/Cubical/Categories/Equivalence/Properties.agda
@@ -141,7 +141,7 @@ module _ {C : Category ℓC ℓC'} {D : Category ℓD ℓD'}
         comp
         (λ j → D [ iso-ob .sec x (~ j) , iso-ob .sec x (~ j) ])
         (λ j → λ
-          { (i = i0) → iso-hom .sec _ (D .id {x = x}) (~ j)
+          { (i = i0) → iso-hom .rightInv _ (D .id {x = x}) (~ j)
           ; (i = i1) → D .id {x = iso-ob .sec x (~ j)} })
         (D .id {x = x})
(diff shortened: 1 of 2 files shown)


msyds:
The [first commit](https://github.com/agda/cubical/pull/1274/commits/cfa1208eb55b6aa5d59a2fa1e60e3318a6966841) is nothing but the application of a straightforward search-and-replace and does not typecheck. The [second commit](https://github.com/agda/cubical/pull/1274/commits/860e5164a7d77e71beacbeb719f3e11bcc79bcf9) manually resolves the handful of scope errors introduced and gets everything compiling again. I'm currently combing through the diff to undo any false-positives and will follow up with one last commit }:). By that point, this should be mergable, assuming this is even a welcome change.

msyds:
(A few of said false-positives (`BiInjective`, `IsoOver`) should probably be renamed to match `Iso` and `isIso` anyway, but that can be done in a separate PR.)

mortberg:
I have no strong opinions on this, so if @maxsnew thinks it's better I'm fine with merging

PR #1276, opened 2025-11-19
Update to `Int.Fast`

LorenzoMolena:
This PR introduces the `Fast` counterparts of the modules in `Cubical.Data.Int`, together with the relevant order and algebraic instances.
The code is complete and typechecks, but I'm marking it as _draft_ until a few design questions are clarified:
- **Folder organization**: currently the `Fast` modules live inside the `Int` directory, but placing a `Fast` folder directly under `Cubical.Data` is also an option, and the same question applies to the instances.  
- **Upcoming PR for fast rationals**: I plan to submit a follow-up PR implementing fast rationals. Since rationals have fewer dependencies than integers, with little effort I should be able to _replace_ the current implementation so that it uses the fast integer operations. Would such a replacement be welcome, or would you prefer a separate `Fast` namespace? In the latter case, I could also experiment with defining rational operations that prenormalize their arguments.
- **Implicit arguments in `Int.Fast.Order`**: I noticed that many properties in `Int.Fast.Order` have implicit variables that are harder for Agda to infer compared to their non-fast counterparts. Should I make these arguments explicit, or keep the signatures identical for better drop-in compatibility when switching between `Int` and `Int.Fast`?

I should also note that this PR depends on #1270, #1271, and #1272.

I would greatly appreciate the maintainers' perspectives on these points. @mortberg @felixwellen @ecavallo

### DIFF
--- a/Cubical/Algebra/AbGroup/Instances/Int/Fast.agda
+++ b/Cubical/Algebra/AbGroup/Instances/Int/Fast.agda
@@ -1,0 +1,9 @@
+module Cubical.Algebra.AbGroup.Instances.Int.Fast where
+
+open import Cubical.Foundations.Prelude
+open import Cubical.Data.Int.Fast
+open import Cubical.Algebra.AbGroup.Base
+open import Cubical.Algebra.Group.Instances.Int.Fast
+
+ℤAbGroup : AbGroup ℓ-zero
+ℤAbGroup = Group→AbGroup ℤGroup +Comm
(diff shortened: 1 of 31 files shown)


LorenzoMolena:
To keep things manageable, I’ve decided to split this PR according to the following plan:
- First, I'll submit a PR without the `Order` module (and all the modules depending on it).
- Then, @marcinjangrzybowski will follow up with a PR introducing solvers for ring equations on fast integers (already present in #1182, and briefly discussed during last month's AIM)
- Finally, I'll submit a PR adding the `Order` module (and the related instances for order structures), using @marcinjangrzybowski's solver to simplify many proofs.

Issue #1264, opened 2025-10-16
Merge Fin.Data and Fin.Data.Inductive


mortberg:
There is currently some unnecessary duplication between `Data.Fin` and `Data.Fin.Inductive`. These two definitions of `Fin` are almost identical, but parts of the library uses one definition and some uses the other. 

@aljungstrom has convinced me that version one in `Data.Fin.Inductive` is better as it satisfies more definitional equalities, so I suggest we put that version in `Data.Fin` and use it everywhere. @croos90 is currently working on implementing this, so this issue is mostly here to avoid someone else starting to work on the same thing.

PR #1279, opened 2025-12-16
Update fin

croos90:
Merge Data.Fin and Data.Fin.Inductive as per issue #1264


LorenzoMolena:
May I suggest a change? 
The relation `_<ᵗ_` in `Data.Nat.Order.Inductive` reduces in the same way as `Bool→Type(m <ᵇ n)`, so if we redefine it like this, all proofs should work unchanged. 
I'm suggesting this change because then we could use the new `_<ᵗ_` for the `Data.Fin.Literals` module, instead of relying on `fromℕ≤`, leading to a noticeably faster instance search for numeric literals. 

Concretely, on my machine the following example get stuck during typechecking:

```
test : Path (Fin 1234567890) 123456789 123456789
test = refl
```
But typechecks instantly with the change I am suggesting.

Please feel free to ignore my comment if this suggestion doesn’t fit well with the current scope of the PR, or if you prefer to leave it for future work.

mortberg:
I think it would be better to merge this now, but it seems like I've lost my merge rights... Will merge once I've got my rights back. You can make a PR with the changes you suggest once this is merged this is merged and I'll try to merge that one fast

mortberg:
Merged! Please make a small follow-up PR @LorenzoMolena and I'll merge it quickly

PR #1280, opened 2025-12-16
`Fast.Int` Update

LorenzoMolena:
See #1276.
I have decided to change the organization of the modules by introducing a common `Fast` directory, hosting the fast `Int` modules and, in the future, fast rationals as well. The same change has been made for instances of algebraic structures.  
In my opinion, this organization seems conceptually clearer. 
For example, in #1276 the instances of fast integers as `AbGroup` are in `Cubical.Algebra.AbGroup.Instances.Int.Fast`, with the file named `Fast`, whereas in this PR we have `Cubical.Algebra.AbGroup.Instances.Fast.Int`, with the file named `Int`. 

Moreover, since in these PRs they are consistently referred to as "fast integers", having them under `Fast.Int` appears more natural than the current `Int.Fast`.

Finally, the only fast `Int` module already merged is the `Base` file, so this reorganization is minimal.

### Update

After discussing with @marcinjangrzybowski, we agreed to adopt a different notion of order than the one currently used in the draft of #1276, and to update this PR accordingly, including the order module together with the related order instances.
In particular, I am not using Marcin’s solver for fast integers, since it is not required for this part of the development. There will still be a separate PR for the solver, which will be especially useful for the development of Fast Rationals, where it can help avoid long chains of equational reasoning.
 
This implementation of the order is inspired by Agda’s standard library and 1Lab, but relies on `_<ᵗ_` from `Cubical.Data.Nat.Order.Inductive` rather than the order from `Cubical.Data.Nat.Order` (which is defined as a sigma type).

The main advantages are that implicit arguments are easier to infer and that the definitions reduce efficiently on concrete numerals, as can be seen by normalizing the proofs of decidability

I would also appreciate feedback on a few design choices in this PR. In particular, I experimented with the `DISPLAY` pragma to obtain a more informative printing of comparison results. I found it convenient, but I do not have strong feelings about keeping it, so I am happy to adjust if preferred.

### DIFF
--- a/Cubical/HITs/Join/Properties.agda
+++ b/Cubical/HITs/Join/Properties.agda
@@ -10,6 +10,10 @@ This file contains:
 
 - Ganea's theorem
 
+- eliminator for join
+- Join is contractible if either side is contractible
+- Join is prop if both sides are prop
+
 -}
 
 
@@ -17,6 +21,7 @@ module Cubical.HITs.Join.Properties where
 
 open import Cubical.Foundations.Prelude
 open import Cubical.Foundations.Equiv
+open import Cubical.Foundations.HLevels
 open import Cubical.Foundations.Isomorphism
 open import Cubical.Foundations.GroupoidLaws
 open import Cubical.Foundations.Function
@@ -27,13 +32,13 @@ open import Cubical.Data.Sigma renaming (fst to proj₁; snd to proj₂)
 open import Cubical.Data.Unit
 
 open import Cubical.HITs.Join.Base
-open import Cubical.HITs.Pushout
+open import Cubical.HITs.Pushout hiding (elimProp)
 
 open import Cubical.Homotopy.Loopspace
 
 private
   variable
-    ℓ ℓ' : Level
+    ℓ ℓ' ℓ'' : Level
 
 -- the inclusion maps are null-homotopic
 join-inl-null : {X : Pointed ℓ} {Y : Pointed ℓ'} (x : typ X)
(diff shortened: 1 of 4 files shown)


anshwad10:
"Fast" doesn't feel like a word to me anymore, after reading that 😄

Cubical/Data/Fast/Int/Order.agda
```diff
@@ -0,0 +1,741 @@
...
+¬<ℕ→¬pos<pos : ∀ {x y} → ¬ (x ℕ.<ᵗ y) → ¬ pos x < pos y
+¬<ℕ→¬pos<pos ¬p (pos<pos p) = ¬p p
+
+¬≥ℕ→¬negsuc≤negsuc : ∀ {x y} → ¬ (x ℕ.≥ᵗ y) → ¬ negsuc x ≤ negsuc y
+¬≥ℕ→¬negsuc≤negsuc ¬p (negsuc≤negsuc p) = ¬p p
+
+¬>ℕ→¬negsuc<negsuc : ∀ {x y} → ¬ (x ℕ.>ᵗ y) → ¬ negsuc x < negsuc y
+¬>ℕ→¬negsuc<negsuc ¬p (negsuc<negsuc p) = ¬p p
+
+{-# DISPLAY negsuc≤negsuc {x} {y} _ = negsuc x ≤ negsuc y #-}
```
mortberg: What do these DISPLAY pragmas do

LorenzoMolena: They are used to show more informative outputs when running the properties. 
I found it useful while writing the code, in particular for the proofs of decidability of order and of trichotomy:

- running `123456 <? 123455` shows `no (¬ pos 123456 < pos 123455)`  instead of `no (¬<ℕ→¬pos<pos (λ x → x))`
- running `-123455 ≟ -123456` shows `gt (negsuc 123455 < negsuc 123454)`  instead of `lt (negsuc<negsuc tt)`

mortberg:
> > Only a question, otherwise everything looks good. Is it ready to be merged?
> 
> Yes, it is ready to be merged

PR #1284, opened 2026-01-05
Use the faster quotient and remainder in the whole library

LorenzoMolena:
This PR changes the names of `quotient_/_`, `remainder_/_`, and `mod`, so that the more efficient implementations become the new default, while the less efficient ones are written with a `'`.
To achieve this, I ported the [proofs](https://agda.github.io/agda-stdlib/master/Data.Nat.DivMod.Core.html) from the standard library, and I also proved the equality of the two implementations by showing the uniqueness of the quotient-remainder pair under some constraints.
I left the old proofs for reference, but if it seems reasonable, I can remove them, as they are no longer used.
Additionally, two more tests have been added to the `Data.Fin.Arithmetic` module.


mortberg:
Is this ready to be merged now?

PR #1291, opened 2026-02-02
Nix flake: pin agda version to 2.8.0

flgrubm:
Right now, the nix flake is tracking the `master` branch of the Agda repo as of July 30, 2025. This pull request pins the used Agda version specifically to the `v2.8.0` tag instead.

The current setup actually builds Agda version `2.9.0` (at least in name), which confuses my Emacs agda-mode and makes it complain about mismatched Agda versions. I suspect that this is due to the fact that after the 2.8.0 release (on July 5) the upstream Agda team increased the version identifier on their master branch before our lock file was updated.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -11,6 +11,12 @@ Start by making a branch for the release. Change the following things:
 
 * If needed, Increment the `version` field in `flake.nix`.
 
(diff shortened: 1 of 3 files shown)


felixwellen:
Yes, good point! I would call that file release intructions though.

PR #1293, opened 2026-02-17
Comm ring solver improvement

marcinjangrzybowski:
depends on #1280 and #1260


marcinjangrzybowski:
for now typechecks with `master` branch of agda (some instances fail to resolve with 2.8.0) , i changed branch of agda in CI for this PR temporarly to make it go thru.
Depending on the pace of the rest of work with those solvers, it can wait to new relase of agda, or I can tweak code so it will work with 2.8.0.

PR #1297, opened 2026-03-03
BooleanAlgebraStr now infers the underlying set

Freek98:
BooleanAlgebraStr was based on a Boolean Ring, which is a tuple consisting of a type and a witness that this type is a Boolean Ring. The problem with this is that instances get difficult to resolve. Now instead, the type (first element of the tuple) is an implicit argument and the structure (second element of the tuple) is explicit, and instances are resolved without hassle.


Freek98:
@felixwellen is this a standard way to work with such instance resolution in the algebra libraries (Evan advised me to ask you).

felixwellen:
Looks good to me - you might get problems with type inference for homomorphisms (e.g. when you use composition).  But you can still fix that later by making an extra record type for homs of boolean rings.

PR #1298, opened 2026-03-03
Proof of equality that Quo rationals are cubically equal to Rationals

reswatson:
This proof means that many functions and definitions in Data/Rationals.agda can be transported over more easily not just to Data/Rationals/MoreRationals/QuoQ but also Data/Rationals/MoreRationals/SigmaQ, since there is already a proof of equality bewteen QuoQ and SigmaQ. And vice versa if we wish.


reswatson:
Is this good to merge? Do I need to make any changes?

mortberg (approved):
Looks good to me! Thanks for the contribution.

PR #1302, opened 2026-04-08
[chore] remove unnecessary --safe --cubical and --no-import-sort flags

maxsnew:
These flags are included in the agda-lib file so they aren't needed. I also removed a related outdated comment in the CONTRIBUTING.md file.

The `--guardedness` flag is used extensively in the `Cubical/Codata` modules even though it is also included in the `agda-lib` flag, but I left those in because it says in `RELEASE.md` that this is intended to be removed from `agda-lib` eventually.


mortberg:
Perfect, thank you for cleaning this up!

PR #1303, opened 2026-04-21
Properties of `OrderedCommRings`

LorenzoMolena:
This PR introduces properties of Ordered Commutative Rings, including, but not limited to:
- Morphisms between OCRs (homomorphisms, monomorphisms, and isomorphisms), univalence for OCRs, and proof that `OrderedCommRing` is a 1-Groupoid
- An `OrderedCommRingReasoning` module, derived from `QuosetReasoning`, with additional helpers for adding/multiplying the same element on both sides of inequalities
- Positive and non-negative cones, with various algebraic structure instances
- Canonical monomorphism from the fast integers, with proof of uniqueness

To prove the last point, this PR also includes the proof of the existence and uniqueness of a ring homomorphism from the fast integers into any other commutative ring.

The notion of OCR homomorphism is derived from *homomorphisms of Ordered Heyting Fields*, following Definition 4.3.1 of [A.Booij PhD Thesis](https://ncatlab.org/nlab/files/Booij-AnalysisInUF.pdf#page=87). In addition to requiring to reflect `<`, we also ask to preserve `≤`. 
Although the latter is logically redundant, in specific cases it could be useful to provide that proof directly, rather than relying on the one derived from `reflect<`. If not needed, this can be avoided using the helper `makeIsOrderedCommRingHom` 

If this addition is welcome, this PR would need to be merged before #1293, which will be updated to depend on these changes.


mortberg (approved):
Sorry for the slow review. The code looks good to me and can be merged. However, before doing can you please add the comment with the reference to Ayberk's thesis as a comment in the code as well so that it's not lost after merging?

LorenzoMolena:
> Sorry for the slow review. The code looks good to me and can be merged. However, before doing can you please add the comment with the reference to Ayberk's thesis as a comment in the code as well so that it's not lost after merging?

Thank you. I just added a comment at the top of the `Morphisms` module. Is it enough? Or were you thinking of a different format to cite the thesis?

mortberg:
Perfect, thanks for the contribution!

PR #1304, opened 2026-04-23
core code missing from Nat and Int

reswatson:


Cubical/Data/Nat/GCD.agda
```diff
@@ -214,6 +232,43 @@ isGCD-multʳ {m} {n} {d} k dGCD = gcd≡→isGCD (gcd-factorʳ m n k ∙ cong (_
...
+-- Core properties of gcd
+
+gcd[m,n]∣m : (m n : ℕ) → (gcd m n) ∣ m
+gcd[m,n]∣m m n = fst (fst (gcdIsGCD m n))
+
+gcd[m,n]∣n : (m n : ℕ) → (gcd m n) ∣ n
+gcd[m,n]∣n m n = transport (cong (λ a → a ∣ n) (gcdSym n m) ) (gcd[m,n]∣m n m)
+
+gcd-greatest : c ∣ m → c ∣ n → c ∣ gcd m n
+gcd-greatest c∣m c∣n =
```
LorenzoMolena: Is there any reason why it is proved in this way, instead of currying `snd (gcdIsGCD m n) c`? 
I tested it, and you could even write `gcd-greatest = curry (snd (gcdIsGCD _ _) _)`, with all implicit varibles solved

Cubical/Data/Nat/GCD.agda
```diff
@@ -214,6 +232,43 @@ isGCD-multʳ {m} {n} {d} k dGCD = gcd≡→isGCD (gcd-factorʳ m n k ∙ cong (_
...
+gcd[m,n]≢0 m n (inr n≢0) gcd0 =
+  n≢0 (sym (∣-zeroˡ (transport ((cong (λ a → a ∣ n) gcd0)) (gcd[m,n]∣n m n))))
+
+gcd[m,n]≡0⇒m≡0 : gcd m n ≡ 0 → m ≡ 0
+gcd[m,n]≡0⇒m≡0 {zero} {n} gmn = refl
+gcd[m,n]≡0⇒m≡0 {suc m} {n} gmn =
+  ⊥.elim {A = λ bot → suc m ≡ 0} (gcd[m,n]≢0 (suc m) n (inl snotz) gmn)
+
+gcd[m,n]≡0⇒n≡0 : ∀ {m n} → gcd m n ≡ 0 → n ≡ 0
+gcd[m,n]≡0⇒n≡0 {m}{n} gmn = gcd[m,n]≡0⇒m≡0 {n}{m}
```
LorenzoMolena: transporting on `cong (λ a → a ≡ 0)` is the same as composing with the reverse path. In my opinion, it would be better to write `gcd[m,n]≡0⇒m≡0 {n} {m} (gcdSym n m ∙ gmn)`, avoiding unnecessary transports.
More generally, I see you are using multiple times the pattern `transport` + `cong`, where you could have used `subst` instead.
However, I'm not a maintainer; I just noticed a couple of places where the code could be made shorter. Feel free to ignore these comments if not relevant

reswatson: Thanks. I'll look at this shortly.

Cubical/Data/Nat/GCD.agda
reswatson: Done

Cubical/Data/Nat/GCD.agda
reswatson: Done

mortberg (approved):
Looks good to me. Will merge

PR #1300, opened 2026-03-24
Add Iterative Sets

flgrubm:
This pull request adds the implementation of iterative sets, which is part of my master's thesis[^1]. In particular, it contains the following:

- definitions of iterative multisets (V∞), iterative sets (V⁰) as well as several other related definitions
- several important properties
- a proof of closure properties of  V⁰ under several set and type-theoretic properties (e.g. (unordered/ordered) pairing, Π, Σ-types)

The formalization follows the proof ideas of [^2] and the corresponding Agda Unimath formalization[^3] more or less, as detailed in my thesis. There is also a second part about showing that V⁰ is a CwF that supports various structures, but this is still a work in proress, particularly the instantiation of Σ-structure, and another pull request will be made for that in the future.

This pull request is currently a draft because I've run into a few issues. I would really appreciate any feedback or guidance on them:
- There are some Performance issues: The module `Cubical.Data.IterativeSets.UnorderedPair` as well as `Cubical.Data.IterativeSets.OrderedPair` only typecheck within a reasonable amount of time if the option `--lossy-unification` is enabled. The file `Cubical.Data.IterativeSets.Pi` does not typecheck within a reasonable amount of time at all, not even with lossy unification (for this, I typechecked individual parts of the file, so I assume the whole thing should be correct as well).
- I have several "helper" lemmas that might be useful to have, e.g. `⊥≢Unit` (which I particularly need in the general version `⊥*≢Unit*`). I don't quite know where to put them.

Thank you for your help!

[^1]: https://grubmueller.dev/publications/2026-mt-iterative-sets/
[^2]: D. Gratzer, H. R. Gylterud, A. Mörtberg, and E. Stenholm, “The Category of Iterative Sets in Homotopy Type Theory and Univalent Foundations”, https://doi.org/10.1017/S0960129524000288
[^3]: https://elisabeth.stenholm.one/category-of-iterative-sets/iterative.set.html

PR #1305, opened 2026-04-23
Category/Presheaf notations, univalent displayed categories, Displayed categories in BinProduct and Elements

maxsnew:
This looks big but it's mostly a re-org to clean up some redundancies in the category theory library:

1. Merge `ΣPropCat` and `FullSubcategory`
2. Implement the binary product of categories as a total category of a weakened displayed category
3. Implement the category of elements of a presheaf as a total category of a displayed category of elements.
4. In updating that, add a definition of univalent displayed categories and prove that the total category of a univalent displayed category over a univalent category is univalent.
5. Add some congruence notation to Category and add PresheafNotation (that we use heavily downstream in `cubical-categorical-logic`).
6. Some minor definitions we have in `cubical-categorical-logic` like `Functorⱽ` and `isSetDep`...

Changing the definition of binary product looks a bit scary but I put a display pragma on it so that I don't think it should ever leak out. If we didn't have `no-eta-equality` for categories, it would be `refl` equal to the old definition, so no code broke with the change.

The most substantive change is that in porting the category of elements to be defined using a total category, I defined univalent displayed categories.

### DIFF
--- a/Cubical/Categories/Category/Base.agda
+++ b/Cubical/Categories/Category/Base.agda
@@ -130,7 +130,8 @@ pathToIso-refl {C = C} {x} = JRefl (λ z _ → CatIso C x z) (idCatIso)
 eqToIso : {C : Category ℓ ℓ'} {x y : C .ob} (p : x Eq.≡ y) → CatIso C x y
 eqToIso {C = C} Eq.refl = idCatIso
 
--- eqToIso-refl would just be refl
+eqToIso-refl : {C : Category ℓ ℓ'} {x : C .ob} → eqToIso {C = C} {x} Eq.refl ≡ idCatIso
+eqToIso-refl = refl
 
 eqToMorphism : {C : Category ℓ ℓ'} {x y : C .ob} (p : x Eq.≡ y) → C [ x , y ]
 eqToMorphism {C = C} p = eqToIso {C = C} p .fst
--- a/Cubical/Categories/Category/Base.agda
+++ b/Cubical/Categories/Category/Base.agda
@@ -4,7 +4,9 @@ open import Cubical.Foundations.Prelude
 open import Cubical.Foundations.HLevels
 open import Cubical.Foundations.Equiv
 open import Cubical.Foundations.Powerset
+
 open import Cubical.Data.Sigma
+import Cubical.Data.Equality as Eq
 
 private
   variable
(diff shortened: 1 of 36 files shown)


mortberg:
This seems like a welcome change even though I haven't looked at it closely yet. How about making sure that we merge https://github.com/agda/cubical/pull/1300 first and then you fix the potential conflicts in this PR?

I plan to take a look at https://github.com/agda/cubical/pull/1300 ASAP and merge once me and the CI are happy

Cubical/Categories/Category/Base.agda
```diff
@@ -108,13 +118,22 @@ idCatIso {C = C} = C .id , isiso (C .id) (C .⋆IdL (C .id)) (C .⋆IdL (C .id))
...
+pathToMorphism : {C : Category ℓ ℓ'} {x y : C .ob} (p : x ≡ y) → C [ x , y ]
+pathToMorphism {C = C} p = pathToIso {C = C} p .fst
+
 pathToIso-refl : {C : Category ℓ ℓ'} {x : C .ob} → pathToIso {C = C} {x} refl ≡ idCatIso
 pathToIso-refl {C = C} {x} = JRefl (λ z _ → CatIso C x z) (idCatIso)
 
+eqToIso : {C : Category ℓ ℓ'} {x y : C .ob} (p : x Eq.≡ y) → CatIso C x y
+eqToIso {C = C} Eq.refl = idCatIso
+
+-- eqToIso-refl would just be refl
```
mortberg: Should this comment be removed?

Cubical/Categories/Presheaf/Morphism.agda
```diff
@@ -45,38 +50,38 @@ module _ {C : Category ℓc ℓc'}{D : Category ℓd ℓd'}
...
-    pushEltF .F-hom {x}{y} (f , commutes) .snd =
-      pushElt _ .snd ∘ᴾ⟨ Q ⟩ F .F-hom f
-        ≡⟨ pushEltNat y f ⟩
-      pushElt (_ , y .snd ∘ᴾ⟨ P ⟩ f) .snd
-        ≡⟨ cong (λ a → pushElt a .snd) (ΣPathP (refl , commutes)) ⟩
-      pushElt x .snd ∎
-    pushEltF .F-id = Σ≡Prop (λ x → (Q ⟅ _ ⟆) .snd _ _) (F .F-id)
-    pushEltF .F-seq f g =
-      Σ≡Prop ((λ x → (Q ⟅ _ ⟆) .snd _ _)) (F .F-seq (f .fst) (g .fst))
+    -- -- This should define a functor on the category of elements
```
mortberg: The commented code can probably be deleted?

mortberg (changes requested):
Just two minor comments about comments. Once these have been resolved I would be happy to merge

Cubical/Categories/Category/Base.agda
maxsnew: Because it's too obvious? I could also add the trivial proof to document it

maxsnew: I pushed an update where I include the trivial proof

Cubical/Categories/Presheaf/Morphism.agda
maxsnew: done

Cubical/Categories/Category/Base.agda
mortberg: Ah ok, now I understand! It's indeed good to include it so that if someone tries to change something we will notice if this breaks

PR #1307, opened 2026-05-13
add List monoid instance

eiais:
List A is a monoid when A is a set


mortberg:
Thank you for the contribution!

PR #1308, opened 2026-05-14
add foldl/map Monoid Homomorphisms

eiais:
foldl is a monoid homomorphism from List A to any monoid on A

### DIFF
--- a/Cubical/Algebra/Monoid/Instances/List.agda
+++ b/Cubical/Algebra/Monoid/Instances/List.agda
@@ -55,3 +55,6 @@ foldlMonHom m = fn , monoidequiv refl respects∙
                                         (x ∙m (fn xs)) ∙m (fn ys)
                                           ≡⟨ cong (_∙m (fn ys)) (fnCons x xs) ⟩
                                         (fn (x ∷ xs)) ∙m (fn ys) ∎
+
+mapMonHom : (A B : Σ Type isSet) → (f : (fst A) → (fst B)) → MonoidHom (ListMonoid A) (ListMonoid B)
+mapMonHom A B f = map f , monoidequiv refl (λ x y → sym (map++ f x y))
--- a/Cubical/Algebra/Monoid/Instances/List.agda
+++ b/Cubical/Algebra/Monoid/Instances/List.agda
@@ -2,6 +2,7 @@ module Cubical.Algebra.Monoid.Instances.List where
 
 open import Cubical.Foundations.Prelude
 open import Cubical.Foundations.HLevels
+open import Cubical.Foundations.Structure
 
 open import Cubical.Data.List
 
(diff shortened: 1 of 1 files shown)


Cubical/Algebra/Monoid/Instances/List.agda
```diff
@@ -14,3 +15,46 @@ MonoidStr.ε (snd (ListMonoid _)) = []
 MonoidStr._·_ (snd (ListMonoid _)) = _++_
 MonoidStr.isMonoid (snd (ListMonoid (_ , pf))) = makeIsMonoid (isOfHLevelList 0 pf)
                                     (λ x y z → sym (++-assoc x y z)) ++-unit-r (λ _ → refl)
+
+foldlMonHom : (M : Monoid ℓ-zero) → MonoidHom (ListMonoid (fst M , MonoidStr.is-set (snd M))) M
```
mortberg: Why is this only for level zero? Can't it be universe polymorphic?

mortberg: Also, why decompose M? Shouldn't `ListMonoid M` work?

Cubical/Algebra/Monoid/Instances/List.agda
```diff
@@ -14,3 +15,46 @@ MonoidStr.ε (snd (ListMonoid _)) = []
...
+                                          ≡⟨ sym (fnCons x (xs ++ ys)) ⟩
+                                        x ∙m (fn (xs ++ ys))
+                                          ≡⟨ cong (x ∙m_) (respects∙ xs ys) ⟩
+                                        x ∙m ((fn xs) ∙m (fn ys))
+                                          ≡⟨ MonoidStr.·Assoc (snd m) x (fn xs) (fn ys) ⟩
+                                        (x ∙m (fn xs)) ∙m (fn ys)
+                                          ≡⟨ cong (_∙m (fn ys)) (fnCons x xs) ⟩
+                                        (fn (x ∷ xs)) ∙m (fn ys) ∎
+
+mapMonHom : (A B : Σ Type isSet) → (f : (fst A) → (fst B)) → MonoidHom (ListMonoid A) (ListMonoid B)
```
mortberg: Please use `hSet` instead of `Σ Type isSet`

mortberg (changes requested):
Thanks a lot. I have a few small comments before this can be merged

Cubical/Algebra/Monoid/Instances/List.agda
eiais: Made it universe polymorphic :). I dont think ```ListMonoid M``` works because I only need fact that it's a set from the structure.

mortberg: Ah, of course!

PR #1311, opened 2026-06-02
Truncated logarithm over the natural numbers

LorenzoMolena:
This PR introduces the truncated logarithm of `x` in base `b`, as the unique natural number `l` such that `b ^ l ≤ x < b ^ (suc l)`.
The base `b` and argument `x` are also natural numbers subject to `2 ≤ b` and `1 ≤ x`; these assumptions are enforced by implicit arguments of type `2 ≤ᵗ b` and `1 ≤ᵗ x`, to make computing with `log` easier. 
However, for stating and proving the properties, it was more straightforward to work with a base of the form `suc (suc m)`, rather than using an implicit argument.

Some auxiliary lemmas were also added in the `Nat.Mod` and `Nat.Order` modules; in particular, `≤-·sk-cancel` and `<-·sk-cancel` have been moved from `Fin.Properties` to `Nat.Order`, using shorter inductive proofs.

### DIFF
--- a/Cubical/Data/Nat/Order.agda
+++ b/Cubical/Data/Nat/Order.agda
@@ -147,11 +147,22 @@ suc-< p = pred-≤-pred (≤-suc p)
            (d + m) · k   ≡⟨ cong (_· k) r ⟩
            n · k         ∎
 
+≤-k· : m ≤ n → k · m ≤ k · n
+≤-k· {k = k} = subst2 _≤_ (·-comm _ k) (·-comm _ k) ∘ ≤-·k
+
 <-k+-cancel : k + m < k + n → m < n
 <-k+-cancel {k} {m} {n} = ≤-k+-cancel ∘ subst (_≤ k + n) (sym (+-suc k m))
 
 ¬-<-zero : ¬ m < 0
 ¬-<-zero (k , p) = snotz ((sym (+-suc k _)) ∙ p)
+
+≤-·sk-cancel : m · suc k ≤ n · suc k → m ≤ n
+≤-·sk-cancel {zero}  {n = n}     = λ _ → zero-≤
+≤-·sk-cancel {suc m} {n = zero}  = ⊥.rec ∘ ¬-<-zero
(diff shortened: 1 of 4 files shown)


Cubical/Data/Nat/Order.agda
```diff
@@ -238,6 +252,17 @@ predℕ-≤-predℕ {suc m} {suc n} ineq = pred-≤-pred ineq
            (d + suc m) · suc k               ≡⟨ cong (_· suc k) r ⟩
            n · suc k                         ∎
 
+<-sk· : m < n → suc k · m < suc k · n
```
mortberg, line 259: Is it really so good that the `k` is part of the name? The same goes for many other results in the file. The variable name is arbitrary and can be changed to anything...

mortberg (changes requested):
Looks good to me, modulo some renaming

Cubical/Data/Nat/Order.agda
LorenzoMolena: Yes, I agree that it is better to avoid using variables in the names of lemmas.
Here I did it anyway, as I thought that was the implicit convention in the rest of the module: `<-·sk` was already present with this name, and `<-sk·` is the same lemma but with multiplication of `suc k` on the left instead of on the right.
Is it okay to leave it as is, or do you have some suggestions?

mortberg: I guess the ideal thing would be fix all the names of the lemmas. Do you want to do it or should we leave a TODO-comment for it?

LorenzoMolena: I think there are quite a lot of names to fix in this module. 
I could try to do it, but I personally prefer to leave a TODO-comment and fix the naming in a future PR

mortberg: Ok, please add a TODO and then I'll merge!

LorenzoMolena: Done. Thank you for the review! Soon I'll try to look at these lemmas and open a PR fixing the naming

PR #1312, opened 2026-06-05
General results about set quotients

cangiuli:
This PR adds several general-purpose lemmas about set quotients:
- HoTT Book Lemma 6.10.8: If we have a function r sending each element of A to its "canonical representative" under the relation R, then the set of fixed points of r (i.e., the set of canonical representatives) satisfies the universal property of the quotient A/R and is equivalent to it.
- _(added in 14019af)_ If `p : A -> B` is a surjection into a set, then B is the quotient of A by `x ~ y := p x = p y`.
- HoTT Book Corollary 6.10.10: If `p : A -> B` is a retraction into a set, then B is the quotient of A by `x ~ y := p x = p y`.
- Every set `A` is equivalent to `A / _≡_`.

If you can think of any better names for these results, please let me know. :)

### DIFF
--- a/Cubical/HITs/SetQuotients/Properties.agda
+++ b/Cubical/HITs/SetQuotients/Properties.agda
@@ -400,20 +400,25 @@ module _ (isSetA : isSet A)
   equivQuotCR : (A / R) ≃ (CanonicalReprs r)
   equivQuotCR = isoToEquiv equivQuotCRIso
 
--- Corollary 6.10.10 in HoTT Book
-quotRetractIso : isSet B
-  → (p : A → B) → (s : B → A) → retract s p
-  → Iso (A / (λ x y → p x ≡ p y)) B
-Iso.fun (quotRetractIso isSetB p s ret) = rec isSetB p (λ _ _ q → q)
-Iso.inv (quotRetractIso isSetB p s ret) b = [ s b ]
-Iso.sec (quotRetractIso isSetB p s ret) = ret
-Iso.ret (quotRetractIso isSetB p s ret) =
-  elimProp (λ _ → squash/ _ _) λ a → eq/ _ _ (ret (p a))
+quotSurjectionEquiv : isSet B
+  → (p : A → B) → isSurjection p
+  → (A / (λ x y → p x ≡ p y)) ≃ B
+quotSurjectionEquiv isSetB p surj = fun , isEmbedding×isSurjection→isEquiv (funEmb , funSurj)
+  where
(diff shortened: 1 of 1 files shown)


anshwad10:
About `quotRetractEquiv`: Doesn't it suffice for `p` to be a surjection, not necessarily a retraction?

cangiuli:
@anshwad10 Yes, good point!

I'll note that the statement I formalized is already stronger than what's in the HoTT Book in that it doesn't require `isSet A`, since it doesn't go through `equivQuotCR` (Lemma 6.10.8). (Actually, I tried going through that result, but the proof was quite a bit longer than the current "direct" one, because a bunch of the details are fiddly.)

I'm not sure whether I should actually add the even stronger statement that `isSurjection p` suffices to this PR, because the proof is more complicated (since we need to fiddle with truncations) and I don't have a use case in mind at the moment...

Cubical/HITs/SetQuotients/Properties.agda
```diff
@@ -353,6 +355,78 @@ descendMapPath f g isSetM path i x =
...
+-- Lemma 6.10.8 in HoTT Book: set quotients via canonical representatives
+module _ (isSetA : isSet A)
+  (r : A → A)
+  (ridem : (a : A) → r (r a) ≡ r a)
+  (r≃R : (a b : A) → (r a ≡ r b) ≃ R a b)
+  where
+
+  canonicalReprsUniversalIso : isSet B →
+    Iso (CanonicalReprs r → B) (Σ[ f ∈ (A → B) ] ((a b : A) → R a b → f a ≡ f b))
+  Iso.fun (canonicalReprsUniversalIso isSetB) g =
```
mortberg, line 374: If you want you can `open Iso` in the module and remove the `Iso.`s

mortberg (approved):
Good stuff! I would be happy to merge as is and you can add the suggested result in a future PR, or you can do it now. Let me know what you want me to do

Cubical/HITs/SetQuotients/Properties.agda
cangiuli: Hmm. I think I prefer as-is, just to stay consistent with the rest of the file (and it doesn't buy much).

cangiuli:
OK, I found an even cleaner way to prove the more general result that @anshwad10 suggested (when `p` is a surjection), and I now derive the previous result (when `p` is a retraction) from it. I'm happy for you to merge this as soon as you approve.

mortberg:
Very nice!

PR #1313, opened 2026-06-17
Normalised Q plus related

reswatson:
A normalised rational Q (Base.agda), along with related additions to other numerical Data.


anshwad10:
What's the difference between this and [`Cubical.Data.Rationals.MoreRationals.SigmaQ`](https://agda.github.io/cubical/Cubical.Data.Rationals.MoreRationals.SigmaQ.Base.html#423)?

reswatson:
Here the non-experimental well-developed standard 'Int' is used, rather than the under-developed and novel 'QuoInt'. The amount of library code available for Int forces this choice. Further, 'Int' is the standard for Cubical Agda and should be used in preference anyway.

reswatson:
@anshwad10: Any suggestions or corrections? I'm thinking for example (i) that maybe 'useful lemmas' should be put in its own file and (ii) that converse and contraposition will either have to be left where they are for now or inlined, but I notice you are working on a library for classical logic - logical relations of any sort should perhaps be put in a 'logic' library. 

@all: Also are there any organisational changes suggested, for example splitting Base into smaller parts? Personally I wouldn't like this as these are mostly minimal essentials that a Properties or an Order might use, but it may be a requested change.

Cubical/Data/Int/Properties.agda
```diff
@@ -1544,3 +1549,68 @@ sumFinℤHom {n = n} = sumFinGenHom _+_ 0 (λ _ → refl) +Comm +Assoc n
 clamp : ℤ → ℕ
 clamp (pos n) = n
 clamp (negsuc n) = zero
+
+
+-- useful lemmas for _·_
```
mortberg, line 1547: The naming here is not very good. The variable names are quite arbitrary and should hence not be part of the name of the lemmas. Can you please come up with more informative names? (You can look at how things are named for the algebraic structures if you want some inspiration)

Cubical/Data/Nat/Properties.agda
```diff
@@ -264,18 +264,25 @@ inj-·sm {suc l} {m} {suc n} p = cong suc (inj-·sm (inj-m+ {m = suc m} p))
 inj-sm· : suc m · l ≡ suc m · n → l ≡ n
 inj-sm· {m} {l} {n} p = inj-·sm (·-comm l (suc m) ∙ p ∙ ·-comm (suc m) n)
 
+¬x≡0→NonZero : ∀ {x : ℕ} → ¬ x ≡ 0 → NonZero x
```
mortberg: Naming

Cubical/Data/Nat/Properties.agda
```diff
@@ -264,18 +264,25 @@ inj-·sm {suc l} {m} {suc n} p = cong suc (inj-·sm (inj-m+ {m = suc m} p))
...
+¬x≡0→NonZero {zero} nx0 = nx0 refl
+¬x≡0→NonZero {suc x} nx0 = tt
+
 integral-domain-· : {k l : ℕ} → (k ≡ 0 → ⊥) → (l ≡ 0 → ⊥) → (k · l ≡ 0 → ⊥)
 integral-domain-· {zero} {l} ¬p ¬q r = ¬p refl
 integral-domain-· {suc k} {zero} ¬p ¬q r = ¬q refl
 integral-domain-· {suc k} {suc l} ¬p ¬q r = snotz r
 
+¬k·l≡0→¬k≡0 : ∀ k l → (k · l ≡ 0 → ⊥) → (k ≡ 0 → ⊥)
```
mortberg: Naming

Cubical/Data/NatPlusOne/Properties.agda
```diff
@@ -40,3 +42,19 @@ private
 
 ·₊₁-identityʳ : ∀ n → n ·₊₁ 1 ≡ n
 ·₊₁-identityʳ (1+ n) = cong 1+_ (injSuc (·-identityʳ (suc n)))
+
+-- Useful lemmas
```
mortberg: Naming

Cubical/Data/Rationals/MoreRationals/NormalisedQ/Base.agda
```diff
@@ -0,0 +1,716 @@
...
+open import Cubical.Data.Nat.Coprime as ℕ
+open import Cubical.Data.NatPlusOne using (1+_; ℕ₊₁; -1+_; ℕ₊₁→ℕ; _·₊₁_)
+open import Cubical.Data.NatPlusOne.PropertiesWithInt
+open import Cubical.Data.Int as ℤ
+  renaming (_+_ to _ℤ+_; _-_ to _ℤ-_; -_ to -ℤ_; _·_ to _ℤ·_; abs to absℤ)
+open import Cubical.Data.Int.GCD as ℤ using (gcd-def; ℕ₊₁→ℤ-gcd-def; gcdSucNot0)
+open import Cubical.Data.Int.Order as ℤ using ()
+
+converse : {ℓ : Level} {a b : Type ℓ} →
+  (a → b) → ¬ b → ¬ a
```
mortberg: Please make this private

Cubical/Data/Rationals/MoreRationals/NormalisedQ/Base.agda
```diff
@@ -0,0 +1,716 @@
...
+↧₋₁ ((z , n) , copr) = n
+
+-- denominator as ℕ₊₁
+↧₊₁_ : ℚ → ℕ₊₁
+↧₊₁ ((z , n) , copr) = 1+ n
+
+↧≡↧→↧₋₁≡↧₋₁ : ∀ {x y : ℚ} → ↧ x ≡ ↧ y → ↧₋₁ x ≡ ↧₋₁ y
+↧≡↧→↧₋₁≡↧₋₁ d = ℕ.injSuc (ℤ.injPos d)
+
+x≡y→↥x≡↥y : ∀ {x y : ℚ} → x ≡ y → ↥ x ≡ ↥ y
```
mortberg: Naming here and at more places in this file

mortberg (changes requested):
I started writing some comments about the code, but then I saw the comment about the relationship to https://agda.github.io/cubical/Cubical.Data.Rationals.MoreRationals.SigmaQ.Base.html#423 and now I'm confused what the point of this code really is?

reswatson:
> I started writing some comments about the code, but then I saw the comment about the relationship to https://agda.github.io/cubical/Cubical.Data.Rationals.MoreRationals.SigmaQ.Base.html#423 and now I'm confused what the point of this code really is?

Sigma uses QuoInt and not the standard Int. It was much easier to start again using the standard Int.

If you want me to change the integer used in Sigma to the standard one and use Sigma instead I can do that, but the reason I didn't do that is because some people may already be using Sigma with QuoInt.

mortberg:
> > I started writing some comments about the code, but then I saw the comment about the relationship to https://agda.github.io/cubical/Cubical.Data.Rationals.MoreRationals.SigmaQ.Base.html#423 and now I'm confused what the point of this code really is?
> 
> Sigma uses QuoInt and not the standard Int. It was much easier to start again using the standard Int.
> 
> If you want me to change the integer used in Sigma to the standard one and use Sigma instead I can do that, but the reason I didn't do that is because some people may already be using Sigma with QuoInt.

Aha I see, wouldn't it be better to just replace the old code with this then if it is better to use the standard integers?

reswatson:
> Aha I see, wouldn't it be better to just replace the old code with this then if it is better to use the standard integers?

Yes, OK. I'll do that along with the name changes.

Cubical/Data/Int/Properties.agda
reswatson: Done

Cubical/Data/Nat/Properties.agda
reswatson: done

Cubical/Data/Nat/Properties.agda
reswatson: done

Cubical/Data/NatPlusOne/Properties.agda
reswatson: done

Cubical/Data/Rationals/MoreRationals/NormalisedQ/Base.agda
reswatson: done (though it was added as a result to other files as well)

Cubical/Data/Rationals/MoreRationals/NormalisedQ/Base.agda
reswatson: done

reswatson:
All requested changes done (in particular naming). Code replaces old SigmaQ stub. Had a further problem with a conflict caused by an instance - resolved by adding a 'hiding' and a 'using ()' to import+open in Papers/RepresentationIndependence.agda.

mortberg (approved):
Thanks, this looks good to me now and I would be happy to merge. Any final comments by @LorenzoMolena or @anshwad10 ?

LorenzoMolena:
> Thanks, this looks good to me now and I would be happy to merge. Any final comments by @LorenzoMolena or @anshwad10 ?

just a couple:
- Merging this might get in conflict with #1324. I don't know what is the best order to merge the two; worst case I will try to adapt mine. The conflict could appear in the conversion between normalised and standard rationals
- In my opinion, there are some proofs that could be made shorter/simpler. should I leave some comment with some suggestions?

Cubical/Data/Rationals/MoreRationals/SigmaQ/Properties.agda
```diff
@@ -1,77 +1,737 @@
...
-open import Cubical.Data.Rationals.MoreRationals.QuoQ as Quo using (ℕ₊₁→ℤ)
-import Cubical.Data.Rationals.MoreRationals.SigmaQ.Base as Sigma
+_-_ : ℚ → ℚ → ℚ
+p - q = p + (- q)
 
+-- Multiplication
+_·_ : ℚ → ℚ → ℚ
+p · q = [ (↥ p) ℤ.· (↥ q) , (↧₊₁ p) ·₊₁ (↧₊₁ q) ]
 
-reduce : Quo.ℚ → Sigma.ℚ
```
LorenzoMolena, line 23: If I'm not mistaken, this PR removes this function (`reduce`) and also any imports of the other kinds of rationals. 
I guess this function would not work anymore as it is written here, but is it ok to remove it completely? 
I think we might have accidentally removed other properties as well while merging @reswatson's work into `SigmaQ`

Cubical/Papers/RepresentationIndependence.agda
```diff
@@ -27,7 +27,7 @@ import Cubical.HITs.Cost.Base                  as CostMonad
 import Cubical.HITs.SetQuotients               as SetQuotients
 import Cubical.Data.Rationals                  as SetQuoQ
 import Cubical.Data.Rationals.MoreRationals.SigmaQ
-                                               as SigmaQ
+                                               as SigmaQ hiding (nonZero-1/')
```
LorenzoMolena, line 30: Are there any name conflicts? I thought it was suggested not to change anything in `Papers`

Cubical/Data/Int/Order.agda
```diff
@@ -181,6 +181,13 @@ isAntisym≤ {m} {n} (i , p) (j , q)
 ≤Monotone+ : m ≤ n → o ≤ s → m ℤ.+ o ≤ n ℤ.+ s
 ≤Monotone+ {o = o} p q = isTrans≤ (≤-+o {o = o} p) (≤-o+ q)
 
+weaken≡→≤ : ∀ {m}{n} → m ≡ n → m ≤ n
+weaken≡→≤ {m}{n} mn =
+  subst (λ x → x) (cong (λ x → x ≤ n) (sym mn)) (isRefl≤ {n})
```
LorenzoMolena: `subst (λ x → x) (cong (λ x → x ≤ n) (sym mn))` could be changed into `subst (λ x → x ≤ n) (sym mn)`

Cubical/Data/Int/Order.agda
```diff
@@ -283,6 +290,10 @@ isTrans< p = ≤<-trans (<-weaken p)
 isAsym< : m < n → ¬ n ≤ m
 isAsym< m<n = isIrrefl< ∘ <≤-trans m<n
 
+<→≢ : ∀ {m n : ℤ} → m < n → ¬ m ≡ n
+<→≢ {m}{n} m<n m≡n =
+  ⊥.elim (isAsym< {m}{n} m<n (subst (λ u → u ≤ m) m≡n (isRefl≤ {m})))
```
LorenzoMolena: `isAsym<` already returns a value of type `⊥`; there shouldn't be any reason to use `⊥.elim`.

Cubical/Data/NatPlusOne/PropertiesWithInt.agda
```diff
@@ -0,0 +1,24 @@
...
+open import Cubical.Data.NatPlusOne.Base
+open import Cubical.Data.NatPlusOne.Properties
+
+ℕ₊₁→ℤ : ℕ₊₁ → ℤ
+ℕ₊₁→ℤ n = pos (ℕ₊₁→ℕ n)
+
+ℕ₊₁→ℤ-inj : ∀{n}{n'} → ℕ₊₁→ℤ n ≡ ℕ₊₁→ℤ n' → n ≡ n'
+ℕ₊₁→ℤ-inj {1+ n} {1+ n'} zz' = ℕ₊₁→ℕ-inj (injPos zz')
+
+ℕ₊₁→ℤ-1+pred-def : ∀ (x : ℕ) → {{px : NonZeroℕ x}} →
```
LorenzoMolena, line 13: Do we need this -`def` lemma that just returns `refl`?

Cubical/Data/Int/Properties.agda
```diff
@@ -1544,3 +1543,66 @@ sumFinℤHom {n = n} = sumFinGenHom _+_ 0 (λ _ → refl) +Comm +Assoc n
...
+    (a → (b → c)) → ¬ c → (b → ¬ a)
+  contraposition = λ z z₁ z₂ z₃ → z₁ (z z₃ z₂)
+
+·≢0 : ∀ {x}{y} → ¬ x ≡ 0 → ¬ y ≡ 0 → ¬ x · y ≡ 0
+·≢0 {x}{y} nx0 ny0 = contraposition (isIntegralℤ x y) ny0 nx0
+
+·-rightComm : ∀ a b c → a · b · c ≡ a · c · b
+·-rightComm a b c = sym (·Assoc a b c) ∙
+  cong (λ u → a · u) (·Comm b c) ∙ ·Assoc a c b
+
```
LorenzoMolena, line 1560: Do we need all these lemmas? Couldn't we use the ring solver instead?

Cubical/Papers/RepresentationIndependence.agda
reswatson: If possible. I've done this minimally. I can, however, just remove the instance though, or put it in a file not imported by papers.

Cubical/Data/Int/Properties.agda
reswatson: I personally prefer having properties ready to use.

Cubical/Data/Rationals/MoreRationals/SigmaQ/Properties.agda
reswatson: This was why I originally had Normalised separate from Sigma. These are the properties that link Sigma with other rationals. I can put in a proof but it passes through the old Sigma definitions. It's easier just to leave it for now.

reswatson:
@anshwad10 

>     * Merging this might get in conflict with [Faster operations over `Rationals` #1324](https://github.com/agda/cubical/pull/1324). I don't know what is the best order to merge the two; worst case I will try to adapt mine. The conflict could appear in the conversion between normalised and standard rationals

No idea. But see my comment to Lorenzo's comment immediately following yours.
 
>     * In my opinion, there are some proofs that could be made shorter/simpler. should I leave some comment with some suggestions?

Please go ahead with suggestion, or make any changes you think appropriate, or paste any changes here and I'll update them...as you prefer.

Cubical/Data/Rationals/MoreRationals/SigmaQ/Properties.agda
LorenzoMolena: On second thought, since this line is missing and there are no imports of the standard Rationals, there might be fewer conflicts (or even none at all). 
By coincidence, #1324 also introduces a `reduce` function, so I could use it to replace that line with a `reduce : ℚ → Sigma.ℚ` in a future PR

Cubical/Data/NatPlusOne/PropertiesWithInt.agda
reswatson: There may be a way to use refl in ℕ₊₁→ℤ-gcd-def in Cubical.Data.Int.GCD, but I didn't find it easily. A helper function can help. Let me know if there is a better way.

Cubical/Data/Int/Order.agda
reswatson: A simplification made

Cubical/Data/Int/Order.agda
reswatson: Done

reswatson:
**I have made some changes that should now satisfy everybody.**

Have added the equivalent of a reduce in Extras.agda (whether needed or not it is nice to have), this includes cubical equalities between SigmaQ and RationalQ and QuoQ. Other changes too, in response to your suggestions @LorenzoMolena and @anshwad10 (example, I have made Papers pristine by making the instance not an instance and also making it private - and I have put the instance in Extras, so Papers never sees it)

@anshwad10, please feel free to contact me at reswatson/Conversation (open an issue). Maybe we can collaborate.

Cubical/Papers/RepresentationIndependence.agda
reswatson: Resolved. I have returned this to the original and found a workaround.

Cubical/Data/Rationals/MoreRationals/SigmaQ/Properties.agda
reswatson: Doesn't matter. I resolved it anyway. Full functionality returned via Extras.agda file.

Cubical/Data/Int/Order.agda
reswatson: Resolved

Cubical/Data/Int/Order.agda
reswatson: Resolved

Cubical/Data/NatPlusOne/PropertiesWithInt.agda
reswatson: Nothing to resolve  - code needed as is

Cubical/Data/Int/Properties.agda
reswatson: Any further comments and/or actionables on this one?

LorenzoMolena: have you tried to see if similar lemmas were already proved for general (commutative) rings? Just to avoid code duplication.

Cubical/Data/NatPlusOne/PropertiesWithInt.agda
LorenzoMolena: now I understand that this is not really "just `refl`": the instance argument allows us to consider only the successor case, skipping the zero case. 
In my opinion, a more informative name would be better, but that's only a matter of personal taste; I'm also okay with the current one.

reswatson: Resolved

Cubical/Data/Rationals/MoreRationals/SigmaQ/Properties.agda
reswatson: Resolved

Cubical/Data/Int/Properties.agda
reswatson: Thanks Lorenzo, I understand what you mean now. Some changes made. Please review.

LorenzoMolena: Please don't leave the CommMonoid instances in the `Int.Properties` module; the correct place is the `CommMonoid.Instances` subdirectory.
I know that doing so would create a cyclic dependency, but that shouldn't be a problem: 
I was suggesting *removing* these properties from here, and then in the module where you need them, you could just import the general lemmas for commutative monoids/rings instantiated for the integers.
But if this is too much work, I would be okay with leaving the lemmas there, as they were in the previous commits.

Also, I remind you that I am not a maintainer; these are just personal suggestions. Feel free to consider them or not, as I don't want to stall this PR

reswatson: I was wondering about that. I'll do it. I've just seen your PR too. And it has conflicts now as well. I'll look at it all this week if I can.

reswatson: PS Lorenzo: I definitely won't be changing Cubical/Primitives/Core and Papers again as you suggest in your PR on my fork. Anything like that should be a separate PR. Some of the other good changes you suggest are also extraneous to what I am doing here, and should also be done separately.

reswatson: @LorenzoMolena : some of your suggestions in the fork PR should be done as a separate PR. I'll cherry pick the ones that are relevant here.

reswatson: @LorenzoMolena : I think I ended up doing everything you suggested. Please take a look and check it's all OK though.

Cubical/Data/Rationals/MoreRationals/SigmaQ/Base.agda
```diff
@@ -1,53 +1,687 @@
...
+  renaming (-_ to -ℤ_; abs to absℤ)
+open import Cubical.Data.Int.GCD as ℤ
+  using (gcd-def; ℕ₊₁→ℤ-gcd-def; gcdSucNot0)
+open import Cubical.Data.Int.Order as ℤ using
+  ( Positiveℤ ; Negativeℤ ; NonNegativeℤ ; NonPositiveℤ
+  ; isDecPositiveℤ ; isDecNegativeℤ ; isDecNonNegativeℤ ; isDecNonPositiveℤ)
+open import Cubical.Algebra.CommMonoid
 
-open import Cubical.Data.Int.MoreInts.QuoInt
+ℤ·CommMonoid : CommMonoid ℓ-zero
```
LorenzoMolena: Please, move this instance to the appropriate directory, creating a module `Integer` in the path `Algebra.CommMonoid.Instance`.
Otherwise, if you don't mind relying on more complicated algebraic structures, you could use the `RingTheory`/`CommRingTheory` modules, with the `Integer` instance of these structures, as it should already be defined.

Cubical/Data/Int/Properties.agda
LorenzoMolena: I just left a comment about the `CommMonoid` Instance of the integers: it is better to move it to an appropriate module. Other than that, this looks okay to me.

Cubical/Data/Rationals/MoreRationals/SigmaQ/Base.agda
reswatson: DONE - and thanks Lorenzo. I'll shortly mark this thread as resolved.

LorenzoMolena: You're welcome!

PR #1316, opened 2026-06-25
added trivialQuotient->1\inI

Freek98:
added the result that if a quotient is trivial, 1 was in your ideal.


mortberg:
Thanks for the contribution

PR #1317, opened 2026-06-30
added a decomposition of the generated ideal.

Freek98:
In working with ideals, it is sometimes useful to do induction over how a term could have gone into the ideal (for example in  [this proof of Markov's principle](https://github.com/Freek98/FormalizationSSD/blob/main/OmnisciencePrinciples/Markov.agda), or [this proof that quotienting by the empty function doesn't change the ring](https://github.com/Freek98/FormalizationSSD/blob/main/CommRingQuotients/EmptyQuotient.agda).


mortberg:
Looks good to me, so will merge

PR #1318, opened 2026-07-30
Add union of iterative sets

elisesouche:
This pull request adds a construction of unions of iterative sets: from a set x, we can build a set `∪ x` such that `z ∈ ∪ x ≃ ∃[ a ∈ index x ] z ∈ elements x a`.

The main difficulty is related to universe levels. We need to consider the image of a `elements x`, but this is a subset of V⁰, at level ℓ-suc ℓ instead of ℓ. Therefore I introduce machinery to show that this type is actually equivalent to one at level ℓ. This notion of smallness is more general and should maybe be separated from this particular construction.

### DIFF
--- a/Cubical/Data/IterativeSets/Union.agda
+++ b/Cubical/Data/IterativeSets/Union.agda
@@ -1,0 +1,148 @@
+-- from a set x we can build the set ∪ x.
+module Cubical.Data.IterativeSets.Union where
+
+open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.Univalence
+open import Cubical.Foundations.Equiv
+open import Cubical.Foundations.Isomorphism
+open import Cubical.Foundations.HLevels
+open import Cubical.Foundations.Function
+open import Cubical.Functions.Image
+open import Cubical.Functions.FunExtEquiv
+open import Cubical.HITs.Replacement
+open import Cubical.HITs.PropositionalTruncation
+open import Cubical.Displayed.Base
+open import Cubical.Data.Sigma
+open import Cubical.Data.W.W
+open import Cubical.Data.IterativeSets.Base
+open import Cubical.Data.IterativeSets.UnorderedPair.Base
+open import Cubical.Data.IterativeMultisets.Base renaming (index to index∞ ; elements to elements∞)
+open import Cubical.Functions.Embedding
+
+-- TOOD move this notion of smallness somewhere else
+module Smallness where
+  is[_]Small : (ℓ : Level) {ℓ' : Level} (A : Type ℓ') → Type (ℓ-max (ℓ-suc ℓ) ℓ')
+  is[_]Small ℓ A = Σ (Type ℓ) λ B → B ≃ A
+
+  isLocally[_]Small : (ℓ : Level) {ℓ' : Level} (A : Type ℓ') → Type (ℓ-max (ℓ-suc ℓ) ℓ')
+  isLocally[_]Small ℓ A = (x y : A) → is[ ℓ ]Small (x ≡ y)
+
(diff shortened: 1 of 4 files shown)


elisesouche:
@mortberg  @flgrubm

Cubical/Data/IterativeSets/Union.agda
```diff
@@ -0,0 +1,148 @@
...
+open import Cubical.HITs.PropositionalTruncation
+open import Cubical.Displayed.Base
+open import Cubical.Data.Sigma
+open import Cubical.Data.W.W
+open import Cubical.Data.IterativeSets.Base
+open import Cubical.Data.IterativeSets.UnorderedPair.Base
+open import Cubical.Data.IterativeMultisets.Base renaming (index to index∞ ; elements to elements∞)
+open import Cubical.Functions.Embedding
+
+-- TOOD move this notion of smallness somewhere else
```
mortberg: How about making a dedicated file in Foundations for Smallness? It seems like a very fundamental concept

Cubical/Data/IterativeSets/Union.agda
```diff
@@ -0,0 +1,148 @@
...
+  isSmallΠ : ∀ {ℓ ℓ' ℓ''} {A : Type ℓ'} {B : A → Type ℓ''}
+    → is[ ℓ ]Small A
+    → ((a : A) → is[ ℓ ]Small (B a))
+    → is[ ℓ ]Small ((a : A) → B a)
+  isSmallΠ sA sB .fst = (a' : sA .fst) → sB (sA .snd .fst a') .fst
+  isSmallΠ sA sB .snd = equivΠ (sA .snd) (λ a' → sB (sA .snd .fst a') .snd)
+
+open Smallness
+
+-- Some results about smallness of iterative sets
```
mortberg: This could then go in IterativeSets.Base (or maybe there should be Properties file where many things from Base gets moved?)

Cubical/Data/IterativeSets/Union.agda
```diff
@@ -0,0 +1,148 @@
...
+            (λ e → isSmall-≃-isSmall
+              (isSmallΠ (isℓSmallℓ A) (λ a → IH a (e .fst a)))
+              funExtEquiv))
+          (invEquiv ≡V∞-≃-≡Equiv)
+
+  isLocallySmallV : {ℓ : Level} → isLocally[ ℓ ]Small (V⁰ {ℓ})
+  isLocallySmallV {ℓ} x y =
+    isSmall-≃-isSmall (isLocallySmallV∞ (x .fst) (y .fst)) (invEquiv ≡V⁰-≃-≡V∞)
+
+-- Key result: the image of a map of levels ℓ → ℓ' is ℓ-small
```
mortberg: This should go to the HITs.Replacement

mortberg (changes requested):
Some suggestions for reorganization, otherwise this looks great!

Cubical/Data/IterativeSets/Union.agda
elisesouche: I put it in Cubical/Foundations/Smallness.agda

Cubical/Data/IterativeSets/Union.agda
elisesouche: I put it in Cubical/Data/IterativeSets/Base.agda. It would be nice to move things to a Properties file, but in a separate PR?

Cubical/Data/IterativeSets/Union.agda
elisesouche: Done, I kept the name Replacement'

mortberg (approved):
Perfect! Will merge once the CI is finished

PR #1319, opened 2026-08-20
Fix naming in `Nat.Order`

LorenzoMolena:
This PR fixes the naming of lemmas in `Nat.Order` so that they don't contain any free variables, as asked in #1311 (in [this](https://github.com/agda/cubical/pull/1311#discussion_r3813064546) comment). 
The actual changes are only in that module; all the others are edited just to use the new names.
@mortberg please, let me know if you find the new name ok, or if you have better options in mind.


mortberg (approved):
Very good! Thanks for fixing this so fast

PR #1320, opened 2026-08-26
Equivalence between iterative sets and IR cumulative hierarchy

ecavallo:
A favor for @mortberg. Also a few general lemmas that I couldn't find.


mortberg:
Very nice, thanks a lot! It's cool the proof is so short. For some reason I thought it would be harder... 

FYI Håkon's thesis contains a pen-and-paper proof in section B.8 which at a quick glance seems different from this direct approach: https://hakon.gylterud.net/mathematics/gylterud-thesis.pdf

PR #1322, opened 2026-08-27
remove `UsingEq`

### LINKED ISSUE #1321: `UsingEq` might be accidentally exploiting a bug (agda issue 8689)
As reported in the Agda issue [#8689](https://github.com/agda/agda/issues/8689):
> The hole in this issue is accidentally exploited by the development branch of the cubical library, which will need a small patch to minSuc, maxSuc, maxLUB, and minGLB in Cubical.Data.Nat.
> 
> The problem is that Cubical.Foundations.Prelude defines _UsingEq : (a : A) → singl a (computing to (a , refl)) for use in with-abstractions, and the previously mentioned functions are defined by with e UsingEq' which rewrites the refl component of the UsingEq pair in the goal to an ill-typed path (λ _ → w : e' ≡ w). Agda accepted this only because its internal checker did not compare the endpoints of path lambdas. With that fixed, these four definitions are rejected as ill-typed with-abstractions.

@mortberg could you please confirm that this is actually the case? I don't know very well how Agda type-checking actually works.

If that's actually the case, as the author of the PR that introduced `UsingEq`, I first of all apologize for having missed this; second of all, as I explained in a comment of [#8689](https://github.com/agda/agda/issues/8689), I think this could be fixed: "with abstraction + `UsingEq`" has been used only in relation with `<ᵇ` and `≡ᵇ`, so I could refactor the affected modules by using decidability of the types `Bool→Type (m <ᵇ n)` and `Bool→Type (m ≡ᵇ n)`, instead.

LorenzoMolena:
This PR closes #1321 by introducing the decision procedures `_<ᵇ?_` and `_≡ᵇ?_`, and using them instead of `UsingEq` in with-abstractions.

To achieve this, `DecBool→Type` was moved from `Data.Bool.Properties` to `Data.Bool.Base`, as otherwise there would be a cyclic dependency; the alternative would be to define a copy of it locally, but I prefer to avoid code duplication.
Since `_≟_`  was also defined with `UsingEq`, I took this opportunity to rewrite it more easily, but with comparable performance, by introducing `¬<ᵇ→≥ᵇ`.

Finally, `_UsingEq` and `_i0:>_UsingEqP` has been removed from `Foundations.Prelude`.

### DIFF
--- a/Cubical/Data/Bool/Base.agda
+++ b/Cubical/Data/Bool/Base.agda
@@ -60,6 +60,15 @@ Bool→Type* : Bool → Type ℓ
 Bool→Type* true = Unit*
 Bool→Type* false = ⊥*
 
+-- this is proved here to avoid cyclic dependency, as it is needed in `Nat.Properties`
+DecBool→Type : {a : Bool} → Dec (Bool→Type a)
+DecBool→Type {a = true} = yes tt
+DecBool→Type {a = false} = no (λ x → x)
+
+DecBool→Type* : {a : Bool} → Dec (Bool→Type* {ℓ} a)
+DecBool→Type* {a = true} = yes tt*
+DecBool→Type* {a = false} = no (λ (lift x) → x)
+
 True : Dec A → Type₀
 True Q = Bool→Type (Dec→Bool Q)
 
--- a/Cubical/Data/Bool/Properties.agda
+++ b/Cubical/Data/Bool/Properties.agda
@@ -317,14 +317,6 @@ isPropBool→Type* : {a : Bool} → isProp (Bool→Type* {ℓ} a)
 isPropBool→Type* {a = true} = isPropUnit*
 isPropBool→Type* {a = false} = isProp⊥*
 
-DecBool→Type : {a : Bool} → Dec (Bool→Type a)
-DecBool→Type {a = true} = yes tt
-DecBool→Type {a = false} = no (λ x → x)
-
-DecBool→Type* : {a : Bool} → Dec (Bool→Type* {ℓ} a)
-DecBool→Type* {a = true} = yes tt*
-DecBool→Type* {a = false} = no (λ (lift x) → x)
-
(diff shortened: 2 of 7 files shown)


Cubical/Data/Bool/Base.agda
```diff
@@ -60,6 +60,15 @@ Bool→Type* : Bool → Type ℓ
 Bool→Type* true = Unit*
 Bool→Type* false = ⊥*
 
+-- this is proved here to avoid cyclic dependency, as it is needed in `Nat.Properties`
+DecBool→Type : {a : Bool} → Dec (Bool→Type a)
```
mortberg, line 64: Why is a implicit here and in the next definition? I would assume that one always have to give the explicitly, so it would better to have them explicit

mortberg (changes requested):
Thanks for removing this so fast. 

It is unfortunate that such an innocent addition can be used to prove false... It would be good to fix the issue in Agda

Cubical/Data/Bool/Base.agda
LorenzoMolena: I just moved the definition here without any changes. 
Since `Data.Bool` exports both `Data.Bool.Base` and `Data.Bool.Properties`, there might be multiple other cases where it has been used before.
But if needed, I can check whether the implicit argument can be inferred often; if not, I could turn it explicit.

LorenzoMolena: I checked, and `DecBool→Type` is also used in `Data.Bool.Properties`, `Data.Fast.Int.Order`, `Data.Finset.Quotients` and `Relation.Nullary.DecidablePropositions`; only in the first module the implicit argument is given explicitly. 
That said, these are not really a lot of modules, so I am not opposed to making it explicit. 
Please let me know what approach you would prefer.

mortberg: Ok, then I guess we can leave it as is

PR #1323, opened 2026-09-05
Unify the Inductive and Recursive Order

LorenzoMolena:
There were only a few modules using `Nat.Order.Recursive` instead of `Nat.Order.Inductive`.
Indeed, in the latter module, there was a comment saying "TODO: unify with recursive.agda", and this PR does exactly that.
In addition, some names have been changed, similarly to PR #1319, so that there is no mention of free variables.

Finally, this PR changes the implementation of the `search` algorithm for finding the least natural number satisfying a decidable property, and introduces an alternative and more efficient proof of trichotomy of `<ᵗ`. In more detail: 
- Currently, to compute `→Least dec (n , Pn)`, it takes `n` recursive calls, whilst with the change introduced here, it takes `fst(→Least dec (n , Pn))` recursive calls, i.e., the least number satisfing the property is also the number of recursive calls, making it more efficient when the least value is much smaller than the upper bound supplied. This is inspired by [1Lab](https://github.com/the1lab/1lab/blob/b65d3e28660289cac0ffbb2b9eec83334dda8c37/src/Data/Nat/Order.lagda.md#L262)
- The new proof of trichotomy, named `_≟ᶠ_`, computes by relying on the **f**ast decidability of boolean builtins. However, multiple modules were using a specific definitional identity of the current (and slower for concrete big natural numbers) implementation. For this reason, `_≟ᶠ_` is added as an alternative, without replacing the previous one, which remains the default.

This merge happens in `Nat.Order.Inductive`, deleting `Nat.Order.Recursive`, and changing an import in `Bool.Properties` in order to avoid cyclic dependencies. 
All the other modules have been edited for compatibility with the new names and/or to import the `Inductive` order rather than the `Recursive` one.


mortberg (approved):
Looks good to me! Unless someone has something against this I will merge later today

mortberg:
> Looks good to me! Unless someone has something against this I will merge later today

Ping @felixwellen @aljungstrom @ecavallo

LorenzoMolena:
@mortberg I fixed the merge conflicts with the master. No new changes have been added. 
This is ready to review and to be merged, in case it is okay to you too, of course.

PR #1324, opened 2026-09-06
Faster operations over `Rationals`

LorenzoMolena:
This PR redefines the operations and relations in `Data.Rationals`, making them rely on `Data.Fast.Int`; it also includes several changes in the `Rationals.Order` subdirectory.
I chose to edit the standard implementation of the rationals, instead of introducing yet another folder with a new encoding, for the following reasons:
- the type of integers is the same (only the implementation of the operations is changed), and other than the import with the keyword `Fast`, there are not many differences when using the fast integers
- Rational numbers are not used nearly as much in the library, compared to the integers, so it is possible to change the implementation without fixing a lot of proofs.

Other than that, this PR introduces records for set-quotient eliminators/recursors. The old eliminator/recursor functions have been redefined to call `go` from the associated record (for backward compatibility and to allow users to use functions instead of the new records, if they prefer). 
Multiple display pragmas have been used to print shorter names in goals.
Similarly, the functions `onCommonDenom` and `onCommonDenomSym` have been turned into records.
 
Other than the use of the orderings in fast rationals, the changes in `Rationals.Order` are:
- use the new record `Rec2SymHProp` to simplify the definitions of the order relations
- use of records wrapping the underlying type of `_<'_` and `_≤'_`, so that the arguments can be inferred more easily
- added `recompute<` and similar helper functions

This solves #1124, but I suggest to @onestruggler to define the operations on `ℚ[ω]` with some application of `reduce`: with this PR, their example will compute instantly, but the result is a huge unnormalized fraction.
For this reason, the `reduce` function has also been introduced, together with the proof that `reduce` is extensionally equal to the identity function

The choice of "redefinition with records" comes from @marcinjangrzybowski, and eventually, it will allow us to get more reliable solvers for rationals.

### DIFF
--- a/Cubical/Algebra/Field/Instances/Rationals.agda
+++ b/Cubical/Algebra/Field/Instances/Rationals.agda
@@ -14,7 +14,7 @@ open import Cubical.Algebra.CommRing.Instances.Rationals
 open import Cubical.Algebra.Field
 
 open import Cubical.Data.Empty as Empty
-open import Cubical.Data.Int as ℤ
+open import Cubical.Data.Fast.Int as ℤ
 open import Cubical.Data.Nat as ℕ using (ℕ ; zero ; suc)
 open import Cubical.Data.NatPlusOne
 open import Cubical.Data.Rationals as ℚ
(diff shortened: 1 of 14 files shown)


mortberg (approved):
This seems like a welcome change to me. Should I merge? @felixwellen @aljungstrom @ecavallo

PR #1325, opened 2026-09-08
Add fixity declaration to imported primitive interval operations

andreasabel:
The fixity declarations in `Agda.Primitive.Cubical` will likely disappear in the future, so the fixities should be established in the `renaming` directive.  (They also make more sense there because e.g. `primIMin` is not syntactially an operator, while `_∧_` is one.)

See agda/agda#1438 .

Note: the syntax used in this PR goes back to Agda 2.6.1, so should not be a limitation for the use of the `cubical` library with relevant older versions of Agda.


mortberg:
Thank you Andreas! I'll merge

PR #1327, opened 2026-09-16
Make some implicit arguments explicit for agda/agda#8749

andreasabel:
Tweaking the type checker for agda/agda#8749 , these cosmetic modifications (generated by Claude Code) are needed to pass.

I tested to make sure they do not break build with agda-2.8.0 or current agda master.


mortberg:
Looks good to me

